PR #346, opened 2014-12-10
[ADD] Odoo official image

aab-odoo:


aab-odoo:
Thanks for the review. I agree with your comments. Requested changes are applied.

library/odoo
```diff
@@ -0,0 +1,4 @@
+# maintainer: Aaron Bohy <aab@odoo.com> (@aab-odoo)
+
+8.0: git://github.com/odoo/docker@2aae53084d32aaf44d3630e2a09d685bf971d81c 8.0
+8: git://github.com/odoo/docker@2aae53084d32aaf44d3630e2a09d685bf971d81c 8.0
```
aab-odoo: Actually, v8 is not the latest version. I put the tag `stable` instead. I also fixed the reference.

tianon: "latest" is actually a misnomer, unfortunately.  "latest" doesn't actually
mean "latest".  It really should've been "default" or something more to
that effect.  That's just the tag that will be used when someone does
"docker run odoo" without a tag -- it'll default to running "odoo:latest",
which should be the default recommended version.  For example, in the
Ubuntu image, "latest" is set to the latest LTS, not the latest release
(14.04 vs 14.10).

aab-odoo: Ok I changed it to latest then.

yosifkit:
Build test of #346; f43ef5b4662cdb6169cb96ba6fb0396ec436275c (`odoo`):

```console
$ ./bashbrew/bashbrew.sh build https://raw.githubusercontent.com/docker-library/official-images/f43ef5b4662cdb6169cb96ba6fb0396ec436275c/library/odoo
Processing odoo:8.0 ...
Processing odoo:8 ...
Processing odoo:latest ...
```

tianon:
:tada: :cake:

PR #352, opened 2014-12-16
Add WebSphere Liberty image

davidcurrie:
Adds an image for IBM WebSphere Application Server for Developers Liberty Profile.

### DIFF
--- a/library/websphere-liberty
+++ b/library/websphere-liberty
@@ -1,4 +1,4 @@
-# maintainer: IBM <https://developer.ibm.com/wasdev/help/> (@WASdev)
+# maintainer: David Currie <david_currie@uk.ibm.com> (@dcurrie)
 
-8.5.5: git://github.com/WASdev/ci.docker@1b66f8e599f11b45d6b0db65d5e71fb853c63b9a websphere-liberty/8.5.5
-latest: git://github.com/WASdev/ci.docker@1b66f8e599f11b45d6b0db65d5e71fb853c63b9a websphere-liberty/8.5.5
+8.5.5: git://github.com/WASdev/ci.docker@f7c3cf46801e5748b6a9227a8c8e5ad3f544fc4a websphere-liberty/8.5.5
+latest: git://github.com/WASdev/ci.docker@f7c3cf46801e5748b6a9227a8c8e5ad3f544fc4a websphere-liberty/8.5.5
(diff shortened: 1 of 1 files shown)


davidcurrie:
Sorry about the delay. I believe I've now factored in all of the feedback related to the Dockerfile and docs including addressing the issues @psftw had with the usage examples.

yosifkit:
Build test of #352; aac9f61533f3dd66835133e08c0f7d45bb499fb6 (`websphere-liberty`):

```console
$ ./bashbrew/bashbrew.sh build https://raw.githubusercontent.com/docker-library/official-images/aac9f61533f3dd66835133e08c0f7d45bb499fb6/library/websphere-liberty
Cloning websphere-liberty (git://github.com/WASdev/ci.docker) ...
Processing websphere-liberty:8.5.5 ...
Processing websphere-liberty:latest ...
```

yosifkit:
Well, I seem to have forgotten this was still here.  I think the big question is if those yml files will be consistent enough to keep say version 8.5.5 on it when 8.5.6 comes out so it does not immediately break builds here until a new PR is merged to fix it (same for the IBM Java version).

davidcurrie:
@yosifkit - you must have heard me typing my 'any update?' comment! For new releases (e.g. 8.5.0 to 8.5.5) I don't believe it should be a problem to ensure that the previous version remains at least until a new PR is merged. (The 8.5.0 release stream is still in support for example.) For fix packs (e.g. 8.5.5.3 to 8.5.5.4) it is true that, up until now we have taken away the old version when the new version is added. This is because our other current usage scenarios just look for the latest fix pack for a given release. I can ensure that the update process is changed so that this does not happen. Alternatively, I could add this behavior back in to the Dockerfile (which would then mean it more accurately reflected the tag) but would then need some other mechanism to trigger a rebuild when a new fix pack comes along. Thoughts?

yosifkit:
There would not be an easy way to trigger an update if the line looks for any 8.5.5.* version in the yml since that specific RUN line would not have changed and would still be in docker cache, so what you have looks correct.  If the builds break here because the yml changed for an update (if ubuntu was updated also, breaking the docker cache), I don't think that will cause too much pain as long as the PR updates are posted in a timely fashion.  What do you think @tianon, does this seem reasonable?

tianon:
The part that worries me is "... would then need some other mechanism to trigger a rebuild when a new fix pack comes along."

A PR to this repo is the only real "rebuild trigger".  We're not going to notice some remote URL updated and automatically rebuild based on that.  Does that make more sense?

tianon:
Although, if the builds break, I suppose that's not a huge tragedy, especially since we'll still have the old tags up.  Fair enough.

davidcurrie:
OK - so sounds like we should stick with the current plan. I've just checked and it looks like we do keep the binaries up - I just need to make sure that we keep at least N and N-1 referenced in the YML.

tianon:
:heart: LGTM

yosifkit:
Build Test is still good. LGTM

PR #375, opened 2014-12-27
Add image for rethinkdb

stuartpb:
I've put a lot of effort into this to produce a set of self-contained Dockerfiles for RethinkDB with a tag for every patch release going back three versions.

No docs yet - PR for that will be incoming.

### DIFF
--- a/library/rethinkdb
+++ b/library/rethinkdb
@@ -1,18 +1,7 @@
 # maintainer: Stuart P. Bentley <stuart@testtrack4.com> (@stuartpb)
 
-1.13.0: git://github.com/stuartpb/rethinkdb-dockerfiles@947a3f157aa355550efa3334f17ffc0e813c4165 1.13.0
-1.13.1: git://github.com/stuartpb/rethinkdb-dockerfiles@947a3f157aa355550efa3334f17ffc0e813c4165 1.13.1
-1.13.3: git://github.com/stuartpb/rethinkdb-dockerfiles@947a3f157aa355550efa3334f17ffc0e813c4165 1.13.3
-1.13.4: git://github.com/stuartpb/rethinkdb-dockerfiles@947a3f157aa355550efa3334f17ffc0e813c4165 1.13.4
-1.13: git://github.com/stuartpb/rethinkdb-dockerfiles@947a3f157aa355550efa3334f17ffc0e813c4165 1.13.4
-
-1.14.0: git://github.com/stuartpb/rethinkdb-dockerfiles@947a3f157aa355550efa3334f17ffc0e813c4165 1.14.0
-1.14.1: git://github.com/stuartpb/rethinkdb-dockerfiles@947a3f157aa355550efa3334f17ffc0e813c4165 1.14.1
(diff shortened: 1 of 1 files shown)


psftw:
I took a look at the latest Dockerfile for 1.15.2 in the master branch which includes the changes from yosifkit, and it LGTM.  This PR looks to pull a different branch and could be updated.

stuartpb:
Whoops, I forgot how PRs for this repo work, what with the explicitly-coded hashes.

stuartpb:
The PR is pulling from the tip of `master` again.

yosifkit:
Whoops, 13.*, 14.*, and 15.0 all fail to build.  It seems these versions are based on `ubuntu:trusty` but are adding `jessie` in the `echo > apt/sources` line.  Is there a reason they are all not just the same distro (either all debian or all ubuntu)?

15.1 and 15.2 build just fine. :smile:

stuartpb:
Whoops, I'll fix that.

The ones that are packaged upstream for Debian are Debian, while the ones that are only packaged upstream for Ubuntu are Ubuntu.

yosifkit:
I see debian jessie (or wheezy) only has these [versions](http://download.rethinkdb.com/apt/pool/jessie/main/r/rethinkdb/), while ubuntu trusty has these [ones](http://download.rethinkdb.com/apt/pool/trusty/main/r/rethinkdb/).  With ubuntu having all of the versions, I would think it would be better to either move all of the versions to ubuntu, or just only have the debian versions (my preference).

With it being 2 distros I am very hesitant on having that within one repo; some people start could using an old version and then migrating might be confused when they try to add a package (unlikely, but a possibility that should be accounted for).

stuartpb:
I'd be fine providing only the debian versions in the official images. I could still provide the Ubuntu-based versions, for cases where old migrations are necessary, as stuartpb/rethinkdb.

stuartpb:
I've updated the PR to include the latest version of RethinkDB, and not include the Ubuntu-exclusive versions.

yosifkit:
This looks good, but you should also tag `1.15.3` :wink:

stuartpb:
:facepalm: Can't believe I missed that. I've been a little distracted this evening/morning...

tianon:
LGTM

tianon:
:heart:

yosifkit:
It's all good. You could copy and modify one of our `generate-stackbrew-library.sh` scripts for use next time.

LGTM
Build test of #375; 2e69422ac7a0c63f958df02304e9ff588057b839 (`rethinkdb`):

```console
$ ./bashbrew/bashbrew.sh build https://raw.githubusercontent.com/docker-library/official-images/2e69422ac7a0c63f958df02304e9ff588057b839/library/rethinkdb
Processing rethinkdb:1.15.1 ...
Processing rethinkdb:1.15.2 ...
Processing rethinkdb:1.15.3 ...
Processing rethinkdb:1.15 ...
Processing rethinkdb:1 ...
Processing rethinkdb:latest ...
```

PR #382, opened 2015-01-01
Adjust bashbrew to fail on dups

tianon:
This is going to fail in Travis until I fix Debian, but now Travis will actually fail for this kind of crap giving us a heads up that we're Doing It Wrong.


yosifkit:
LGTM

tianon:
Rebased now that #383 is merged and it passes like a champ. :+1:

PR #397, opened 2015-01-13
Remove the explicit "latest-" from perl's "latest-threaded"

tianon:
For all the images, "latest" is a convenience and even a misnomer.  It's really "default", and only exists such that users doing "just-the-image" get a useful tag, so users shouldn't be explicitly using "the-image:latest" and thus "the-image:latest-variant" doesn't make much sense, but "the-image:variant" does and can (and this matches the pattern of all the other official image tags too).

ping @PeterMartini


PeterMartini:
Noted for the future; thanks for the clarification

tianon:
So you're not opposed to this change? :)

PeterMartini:
Hehe of course not, feel free and thanks for the help :-)

tianon:
:+1: :heart:

cc @yosifkit

yosifkit:
LGTM

PR #404, opened 2015-01-16
add BASHBREW_NAMESPACES support

psftw:
motivation: on my local machine, I'm not interested in any extra namespaces, and would rather set an ENV variable than write long shell incantations.

### DIFF
--- a/bashbrew/bashbrew.sh
+++ b/bashbrew/bashbrew.sh
@@ -9,7 +9,7 @@ dir="$(dirname "$(readlink -f "$BASH_SOURCE")")"
 library="$dir/../library"
 src="$dir/src"
 logs="$dir/logs"
-namespaces='_ library stackbrew'
+namespaces=${BASHBREW_NAMESPACES-_ library stackbrew}
 docker='docker'
 
 library="$(readlink -f "$library")"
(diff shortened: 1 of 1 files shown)


tianon:
I've got some mixed feelings here.  While I appreciate the need/desire to configure the script with saner per-env defaults, I'm just not a fan of proliferating environment variables to do it, because then where does it stop?  Where's the line?

psftw:
It doesn't have to be an environment variable, so long as I can avoid typing ``--namespaces="_"`` to avoid the extra tags.  This just scratches an itch for me, so I won't be too sad if you reject it.  I would imagine that most people who are testing their image builds using bashbrew aren't actually interested in the extra tags either.

tianon:
I wouldn't be opposed to changing the default either, as long as we have a sane way to configure the official build server in the other direction.  This just needs more thought, I think, since environment variables like this are untenable IMO. :+1:

psftw:
Yes, even better in my mind if we just make the default ``--namespaces="_"``.

yosifkit:
I am +1 with changing the default to just the empty namespace. It would make more sense that namespaces is the extra namespaces that I want auto-tagged or listed, or the namespaces I want pushed.  Notice that there is a subtle difference on push, but it has always been there and it is required.

yosifkit:
Wait maybe I am wrong, it should probably be all the namespaces, not just the extra.

psftw:
Closing, another PR incoming..

PR #407, opened 2015-01-17
change default --namespaces '_'

psftw:
motivation: I don't want to have to specify this to avoid additional namespace tags.

### DIFF
--- a/bashbrew/bashbrew.sh
+++ b/bashbrew/bashbrew.sh
@@ -9,7 +9,7 @@ dir="$(dirname "$(readlink -f "$BASH_SOURCE")")"
 library="$dir/../library"
 src="$dir/src"
 logs="$dir/logs"
-namespaces='_ library stackbrew'
+namespaces='_'
 docker='docker'
 
 library="$(readlink -f "$library")"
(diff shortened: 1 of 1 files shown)


yosifkit:
:cake: :tada: LGTM

tianon:
https://github.com/docker-library/official-images/pull/404#issuecomment-70316179

> ... as long as we have a sane way to configure the official build server in the other direction.

So, what ideas do you have? :wink:

yosifkit:
Configure? Isn't it just: `./bashbrew.sh build --namespaces='_ library stackbrew' ...`

tianon:
For every time we run it?

tianon:
I suppose we could just keep the opposite of this patch as an unstaged change in Git on the server.  That'll make updates to this script more annoying to apply (since we'll have to do the `stash`, `pull`, `stash pop` dance), but I'm OK with that if you are.

yosifkit:
What do you have in mind?  Would we really want an ENV or config file to change it?

tianon:
Definitely not an environment variable.  I was thinking about a config file, but that feels like overkill for now.  Maybe we go with this for now and revisit the debate later when/if it turns out to be annoying? :smile:

yosifkit:
That works :+1:

tianon:
:smiling_imp:

LGTM

PR #409, opened 2015-01-19
Add image for io.js

pesho:
Adds an [io.js](https://iojs.org/) official image.

I hope it's acceptable to have a http link instead of an email in the `maintainer` line.

### DIFF
--- a/library/iojs
+++ b/library/iojs
@@ -1,0 +1,16 @@
+# maintainer: io.js Docker team <https://github.com/iojs/docker-iojs> (@iojs)
+
+1.0.2: git://github.com/iojs/docker-iojs@0dccce1ac9338bd36678d81d8a08a9b3ac381645 1.0
+1.0: git://github.com/iojs/docker-iojs@0dccce1ac9338bd36678d81d8a08a9b3ac381645 1.0
+1: git://github.com/iojs/docker-iojs@0dccce1ac9338bd36678d81d8a08a9b3ac381645 1.0
+latest: git://github.com/iojs/docker-iojs@0dccce1ac9338bd36678d81d8a08a9b3ac381645 1.0
+
+1.0.2-onbuild: git://github.com/iojs/docker-iojs@269b813b6843c20cd93cc901bfd7f59640078e12 1.0/onbuild
+1.0-onbuild: git://github.com/iojs/docker-iojs@269b813b6843c20cd93cc901bfd7f59640078e12 1.0/onbuild
+1-onbuild: git://github.com/iojs/docker-iojs@269b813b6843c20cd93cc901bfd7f59640078e12 1.0/onbuild
+onbuild: git://github.com/iojs/docker-iojs@269b813b6843c20cd93cc901bfd7f59640078e12 1.0/onbuild
+
+1.0.2-slim: git://github.com/iojs/docker-iojs@0dccce1ac9338bd36678d81d8a08a9b3ac381645 1.0/slim
+1.0-slim: git://github.com/iojs/docker-iojs@0dccce1ac9338bd36678d81d8a08a9b3ac381645 1.0/slim
+1-slim: git://github.com/iojs/docker-iojs@0dccce1ac9338bd36678d81d8a08a9b3ac381645 1.0/slim
+slim: git://github.com/iojs/docker-iojs@0dccce1ac9338bd36678d81d8a08a9b3ac381645 1.0/slim
--- a/library/iojs
+++ b/library/iojs
@@ -1,16 +1,16 @@
 # maintainer: io.js Docker team <https://github.com/iojs/docker-iojs> (@iojs)
 
-1.0.2: git://github.com/iojs/docker-iojs@ebe290918a3f5f8eb8e74bd4ea9fdf8fcabbb2d8 1.0
-1.0: git://github.com/iojs/docker-iojs@ebe290918a3f5f8eb8e74bd4ea9fdf8fcabbb2d8 1.0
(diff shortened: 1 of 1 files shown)


jfrazelle:
Awesome!! :dancer:

josephpage:
+1

jfrazelle:
The only thing I am confused by is why python is a dependency for iojs, but not node https://github.com/iojs/docker-iojs/blob/master/1.0/Dockerfile#L7 & https://github.com/joyent/docker-node/blob/master/0.10/Dockerfile

pesho:
It's a requirement for [node-gyp](https://github.com/TooTallNate/node-gyp), which is Python-based and is needed for building native modules.

Note that the node image also includes Python, it's included in `buildpack-deps:jessie`. Our `Dockerfile` is just more explicit about it. 

cc @wblankenship (if I missed something)

tianon:
Could we use the full fingerprint instead of just the long ID in the `gpg --recv-keys` line? (ie, `DD8F2338BAE7501E3DD5AC78C273792F7D83545D` instead of just `C273792F7D83545D`)

Also, I love that you've got a slim version for people who want to bring their own dependencies by hand, but I'm confused why it still explicitly has `build-essential`, `pkg-config`, `git`, and `python` included in the image by default -- it looks like it might be because the `node` slim variant did, which is valid, but it does seem fatter than it needs to be (since people who want packages with native deps could install `python` and `gcc` themselves, but for people who don't need them, it's already too late to remove them).  Since `node:slim` is already the same, this won't be a blocker for inclusion, but I want to make sure we start the discussion and maybe even get some improvement going. :+1:

tianon:
I'm also not thrilled about having a URL in the maintainer line, but there is precedent for it from @aws (https://github.com/docker-library/official-images/blob/master/library/glassfish), so I won't let that be a blocker either.

pesho:
> Could we use the full fingerprint instead of just the long ID in the gpg --recv-keys line? (ie, DD8F2338BAE7501E3DD5AC78C273792F7D83545D instead of just C273792F7D83545D)

For some reason that didn't work with @rvagg's key initially (probably the keyserver pool wasn't properly updated yet, the key was freshly published). But it works now, and I will change it.

> why it still explicitly has build-essential, pkg-config, git, and python included in the image by default

`build-essential` - installing a native module from npm will definitely need this
`git` - npm supports fetching from git/GitHub repositories, it's a common usecase
`python` - most native modules will require this to build properly, because of `node-gyp`
`pkg-config` - this one I'm not sure about

Installing native/binary modules is very common. Most database drivers need native modules. Requiring users to install these themselves would cause a lot of questions.

We could probably use a smaller image as a base for the default variant, e.g. `buildpack-deps:jessie-micro`. Experimenting with that is definitely on the roadmap.

wblankenship:
If your package has native modules in it's dependency tree, python will be needed to install them (same with gcc). `git` is a dependency of npm as `package.json` can specify git urls as dependencies.

Perhaps moving these out of slim would be fine (and desirable), but would need a disclaimer that `npm install` could fail on that image.

pesho:
> But it works now, and I will change it.

Updated the PR with full GPG key fingerprints.

yosifkit:
Recently learning that `build-essential` is not what I had thought it was (https://github.com/docker-library/buildpack-deps/pull/15#issuecomment-70310120), I am reluctant to include it anywhere since it is really a set of packages specificially for building debian packages.   This is most important on the slim versions since they are supposed to be the bare minimum for running the software.  It is designed to be for advanced users that are supposed to add only the packages that they need.  If we add too much, then the user doesn't gain much by going slim.

pesho:
Thanks for the good suggestions regarding the image size. I've opened [a separate issue](https://github.com/iojs/docker-iojs/issues/9) to discuss them.

Is this considered a blocker for this PR? The current dependencies are exactly the same as for `node` (except moving from wheezy to jessie), and `node` is already in the library.

yosifkit:
Thanks @pesho, that sounds reasonable to me.

@tianon, should we do a build test and move to merge?

tianon:
Yeah, I think so.  Just for reference, these commit hashes are from https://github.com/iojs/docker-iojs/pull/8, which I'm guessing is only not merged into `master` yet in case there were more changes that came from this review? :smile:

pesho:
@tianon yup, that PR is intended to be merged only after (and if) the image becomes official :)

tianon:
Nice, LGTM then

yosifkit:
LGTM

Build test of #409; 5acac20fba28d95f21c6c9e60176e1e3f51f8523 (`iojs`):

```console
$ ./bashbrew/bashbrew.sh build https://raw.githubusercontent.com/docker-library/official-images/5acac20fba28d95f21c6c9e60176e1e3f51f8523/library/iojs
Cloning iojs (git://github.com/iojs/docker-iojs) ...
Processing iojs:1.0.2 ...
Processing iojs:1.0 ...
Processing iojs:1 ...
Processing iojs:latest ...
Processing iojs:1.0.2-onbuild ...
Processing iojs:1.0-onbuild ...
Processing iojs:1-onbuild ...
Processing iojs:onbuild ...
Processing iojs:1.0.2-slim ...
Processing iojs:1.0-slim ...
Processing iojs:1-slim ...
Processing iojs:slim ...
```

rvagg:
FYI there _will_ be other releasers so there will be more than just my GPG key to fetch, but we will keep the io.js Docker team updated on this and will probably list them all on the io.js README

tianon:
@rvagg SGTM :+1:

PR #414, opened 2015-01-21
Update for node 0.11.15 and npm 2.2.0

dcrudgington:


chorrell:
Looks like we need to add another gpg fingerprint to the dockerfiles for the 0.11.15 release.

Lets drop this and #415 (which was a semi-fix for this pull) since we'll need the above resolved first

PR #421, opened 2015-01-23
docker-library/php versions bumped

rogeriopradoj:
https://github.com/docker-library/php/commit/c871678cc7c8dbc9277832e35f14f3534f2dd0fb and https://github.com/docker-library/php/issues/65


yosifkit:
I do not get this same file when I run `./generate-stackbrew-library.sh` in the php repo.  Also, I just pushed the bump for 5.4.

tianon:
See #425

PR #422, opened 2015-01-23
Add biicode image to dockerhub

biicode:
Hi Docker, I'm David and I write you from [biicode](https://www.biicode.com/) a small start-up that’s developing a C and C++ dependency manager.

We are interested in creating a official repository so we have done 2 MR, One to the docs repo with the biicode image info and another MR to the official-images repo with the image commit. Is this all we need to do?.

If anything is wrong or we skipped something in the way please let us know and we will fix it at soon as possible.

[Dockerfile Repo](https://github.com/biicode/docker-biicode)

### DIFF
--- a/library/biicode
+++ b/library/biicode
@@ -1,0 +1,4 @@
+# maintainer: David Sanchez <david.sanchez@biicode.com> (@davidsanfal)
+
+latest: git://github.com/biicode/docker-biicode@67dc087553e185f9834fbeb03fe82a753d77a2bb
+2.4.1: git://github.com/biicode/docker-biicode@67dc087553e185f9834fbeb03fe82a753d77a2bb


psftw:
I sent over some minor changes via https://github.com/biicode/docker-biicode/pull/1
I also see that there has been a new release since this PR was created.  You'll need to think about the versions you'll support and be explicit about at least one in addition to latest (which can be the same thing).  The idea is that every image build refers to a specific version and users can choose between "latest" or a specific version.

I see that you have passwordless sudo for the biicode user.  Is that strictly required by bii, or could the Dockerfile be updated to not require it?

davidsanfal:
Hi, thanks for your changes, I had never  thought of the memory consumption.

We usually launch a new version every one or two weeks with bugs fixes and with minor or even major improvements, so we prefer to only support the last one.

However, if you prefer to have all the versions, we can do it because we store all the previous packages, but they could have bugs nonetheless.

Of course, after launching a release, we will try to update the library repository as soon as possible.

About the necessity of a passwordless sudo for the biiuser. We decided to include it because the tools installed with the bii cpp:setup comand need to be installed as sudo by a regular user because it is installed in the ~/ folder, Is there a better way to do this?

tianon:
I'm definitely +1 to only including the versions that are supported, and would ask you to remove versions that won't ever actually be updated later anyhow.

As for `sudo`, why do you need `sudo` in a `Dockerfile`?  The `USER` instruction accomplishes the same goal, but without an extra package installed in the image itself.

davidsanfal:
I just updated the Dockerfile with your recommendations. I hadn't work before with users, so this was new for me, thanks for the info.

In the current version we just create the `home/biiuser` folder and add it to the `ENV HOME` and use root user as default user. 

The real problem with *'act as sudo'* is that there are some blocks in biicode using hooks to install some packages with `sudo apt-get install ...` and we need to support *'act as sudo'* to use these blocks inside the Docker's image.

psftw:
I don't have any major issues with the latest version.  In my opinion there isn't that much difference between having passwordless sudo and running as root.  BTW, with ubuntu:14.04 ``sudo`` comes out of the box.  As part of updating the image with new releases, you may want to check in to using a new naming scheme for your downloads so that you could extract out the version into one ENV variable.  Not a big deal, just a suggestion in case you are lazy like me.

tianon:
Thanks for taking out the `sudo` bits!  I feel much better about how this Dockerization looks now.  Running a build test!

tianon:
Arg, looks like we took too long to review.  Can you update?

```
Step 6 : RUN bii setup:cpp
 ---> Running in b2e17713d240
INFO: There is a new version of biicode. Download it at https://www.biicode.com/downloads
ERROR: Your current version is deprecated and won't work any longer
time="2015-02-13T11:52:31-07:00" level="info" msg="The command [/bin/sh -c bii setup:cpp] returned a non-zero code: 1" 
```

If you've got a good clean way to get the "latest version number" of biicode, I'd be happy to write an `update.sh` script similar to what we've got in some of our own repos to help keep this up to date more easily! (https://github.com/docker-library/wordpress/blob/master/update.sh, https://github.com/docker-library/python/blob/master/update.sh, https://github.com/docker-library/mongo/blob/master/update.sh)

tianon:
https://webapi.biicode.com/v1/downloads certainly looks promising! :+1:

Moghedrin:
I'm going to go ahead and close this. If anything changes, let us know here and we can reopen as needed. :)

PR #430, opened 2015-01-27
Aerospike

volmarl:
adding Aerospike official images.


volmarl:
will squash this and resubmit.

PR #432, opened 2015-01-27
Update buildpack-deps URL

tianon:
Was looking through Dockerfiles and realized we'd moved this repo quite some time ago.


tianon:
Updated to include `mysql` too.

tianon:
Since this is just a URL change, Travis is sufficient and I'm not doing a full build-test. :wink:

yosifkit:
LGTM

PR #440, opened 2015-01-29
Add swarm to official images.

aluzzardi:
This is in order to distribute `swarm` as an official image so that it can be used as simply as:

`docker run swarm`

### DIFF
--- a/library/swarm
+++ b/library/swarm
@@ -1,0 +1,5 @@
+# maintainer: Andrea Luzzardi <al@docker.com> (@aluzzardi)
+# maintainer: Victor Vieux <vieux@docker.com> (@vieux)
+
+0.1.0-rc3: git://github.com/docker/swarm-library-image@c0f8204b2a1bb34233cd2144df51a4de9572440c
+latest: git://github.com/docker/swarm-library-image@c0f8204b2a1bb34233cd2144df51a4de9572440c


aluzzardi:
Note that the `0.1.0` image tag currently points to `0.1.0-rc2`.

Eventually, we are going to update that to `0.1.0` as well but didn't want to create `rc` image tags.

Thoughts?

vieux:
@aluzzardi I agree.

psftw:
Really excited to see this.  I sent over some changes via https://github.com/docker/swarm/pull/330
In general, the image that is produced is still a bit large since it brings in the whole stack in order to build it.  It would be worth looking into a smaller image that has just enough OS and downloads the release binary?

Also, while I'm not the official word on it, I don't like the idea of overwriting the 0.1.0 release tag with the actual 0.1.0 release when it is ready.  In my mind, we should call this explicitly an -rc2 tag.  Users will likely just run latest anyway (at least I plan to) unless there is a reason to support multiple versions at some point.

yosifkit:
+1 on @psftw's changes.  If we want a really tiny dockerfile for releases (the current one is great for building), we could do something like the following:

```Dockerfile
FROM debian:jessie

RUN apt-get update \
	&& apt-get install -y --no-install-recommends ca-certificates \

ENV SWARM_VERSION 0.1.0-rc2

RUN apt-get update \
	&& apt-get install -y --no-install-recommends curl \
	&& rm -rf /var/lib/apt/lists/* \
	&& curl -SL "https://github.com/docker/swarm/releases/download/v${SWARM_VERSION}/docker-swarm_linux-amd64" -o /usr/local/bin/swarm \
	&& chmod +x /usr/local/bin/swarm \
	&& apt-get purge -y --auto-remove curl

CMD [ "swarm" ]
```

This comes in at ~142 MB.

vieux:
Why is `CMD [ "swarm" ]` better than entrypoint ?

tianon:
> The best use for `ENTRYPOINT` is as a helper script. Using `ENTRYPOINT` for other tasks can make your code harder to understand. For example,
> 
>     docker run -it official-image bash
> 
> is much easier to understand than
> 
>     docker run -it --entrypoint bash official-image -i
> 
> This is especially true for new Docker users, who might naturally assume the above command will work fine. In cases where an image uses `ENTRYPOINT` for anything other than just a wrapper script, the command will fail and the beginning user will then be forced to learn about `ENTRYPOINT` and `--entrypoint`.

aluzzardi:
Re: Large images. /cc @psftw @yarikoptic 

I entirely agree, the current image is "fat" for what it's supposed to do and downloading a binary would make it so much smaller.

However, doing so in our repo would have a few drawbacks:

1/ `docker build` on our repo wouldn't do what it's supposed to do anymore.

If you checkout master (or another branch, or make local changes) and run a build, it would build an image with the latest version instead of the current one. This is super confusing.

2/ Chicken and egg problem.

In order to update the release in the Dockerfile, we would need to push a release. In order to push a release, we would need to tag it. If we tag the release **before** updating the Dockerfile, then `git co v0.1.0 && docker build` will result in a build of the **previous** release.

This could be fixed in a few different ways (build the binaries before tagging, update the Dockerfile with a yet-non-existing version, ...) but it's prone to errors and far from perfect.

How about this:

We keep the Dockerfile in swarm as is (build from source) and we create a `docker-library/swarm` repo (in the same way we do for golang, rails, etc) that contains a Dockerfil` that simply downloads a static binary?

Our release process would be: 1/ release `docker/swarm` and push the binaries 2/ update `docker-library/swarm` and to fetch the new binary and tag it 3/ update `docker-library/official-images` to point to the new tag.

Thoughts?

Re: overwriting the 0.1.0 release tag /cc @psftw 

Actually I would highly recommend our users to run with a specific tag.

1/ We have no guarantees that `1.0.0` will work exactly like `0.1.0`. It would be a problem if in between runs of `docker run swarm`, the version changes.

2/ Eventually, swarm will run in a multi manager setup (with leader election). If you bring new manager slaves in the cluster later on, they better be running the exact same version as the others. We cannot guarantee that the leader election protocol between, say, `0.6.0` and `1.2.0` will be the same. This is the same thing with any distributed system (you could replace the word "swarm" by "mongodb" in the previous example and it would be the same).

Regarding overwriting the tag, I agree it's bad practice but I really don't want to push `0.1.0-rc2` as a tag. This is a one time thing: We want to add swarm to the library now since it's an entirely new image and we know that the first review will take time. In the future, we are not planning to distribute rc versions as library images.

Re: CMD vs ENTRYPOINT /cc @psftw @tianon

I would find that super confusing: `swarm` by itself doesn't do anything beside printing the help message.

Which means that:

`docker run swarm` will print the help message
`docker run swarm manage` will output something like: `manage: No such file or directory`.

This is even more confusing than the `--entrypoint` thing, especially since I see no reason a user would want to run a `bash` inside the `swarm` image (and ideally, there wouldn't be any `bash` to run, just a static binary).

psftw:
Yes, I like @yosifkit's Dockerfile and your suggestion to have a separate repository for it.  The size difference is significant: 142M vs. 464M.  At some point if we have a slimmed down static binary release, even better.  I'm also on board to overwrite the image tag *in this case only*.  I tend to agree with your reasoning behind ENTRYPOINT vs CMD, but I still have an basic expectation that general-purpose official images should allow me to specify the command being executed.  In some cases, an ENTRYPOINT can trick me into thinking that's what happened (i.e. the postgres entrypoint script in the docs), which is also fine.  In the official images, this default behavior is a strong convention, so if that gets relaxed starting with ``swarm``, we'd need to come up with a better policy about it.  My default would be to type ``docker run swarm swarm <args>`` if I hadn't just read the docs.  Totally agree that a docker error message could confuse users, but this may also confuse them if they are unaware of ENTRYPOINT magic:

    $ docker run swarm:latest swarm -h
    No help topic for 'swarm'

I'm hesitant to suggest an ENTRYPOINT which allows either ``docker run swarm:tag swarm <args>`` or ``docker run swarm:tag <args>`` by checking the first arg.  :cold_sweat:

vieux:
My problem is, unlike @tianon's example with bash, it makes no sense to call `swarm` without argument.
So why even bother with a `CMD` ? Let's use nothing.
I'm kidding.

`ENTRYPOINT` ftw

aluzzardi:
@psftw @yosifkit: I could definitely rebuild the release as a static binary to even slim down further the image. Any suggestion on what the Dockerfile should look like to avoid the extra dependencies (as in, what to use instead of debian:jessie?)

@psftw: Great, it'd be awesome if we could overwrite the image tag in this case only.

Again, the whole point is to just speed up the review process when we have our release (basically, this time we are "bootstrapping" the release process and next time we'll come up with a 2 lines diff PR when `0.1.0` is ready).

I would also be okay with only tagging **latest** this time (pointing to 0.1.0-rc2) and next time adding **0.1.0** and updating **latest** to point to `0.1.0` (git tag). @vieux @psftw: Thoughts?

@psftw @tianon: A bit of background behind the `ENTRYPOINT` vs `CMD` debate. The only reason we are doing an official image is because docker doesn't support plugins yet.

Someday you would be able to run:
```
docker swarm manage <foo>
```

Today, our workaround is to do:
```
docker run swarm manage <foo>
```

`swarm` requires arguments and is useless without any (it just prints the help).

I think the difference between `swarm` and the other official images (where we use `CMD`) is that `swarm` is meant to be used as a tool whereas the other images are meant to be used as `FROM` images.

For instance, in the same way it doesn't make much sense to `docker run go`, it doesn't make sense either to write `FROM swarm`.

Thoughts?

yosifkit:
@aluzzardi, I like the idea of the slimmer image, but I am not sure we want to go thinner than `debian:jessie`.  I would argue that we need to add pgp signing to those binary releases with a related `.asc` file and then verify that in the Dockerfile like [many](https://github.com/docker-library/mongo/blob/master/2.6/Dockerfile#L19-25) of the [other](https://github.com/docker-library/httpd/blob/master/2.4/Dockerfile#L23-46) [official](https://github.com/docker-library/tomcat/blob/master/8-jre7/Dockerfile#L9-29) [images](https://github.com/docker-library/php/blob/master/5.6/Dockerfile#L15-33).

Moghedrin:
I'm very much with @tianon and @psftw on using CMD instead of ENTRYPOINT. There's no reason to depart from the established [best practices](https://docs.docker.com/articles/dockerfile_best-practices/#entrypoint). As far as issues of confusing the users go, that is precisely why there is documentation for all the official images.

@aluzzardi The fact that this official image is only a way to workaround the fact that docker doesn't yet support the plugins is, IMO, irrelevant.

psftw:
The idea behind an official image is that it will be maintained over time.  If this is just a temporary workaround to release swarm as a container with no longer term plans to support it, then I'm not sure we want to add the confusion for the short-term benefit ?

aluzzardi:
@psftw @tianon 

How about this:

1/ We're going to have a separate repository (with a different `Dockerfile`) for the `swarm` image.
2/ In that repository, we are going to upload the binary release.
3/ The `Dockerfile` will pretty much be:
```
FROM scratch
COPY swarm /swarm
ENTRYPOINT ["/swarm"]
```

This will result in an image of a few megabytes.

There's no point in using `CMD` - there's not a single other file in the image beside the binary, there's no bash to run or whatever and it doesn't make any sense to override the entrypoint since there's no other target, so it's not confusing to the user.

vieux:
@aluzzardi :heart:

psftw:
Totally agree that ENTRYPOINT is the way to go with a single-binary image.  If you run swarm in this fashion though, you'll quickly discover that without the usual OS-supplied CA certificates, you run in to trouble:

``Post https://discovery-stage.hub.docker.com/v1/clusters: x509: failed to load system roots and no roots provided``

Maybe if an application only ever talks to one known service, you might be able to get away with adding the certs manually, but I'm not sure of the consequences of this to be honest.  In the general case I would prefer to ship blessed sets of CA certs from trusted upstream sources (i.e. ca-certificates Debian package).  Thoughts?

I'm still a fan of @yosifkit's Dockerfile example above :-)

tianon:
What if we used @yosifkit's Dockerfile for now, but made the default
command be ["swarm", "join"] ?  Isn't that the one that stays running as a
daemon if you have the cluster environment variable set appropriately? (ie,
for the long-running mode of swarm, it's going to be the most common?)

tianon:
It's also worth noting that your current amd64 binary isn't fully static,
so won't work inside scratch as-is:

$ curl -sSL '
https://github.com/docker/swarm/releases/download/v0.1.0-rc2/docker-swarm_linux-amd64'
> swarm
$ chmod +x swarm
$ ldd ./swarm
linux-vdso.so.1 (0x00007fff9cd2a000)
libpthread.so.0 => /lib64/libpthread.so.0 (0x00007fecfb8b0000)
libc.so.6 => /lib64/libc.so.6 (0x00007fecfb500000)
/lib64/ld-linux-x86-64.so.2 (0x00007fecfbad0000)

aluzzardi:
There are actually two commands that need to run as daemon:

`swarm manage` to start the master
`swarm join` on every cluster node

tianon:
Plus an environment variable being passed in on every invocation, right?

aluzzardi:
This is the general workflow:

```
# create a cluster
$ swarm create
6856663cdefdec325839a4b7e1de38e8 # <- this is your unique <cluster_id>

# on each of your nodes, start the swarm agent
#  <node_ip> doesn't have to be public (eg. 192.168.0.X),
#  as long as the other nodes can reach it, it is fine.
$ swarm join --addr=<node_ip:2375> token://<cluster_id>

# start the manager on any machine or your laptop
$ swarm manage -H tcp://<swarm_ip:swarm_port> token://<cluster_id>
```

jfrazelle:
I put a comment here https://github.com/docker/swarm/pull/330#discussion_r23880448 on the entrypoint cmd debacle

vieux:
I'm just adding @jfrazelle comment here:

> I'm just going to give my 2 cents here, because isn't that what open source is all about ;)
>
> I think it should be `ENTRYPOINT` because we should use the best practices for our Dockerfiles. So:
```console
$ docker run -d swarm swarm --version
```
is kinda weird when it should be:
```console
$ docker run -d swarm --flag my --cluster is --awesomer than --everyone elses
```
> 
> Also if you are concerned with people not knowing the difference between `ENTRYPOINT` and `CMD` this would be a great example to show if you did something like:
```
ENTRYPOINT ["swarm"]
CMD ["--help"]
```

vieux:
and @crosbymichael's one: https://github.com/docker/swarm/pull/330/files#r23880656

aluzzardi:
update-certs.sh
```
ID=$(docker run debian:jessie bash -c "apt-get update && apt-get install -y --no-install-recommends ca")
docker cp $ID:/etc/ssl/certs/ca-certificates.crt ca
```

Dockerfile:
```
FROM scratch
COPY ./docker-swarm /swarm
COPY ca/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
ENTRYPOINT ["/swarm"]
CMD ["--help"]
```

```
$ docker run -t swarm manage token://63fe64c471e7d76be96a625350468dfc65c06c31                                                                                 INFO[0000] Listening for HTTP                            addr=127.0.0.1:2375 proto=tcp
```

```
$ docker images | head -2
REPOSITORY          TAG                 IMAGE ID            CREATED             VIRTUAL SIZE
swarm               latest              19e88684b9d3        38 seconds ago      9.201 MB
```

psftw:
@aluzzardi :heart: the 9MB image.  I hereby concede on the ENTRYPOINT vs CMD argument.  :smirk_cat:

tianon:
Indeed, that's an approach I can be happy about, even with ENTRYPOINT.
:heart:

I just ask that you make that ca.crt handling a script in the repo for both
transparency and your own repeatability. :smile:

aluzzardi:
PTAL

/cc @tianon

library/swarm
```diff
@@ -0,0 +1,4 @@
+# maintainer: Andrea Luzzardi <al@docker.com> (@aluzzardi)
+# maintainer: Victor Vieux <vieux@docker.com> (@vieux)
+
+latest: git://github.com/docker/swarm-library-image@c0f8204b2a1bb34233cd2144df51a4de9572440c
```
tianon: Did you want to add a `0.1.0-rc3` tag too?

aluzzardi: Depends - if we can point to it in here (`swarm-library-image@v0.1.0-rc3`) then it would make sense, otherwise not.

What do you think?

/cc @vieux

tianon: I mean a tag on the image, not on the Git repo.

tianon: Next release, you'd remove it from here, since it wouldn't be
"officially supported" anymore, and thus we wouldn't rebuild it, but
it would still be available on the Hub for folks who need it for
whatever reason.

vieux: I don't think so, as long as we are going to release the next rc or the beta we would discard the v0.1.0-rc3 tag. So let's just use latest.

tianon: For an example of what I mean, see "3.0.0-rc7" on
https://registry.hub.docker.com/_/mongo/ vs
https://registry.hub.docker.com/_/mongo/tags/manage/ which has "3.0.0-rc6"
and "2.8.0-rc5" etc.

aluzzardi: Oh, right!

So, before we mentioned that were planning to update the `0.1.0` tag to point the final release as soon as it gets there (and in the meantime, point to the `rc`).

Since "retagging" seems odd, we thought we could instead only have latest, and tag `0.1.0` as soon as it's official.

This is a one off: In the future we won't put `rc`s in the library.

tianon: As a user, I personally really appreciate being able to grab an older image.

As a maintainer, I find it especially nice for being able to grab that
older version and compare to see if the current bug was in that older
version too.

aluzzardi: We are planning to keep older versions - `rc`s are kinda different in the sense that they are not really releases.

We could go either way - I'm not entirely sure.

We're not planning to push rc's in the library (at least, we don't think so unless we find a compelling reason).

tianon: On 4 February 2015 at 16:13, Andrea Luzzardi <notifications@github.com> wrote:
> We're not planning to push rc's in the library (at least, we don't think so unless we find a compelling reason).

I don't really see a compelling reason not to, especially if what
@vieux is telling me about you guys not doing official binaries
anymore is true (which also makes me kind of sad, but it's
understandable).

tianon:
Appears to run well, and is appropriately tiny:
```console
$ docker run -it --rm swarm
NAME:
   swarm - a Docker-native clustering system

USAGE:
   swarm [global options] command [command options] [arguments...]

VERSION:
   0.1.0 (e87c276)

COMMANDS:
   create, c	create a cluster
   list, l	list nodes in a cluster
   manage, m	manage a docker cluster
   join, j	join a docker cluster
   help, h	Shows a list of commands or help for one command
   
GLOBAL OPTIONS:
   --debug			debug mode [$DEBUG]
   --log-level, -l "info"	Log level (options: debug, info, warn, error, fatal, panic)
   --help, -h			show help
   --version, -v		print the version
   
$ docker images swarm
REPOSITORY          TAG                 IMAGE ID            CREATED             VIRTUAL SIZE
swarm               latest              615466d45221        2 minutes ago       7.111 MB
```

library/swarm
tianon: I guess what I'm getting at is that when `0.1.1` comes out, you won't still have `0.1.0` in this file either, since it won't be officially supported anymore.  So my question is, what's the harm in tagging this release too?

aluzzardi:
Updated. PTAL

tianon:
LGTM

Build test of #440; 2fc5161d2ddaff91d1c96ec66cbe124a0f65fc5c (`swarm`):

```console
$ bashbrew build https://raw.githubusercontent.com/docker-library/official-images/2fc5161d2ddaff91d1c96ec66cbe124a0f65fc5c/library/swarm
Processing swarm:0.1.0-rc3 ...
Processing swarm:latest ...
```

aluzzardi:
:fireworks:

yosifkit:
LGTM

thaJeztah:
:clap: great job here! :heart: the small image

bfirsh:
Holy crap. This is brilliant. :sparkles: :cake: :sparkles:

Issue #390, opened 2015-01-09
UTC ALL THE THINGS


tianon:
`date +%Z` → `UTC`

- [ ] `cirros`: `MST` (@ewindisch)
- [ ] `fedora:21`: `EST` (@lsm5)
- [x] `oraclelinux:6`: `EST`

Created an issue for it so we can track completion. :+1:

vaijab:
+1

tianon:
ping @ewindisch @lsm5 

Status update?:no_mouth:

lsm5:
@tianon @kushaldas built an image with tz corrected, but getting that out via fedora's official release process might take a bit longer (I think). correct me if i'm wrong @kushaldas

tianon:
Is there a strong reason for the official builds to be on `EST`?  It seems really random. :smile:

lsm5:
afaik nope, probably just something that crept in :|

subesokun:
+1
It would be awesome if you allow passing in a configuration parameter when starting the docker container in order to set the time zone. Reading out the time zone from the host might not be always sufficient e.g.:
> If your Jenkins instance is running in a different location than the user's (for example: the server is in NY but users are in LA), then the NY time zone will most probably be used, which may be quite annoying if you need to compare build dates.
> https://wiki.jenkins-ci.org/display/JENKINS/Change+time+zone

An implementation similar to [GITLAB_TIMEZONE](https://github.com/sameersbn/docker-gitlab#available-configuration-parameters) would be great.

cpuguy83:
Because EST is superior to all other time zones, of course :)

tianon:
ping @lsm5 @ewindisch

lsm5:
@tianon, @kushaldas has committed a patch, I'll send out an update to you once it's built in fedora's build system.

ewindisch:
@tianon - the MST setting is imported directly from the upstream cirros project. All CirrOS images for VMs and containers both have this set. Changing it just for Docker could potentially create incompatibilities. I'll file a bug upstream and see what they say.

tianon:
:heart: -- I'd love to at the very least get the rationale for it :+1:

ewindisch:
@tianon pretty sure it's, "Scott Moser lives in MST"

tianon:
hahaha, I can certainly respect that level of narcissism :laughing:

(but I want it from the horse's mouth before I give up, so I appreciate your filing the issue)

ewindisch:
[...]

PR #450, opened 2015-02-02
Update Cirros to 1.3.3

ewindisch:


library/cirros
```diff
@@ -1,4 +1,5 @@
 # maintainer: Eric Windisch <ewindisch@docker.com> (@ewindisch)
 
-latest: git://github.com/ewindisch/docker-cirros@1cded459668e8b9dbf4ef976c94c05add9bbd8e9
+latest: git://github.com/ewindisch/docker-cirros@5ef3f5024b0aa80553cc34be9eff6685cb31b458
+0.3.3: git://github.com/ewindisch/docker-cirros@5ef3f5024b0aa80553cc34be9eff6685cb31b458
 0.3.0: git://github.com/ewindisch/docker-cirros@1cded459668e8b9dbf4ef976c94c05add9bbd8e9
```
yosifkit: Is there a reason for keeping the old version, 0.3.0, around?  Is it going to be a [supported version](https://registry.hub.docker.com/_/cirros/)?

ewindisch: It seems reasonable to keep the N-1 version around, if users might depending on it run into issues/bugs.

tianon: We don't delete removed tags -- they just don't won't be rebuilt and aren't
officially supported.

tianon: See https://registry.hub.docker.com/_/redis/ vs
https://registry.hub.docker.com/_/redis/tags/manage/ for a good
example of this in practice.

ewindisch: That's fine & reasonable. I'll remove 0.3.0 then.

I had seen other official images keeping clearly-not-supported versions tagged here, so this wasn't clear. I'll follow-up on that and see about having those images remove their unsupported versions, too.

tianon: Indeed -- if you have specific examples, that would be helpful.  Some of
them have legitimate reasons for it that we've discussed with them.

tianon:
LGTM

Build test of #450; c0633f08921f4e8f945190da1836999012d9e9b1 (`cirros`):

```console
$ bashbrew build https://raw.githubusercontent.com/docker-library/official-images/c0633f08921f4e8f945190da1836999012d9e9b1/library/cirros
Fetching cirros (git://github.com/ewindisch/docker-cirros) ...
Processing cirros:latest ...
Processing cirros:0.3.3 ...
```

tianon:
Do you happen to have any insight into #390? (ie, why does CirrOS use `MST`?)

ewindisch:
@tianon I've updated #390 directly. Thanks for the heads up

yosifkit:
LGTM

PR #476, opened 2015-02-10
updated crate version to 0.46.5-2

chaudum:
we have a new dockerfile of the same crate version
therefore the `-2` suffix (build number)

### DIFF
--- a/library/mageia
+++ b/library/mageia
@@ -1,5 +1,5 @@
 # maintainer: Juan Luis Baptiste <juancho@mageia.org> (@juanluisbaptiste)
 
-latest: git://github.com/juanluisbaptiste/docker-brew-mageia@603a75ec98dfd7c26ed85108b309d4e5f39ded91 4
-cauldron: git://github.com/juanluisbaptiste/docker-brew-mageia@603a75ec98dfd7c26ed85108b309d4e5f39ded91 cauldron
-4: git://github.com/juanluisbaptiste/docker-brew-mageia@603a75ec98dfd7c26ed85108b309d4e5f39ded91 4
+latest: git://github.com/juanluisbaptiste/docker-brew-mageia@ed9534c591e9b03593e8111990395f77bbd256c0 4
+cauldron: git://github.com/juanluisbaptiste/docker-brew-mageia@ed9534c591e9b03593e8111990395f77bbd256c0 cauldron
+4: git://github.com/juanluisbaptiste/docker-brew-mageia@ed9534c591e9b03593e8111990395f77bbd256c0 4


tianon:
https://github.com/crate/docker-crate/compare/0.46.5...0.46.5-2

This has some pretty hefty changes in it compared to previous versions.  What's the rationale for having such a large amount of custom code for Mesos and Weave?

chaudum:
what we try to achieve is to make crate as easy to use as possible on different platforms and environments. when running the crate docker container, you quickly have a lot of arguments to set. we want to decrease them.

the actual amount of code for mesos and weave is not that large I think. instead of running the crate binary directly, we've wrapped it into a run script which detects mesos and weave (or any other env in the future) and includes either `scripts/mesos.in.sh` or `scripts/weave.in.sh`. The included scripts set environment variables and command line arguments for the crate binary to remove the amount of docker run arguments a user needs to specify.

the rest of the code (`bootstrap.py`, `buildout.py`, all files in `tests/`) is only for integration testing the dockerfile.

tianon:
The testing bits are really neat. :+1:

However, I'm not comfortable with an official image that has specific Mesos or Weave support (that isn't part of the upstream software itself and was custom-built and separately-maintained specifically for the Dockerization), especially since the support added here looks like something that could be trivially added as a simple wrapper around the software either in an image `FROM crate` or even in a script that wraps up the invocation of this image as-is.  All of the official images strive to be as simple and upstream-focused a Dockerization as possible, and this is a firm departure from that.

psftw:
Agree.  It may seem like nitpicking, but the distinction tianon brings up is important.  We should expect power users to read up on how the image works and I would be turned off by the extra magic going on.

chaudum:
I can understand that official images should only contain the upstream software itself. This makes total sense for images that are re-used as base image for other Dockerfiles, such as `java` for java applications, because they are usually not used directly by end-users.

However, in case of Crate - or any other images that are not re-used as base for other images, which is the case for most databases I guess - I find it a bit too nitpicky not to allow making the the Dockerfile simpler to use by trying to reduce the number of arguments a user needs to pass to the run command.
In the end it's the Crate users who want/need to use the image, and we should try to make it as easy as possible for them. Power users can still control everything by overriding set defaults if they want to.

I can see a solution where the starter script accepts an argument for different supported environments if that is more aligned with the Docker philosophy; something like that:

```sh
docker run -d -p 4200:4200 crate:0.46.5-2 cratectl --env MESOS
```

However, then we are not backwards compatible and the command line arguments are different than on non-Docker environments.

Up for discussion ... ;)

yosifkit:
Although I agree that it would be nice for a user to have their database autoconfigured depending on the environment that it is starting in, I don't think that this is a realistic goal for official images.  First it is just mesos and weave integration, then there will be other providers and tools (kubernetes, project atomic, yarn, etc) that will want the official images to just work in their setup, so we would have to add complicated startup scripts to every image.  This does not seem like a sustainable system. With automated builds (https://docs.docker.com/docker-hub/builds/) and repository links (https://docs.docker.com/docker-hub/builds/#repository-links), it's reasonably easy to have an up-to-date image built `FROM` this one with custom per-service items enabled.

chaudum:
ok, so we'll revert the Dockerfile to what we had before and come up with a cleaner solution.

PR #468, opened 2015-02-06
Node.js 0.12.0 and npm 2.5.0

chorrell:
This adds Node.js 0.12.0, setting it to the "latest" and "0" tags and also includes npm 2.5.0

PR #477, opened 2015-02-10
Remove the 0.11 tag now that 0.12 is available

chorrell:
This removes the 0.11 tag per our discussion in #468 . I also removed 0.11 from https://github.com/joyent/docker-node


tianon:
LGTM

yosifkit:
LGTM

PR #489, opened 2015-02-14
Rename "shellshock" to "cve-2014--shellshock"

tianon:
This makes room for more "cve-xxxx-yyyy--zzzzz" style test names in the future, making it more clear what the purpose is, especially as we grow our test suite.

(Like the "cve-2014-9680--sudo-tz" I'm contemplating if it ever gets fixes released.)


yosifkit:
LGTM

PR #490, opened 2015-02-14
Fix some minor issues in bashbrew.sh

tianon:
Double quotes and indented heredocs!

Noticed these while doing some Bash tutorializing with @knickers, using `bashbrew` itself as an example of some of the concepts we were discussing. :+1:

### DIFF
--- a/bashbrew/bashbrew.sh
+++ b/bashbrew/bashbrew.sh
@@ -19,47 +19,47 @@ logs="$(readlink -f "$logs")"
 self="$(basename "$0")"
 
 usage() {
-	cat <<EOUSAGE
-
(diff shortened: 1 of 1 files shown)


bashbrew/bashbrew.sh
```diff
@@ -97,10 +97,9 @@ done
 
 # which subcommand
 subcommand="$1"
+shift || { usage >&2 && exit 1; }
 case "$subcommand" in
-	build|push|list)
-		shift
```
tianon: This one led to fun messages like `unknown subcommand: ` when you don't supply any arguments at all.

yosifkit:
LGTM

PR #496, opened 2015-02-17
updated crate to 0.47.3

bogensberger:


tianon:
I would consider this a duplicate of https://github.com/docker-library/official-images/pull/476.  I don't see any reason to splinter that discussion even further.

bogensberger:
sorry, didn't know that there is an ongoing discussion.

PR #500, opened 2015-02-18
updated crate version

chaudum:
sorry for the confusion caused by the last pull request. now we've reverted the support for weave and mesos within the docker file.
we're also using jre instead of jdk now, to reduce the size of the image.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,83 +4,59 @@
 
 ## Contributing to the standard library
 
-Thank you for your interest in the Docker official images project! We
-strive to make these instructions as simple and straightforward as
-possible, but if you find yourself lost, don't hesitate to seek us out
-on Freenode IRC in channel `#docker-library`, or by creating a GitHub
-issue here.
+Thank you for your interest in the Docker official images project! We strive to make these instructions as simple and straightforward as possible, but if you find yourself lost, don't hesitate to seek us out on Freenode IRC in channel `#docker-library`, or by creating a GitHub issue here.
 
-Be sure to familiarize yourself with the [Guidelines for Creating and
-Documenting Official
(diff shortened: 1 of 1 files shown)


tianon:
https://github.com/crate/docker-crate/compare/0.46.5...0.47.4

I'm very +1 on using the JRE appropriately in place of the JDK, especially for size! :+1:

Some of the new bits were still kept:
```Dockerfile
ADD scripts/run.sh /usr/bin/crate
```

The [`run.sh` script](https://github.com/crate/docker-crate/compare/0.46.5...0.47.4#diff-dcd8a73331aba627440c247763086b99) itself is a little puzzling in several ways:
- `exec` near the end which terminates the script followed by some `wait` logic which won't ever run thanks to the `exec` before it
  - was the `trap`+`exec`+`wait` logic attempting to keep the shell wrapper around as a resident process to capture signals and kill crate since it's not capturing signals by itself?
- very verbose in telling the user what they asked for or didn't ask for (but at the same time, not even making recommendations about what the user _should_ be doing)
- assumes output is always going to a terminal

Regardless of those however, it being added as `/usr/bin/crate` when we have a `crate` binary on the path already is inevitably going to be a source of confusion for users, especially since `/crate/bin/crate` accepts arguments (`-v`, `-h`, `-D`, `-X` for some that seem like they might potentially be useful to Docker users) but `/usr/bin/crate` does not pass arguments through to `/crate/bin/crate`.

chaudum:
thanks for the feedback! I guess it is best to only expose the crate binary at `/crate/bin/crate` and get rid of the run script. think I will need to do a new PR then because we'd need to make a new version tag.

PR #507, opened 2015-02-20
Remove EOL'd Jetty versions 7 and 8

md5:
The Jetty 7 and 8 tags will no longer be considered supported tags. They will not likely receive any future version updates from upstream, but I've created the [`eol/jetty-7`](https://github.com/md5/docker-jetty/tree/eol/jetty-7) and [`eol/jetty-8`](https://github.com/md5/docker-jetty/tree/eol/jetty-8) branches just in case I need to do any future updates. Those branches may also be useful to people who want to build the Jetty 7 and 8 images themselves.

See md5/docker-jetty#2 for background


tianon:
LGTM :heart:

Moghedrin:
LGTM

PR #511, opened 2015-02-20
Call out explicitly that Git tags are discouraged

tianon:
Also, clarified some of the language around the file format itself, including concrete real-world examples of how it looks/works in practice, and the recommendation to go browse other existing files too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,26 +12,41 @@ Also, the Hub descriptions for these images are currently stored separately in t
 
 ## Library definition files
 
-The library definition files are plain text files found in the `library/` subfolder of the official-images repository.
+The library definition files are plain text files found in the [`library/` directory of the `official-images` repository](https://github.com/docker-library/official-images/tree/master/library).
 
-### File names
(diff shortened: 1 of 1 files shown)


tianon:
Of course, https://github.com/docker-library/official-images/pull/511/files?w=1 is easier to look at. :+1:

md5:
:+1: I was wondering about this recently. Glad to see it clarified, though I'd be gladder to see tags banned.

Would you guys consider banning tags for any new images going forward and grandfathering the ones currently using tags?

tianon:
That's effectively what this commit was intended to be. :+1:

I realized somewhat recently that even if we trust the maintainer of the image not to change a tag, that doesn't mean someone else with access to that repo (or someone who maliciously gains access) isn't going to change that tag either inadvertently or maliciously.

Also, a full Git commit reference is a hash of the contents it represents, further solidifying the repeatability in the same way we use full GPG fingerprints and embedded hash verification in our `Dockerfile`s wherever possible.  Once I fully realized the implications of this, it was clear that we needed to make tags no longer acceptable.

md5:
:metal:

yosifkit:
LGTM

PR #513, opened 2015-02-20
Make it clear that there is a review process

tianon:
This also adds a little more background about what the review process might include and the kinds of things that will require changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,11 +4,22 @@
 
 ## Contributing to the standard library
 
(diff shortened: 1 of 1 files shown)


tianon:
cc @psftw :wink:

yosifkit:
LGTM

psftw:
LGTM, I'll likely add more to this soon as well.

PR #516, opened 2015-02-22
New Alpine Linux base image

andyshinn:
This is a new Alpine Base Linux image maintained by Glider Labs. Check our the extended build docs at http://gliderlabs.viewdocs.io/docker-alpine/build. Some notable differences between this and other Alpine Linux images out there:

* We (Glider Labs) are making an effort to maintain and support the image. We hang out in #docker-library and our own channel at #gliderlabs to answer questions.
* We build the images on CircleCI in transparent manner so everyone can see the build process.
* Every build has images as artifacts to make for easy testing from CI.
* Even though very basic, some tests (install a package).

To support Automated Builds on Docker Hub we also keep the `rootfs.tar.xz` around and force-push a commit to a branch for the version name (such as https://github.com/gliderlabs/docker-alpine/tree/rootfs-v3.1). We do a separate branch since builds are parallelized and force-pushes to a single branch would overwrite each other.

### DIFF
--- a/library/alpine
+++ b/library/alpine
@@ -1,0 +1,10 @@
+# maintainer: Glider Labs <team@gliderlabs.com> (@gliderlabs)
+
+edge: git://github.com/gliderlabs/docker-alpine@09c2fde27b1aa585009a070dc9cc13a020a30cf9 versions/library-edge
+
+3.1: git://github.com/gliderlabs/docker-alpine@337bd45a6de1404ceb7c7a0a2fd01744b488eaef versions/library-3.1
+latest: git://github.com/gliderlabs/docker-alpine@337bd45a6de1404ceb7c7a0a2fd01744b488eaef versions/library-3.1
+
+2.7: git://github.com/gliderlabs/docker-alpine@2a7b889be5868aee4706fd8498e6b11630e8ace2 versions/library-2.7
+
+2.6: git://github.com/gliderlabs/docker-alpine@12cf56252dcb1535e0fbeb9c3e3586551af671ea versions/library-2.6


md5:
Nice! This looks like a great addition to the available base images.

It looks like the `edge` and `3.1` versions are using `GMT` as the timezone instead of `UTC`. The `2.6` and `2.7` versions are using `UTC`:

```
$ ./bashbrew/bashbrew.sh build alpine
Cloning alpine (git://github.com/gliderlabs/docker-alpine) ...
warning: expected SRV RR, found RR type 1
Processing alpine:edge ...
Processing alpine:3.1 ...
Processing alpine:2.7 ...
Processing alpine:2.6 ...
$ ./bashbrew/bashbrew.sh list alpine | xargs ./test/run.sh
testing alpine:edge
	'utc' [1/2]...failed; unexpected output:
--- /work/test/tests/utc/expected-std-out.txt	2015-02-06 08:29:29.000000000 +0000
+++ -	2015-02-22 21:23:03.539309424 +0000
@@ -1 +1 @@
-UTC
+GMT
testing alpine:3.1
	'utc' [1/2]...failed; unexpected output:
--- /work/test/tests/utc/expected-std-out.txt	2015-02-06 08:29:29.000000000 +0000
+++ -	2015-02-22 21:23:05.137367838 +0000
@@ -1 +1 @@
-UTC
+GMT
testing alpine:2.7
	'utc' [1/2]...passed
testing alpine:2.6
	'utc' [1/2]...passed
```

andyshinn:
Looks like BusyBox on uClibc and musl get timezone differently. Is it fair to solve this by setting the `TZ` env var in the profile? I'd rather avoid installing `tzdata` which is 3.2 MB (massive in the Alpine world) :(

md5:
We'll have to see what @tianon has to say, but I personally feel like `GMT` should be acceptable if the alternative is doubling the image size. As far as I understand it, `UTC` and `GMT` are functionally equivalent (with the possible exception of historical date handling).

andyshinn:
I pushed a change that builds with the tzdata package, copies over the `UTC` timezone file to `/etc/localtime`, and removes the package before saving the rootfs.

library/alpine
```diff
@@ -0,0 +1,9 @@
+# maintainer: Glider Labs <team@gliderlabs.com> (@gliderlabs)
+
+edge: git://github.com/gliderlabs/docker-alpine@8f429d1283f6897a1c21d5745dca8fa25d394a8b versions/edge
+
+3.1: git://github.com/gliderlabs/docker-alpine@e119cc4675afe3abe221b8a23144888c02601f4b versions/v3.1
```
tianon: Didn't you want to add a `latest` alias for `3.1` here?  (functionally, this is accomplished by duplicating the line and changing the `3.1:` at the start to be `latest:`)

tianon:
Nice trick -- that's a pretty clean solution to get a consistent timezone, IMO! :+1:

library/alpine
andyshinn: D'oh, you are correct. I'll fix this up.

tianon:
As we discussed before, the `apk-install` script still makes me a bit queasy, since it's not obviously Docker-specific, and what it provides doesn't really add much except the convenience of not having to remember exact arguments (`apk add --update "$@" && rm -rf /var/cache/apk/*`).

My fears are that this will potentially alienate other core Alpine Linux contributors, and confuse newcomers.

andyshinn:
I meant to remove this from the main `content.md`. Do you feel like it is still confusing or alienating even if it is outside docs for more "power user" usage? We really like it as a shortcut. But we don't need to call it out in the docs.

tianon:
I wasn't even looking at the docs -- I simply mean the name of the command is itself misleading, and not obviously a Docker-specific addition.

andyshinn:
What if we built a slightly different rootfs (san `apk-install`) for the library than we do for `gliderlabs/alpine`? My idea is to rework the build scripts a bit so that there are builds for library that are pure upstream (including the TZ change) and then our own builds which just flip build time switches to enable `apk-install` script and other conveniences.

A sort of separate but related question, is it possible to get the `gliderlabs/alpine` image switched to an Automated Build to preserve everything (stats and whatnot) without having to delete and re-add?

yosifkit:
That sounds reasonable to me :+1:.  I don't see any way to change the docker repo to an automated repo, but maybe support@docker.com might have a way to do it, since the stats are important.

md5:
I just noticed this image doesn't have a default command. How about adding `CMD ["/bin/ash"]` or `CMD ["/bin/sh"]`?

andyshinn:
The motivation for the image is very minimalist type stuff. So adding an extra layer for the default command goes against this. But this could be added now that the official image is separated out and has different build options. I'm curious if others think there should be a default as well.

md5:
Got it. I think we're talking about 2k or so to add that layer, but I understand where you're coming from.

andyshinn:
I pushed up new rootfs that now build without the `apk-install` script for the `version/library-*` images. For the time being, we'll just maintain the more vanilla upstream official version (only the timezone modifications) and our own which we add `apk-install` to and maybe even package CDN and some other non-standard things.

The library docs were also rebased and don't reference `apk-install`. Once merged in, I'll clean up the main repository docs to point out the differences between the two images.

yosifkit:
LGTM
Build test of #516; 398decb0d74e1ce3e9ae9f852bab0172cb7722c7 (`alpine`):

```console
$ url="https://raw.githubusercontent.com/docker-library/official-images/398decb0d74e1ce3e9ae9f852bab0172cb7722c7/library/alpine"
$ bashbrew build "$url"
Cloning alpine (git://github.com/gliderlabs/docker-alpine) ...
Processing alpine:edge ...
Processing alpine:3.1 ...
Processing alpine:latest ...
Processing alpine:2.7 ...
Processing alpine:2.6 ...
testing alpine:edge
	'utc' [1/2]...passed
testing alpine:3.1
	'utc' [1/2]...passed
testing alpine:latest
	'utc' [1/2]...passed
testing alpine:2.7
	'utc' [1/2]...passed
testing alpine:2.6
	'utc' [1/2]...passed
```

tianon:
:+1: LGTM!

PR #524, opened 2015-02-27
Adding file for nevergreen official docker image

manasik:
Creating file for nevergreen

### DIFF
--- a/library/nevergreen
+++ b/library/nevergreen
@@ -1,0 +1,6 @@
+# maintainer: Manasi Kulkarni <mkulkarn@thoughtworks.com>
+# maintainer: Stephen Cowley <ste@thoughtworks.com>
+# maintainer: Joe <joe@joejag.com>
+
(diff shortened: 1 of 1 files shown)


psftw:
Built and ran well for me.  I don't have any technical reservations on this image, just simply philosophical.  I'm on the fence as to whether this is a good candidate for an Official Repo, but I'm not the final decision maker on the matter.  It is a fairly new and minimalist project that solves a specific problem -- displaying build statuses in a particular way.  I've commented in a longer form about my hesitations earlier today to provide more context: https://github.com/docker-library/official-images/pull/517#issuecomment-76439816

Moghedrin:
I agree with you on this one. Unless @tianon or @yosifkit feel otherwise, or @manasik can provide a strong case for why it *should* be an official image, I'm okay with closing this for now.

yosifkit:
Closing, feel free to ping if you feel we should revisit this.

PR #530, opened 2015-03-04
Add library file for logstash

### LINKED ISSUE #488: [New Image] logstash
We are working on a new image for logstash.  Look for the PR within a the next few days.

Moghedrin:
Add Logstash to official images
Fixes #488

### DIFF
--- a/library/logstash
+++ b/library/logstash
@@ -1,0 +1,5 @@
+# maintainer: InfoSiftr <github@infosiftr.com> (@infosiftr)
+
+1.4.2: git://github.com/docker-library/logstash@d87d3511533cdf9074a3fddeb81b770cbb6bf05c
+1.4: git://github.com/docker-library/logstash@d87d3511533cdf9074a3fddeb81b770cbb6bf05c
+latest: git://github.com/docker-library/logstash@d87d3511533cdf9074a3fddeb81b770cbb6bf05c


md5:
How come you're using the tar.gz install instead of the debs?

I was actually looking into the existing `logstash` images yesterday and I was inclined to use the debs as indicated by this page: http://www.elasticsearch.org/overview/elkdownloads/

I know that [the docs](http://logstash.net/docs/1.4.2/tutorials/getting-started-with-logstash) don't currently recommend the debs, but I believe the ELK page is also official since Logstash is part of the Elasticsearch family of products now.

Moghedrin:
I was going with the docs. As far as the ELK page indicating to use the debs, I see no such indication at all. I see that they have the debs to download, but zero indication that doing so is the preferred way of installation, unless I am missing something. If I am missing something, I see no problems with switching in favor of using the debs - otherwise, it makes more sense to stick with what the documentation does.

md5:
It means that if someone starts using the `/usr/local`-based version and the image is switch the the deb-based version, any customizations they've made through adding or mounting config files will have to be updated. The bad part of that is that they likely will have no idea when a path-changing update has happened, so they may pull `logstash:latest` onto a new machine or something and wonder why their container has stopped working.

tianon:
"Download and unzip the latest Logstash release" does seem to _imply_ that they prefer binary-based installation, even though they've got debs and rpms. :confused:

On the other hand, [`elasticsearch` does use the debs](https://github.com/docker-library/elasticsearch/blob/90098dc7c2f34bc1c6a48c1b3ce70b09b83fcb7b/1.4/Dockerfile#L7), so maybe it's better that we're self-consistent in this case, especially since upstream seems to have somewhat mixed messages on this front?

yosifkit:
But the elasticsearch uses the apt repo whereas there is none for logstash, just a deb file.

md5:
@tianon I was just assuming they haven't made all their sites consistent yet. Then again, it looks like the projects combined efforts in August 2013, so perhaps it's intentional.

@yosifkit That's not true: http://packages.elasticsearch.org/logstash/1.4/debian/pool/main/binary-amd64/Packages

md5:
What they *don't* have is one for Kibana, as far as I could tell when I was looking into this yesterday.

Moghedrin:
It makes sense to me to stick with what Logstash has in their documentation. While it's true that if the image switches to a deb-based version after we put it into the official images, it may break the user's images, it doesn't make sense to me to *assume* that this will be a change we make at all, especially if the projects combined efforts in 2013 and Logstash's docs STILL seem to imply the binary-based installation.

md5:
Cool. Just thought I'd ask. :see_no_evil:

md5:
The only thing that sticks in my craw is that you're downloading from downloads.elasticsearch.org anyways. It's clear to me that these things are packaged by the same people, though I suppose that's an assumption as you say.

Moghedrin:
In my mind, How we install Logstash in this image is independent to the fact that they appear to be packaged by the same people. The Logstash docs indicate a binary-based installation, and there doesn't appear to be anything strongly recommending otherwise - therefore, this is the approach that makes sense to me.

If that's the only blocker here, then let's get it merged in, but open an issue so that people can see it's an area of discussion and can give feedback. Sound reasonable?

md5:
SGTM

tianon:
++

library/logstash
```diff
@@ -0,0 +1,3 @@
+# maintainer: InfoSiftr <github@infosiftr.com> (@infosiftr)
+1.4: git://github.com/docker-library/logstash@d87d3511533cdf9074a3fddeb81b770cbb6bf05c
```
tianon: Why only `1.4`?  Why not also `1.4.2`, at least?

yosifkit:
LGTM
Build test of #530; d9df2a8dd5eab81d5a3737f3d37605075be76d53 (`logstash`):

```console
$ url="https://raw.githubusercontent.com/docker-library/official-images/d9df2a8dd5eab81d5a3737f3d37605075be76d53/library/logstash"
$ bashbrew build "$url"
Processing logstash:1.4 ...
Processing logstash:latest ...
testing logstash:1.4
	'utc' [1/2]...passed
testing logstash:latest
	'utc' [1/2]...passed
```

Moghedrin:
@tianon Better? :dancer:

yosifkit:
Build test of #530; 02c0b0b4677ecdf593c0bd96a2e2758943d8f515 (`logstash`):

```console
$ url="https://raw.githubusercontent.com/docker-library/official-images/02c0b0b4677ecdf593c0bd96a2e2758943d8f515/library/logstash"
$ bashbrew build "$url"
Processing logstash:1.4.2 ...
Processing logstash:1.4 ...
Processing logstash:latest ...
testing logstash:1.4.2
	'utc' [1/2]...passed
testing logstash:1.4
	'utc' [1/2]...passed
testing logstash:latest
	'utc' [1/2]...passed
```

tianon:
LGTM

jordansissel:
> "Download and unzip the latest Logstash release" does seem to imply that they prefer binary-based installation, even though they've got debs and rpms. :confused:

The reason our docs focus on "download and unzip" is that those instructions are nearly universal across operating systems - Windows, Solaris, Linux, OSX, etc; instead of targeting just Ubuntu with deb instructions. My preference is helping the most people, so that's what the documentation tends to target first. "Use what is easiest for you" is my general rule, and so if 'yum install' is easier, you should use our yum repo, for example.

Hopefully this clarifies things.

tianon:
@jordansissel :heart_eyes: thanks, that definitely does!

PR #533, opened 2015-03-06
Start adding tests for docker-php-ext-*

md5:
This adds a single test for `docker-php-ext-install pdo_mysql`. More tests should be added for other common modules (e.g. `gd`, `intl`, `mbstring`, etc.).

### DIFF
--- a/test/config.sh
+++ b/test/config.sh
@@ -34,6 +34,9 @@ declare -A imageTests=(
 	[mysql]='
 		mysql-basics
 	'
+	[php]='
+		php-ext-install
+	'
 # example onbuild
 #	[python:onbuild]='
 #		py-onbuild
--- a/test/tests/php-ext-install/container.sh
+++ b/test/tests/php-ext-install/container.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+set -ueo pipefail
+
+docker-php-ext-install pdo_mysql 2>&1
--- a/test/tests/php-ext-install/run.sh
+++ b/test/tests/php-ext-install/run.sh
@@ -1,0 +1,1 @@
+../run-bash-in-container.sh


yosifkit:
I like the test, though I am unsure on the naming scheme (`docker-php-ext`). Currently all the tests do one of the following: start with the name of the image/language, start with a cve number, or are global.  What do you think @tianon?  How do we want to contain the names (esp. if we want a large influx of community made tests)?

md5:
Yeah, I almost named it `php-docker-php-ext`, but that sounded dumb :wink:

tianon:
IMO, `php-ext-install` would be sufficient here, since `docker-` is implied, and it's clear that we're testing extensions.

I think @yosifkit's description of the naming currently is accurate and good for alphabetical sorting (makes logical groups clump together nicely).

md5:
Updated to `php-ext-install`.

The other thing that came to mind while I was working on this was to use directories under `test/tests`, but I didn't try to implement that change.

tianon:
Ahh, nice and simple. :+1:

LGTM

I guess with our config explicitly listing all tests, there's no reason why subdirectories wouldn't work that I can think of.  Certainly an interesting idea.

tianon:
This does bail non-zero if it fails to build or install properly, right?

md5:
I believe so. I tested it with before docker-library/php#73 was merged and it was failing.

tianon:
You should also be aware that we're planning to come up with a list of tests we'd like to see for the Docker birthday celebration: http://docker.party/

(so if you see issues for those rolling in, please don't steal the easy ones :heart:)

tianon:
Nice, SGTM

md5:
> (so if you see issues for those rolling in, please don't steal the easy ones :heart:)

Noted :notes:

yosifkit:
LGTM

tianon:
:tada: :balloon: :birthday: :+1:

md5:
Nice. I had originally wanted to have more than one test in here (each executed in a clean container), but I kept having trouble making it work and decided to start with the single test I had working.

PR #567, opened 2015-03-19
version 3.5.4

volmarl:
Aerospike version 3.5.4

### DIFF
--- a/library/aerospike
+++ b/library/aerospike
@@ -1,4 +1,5 @@
 # maintainer: Lucien Volmar <lucien@aerospike.com> (@volmarl)
 
+3.5.4: git://github.com/aerospike/aerospike-server.docker@54bb45af2df342e2fd4361df1ec3f160327ec827
 3.5.3: git://github.com/aerospike/aerospike-server.docker@001bf913432989f30e19816467b0e0e8430730bc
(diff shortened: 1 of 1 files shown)


yosifkit:
The Dockerfile changes look all fine to me (https://github.com/aerospike/aerospike-server.docker/compare/001bf913432989f30e19816467b0e0e8430730bc...54bb45af2df342e2fd4361df1ec3f160327ec827).

Just a couple points of clarification. What is the reason to keep the `3.5.3` version? Is it to keep it on the Docker hub? The image will stay on the hub regardless of what is in the library file. The library file is what controls the "supported versions" list on the Docker hub page. Also, if it is in the library file, it will be rebuilt when changes are made to the base image (`debian:7`).

tianon:
ping @volmarl :wink:

yosifkit:
@volmarl, we'd love to get this merged and any newer updates you may have now, but we just had a few questions.
- what is the reason to keep the 3.5.3 version?
- is `3.5.3` kept here so that it will appear on the docker hub? The image will stay on the hub regardless of what is in the library file.
  - will it get bug or security fixes?

volmarl:
@yosifkit @tianon sorry missed the updates. I was actually submitting 3.5.8 today. Good point on the older versions. I will remove and update latest pull request

PR #581, opened 2015-03-21
Add "trap" to mysql-basics test for deleting the container even in failure

tianon:
:disappointed:


yosifkit:
Don't we still need the rm at the end too?

tianon:
Nope, since it's just trapping `EXIT`, it happens regardless of how we exit (success or failure).

yosifkit:
LGTM

PR #620, opened 2015-03-27
Add a "hello web" test for php:fpm

md5:


md5:
Just pushed another commit to this branch that makes the test easier to read and extend if necessary.

test/tests/php-fpm-hello-web/run.sh
```diff
@@ -0,0 +1,48 @@
...
+
+set -eo pipefail
+
+dir="$(dirname "$(readlink -f "$BASH_SOURCE")")"
+
+image="$1"
+
+# Build a client image with cgi-fcgi for testing
+client_image="php-fpm-client-$RANDOM-$RANDOM"
+docker build -q -t "$client_image" - > /dev/null <<EOF
```
yosifkit: Is there a way to do this without building an image?  This one is not cleaned up after the fact (but that would get non-trivial when testing all the versions of `php:fpm`).

md5: I'm not aware of any way to test it directly without using `cgi-fcgi`. I thought about the fact that the client image would be left behind, but I didn't make the fix. Should be easy enough to delete it along with the server container in the `trap`.

I can fix that later today. Let me know if you see any other changes you'd like made.

tianon: Well, as @yosifkit pointed out, if we just delete the image it'll have to
be rebuilt from scratch for every single tag we test, which will be really
slow. :cry:

md5: Got it. Aside from some naming convention for test related container that the top-level `run.sh` could clean up, I'm not sure what can be done about that. Container tagging/metadata would be a nice to have for this, assuming Docker ever gets it.

FWIW, building this client image is pretty quick since it's just installing a single package. Perhaps since this image doesn't need to be really tiny, there could be a base image for test client container that has already run `apt-get update`. That way the client install would be the only thing to wait for.

Just throwing out ideas. Not sure what the right approach is.

yosifkit: I had thought of having some sort of 'post-tests' cleanup script, that would see what tests ran and then clean up their stuff.  This would imply that the first run of the test would set up necessary things, since we do not know the complete list of tests until they are run.

Just my 2 :dollar:/100

tianon: Having an image namespace common to all tests doesn't seem like too bad of
an idea, especially if we can come up with a good one.  Then we can have
"run.sh" optionally clean up all images matching "ournamespace/*" (or just
not worry about it, if we can come up with an obvious enough name for the
namespace).

md5: How about `docker-library-test/*`? Then again, using un-namespaced names has the benefit that it can't conflict since it could only conflict with another official image :+1:

md5: Or just `library-test`. Maybe go reserve the namespace too.

tianon: "This value may contain only lower case letters and numbers, at least 4 and
no more than 30 characters" :cry:

tianon: "librarytest" is reserved now, though

md5: `$RANDOM` maxes out at 32767, so I guess I'll use `librarytest/fpm-$RANDOM-$RANDOM`, which will max out at 28 characters...

yosifkit: Hmmm, so an image with all the cli's straight from debian packages to test all the others? (`mysql-client`, `libfcgi0ldbl`, etc?)

tianon: Since we're namespacing, might as well just use the test name now (`librarytest/php-fpm-hello-web:fcgi` or something like that), especially since the image name doesn't need to be amazingly unique (like a container name needs to somewhat reasonably be).

md5: I actually just meant a "fat" image that is a locally built version of `debian:jessie` that has already run `apt-get update`. That way different client images wouldn't need to run it. Another option might be to actually publish the client images under the `librarytest` ns as automated builds.

md5: And now that I spell it out, the "fat" idea seems dumb since it would be some ad-hoc thing for people to make client image `FROM` that is only for official image tests.

md5:
I've updated the test to build the client in the `librarytest/*` namespace.

test/tests/php-fpm-hello-web/run.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+
+set -eo pipefail
+
+dir="$(dirname "$(readlink -f "$BASH_SOURCE")")"
+
+image="$1"
+
+# Build a client image with cgi-fcgi for testing
+client_image="librarytest/php-fpm-hello-web:fcgi"
```
yosifkit: Can we get some :camel:casing where possible?  :pray:

md5: Just here or over on `$cname` down below as well?

tianon: I'm meh on `cname` (that seems sane and reasonable unless you really feel like expanding it to a longer name), but the underscore here for sure. :+1:

tianon: (otherwise this test looks good to me now)

tianon: Oh and you should switch this to single quotes while you're there -- unnecessary interpolation and all that fun.

test/tests/php-fpm-hello-web/run.sh
```diff
@@ -0,0 +1,48 @@
...
+docker build -q -t "$client_image" - > /dev/null <<EOF
```
tianon: As long as you're here, you should `'EOF'` too.

test/tests/php-fpm-hello-web/run.sh
```diff
@@ -0,0 +1,48 @@
...
+	local method="$1"
+
+	local url="$2"
+	local queryString=
+	if [[ "$url" == *\?* ]]; then
+		queryString="${url#*\?}"
+		url="${url%%\?*}"
+	fi
+
+	docker run --rm -i --link "$cname":fpm \
```
tianon: Oh, can `--link` not take the `cid` directly?

test/tests/php-fpm-hello-web/run.sh
md5: So wait, underscore or camel case?

tianon: Camels all the way down!
```bash
clientImage='librarytest/php-....'
```

md5: I finally realized you meant the variable name... I thought you meant the value.

md5:
:heavy_check_mark: it

yosifkit:
LGTM

yosifkit:
On the premise that we need to possibly address cleanup of `librarytest/*` images.

tianon:
LGTM

PR #622, opened 2015-03-27
Add uniq option to bashbrew

yosifkit:
This will help in running tests.


tianon:
Failure here is Fedora...

Change LGTM :+1:

PR #633, opened 2015-04-01
md5's machine is faster at starting up FPM than mine, apparently

tianon:
cc @md5

### DIFF
--- a/test/tests/php-fpm-hello-web/run.sh
+++ b/test/tests/php-fpm-hello-web/run.sh
@@ -8,18 +8,21 @@ image="$1"
 
 # Build a client image with cgi-fcgi for testing
 clientImage='librarytest/php-fpm-hello-web:fcgi'
-docker build -q -t "$clientImage" - > /dev/null <<'EOF'
+docker build -t "$clientImage" - > /dev/null <<'EOF'
 FROM debian:jessie
 
-RUN set -x; apt-get update && apt-get install -y libfcgi0ldbl && rm -rf /var/lib/apt/lists/*
+RUN set -x && apt-get update && apt-get install -y libfcgi0ldbl && rm -rf /var/lib/apt/lists/*
 
 ENTRYPOINT ["cgi-fcgi"]
 EOF
 
 # Create an instance of the container-under-test
-cname="php-fpm-container-$RANDOM-$RANDOM"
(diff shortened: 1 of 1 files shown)


test/tests/php-fpm-hello-web/run.sh
```diff
@@ -8,19 +8,22 @@ image="$1"
...
-RUN set -x; apt-get update && apt-get install -y libfcgi0ldbl && rm -rf /var/lib/apt/lists/*
+RUN set -x && apt-get update && apt-get install -y libfcgi0ldbl && rm -rf /var/lib/apt/lists/*
 
 ENTRYPOINT ["cgi-fcgi"]
 EOF
 
 # Create an instance of the container-under-test
-cname="php-fpm-container-$RANDOM-$RANDOM"
-cid="$(docker run -d -v "$dir/index.php":/var/www/html/index.php:ro --name "$cname" "$image")"
+cid="$(docker run -d -v "$dir/index.php":/var/www/html/index.php:ro "$image")"
```
tianon: This is unrelated, but I realized we weren't even using `cname` anymore so I removed it. :+1:

test/tests/php-fpm-hello-web/run.sh
```diff
@@ -8,19 +8,22 @@ image="$1"
 
 # Build a client image with cgi-fcgi for testing
 clientImage='librarytest/php-fpm-hello-web:fcgi'
-docker build -q -t "$clientImage" - > /dev/null <<'EOF'
+docker build -t "$clientImage" - > /dev/null <<'EOF'
```
tianon: This one is because what `-q` actually does/means upstream isn't super well defined, and we're redirecting the output anyhow.  Just trying to avoid it breaking out from under us if Docker decides to make `-q` mean something.

md5:
LGTM. I actually considered putting a check like `mysql-basics` does and looping, but I wasn't sure what to call in that loop besides `fcgi-request`. Perhaps using the bash `/dev/tcp/fpm/9000` stuff to test whether FPM is listening might work.

tianon:
Yeah, it's not an obvious fix which is why I'm just punting for now. :innocent:

yosifkit:
LGTM

PR #634, opened 2015-04-02
update to fix centos-release package, add systemd

jperrin:


jperrin:
need to hold off on this one for now.

tianon:
It's probably worth reading through the discussion that starts in https://github.com/docker-library/official-images/pull/497#issuecomment-74713699, too. :wink:

PR #639, opened 2015-04-03
Add test/clean.sh script

### LINKED ISSUE #638: Implement cleanup for "librarytest/*" images after tests complete
Adding this issue here to track implementation of the `librarytest/*` image cleanup discussed in #620.

Are we just looking for a `trap _cleanup EXIT` in `run.sh` where `_cleanup` runs a `docker rmi` for all the images in that namespace?

md5:
Fixes #638

### DIFF
--- a/test/clean.sh
+++ b/test/clean.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+set -e
+
+docker images -q 'librarytest/*' | xargs docker rmi -f
--- a/test/clean.sh
+++ b/test/clean.sh
@@ -1,4 +1,4 @@
 #!/bin/bash
 set -e
 
-docker images -q 'librarytest/*' | xargs docker rmi -f
+docker images 'librarytest/*' | awk 'NR>1 { print $1":"$2 }' | xargs -r docker rmi


tianon:
OMG when were wildcards implemented in `docker images`, and why didn't I notice??

md5:
No idea, but I figured I'd test it to see if it worked. It's kind of surprising given that `image:tag` and `ID` don't work.

tianon:
LGTM, but I'm still reeling from the fact that I somehow missed such a great feature that's clearly working in Docker 1.5.0!

tianon:
So far, my digging shows it's been there since at least as far back as 1.2.0; still digging to see how deep this rabbit hole goes.

tianon:
0.10.0 :flushed:

tianon:
This feature goes all the way back to Docker 0.6.2, and was added in https://github.com/docker/docker/pull/1817.

tianon:
BTW, the "trail" I followed to find that PR was `graph/list.go`, `server/image.go`, `server/server.go`, `server.go` (every file that line of code has moved through)

test/clean.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+set -e
+
+docker images -q 'librarytest/*' | xargs docker rmi -f
```
yosifkit: Can we do a little changing here, to get rid of tags instead of images?  Images will go away, but only if nothing else is using it.
```bash
docker images 'librarytest/*' | awk 'NR>1 { print $1":"$2 }' | xargs -r docker rmi
```

tianon: Ah, good point; there is also a possibility (however slim) that one of our
librarytest images is the same layer as another legitimate image that's not
ours, especially since the Dockerfiles will usually be really simplistic.
:innocent:

md5:
Updated to only remove tags instead of images themselves. Too bad there isn't just a `docker untag`.

md5:
Of course, this introduces another GNU-ism that makes it impossible to run the `official-images` scripts easily under OS X.

tianon:
OS X should just adopt the GNU :smiley_cat:

tianon:
(I'm OK with skipping the `-r`, but it does remove the `docker: "rmi" requires a minimum of 1 argument. See 'docker rmi --help'.` error nicely.)

md5:
I love GNU, I really do, but I don't see OS X doing that. I should probably just go back to Linux.

I'll leave the `xargs -r` there since there are `readlink -f` and a few other GNU-ishs and Bash 4 things that make other scripts already not work under OS X.

tianon:
Ok, fair enough. :heart:

LGTM

yosifkit:
LGTM

md5:
I just realized this is going to choke if someone creates an image in the `librarytest/*` namespace without a tag. The `awk` command will return `librarytest/foo:<none>`, which will make `docker rmi` sad :cry:

md5:
Or perhaps Docker just doesn't allow what I'm saying, since it defaults to tagging as `latest`.

tianon:
That's the one :+1:

tianon:
(Sans the new v2 digests thing)

PR #641, opened 2015-04-03
rename to python-imports and add bz2 test

### LINKED ISSUE #560: [test] rename python-zlib to python-imports and add bz2
As pointed out in https://github.com/docker-library/python/issues/39, we need to make sure `bz2` loads correctly (especially since it can't be added/fixed after the fact).

To do this, we should rename the current `python-zlib` test to something like `python-imports` and have it do some very rudimentary round-tripping via `bz2` like it does `zlib`.

denibertovic:
Closes #560

cc @tianon @yosifkit


tianon:
LGTM :+1:

yosifkit:
LGTM

PR #643, opened 2015-04-03
Add official image for regsitry 2.0

stevvooe:
Closes docker/distribution#322

Signed-off-by: Stephen J Day <stephen.day@docker.com>

### DIFF
--- a/library/registry
+++ b/library/registry
@@ -1,6 +1,12 @@
 # maintainer: Joffrey F <joffrey@docker.com> (@shin-)
 # maintainer: Sam Alba <sam@dotcloud.com> (@samalba)
+# maintainer: Stephen Day <stephen.day@docker.com> (@stevvooe)
+# maintainer: Olivier Gambier <olivier.gambier@docker.com> (@dmp42)
 
 latest: git://github.com/docker/docker-registry@0.9.1
 0.8.1: git://github.com/docker/docker-registry@0.8.1
 0.9.1: git://github.com/docker/docker-registry@0.9.1
+
+2: git://github.com/docker/distribution@62b70f951f30a711a8a81df1865d0afeeaaa0169
+2.0: git://github.com/docker/distribution@62b70f951f30a711a8a81df1865d0afeeaaa0169
+2.0.0: git://github.com/docker/distribution@62b70f951f30a711a8a81df1865d0afeeaaa0169


yosifkit:
Finally getting to this, sorry for the delay.  A few comments:

- why is it `registry2`? this is going to cause a lot of user confusion, since it's referred to as "registry v2" but the repo is actually "distribution", so `distribution` or even `registry:2.0.0` or `registry:v2-2.0.0` seem more appropriate
  - the tagging on `registry` would depend on if there would be a "registry v3" or would it be an incremental upgrade of the "registry v2" (ie, registry v2 version 3.0)
- we [prefer commit ids rather than tags](https://github.com/docker-library/official-images#instruction-format); since tags could change reference and will not cause a rebuild
- your `master` tag following a branch will not work in an official image; we only rebuild when a PR is given to this file or when the base image changes, and even then rely heavily on docker cache to minimize changes for users.

stevvooe:
@yosifkit

1.  It is "registry2" to differentiate the product from "registry". They are different products and not necessarily backwards compatible. Placing artifacts into "registry" that are from distribution will cause people to infer direct compatibility, which is not the case.  The artifact "registry2" may come from distribution but this shouldn't matter to users. We'll make this clear in the messaging.
2. We'd like to drop the terminology "registry v2" in written prose. it's imprecise and unclear whether it means the API, a service or something else.
3. The tags are signed and part of releases [here](https://github.com/docker/distribution/releases). These will not be changed.
4. How can we have an official image track `master`?

yosifkit:
I think having `registry` and `registry2` will confuse most users.  I am -1 on having arbitrary numbers in the repo name.  As you note, you are trying to get away from "registry v2" but keeping it as `registry2` will only exacerbate the problem.  A clean cut like you have already done with the name of your repo would be the best approach, ie `distribution`.

Even though they are release tags, we ["heavily discourage"](https://github.com/docker-library/official-images#instruction-format) tags and prefer commits.  We reserve the right to deprecate use of tags :wink:.

We try to do a thorough review of every official image to ensure a level of consistency and quality, as such, tracking a `master` branch is not something feasible for official images.  They are only meant for releases and in some cases release candidates.  So far we have had [bad luck](https://github.com/docker-library/mariadb/pull/7) with supporting beta versions, so they are a case by case basis.  See [swarm](https://registry.hub.docker.com/_/swarm/) for an example of releases vs [master](https://registry.hub.docker.com/u/dockerswarm/swarm/).

stevvooe:
@yosifkit I've updated the PR with the tagged releases using git commit ids. The "master" tag has been removed.

 Unfortunately, calling the image `distribution` would be just as confusing for the users. The term `distribution` is a project but not necessarily an individual product. Users do not run `distribution`. They run a "registry" and I don't want to encourage calling it "distribution" at all.

That said, if you're willing to deal with user fallout in switching the "latest" on an existing image (ie "registry") to something that is not runtime compatible, we can go ahead with this. Otherwise, I suggest this should be called "registry2" to differentiate it. In the future, when the python registry is gone, at that time, it might okay to move "registry2" to "registry".

tianon:
Since we don't have an effective way to rename a repository once it's up, I think it'd be best if we treat this like other jarring transitions in the software world.  For example, the transition from PHP 4 to PHP 5 was a bit rocky, but nobody would contest that in the end, they're both PHP, even if fundamentally incompatible in multiple ways.

As long as we clearly document that this new release is incompatible, I think we're fine with taking over the `registry` namespace right away, especially if we keep the old versions for as long as they're still supported, even if just in "maintenance mode".

In that regard, I think it's probably best if we leave the current `registry:latest` alone until we're ready for "the masses" to switch their systems over.  We always recommend that if someone is relying on an image not changing in disruptive ways that they should either use a specific version, or if they need even stronger assurances, re-tag the image and maintain their own copy (similar to vendoring in Go, which solves a similar problem in a similar way).

dmp42:
Chipping in: the plan is to update the "registry" image so that it does contain *both* python and go registries, plus nginx (similarly to https://github.com/docker/distribution/tree/master/contrib/compose)

That will let people smoothly move forward, and we could safely update registry:latest to point to that, letting users fine tweak their config to use one and/or the other, and is the sanctioned transition path.

Admittedly, though, this will make for a quite "bloated" image.

Now, some (most?) people just want to run the golang registry, and use the "latest" of that.
We can certainly have a `stevvooe/registry` image that is not official, but I would rather avoid that situation (we suffered enough from this with the infamous `samalba/registry` image), which is why we are submitting this one.

What do you think?

Thanks for caring @tianon :)

cpuguy83:
I don't think it makes sense to do that.
The older version can always be updated to include nginx, other tweaks meanwhile `registry:2.0` can be it's own thing.

dmp42:
@cpuguy83 so, should registry:latest point to the latest python registry?
Or to the latest golang registry?
If the former, why would the old registry be the "latest", and if the latter, we will break things suddenly for people.

Wdyt?

cpuguy83:
@dmp42 I'd agree with @tianon that the `latest` tag is kind of special.
Until you are ready to take on the masses using 2.0 it should not be latest, since it is a massively breaking change.

md5:
If you guys are cutting actual RC releases, it might be good to only have very specific tags like `2.0.0-rc1` and not aliases like `2.0` until you want to support a broad user base.

md5:
And I see from the diff that you do indeed have RCs.

yosifkit:
I'll jump back in agree with @cpuguy83.  Keeping `registry:latest` to be the python version will be a nice stopgap to keep those users running, while adding `registry:2.0.0-rc.3` will allow users to get on board with the new developments in the go version.  We'll just need to clearly document the differences in the [readme](https://github.com/docker-library/docs/blob/master/registry/content.md) for docker hub and ensure users can understand which is which and when they would need to use each one as well as get to the respective documentation.

I would definitely be against a single container that had both registries and nginx when that is something that could easily be done with a few nginx configuration files and a `docker-compose.yml`.

jfrazelle:
ping @tianon @yosifkit can we get this in so we can do the release, unsure what are the pending action items

stevvooe:
@jfrazelle @tianon This is updated with the release shas and ready to go!

yosifkit:
I don't want to hold up the release, but I have reservations with these new images, namely I cannot do `docker run -it registry:2 bash` like any other official image allows.  Users like to jump in the image to see what is there and an entrypoint of `registry` makes it less obvious.  They have to know that they instead need `docker run -it --entrypoint bash registry -i`.  I would much prefer consistency across all official images.

bfirsh:
@yosifkit The Swarm image uses an entrypoint like this.

stevvooe:
@yosifkit We've already tagged a release. You're welcome to submit a PR to docker/distribution or create a Dockerfile project to manage this but it is too late at this point.

tianon:
The Swarm image doesn't have a shell at all -- it's literally _just_ the
Swarm binary.

md5:
The `swarm` image also only has the `swarm` binary in it and doesn't write to disk.

tianon:
I'm really quite disappointed that the desire to have this image ready for
the release of Docker 1.6.0 wasn't communicated sooner than today when it's
time for release.

stevvooe:
@tianon This PR was created two weeks ago.

tianon:
Yes, and the review process isn't a free ride.  The whole point of the
official images program is that it is a curated library of images that
users can trust to be reasonably consistent.

tianon:
As a compromise, I'm willing to merge this if we drop all the RC releases and commit to swap `ENTRYPOINT` out for something that allows `docker run <image> bash` before `:latest` changes over to the v2 series.

(or switch to a release method similar to what was done for `swarm` where we have a _much_ smaller image consisting of just this binary and `ca-certificates` if necessary updated-via-script appropriately)

bfirsh:
@tianon That sounds like a reasonable compromise. I think a really small image with just the registry binary is a good idea, so this could be stopgap that behaves the same as that image would do.

The advantage of it behaving like this from the POV of the user, is that they can do things like:

```
$ docker run registry:2.0 --help
$ docker run registry:2.0 path/to/config/file
```

Without this, they'd have to type `docker run registry registry ...`.

tianon:
That pattern can also be accomplished with a shell script, similar to how [the `mongo` image does it](https://github.com/docker-library/mongo/blob/7d09ac9eb83e459dc4c66fd7438e46cc969813f3/3.0/docker-entrypoint.sh#L4-L6) :wink:.

stevvooe:
@tianon RCs are removed.

bfirsh:
@tianon Oh neat, perhaps we could add something like that.

Are we ready to go besides that? So sorry for the short notice. :(

yosifkit:
Running the build test, but LGTM

tianon:
Indeed, thanks.  As soon as @yosifkit's test is complete, we can merge and build/push.

yosifkit:
Build test of #643; 682b2827f1359027107d842457fec4592b6660f7 (`registry`):

```console
$ url="https://raw.githubusercontent.com/docker-library/official-images/682b2827f1359027107d842457fec4592b6660f7/library/registry"
$ bashbrew build "$url"
Processing registry:latest ...
Processing registry:0.8.1 ...
Processing registry:0.9.1 ...
Processing registry:2 ...
Processing registry:2.0 ...
Processing registry:2.0.0 ...
$ bashbrew list "$url" --uniq | xargs test/run.sh
testing registry:latest
	'utc' [1/2]...passed
testing registry:0.8.1
	'utc' [1/2]...passed
testing registry:2
	'utc' [1/2]...passed
```

PR #645, opened 2015-04-04
Let bashbrew tolerate EOL diversity

md5:
This fixes issues where the library file has line endings besides `LF`.

### DIFF
--- a/bashbrew/bashbrew.sh
+++ b/bashbrew/bashbrew.sh
@@ -195,7 +195,7 @@ for repoTag in "${repos[@]}"; do
 	fi
 	
 	# parse the repo library file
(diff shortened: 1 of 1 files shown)


tianon:
Got to be honest with you @md5, I'm torn on this one (as you probably anticipated). :smile:  Part of me wants to say "no" just so that we don't encourage `\r` to exist in these files, but I think it might finally be time instead for a `bashbrew lint` script that could check for things like this and give you a more useful error message instead.

md5:
I think a `bashbrew lint` would be fine. I'm just anticipating that you'll probably have other try to submit library files with `CR` in them and it would be nice for it to either work or given a decent error message.

PR #654, opened 2015-04-13
Add official roma image to dockerhub

banrui:
Hi Docker team, I'm Rui Bando from ROMA team. 
ROMA is one of the data storing systems for distributed key-value stores. 
Refere to [website](http://roma-kvs.org).

We would like to create a official repository. So I sent 2 pull requests.
- docs repo with ROMA image info.
- oficial-images repo with the image commit.

### DIFF
--- a/library/roma
+++ b/library/roma
@@ -1,0 +1,4 @@
+# maintainer: Rui Bando <bando.rui@gmail.com> (@banrui)
+
+latest: git://github.com/roma/docker@95233a26fe743dea843a68f4f8120079915c8f7a
+1.1.0: git://github.com/roma/docker@95233a26fe743dea843a68f4f8120079915c8f7a


tianon:
I've got to be honest, I'm having a hard time wrapping my head around this one.  From the documentation you've written in https://github.com/docker-library/docs/pull/206, it sounds like the way you expect this image to be run is just to get the `roma-client` gem for use in applications that want to talk to a `romad` running elsewhere?  Wouldn't those applications already have `roma-client` in their `Gemfile`, making using this base image completely moot?

I think the much more interesting thing to do here would be to run `romad` in a container, which is included in the `Dockerfile` installation, but isn't even hinted at in the documentation, nor is it obvious how you expect users to do so successfully from the `Dockerfile` (which is really complicated for what from reading http://roma-kvs.org/getting_started.html sounds like should as easy as `gem install roma --version 1.1.0`).

Where this appears to get complicated is that users have to run `mkroute` for all the `romad` instances they want to run in their cluster, and then they have to have that configuration available to each instance of `romad` that is actually run, which is fairly complex orchestration for what purports to be a simple key-value store.  Have I missed something that makes this simpler, especially for running `romad` inside containers?

yosifkit:
Closing, feel free to ping if you feel we should revisit this.

PR #675, opened 2015-04-22
Fix a few (but not all) too-complex expressions that won't fail properly with just "set -e"

tianon:


bashbrew/bashbrew.sh
```diff
@@ -280,6 +278,22 @@ for repoTag in "${repos[@]}"; do
...
+		cd "$1"
+		git reset -q HEAD
+		git checkout -q -- .
+		git clean -dfxq
+		git checkout -q "$2" --
+		cd "$1/$3"
+		"$dir/git-set-mtimes"
+	)
+	return 0
+}
```
tianon: This one still won't "fail" properly, but at least it's abstracted for better readability.

md5:
Would `-o pipefail` help with any of these?

tianon:
Nope, unfortunately not. :cry:

https://github.com/docker/docker/pull/12500 actually has quotes and links to the relevant Bash documentation for where `set -e` falls on its face, which makes me a really sad panda (the gist being that it falls on its face any time an expression gets "complicated" -- inside expressions including `&&`, `||`, `if` context, etc).

md5:
Gotta work with what you're given I suppose.

tianon:
Yeah, I still love Bash, but this knowledge put a serious dent in my Bash-as-a-serious-programming-language-contender argument. :cry: :disappointed:

yosifkit:
LGTM

PR #676, opened 2015-04-22
Add "retries" logic for build and push (since sometimes it just needs a little kick and everything works fine)

tianon:
Depends on #675 (so review that first, and I'll rebase this).


tianon:
Rebased :heart:

tianon:
This will help with both intermittent push failures _and_ intermittent build failures (like an APT mirror having a hiccup, which happens every now and then).

yosifkit:
LGTM

PR #683, opened 2015-04-27
Debian: current latest is jessie

pierrecdn:
Update docker-library/debian to jessie as latest


yosifkit:
Hmm, looks like the current builds were created on 2015-04-21. @tianon, were you going to make a new tarball for the release of Jessie?

tianon:
I am planning to do a rebuild this week, especially to swap `oldstable`, `stable`, and `testing`, and to add `stretch`.

yosifkit:
Sounds good, I'll close this as it is unnecessary.

PR #689, opened 2015-04-28
Add new drupal image

tianon:


tianon:
Build test of #689; 15fc46735ec99d9007d87693f39d19c10fda35d5 (`drupal`):

```console
$ url="https://raw.githubusercontent.com/docker-library/official-images/15fc46735ec99d9007d87693f39d19c10fda35d5/library/drupal"
$ bashbrew build "$url"
Processing drupal:7.36 ...
Processing drupal:7 ...
Processing drupal:latest ...
testing drupal:7.36
	'utc' [1/2]...passed
```

md5:
Why `php:5.5-apache` instead of `php:5.6-apache`? The [Drupal requirements](https://www.drupal.org/requirements) seem to say:

> * Drupal 6: PHP 5.x only (5.2.5 or higher recommended). Warning: support for PHP 4.x has been dropped. Drupal core should work with PHP 5.3.x, but PHP 5.3.x and higher may produce errors or unexpected behavior especially for contributed modules and themes.
> * Drupal 7: PHP 5.2.5 or higher (5.4 or higher recommended).
> * Drupal 8: PHP 5.4.5 or higher

tianon:
That was specifically because the chart at https://www.drupal.org/requirements/php#drupalversions implies that Drupal 7 + PHP 5.6 hasn't received nearly as much attention and testing as PHP 5.5 has (which is a solid green "yes" instead of that orange "??").  Since 5.5 is still supported by upstream PHP, it seemed like a reasonable thing to stick with 5.5 for now since the docs claim pretty strongly that it's well-tested and supported in that combination.

md5:
SGTM

md5:
Wait, the PHP folks are skipping PHP 6 and going straight to PHP 7? :see_no_evil:

tianon:
Yeah...  PHP 6 was mis-advertised so broadly that they're skipping it to
avoid the sigma associated with not delivering what was advertised, IIRC.

tianon:
*stigma, even

yosifkit:
https://wiki.php.net/rfc/php6

md5:
@yosifkit :+1:

yosifkit:
I hate to do this, but: [drupal.org](https://www.drupal.org/documentation/version-info) says "At any given time, there are two major release series of Drupal which are supported".  Sooo... we should add Drupal 6.

tianon:
So, the reason we didn't do that is because anything above PHP 5.2 with Drupal 6 is sketchy (as seen in https://www.drupal.org/requirements/php#drupalversions linked above), and 5.5+ is flat out failure.

yosifkit:
Sounds reasonable to me, PHP does [not support version 5.2](https://php.net/supported-versions.php), so Drupal 6 is out.

yosifkit:
LGTM

PR #681, opened 2015-04-23
Checking sha1 hashes for jruby

cpuguy83:

PR #701, opened 2015-04-30
Comment out "dbm" from the "ruby-standard-libs" test for now

tianon:


tianon:
ie, the fix for the problem in https://github.com/docker-library/official-images/pull/681#issuecomment-96068035

yosifkit:
LGTM

tianon:
338950e :cry:

PR #711, opened 2015-05-02
Remove "latest" tag from "docker-dev"

tianon:
# "latest" intentionally missing, since "docker-dev:latest" implies (IMO) the "latest development image", which these builds never will be
	# if that is what you are after, see "dockercore/docker" (https://registry.hub.docker.com/u/dockercore/docker/)

This is just to further solidify the idea that "docker-dev" is _not_ for the latest and greatest builds of `github.com/docker/docker`.

### DIFF
--- a/library/docker-dev
+++ b/library/docker-dev
@@ -3,6 +3,8 @@
 1.6.0: git://github.com/docker/docker@v1.6.0
 1.6: git://github.com/docker/docker@v1.6.0
 1: git://github.com/docker/docker@v1.6.0
-latest: git://github.com/docker/docker@v1.6.0
+
+# "latest" intentionally missing, since "docker-dev:latest" implies (IMO) the "latest development image", which these builds never will be
+# if that is what you are after, see "dockercore/docker" (https://registry.hub.docker.com/u/dockercore/docker/)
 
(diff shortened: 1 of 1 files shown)


tianon:
This was also partially a result of the discussion in https://github.com/docker-library/docker/issues/2. :+1:

tianon:
It's also worth dropping a pointer to https://github.com/docker-library/docs/pull/195 here, just to tie all the strings together.

yosifkit:
LGTM

yosifkit:
But won't that break people that try to `docker pull docker-dev` when `1.7` comes out and then they get `1.6`?

tianon:
I actually plan to delete the `latest` tag from the hub pending this merge so that it doesn't exist at all.

yosifkit:
That sounds a little dangerous, but I have no alternative.

tianon:
Heh, the alternative would be not to build it at all.  This image exists due to popular demand, but "latest" is really, really misleading, and not at all helpful to anyone.

yosifkit:
Well... OK. But there will be issues of "I can't `docker pull docker-dev`".

tianon:
We'll just have to point them all to the "dockercore/docker" image they're
really looking for. :+1:

PR #721, opened 2015-05-06
Initial test to check for a root password

### LINKED ISSUE #720: [test] root password
Would it be simple for us to detect whether `root` has a password set within the image?  If so, we should add a test for it! :+1:

ref https://botbot.me/freenode/docker-library/msg/38427979/

yosifkit:
Still needs some work. But here is a start to fix #720.

### DIFF
--- a/test/config.sh
+++ b/test/config.sh
@@ -4,6 +4,7 @@ set -e
 globalTests=(
 	utc
 	cve-2014--shellshock
+	no-root-password
 )
 
 declare -A testAlias=(
@@ -110,6 +111,9 @@ declare -A globalExcludeTests=(
 	# single-binary images
 	[hello-world_utc]=1
 	[swarm_utc]=1
+	
+	[hello-world_no-root-password]=1
+	[swarm_no-root-password]=1
 
 	# no "native" dependencies
 	[ruby:slim_ruby-bundler]=1
--- a/test/tests/no-root-password/run.sh
+++ b/test/tests/no-root-password/run.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+set -e
+
+pass="$(docker run --rm --entrypoint awk "$1" -F ':' '$1 == "root" { print $2 }' /etc/passwd)"
+if [ "$pass" = 'x' ]; then
+	pass="$(docker run --rm --entrypoint awk --user root "$1" -F ':' '$1 == "root" { print $2 }' /etc/shadow)"
+fi
+[ -z "$pass" -o "$pass" = '*' ]


tianon:
Seems sane for a first pass; definitely needs more love though. :+1:

test/tests/no-root-password/container.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+set -e
+
+[ "$(grep root /etc/shadow | cut -d: -f2)" = '*' ]
```
tianon: I think `awk` would probably be more well-suited for parsing `/etc/shadow`; something like `awk -F ':' '$1 == "root" { print $2 }' /etc/shadow`, right?

tianon: I also wonder whether we can/should check against the hashed version of the empty string instead of just `*`.

tianon: But `*` is the obvious easy one :smile:

yosifkit:
Updated!  Now we have two problems:
- need root to access `/etc/shadow`
  - add `--user root` to the `docker run` (add to [run-in-container](https://github.com/docker-library/official-images/blob/4e9cd0e7ec2de5fa3c14a0be6c8625161f9b97c4/test/tests/run-in-container.sh)?)
- `exec: "bash": executable file not found in $PATH`
  - need generic solution for no bash, (alpine, busybox, cirros, hello-world, swarm)

tianon:
Hmm, I'm not sure `--user root` is going to be OK for _all_ the `run-in-container` users.

Something like how `cve-2014--shellshock` does "bash detection" should be reasonable, I'd think: https://github.com/docker-library/official-images/blob/master/test/tests/cve-2014--shellshock/run.sh

```bash
if ! docker run --rm --entrypoint bash "$1" -c 'true' &> /dev/null; then
	# die quietly and gracefully if this image doesn't have bash at all
	exit 0
fi
```

(or maybe we go for strict `/bin/sh` so we can cover `busybox` and `alpine` too?)

tianon:
In theory, we could eschew in-container shell entirely and just `docker run ... awk ...` directly.

tianon:
It'd be really super if we had something like `docker cp` but for files in an image... :confused:

test/config.sh
```diff
@@ -110,7 +111,9 @@ declare -A globalExcludeTests=(
 	# single-binary images
 	[hello-world_utc]=1
 	[swarm_utc]=1
-
+	[hello-world_utc_no-root-password]=1
+	[swarm_no-root-password]=1
+
```
yosifkit: woops: hello-world_utc_no-root-password -> hello-world_no-root-password

yosifkit:
Output of test (grep only failing images for brevity):
```console
$ ./bashbrew/bashbrew.sh list --namespaces='_' --all --uniq | xargs ./test/run.sh -t no-root-password  | grep -B1 failed
awk: /etc/shadow: No such file or directory
testing busybox:ubuntu-12.04
	'no-root-password' [1/1]...failed
awk: /etc/shadow: No such file or directory
testing busybox:ubuntu-14.04
	'no-root-password' [1/1]...failed
--
testing centos:latest
	'no-root-password' [1/1]...failed
testing centos:centos6
	'no-root-password' [1/1]...failed
testing centos:centos5
	'no-root-password' [1/1]...failed
testing centos:centos7.1.1503
	'no-root-password' [1/1]...failed
testing centos:centos7.0.1406
	'no-root-password' [1/1]...failed
testing centos:centos6.6
	'no-root-password' [1/1]...failed
testing centos:centos5.11
	'no-root-password' [1/1]...failed
testing cirros:latest
	'no-root-password' [1/1]...failed
--
testing fedora:latest
	'no-root-password' [1/1]...failed
testing fedora:rawhide
	'no-root-password' [1/1]...failed
testing fedora:20
	'no-root-password' [1/1]...failed
--
testing opensuse:13.1
	'no-root-password' [1/1]...failed
testing opensuse:13.2
	'no-root-password' [1/1]...failed
testing opensuse:tumbleweed
	'no-root-password' [1/1]...failed
testing oraclelinux:5
	'no-root-password' [1/1]...failed
--
testing ubuntu-upstart:12.04
	'no-root-password' [1/1]...failed
testing ubuntu-upstart:14.04
	'no-root-password' [1/1]...failed
testing ubuntu-upstart:14.10
	'no-root-password' [1/1]...failed
```

tianon:
LGTM

tianon:
Probably worth opening issues or PRs at the upstream repos that failed.

PR #722, opened 2015-05-06
Update library/registry image to v2.0.1

RichardScothern:


yosifkit:
Changset: https://github.com/docker/distribution/compare/62b70f...134122 (no change to the Dockerfile).

(I am still pulling updated images to my local machine, so I will get a build test done in the morning)

yosifkit:
Build test of #722; 29828141ed440ca1e19e1a790bea012c64eae379 (`registry`):

```console
$ url="https://raw.githubusercontent.com/docker-library/official-images/29828141ed440ca1e19e1a790bea012c64eae379/library/registry"
$ bashbrew build "$url"
Fetching registry (git://github.com/docker/distribution) ...
Processing registry:latest ...
Processing registry:0.8.1 ...
Processing registry:0.9.1 ...
Processing registry:2 ...
Processing registry:2.0 ...
Processing registry:2.0.0 ...
Processing registry:2.0.1 ...
testing registry:latest
	'utc' [1/2]...passed
testing registry:0.8.1
	'utc' [1/2]...passed
testing registry:2
	'utc' [1/2]...passed
testing registry:2.0.0
	'utc' [1/2]...passed
```

library/registry
```diff
@@ -1,12 +1,12 @@
...
 
 latest: git://github.com/docker/docker-registry@0.9.1
 0.8.1: git://github.com/docker/docker-registry@0.8.1
 0.9.1: git://github.com/docker/docker-registry@0.9.1
 
-2: git://github.com/docker/distribution@62b70f951f30a711a8a81df1865d0afeeaaa0169
-2.0: git://github.com/docker/distribution@62b70f951f30a711a8a81df1865d0afeeaaa0169
+2: git://github.com/docker/distribution@1341222284b3a6b4e77fb64571ad423ed58b0d34
+2.0: git://github.com/docker/distribution@1341222284b3a6b4e77fb64571ad423ed58b0d34
 2.0.0: git://github.com/docker/distribution@62b70f951f30a711a8a81df1865d0afeeaaa0169
```
tianon: What's the rationale for keeping `2.0.0` in here still?  We won't delete the tag, so this just controls whether it gets rebuilt (and if it won't receive any changes or fixes in the future, it's hard to call it "supported" :smile:).

library/registry
```diff
@@ -1,12 +1,12 @@
-# maintainer: Joffrey F <joffrey@docker.com> (@shin-)
-# maintainer: Sam Alba <sam@dotcloud.com> (@samalba)
 # maintainer: Stephen Day <stephen.day@docker.com> (@stevvooe)
 # maintainer: Olivier Gambier <olivier.gambier@docker.com> (@dmp42)
+# maintainer: Richard Scothern <richard.scothern@gmail.com>
```
tianon: Can you add ` (@RichardScothern)` to the end here?

tianon:
@stevvooe I'm guessing the answer is yes, but you're OK with @RichardScothern adding himself as an image maintainer here, right? :smile:

library/registry
stevvooe: @tianon It's standard procedure to not get rid of old releases when releasing new software. For example, if someone setup there infrastructure to pin to that version, we should not break it by removing that version. No, it is not supported, but if they want to run that version, it is their choice and not ours.

tianon: Right, which is exactly why we don't delete the tags from the Hub.  This
file specifies which versions are supported and thus rebuilt.

stevvooe: That is an odd model: I would expect this to be the list of currently available versions of the particular software. Is this clarified anywhere?

tianon: It's an odd problem, too.  This file isn't the release artifact, it's just the instructions to the "release artifact builder" for what release artifacts to build.  The actual release artifact is the built image, so those are tracked completely via tags on the Hub.

See tags listed on https://registry.hub.docker.com/_/redis/ vs https://registry.hub.docker.com/_/redis/tags/manage/ for a good concrete example of this.

I'm not seeing this documented well in our `README.md` though, so that's definitely something we need to rectify (it's implied by phrases like `The generated image will be tagged as <manifest-filename>:<docker-tag>.`, but that's not nearly explicit enough to make it obvious). :+1:

stevvooe:
@tianon Yes, please add @RichardScothern.

library/registry
stevvooe: @tianon Does docker library keep track of the git hashes used to build each tag in the past? By removing them from this file, it seems like we may be losing the documentation of what tag is associated with which git hash. This doesn't seem like a problem but if we are removing the old ones on each release, it seems rather pointless to disallow the use of tags over hashes; the immutability doesn't matter if the line item isn't built on the next release.

@RichardScothern I suppose we should drop the 2.0.0 tag, based on this feedback.

tianon: The Git hash is preferred for repeatability of _this_ build (ie, the base
image changes and we rebuild supported tags of this image as a
consequence).  A tag can be easily changed in the upstream repo, leaving us
with inconsistent state of "what's 2.0.0 right now" (since some builders
likely already have it cloned and wouldn't have an obvious indicator that
they need to fetch again).  If we use a direct Git hash, it is guaranteed
to be immutable (ignoring potential hash collisions, but those are rare
enough that I'd love to see one in the wild).

stevvooe: @tianon I'm confused. Why require the commitish to be immutable if the file is not an immutable description of the contents of the repository? Ultimately, it sounds like these are time-based snapshots, either way, be it by git tag or a PR that changes or removes targets.

md5: @stevvooe  How could this file be an immutable description of the contents of the image if the `Dockerfile` contains something like `FROM golang:1.4`, which is a floating tag (which derives from `buildpack-deps:jessie-scm`, another floating tag)? The goal is not to have an immutable description of the image, but rather a repeatable build to whatever extent possible. Mandating an immutable git tag is one of the few things that can reduce the unpredictability of the image build.

library/registry
RichardScothern: Done!

RichardScothern:
ping @tianon

yosifkit:
Another build test for good measure: Build test of #722; 1d11566b4500bdc156cd59a94cfbeccb9608054b (`registry`):

```console
$ url="https://raw.githubusercontent.com/docker-library/official-images/1d11566b4500bdc156cd59a94cfbeccb9608054b/library/registry"
$ bashbrew build "$url"
Processing registry:latest ...
Processing registry:0.8.1 ...
Processing registry:0.9.1 ...
Processing registry:2 ...
Processing registry:2.0 ...
Processing registry:2.0.1 ...
testing registry:latest
	'utc' [1/2]...passed
testing registry:0.8.1
	'utc' [1/2]...passed
testing registry:2
	'utc' [1/2]...passed
```

tianon:
LGTM

PR #724, opened 2015-05-08
point to updated clojure with increased security

Quantisan:
Made some security enhancements on the Clojure image for v2.5.1:

- use non-root user to run
- check SHA of downloaded `lein` package to reduce man-in-the-middle attack
- use the `lein-pkg` script to install as recommended by Leiningen team for downstream packagers

PR #755, opened 2015-05-26
removing lein 2.4 as it's been deprecated by 2.5

Quantisan:
you're right there https://github.com/docker-library/official-images/pull/724#issuecomment-101058913 that 2.4 isn't needed anymore


tianon:
LGTM :+1: :heart:

PR #777, opened 2015-06-01
Adding whalesay image

moxiegirl:
For Dockercon, the WWW team is creating a tutorial that walks users through a demo of the product.   For the demo, users run and later build their own demo from whalesay.  So, it would be better if the image were an "official image" because we want users to search for official images as part of the demo.

Actually, I'm not sure if this is entirely the correct format.  I tried to follow the doc but it was a bit too general for me.  Um, I'm adding fixing that to my todo list.

Signed-off-by: Mary Anthony <mary@docker.com>

### DIFF
--- a/library/whalesay
+++ b/library/whalesay
@@ -1,0 +1,3 @@
+# maintainer: Mary Anthony <mary@docker.com> (@moxiegirl)
+
+latest: git://github.com/git@github.com:moxiegirl/whalesay@a2e4beab921d577835e1ead253002275a84afc93


tianon:
I've provided some technical suggestions for improving the `whalesay` Dockerization in https://github.com/moxiegirl/whalesay/pull/1, but I've got to say I'm generally -1 on it being an official image.  I understand what you're trying to accomplish, but this software doesn't really meet _any_ of the criteria we usually look for in an official image (well-established, reasonable user base, etc).

The [Docker documentation on this subject](https://docs.docker.com/docker-hub/official_repos/#how-do-i-create-a-new-official-repository) has the following (thanks to @psftw):

> There are also subjective considerations during the review process. These subjective concerns boil down to the basic question: "is this image generally useful?" For example, the `python` Official Repository is "generally useful" to the large Python developer community, whereas an obscure text adventure game written in Python last week is not.

See also https://github.com/docker-library/official-images/pull/309 which was a similar request.

moxiegirl:
@tianon I don't know that it is similar. I could have used that image I suppose.  This is a useful image, we are using it as a teaching tool for Docker itself.  Users are already creating hello-world images for themselves and cowsay images as well.  

This is part of an initial experience I'm developing for Docker's website. To provide a detailed teaching piece in the documentation that helps the most people, I need to have a very controlled environment.  I will use this image to teach any number of things:

* how to locate an official image
* how to run an official image
* how to modify an image with a Dockerfile
* how to tag and  push an image

Moghedrin:
I'm strongly with @tianon on this one. I feel like there *might* be a stronger case for making this an official image if the only images you could search for were the official images. Since locating and running official images is **exactly** identical to normal images (except the few characters that would be saved), I don't see a strong reason to include this as an official image.

APhadnis:
Hi guys,

Is this any different from the hello-world Docker Image that is marked as Official Image?
My sense is that there is not but love to hear your thoughts.

Since this image will be experiencing heavy traffic, I’d like to associate it with the Docker project.
Can we post this image under the Docker project but NOT mark it as official? This alternative I think is satisfactory.

Thanks.

tianon:
I don't know who's in control of https://registry.hub.docker.com/repos/docker/, but I definitely think that'd be a great place for this image to live.

In my mind, what makes this different from `hello-world` is that `hello-world` is both the "example" official image (in exactly the same vein as the [`hello` package in Debian and Ubuntu](https://packages.debian.org/sid/hello)), and is a really tiny entrypoint for testing that Docker works successfully.  Additionally, there isn't likely to be a _serious_ project named "hello-world", so there's not much chance for overlap.  It also makes for a good example of effective `FROM scratch` usage.

What we've got with `whalesay` is simply a demonstration of wrapping another project with some minor modifications.  It's a perfect Dockerization example for the docs and training, but it's not a good candidate for the official images as such.

As an additional note, the vast majority of the official images are built explicitly such that `FROM <image>` can be used instead of taking the official image source and modifying it directly to suit your needs, so I'd be wary of anything encouraging folks to fork official images to make minor changes to them since the majority of those changes can be made in a dependent `Dockerfile` instead.  Combine that with automated builds (https://docs.docker.com/docker-hub/builds/) and repository links (https://docs.docker.com/docker-hub/builds/#repository-links) and it's reasonably easy to have an up-to-date image built `FROM` an official one, and not have "version drift" because the fork doesn't receive the same updates the original does.

psftw:
I tend to agree that these kinds of images are not really appropriate for the program, but in this case, if it is an important part of the Docker user documentation, then I would not have a problem with it.

moxiegirl:
Closing. Using docker repo thanks for the pointers folks.

PR #797, opened 2015-06-09
Add IBM DB2 Express-C image

wzymaster:
There is already a repo of IBM DB2 Express-C image at https://registry.hub.docker.com/u/ibmcom/db2express-c/, and now we want to have an Official Repo to help all developers with fast deployment of DB2 app box.

### DIFF
--- a/library/db2express-c
+++ b/library/db2express-c
@@ -1,0 +1,4 @@
+# maintainer: Zhong Yu (leo) Wu <leow@ca.ibm.com> (wzymaster)
+
+10.5.0.5: git://github.com/IMC3ofC/db2express-c.docker@4036c80f318a147f9e8e56fa1ba42e6fc3e5ec9f
+latest: git://github.com/IMC3ofC/db2express-c.docker@4036c80f318a147f9e8e56fa1ba42e6fc3e5ec9f
--- a/library/db2express-c
+++ b/library/db2express-c
@@ -1,5 +1,5 @@
 # maintainer: Zhong Yu (leo) Wu <leow@ca.ibm.com> (wzymaster)
 
-10.5.0.5: git://github.com/IMC3ofC/db2express-c.docker@4036c80f318a147f9e8e56fa1ba42e6fc3e5ec9f
-latest: git://github.com/IMC3ofC/db2express-c.docker@4036c80f318a147f9e8e56fa1ba42e6fc3e5ec9f
+10.5.0.5: git://github.com/IMC3ofC/db2express-c.docker@73546bc03eadbbf94f396c269c47978e07395c5a
+latest: git://github.com/IMC3ofC/db2express-c.docker@73546bc03eadbbf94f396c269c47978e07395c5a
 
--- a/library/db2express-c
+++ b/library/db2express-c
@@ -1,5 +1,5 @@
 # maintainer: Zhong Yu (leo) Wu <leow@ca.ibm.com> (wzymaster)
 
-10.5.0.5-3.10.0: git://github.com/IMC3ofC/db2express-c.docker@4036c80f318a147f9e8e56fa1ba42e6fc3e5ec9f
+10.5.0.5: git://github.com/IMC3ofC/db2express-c.docker@4036c80f318a147f9e8e56fa1ba42e6fc3e5ec9f
 latest: git://github.com/IMC3ofC/db2express-c.docker@4036c80f318a147f9e8e56fa1ba42e6fc3e5ec9f
 
--- a/library/db2express-c
+++ b/library/db2express-c
@@ -2,3 +2,4 @@
 
 10.5.0.5: git://github.com/IMC3ofC/db2express-c.docker@4036c80f318a147f9e8e56fa1ba42e6fc3e5ec9f
 latest: git://github.com/IMC3ofC/db2express-c.docker@4036c80f318a147f9e8e56fa1ba42e6fc3e5ec9f
+
--- a/library/db2express-c
+++ b/library/db2express-c
@@ -1,5 +1,5 @@
 # maintainer: Zhong Yu (leo) Wu <leow@ca.ibm.com> (wzymaster)
 
-10.5.0.5: git://github.com/IMC3ofC/db2express-c.docker@73546bc03eadbbf94f396c269c47978e07395c5a
-latest: git://github.com/IMC3ofC/db2express-c.docker@73546bc03eadbbf94f396c269c47978e07395c5a
+10.5.0.5: git://github.com/IMC3ofC/db2express-c.docker@496f6cc313b7b7dbaaca7717f02c60308289006b
+latest: git://github.com/IMC3ofC/db2express-c.docker@496f6cc313b7b7dbaaca7717f02c60308289006b
 
--- a/library/db2express-c
+++ b/library/db2express-c
@@ -1,5 +1,5 @@
 # maintainer: Zhong Yu (leo) Wu <leow@ca.ibm.com> (wzymaster)
 
-10.5.0.5: git://github.com/IMC3ofC/db2express-c.docker@4036c80f318a147f9e8e56fa1ba42e6fc3e5ec9f
+10.5.0.5-3.10.0: git://github.com/IMC3ofC/db2express-c.docker@4036c80f318a147f9e8e56fa1ba42e6fc3e5ec9f
 latest: git://github.com/IMC3ofC/db2express-c.docker@4036c80f318a147f9e8e56fa1ba42e6fc3e5ec9f
 


psftw:
``User root`` isn't actually doing anything.  The install route for epel needs improvement (maybe just ``yum install epel-release`` is needed?).  You should clear the package cache so that users are forced to ``yum update`` in their derived images.  The package list being installed seems like it may contain too many things and the final image is pretty large at 1.7G.  Are all packages justified?  I think I caught a sample database being prepared during the build - is this necessary and worth the space usage?  Should we pin some of the packages which may have version changes?  Root password being set seems odd and causes a test failure:

    $ ./test/run.sh db2express-c
    testing db2express-c
    	'utc' [1/3]...passed
    	'no-root-password' [3/3]...error: crypt password detected for root

I see potential problems with the install and entrypoint scripts, but don't want to dig too deeply now because there are a lot of basic problems as mentioned above that should be corrected.  More feedback in the docs PR https://github.com/docker-library/docs/pull/254

wzymaster:
I have updated all scripts and README files, and should I make another pull request ?

psftw:
@wzymaster I see you updated the tag name to add the kernel version, but are still referring to the same commit hash.  If you update the commit hash as well, it will be easier to review/test on my end.  I'll check it out when I have a moment, but that could be next week some time.

psftw:
BTW, I don't think we should have the kernel version in the tag name.  I think we should just call out in the docs that kernel 3.10 is the minimum required version to avoid long tag names.

wzymaster:
I have updated tag name and all docs and please check it out. @psftw thanks very much for your help here !

wzymaster:
@psftw do you have any update on this pull request ? When can we have an Official Repo of IBM DB2 Express-C image in Docker Hub ?

psftw:
@wzymaster there is a high bar to acceptance, and not addressing the vast majority of the points that I've raised here and elsewhere is not encouraging.  If you have questions about anything I've suggested, read the docs and let me know if you get stuck on something, but there is nothing new to review in 17 days as of right now (based on https://github.com/IMC3ofC/db2express-c.docker/commits/master/Dockerfile). I will have some time tomorrow to assist if I see a serious attempt to improve the Dockerfile.  Make sure to update the PR here so that it points to the newest commit each time you are ready for review.

wzymaster:
@psftw sorry, there must be some misunderstanding.  I have updated the following points : 

1) remove kernel version in tag name

2) support  daemon mode in DB2 image as MySQL does by changing entrypoint.sh 

3) switch to a different user only if user want to create a sample database

All above updates according to your comments  doesn't require a change to  https://github.com/IMC3ofC/db2express-c.docker/commits/master/Dockerfile but other files. 

https://github.com/IMC3ofC/db2express-c.docker/blob/master/README.md
https://github.com/IMC3ofC/db2express-c.docker/blob/master/docker-entrypoint.sh
https://github.com/IMC3ofC/official-images/blob/master/library/db2express-c

Can you review them again ?

psftw:
I was aware that you had addressed a few mostly cosmetic things outside of the ``Dockerfile`` when writing my last comment.  I also asked you to update the PR here so that it points to the latest code, but it is still referring to the ``4036c80...`` commit, and this is the third time I've made this request here.  Technically, that particular point doesn't prevent me from reviewing your work, but it's an important part of the image maintenance process that you would need to understand, so I'm harping on it.  My motivation to help you through the process (something I'm not obligated to do) decreases each time you ignore my feedback.

wzymaster:
@psftw  Sorry, it's my fault  to misunderstand your meaning.  Please review it .

psftw:
@wzymaster I have been running in to some issues with getting the image to run, maybe you can point me in the right direction?

On system 1 (Ubuntu 14.04, Docker 1.7.0, aufs, 2G mem):

    # uname -a
    Linux drop 3.13.0-52-generic #86-Ubuntu SMP Mon May 4 04:32:59 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux
    # docker run -d -P -e DB2INST1_PASSWORD=dE8e8odAEQJ-e8u -e LICENSE=accept --name db2orig db2:orig -d
    # docker logs db2orig
    Changing password for user db2inst1.
    passwd: all authentication tokens updated successfully.
    SQL1220N  The database manager failed to allocate shared memory.
    New password: Retype new password: 

http://www-01.ibm.com/support/knowledgecenter/SSEPGG_10.5.0/com.ibm.db2.luw.messages.sql.doc/doc/msql01022c.html -- This seems to imply that I don't have enough free memory, which is surprising to me given that there is 1.8G free on the system.

I also noticed the following when building the image on this system:

    A minor error occurred while installing "DB2 Express-C " on this computer.
    Some features may not function correctly.
    
    For more information see the DB2 installation log at
    "/tmp/db2_install_db2inst1.log".

The error appears to be related to starting the database during the install process and also not having enough memory? https://gist.github.com/psftw/36b0b08dfc604429a1f3

On system 2 (Arch Linux, Docker 1.7.0, btrfs, 24G mem), the container runs a little bit longer, and then fails with:

    $ uname -a
    Linux beast 4.0.7-2-ARCH #1 SMP PREEMPT Tue Jun 30 07:50:21 UTC 2015 x86_64 GNU/Linux
    $ docker logs db2orig
    Changing password for user db2inst1.
    passwd: all authentication tokens updated successfully.
    SQL1063N  DB2START processing was successful.
    New password: Retype new password: Redirecting to /bin/systemctl start  sshd.service
    Failed to get D-Bus connection: No connection to service manager.

This looks like the database was started up in the background and then the entrypoint script failed/exited.  I'm happy to ignore the first system and proceed with the second one for now.  I would like to get to the point where I can actually run the database and query it before helping further with the Dockerfile, etc.  Can you provide some assistance to help me get that far?  I'm not sure what I've missed.

psftw:
hmm I'm also unable to try the ibmcom/db2express-c image -- were you aware of this?

    # docker pull ibmcom/db2express-c
    Pulling repository ibmcom/db2express-c
    Repository not found

wzymaster:
we just disable it for our house keeping jobs. You can always pull it from imcloud/db2express-c @psftw Can you give me your account name of Docker Hub ?

wzymaster:
@psftw I have successfully  build this image on CentOS 6.6/7. Would you like to try that ?

psftw:
Another thing I noticed while looking around was this statement at http://www-01.ibm.com/support/docview.wss?uid=swg24038920

> DB2 fix packs can not be used with the Free version of DB2 Express-C. If you purchase the optional 12 Months Subscription license (also known as the Fixed Term License or FTL) you will be able to use DB2 Express with the FTL license and will be able to install the regular fix packs (Universal fix pack or Server fix pack). In addition to the ability to apply fix packs you also get access to IBM's world class 24/7 customer support. 

Doesn't the download from S3 labeled ``v10.5fp5_linuxx64_expc.tar.gz`` include these "fixpacks" which would imply that there may be redistribution issues?

wzymaster:
@psftw  regarding to the license/legal issue, I will check with our team member to confirm.

psftw:
Can you post a gist of running the image and verifying that it is functional?  I think it would be a weird bug in Docker if the underlying OS made a difference.

wzymaster:
@psftw  we have 28000+ downloads of ibmcom/db2express-c so far and I believe it works for the most of people on various platforms.

Please let me know what you are looking for here ? We have disabled ibmcom/db2express-c but have a private repo for it at this moment  (imcloud/db2express-c). Please give me your docker hub account and then I can add you so that you can download to try it out then.

psftw:
@wzymaster it's the same as github - psftw

Thanks

wzymaster:
your account is added and please try it out at imcloud/db2express-c.

wzymaster:
@psftw Unlike for the paid editions of DB2, IBM does not release "fixpacks" for the community edition - DB2 Express-C. However IBM does release full refreshes of DB2 Express-C at certain fixpack levels. E.g. You can get DB2 Enterprise Edition at V10.5.0 and apply fixpack 5 to get to 10.5.0.5 level but you cannot do this with DB2 Express-C. But you can download DB2 Express-C which is already at 10.5.0.5 level i.e. the same level as if FP5 was applied.

psftw:
The imcloud/db2express-c image was built with a different Dockerfile than what is here, and it actually functions in daemon mode for me.  The difference is in the entrypoint script:

    $ diff -crdNB entrypoint.sh entrypoint.sh-proposed 
    *** entrypoint.sh	2015-05-20 17:52:50.000000000 -0400
    --- entrypoint.sh-proposed	2015-07-07 13:38:45.000000000 -0400
    ***************
    *** 43,51 ****
         exit 1
      fi
      
    ! if [[ $1 = "db2start" ]]; then
        su - db2inst1 -c "db2start"
    !   nohup /usr/sbin/sshd -D 2>&1 > /dev/null &
        while true; do sleep 1000; done
      else
        exec "$1"
    --- 44,52 ----
         exit 1
      fi
      
    ! if [[ $1 = "-d" ]]; then
        su - db2inst1 -c "db2start"
    !   service sshd start
        while true; do sleep 1000; done
      else
        exec "$1"

Running ``sshd`` along side the database violates a best practice for Dockerizing an application, so we need a solution which can run the db2 database instance by itself in daemon mode.  Please verify that what you are proposing actually works.  Based on my experience today, it looks like this wasn't done, which led to some failed attempts on my end.

wzymaster:
@psftw  Yes, it's built from different dockerfile because it's originally used for testing by IBM developers.  
Yes, I can definitely remove "ssh start" part in entrypoint.sh and DB2 service has nothing to do with ssh service at all. 

I have made the changes above and all checks have passed and please try it out.

For your convenience,  those are latest build  from my Github:

https://github.com/IMC3ofC/official-images/commit/8746559eec41cb7f53ce9c1578075331eb27055f

https://github.com/IMC3ofC/db2express-c.docker/commit/496f6cc313b7b7dbaaca7717f02c60308289006b

psftw:
seeing some interesting things in what looks like a log file being generated in my running imcloud/db2express-c container: https://gist.github.com/psftw/1a1c648430776a6362f3

should I be concerned about those?

one more thing -- what's the easiest way to perform a simple query of the database so that I can verify that it is functional?

wzymaster:
@psftw you can install a DB client tool like dbvisualizer for mac to establish a connection to it after you create a sample DB in DB2. 

Here are instructions : 

1) Start a container according to the below command : 
```shell
docker run --name db2expc -d -p 50000:50000 -e DB2INST1_PASSWORD=db2inst1-pwd -e LICENSE=accept db2express-c:latest -d
```
2) Create a sample DB

```shell
docker exec -it db2expc  bash
```

```shell
$su - db2inst1
$db2sampl
```

3) Download and install dbvisualizer (https://www.dbvis.com/download/)

4) Start a new connection of DB2 and set up information as below : 

Driver: DB2
Database Server:  <ip> of Docker host
Database Port:  5000
Database: sample
Database userid:  db2inst1
Database Password: db2inst1-pwd

5) Once you have connection established, you can manipulate sample DB of DB2 then.

psftw:
Thanks, it is working now.  I'm not sure about the sleep loop to keep the process alive and wonder if there isn't a cleaner solution to this. hmm...
If you run ``docker diff`` against the running container, you'll see lots of files being added/modified in ``/home/db2inst1/sqllib`` and ``/home/db2inst1/db2inst1``.  For the sake of keeping the data in volume(s), are these the two primary locations that retain state we are concerned with? 

Can you check on the db2diag.log error messages I posted above and confirm that they can be ignored?

wzymaster:
/home/db2inst1/sqllib is the home directory for the default DB2 instance. Yes, some files will be generated dynamically when starting instance, which is expected. We don't plan to support data in volumes in the docker image at this moment.  Probably we will add that later. 

For the error in the db2diag.log, I need to check with our internal DB2 team folks to confirm.  According to my quick look at it,  it looks like an ignorable error.

tianon:
> We don't plan to support data in volumes in the docker image at this moment.

I think there must be some misunderstanding of terminology here, because I'm not sure it'd actually be possible for a database to work at all without a data volume unless it's ephemeral or stores all the data remotely (directly in S3 or something like that).

What directory is the actual data for the database itself stored in?  As an example, for MySQL this is typically `/var/lib/mysql`.

wzymaster:
In fact, /home/db2inst1 is home directory for DB2 data volume. all data are stored there actually.

psftw:
I've pushed up some WIP code on my fork which simplifies things a bit, shrinks the image by 300MB, and makes some improvements based on what I've discussed above.  Please check it out at https://github.com/psftw/db2express-c.docker and let me know what you think.

psftw:
Putting this down for the moment, but the major issues I have been struggling with are mostly around the install process.  Currently it seems to mix what are typically two steps in a database installer: installing the tools, and initializing a database.  I'm not sure if it is possible to split these up so that only the tooling is installed in the image, which would be preferred.  When the image is built, it contains an initialized database and tooling all within the ``/home/db2inst1/sqllib`` folder, so when the image is started, it copies all of this into a data volume which is overkill.  Ideally, just the actual data for the system-wide database would be generated in the volume upon the first start, but it wasn't clear to me if that is possible with how it works yet.  
Another area of concern is the dependencies which are actually required for this image to function.  I removed a lot of them and it still appears to work, which makes me question why they were included.  For example ``system-config-language`` is a GUI program that brings a pile of X11 dependencies with it and is probably never used.
Lastly, I'm not happy with the way the entrypoint works by adding a sleep loop after the ``db2start`` binary forks to the background, but don't have a better solution to drop in place.

wzymaster:
@psftw In fact,  so far we are not considering to optimize this image for now , but have it available for try by general public.  To address two of your concerns here : 

1)  The default instance is initialized and created during DB2 installation by default, which is from DB2 design.  In fact, all requires binaries, tools of DB2 instance haven been  copied during initialization process, and I don't know why you think it's done during starting image. 

2) some additional package like system-config-language is originally from the base official Docker image of CentOS 6/7, which we never touch it. So far, we would like to go with the defaults and consider optimization later.

Please let me know if you have any other concerns with running this image. The focus here is to try to speed up the process of Official Repo since it has been months since we submitted.  We are looking for minimal viable product of DB2 Express-C image and will optimize it later.

psftw:
An official image for ``db2express-c`` would need to meet certain quality standards before being accepted.  The same standards apply to every proposal.  I've resolved most of the smaller issues in my fork, but there are still a couple of major blockers to work through in my mind.  Your latest response is dismissive and does not address them, which is not helping your cause.

tianon:
I agree with @psftw; if the goal is simply publication, then having this
image publicly available in an "IBM" organization on the Hub is sufficient
to accomplish that while the quality is improved, and I wholeheartedly
encourage that even after the image is accepted in the official images
program.

The official images are intended to be a source of high quality software
Dockerizations.  Once the public image is ready to be improved in the
specific ways we require, official status can and should be pursued.  If
the image as it stands today is not going to be modified based on our
review feedback, then I would respectfully request that we close this
proposal until it is.

toli:
@wzymaster, you have definitely put in a lot of work into creating a Dockerized container for IBM DB2-Express, and we appreciate that. 

@psftw made a number of what seems like high-quality suggestions in his fork. 
He's cleaned up the `yum install` line to be slimmer to reduce footprint, he has "unrolled" the `install_db2.sh` installer file to be more Docker-friendly and transparent, and, most importantly, he's created a `volume` so that the DB data doesn't get lost after the container shuts down.

Not having an automated way to test/verify that the image works correctly w/out using a separate GUI makes it harder as well. Is there no command-line to be used so that we can include that as part of the automated testing framework as part of the build process?

Would it be possible for you to incorporate that feedback? We are very interested in having DB2 as part of official images, but the image does have to pass a certain threshold in order to be listed, so that our mutual end-users have a good experience.

Please let us know if you can accommodate the changes, and we'd be happy to review them.

wzymaster:
@psftw @tianon @toli 
 First of all, I believe there must be some misunderstanding.  I never expect to publish a low quality product on Docker Hub for DB2 Express-C.   We definitely are expecting for the best quality of Docker image here for DB2 Express-C.

Regarding to 1),  Once instance is created, all tool will be placed under ```/home/db2inst1/sqllib```. After a new sample DB is created, raw db data home directory will go under account home directory, like '/home/db2inst1' by default, so you will see '/home/db2inst1/db2inst1'.  However, we can't put it under '/var/lib/db2' because db2 never requires instance owner has permission to do that,  and we would like everything is self-contained in owners home directory, like '/home/db2inst'.  So, again, we can't put those data volume to "/var/lib" like other DB products.

I believe you probably don't get better understanding how DB2 instance works. It's from design of how DB2 instance is managed by owner by default,  and don't know why you think it's a problem with quality. You can't expect DB2 to behave the similar to MySQL, Postgres or other database products, because each product has its own design, trait, which has both negative and positive sides.  

Regarding to 2),  I  definitely would like to incorporate for any changes that you guys want, like remove system-config-language packages from docker image to reduce the size of image for sure.

toli:
@wzymaster - glad to hear you are committed to creating a quality DB2 image. That's excellent. 
I think there's some confusion about `volume`. I'm speaking of using a `Docker Volume` command (see more info at https://docs.docker.com/reference/builder/, search for VOLUME)
When you run your DB2 image now, all the data will be created and stored somewhere in `/home/db2inst1` as you mentioned. However, when the container stops and is removed, the next time you run another DB2 container, all of that data will be lost. 
Think of it as `objects` and `classes`. When an object is created, you can modify it, but when the object disappears and another one is instantiated, it's just "vanilla" and doesn't have those changes. 

Same with containers. Unless you persist the data with `docker save` (which creates an entirely new image), the "new data" disappears.

Hence the suggestion to create a `VOLUME`. You designate some directory as a volume with `VOLUME` command, and then you link it with some external persistent storage at startup:
`docker run -it -v /some/storage/db2:/home/db2inst1` (or whatever the mounted directory should be)

Does that make more sense?

Look forward to seeing you incorporate the other feedback as well.

wzymaster:
@toli  thanks for your prompt response !  Yes, I understand what you mean here for Volume.
In fact, DB2 users can mount an external storage as volume anywhere while starting an image, and then they update DB2 configuration pointing to that volume folder.  We don't want to restrict it to /home/db2inst1 or somewhere in Docker image since it is definitely optional for DB2. 

Do we have to pre-define a data volume directory in DB2 Docker image so that user can designate that directory while starting Docker image ?

toli:
@wzymaster - your explanation makes sense. 
However, i'm trying to have your image `work out of the box`. If someone downloads the image and runs it without making changes, the data will be saved to `/home/db2inst1`, won't it? 
So i'm just recommending that you make it work out of the box. 

If you expect people to make configuration changes before they start using the system, then you should specify so in the accompanying README.md file that explains how the image is to be used.

yosifkit:
It is not necessary to define the `VOLUME` in the Dockerfile for them to be able to mount a host folder on run using `-v /local/:/home/db2inst1`, but it is helpful to let users know where to mount without having to go to the readme that will be on the hub.  The `VOLUME` will also allow a user to migrate one instance to another using `--volumes-from old-db2` to reuse the data directory, even when the first container is run without a specific bind mount, since docker uses an automatic host mount on all `VOLUME`s.

bsideup:
@wzymaster @psftw any progress here?

wzymaster:
@bsideup In fact, I have done with the required changes in the latest build and wonder if it's qualified to promote it to official image repo ? 

Please feel free to let me know if there is any further change is needed.

psftw:
Closing due to inactivity, and:

* The binary from S3 (https://s3.amazonaws.com/db2-expc105-64bit-centos/v10.5fp5_linuxx64_expc.tar.gz) appears to be inaccessible.
* The ``ibmcom/db2express-c`` repository on DockerHub appears to be abandoned, despite interest from users and thousands of downloads.
* My access to the private IBM repository (``imcloud/db2express-c``) has also been revoked, so I could only speculate if development has continued there.

I would be happy to work with folks at IBM who have a deep understanding of DB2 Express and are motivated to finish the work that was started here.  I think this is a good candidate for an official image.

PR #802, opened 2015-06-10
Bump jruby 9k to rc1

cpuguy83:
Signed-off-by: Brian Goff <cpuguy83@gmail.com>

PR #812, opened 2015-06-12
Add "BUNDLED WITH" to Gemfile.lock

tianon:
https://github.com/bundler/bundler/pull/3546


tianon:
See also https://github.com/docker-library/official-images/pull/802#issuecomment-110926558.

yosifkit:
LGTM

PR #827, opened 2015-06-16
Update ruby's bundler to 1.10.4

tianon:

PR #831, opened 2015-06-16
Remove Gemfile.lock (it is far too volatile for VCS now)

tianon:


tianon:
See also https://github.com/docker-library/official-images/pull/827#issuecomment-112575038 (and any of the `library/jruby` PRs, which are notorious for breaking on `Gemfile.lock` wanting to be changed by bundler).

yosifkit:
LGTM

PR #848, opened 2015-06-24
Aerospike 3.5.14

volmarl:
New Aerospike-tools container and update on main Aerospike server container.

### DIFF
--- a/library/aerospike
+++ b/library/aerospike
@@ -1,4 +1,4 @@
 # maintainer: Lucien Volmar <lucien@aerospike.com> (@volmarl)
 
-3.5.12: git://github.com/aerospike/aerospike-server.docker@b92075d0b0280b38ac7e2ea11dfe49c3903b83aa
-latest: git://github.com/aerospike/aerospike-server.docker@b92075d0b0280b38ac7e2ea11dfe49c3903b83aa
+3.5.14: git://github.com/aerospike/aerospike-server.docker@6595a46a6fe2d190191985b62ffccac53300dd53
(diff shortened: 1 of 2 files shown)


yosifkit:
I am slightly confused by the new tools image.  I would imagine that the tools would be part of the server image since that would allow a user to `docker exec` to use the tools while also allowing them to run the tools in a new container without downloading another docker image.  Would the installed size of the tools be detrimental to users?

volmarl:
@yosifkit We debated about adding both tools and server on same container. It seems there is a large number of users who would need to use the tools on a client and not require the full server binary. This container was created to give the flexibility to an aerospike devops engineer to run this container and manage either aerospike running on docker or managing any aerospike cluster running on baremetal or a cloud service. We can definitely add the the tools also to the server container and take advantage of docker exec. I having the tools in both an independent container and part of the servers will offer more flexibility to our users.

yosifkit:
I recommend adding the tools to the server version, and then using the same Dockerfile minus the server install to make the tools image (since you'll need the same sha256 checking for it as well).

I would be reluctant to add a new image repository for just the tools and would rather it was a tag in the current `aerospike` repo (ie, `aerospike:tools` and `aerospike:3.5.14-tools`).  It is fine if they stay in separate git repositories.  You'll definitely want to update the `content.md` in the [docker-library/docs](https://github.com/docker-library/docs) repo to explain the new tags.

volmarl:
Thanks for the recommendation. I will cancel this pull request. We will go ahead and add the tools binary to the server's image and give instructions on using docker exec in the readme to use the tools.

PR #859, opened 2015-07-01
Disregard -- testing clean clone size in Travis

tianon:


tianon:
So for future reference, our current "fresh" `src` size is ~1.8GB.

PR #920, opened 2015-07-23
Add a test for new postgres initdb logic

tianon:
A test to verify the functionality implemented in https://github.com/docker-library/postgres/pull/75.


yosifkit:
LGTM

PR #929, opened 2015-07-25
Add image for Joomla!

mbabker:
This PR adds a Docker image for the Joomla! CMS

Docs PR: https://github.com/docker-library/docs/pull/293

### DIFF
--- a/library/joomla
+++ b/library/joomla
@@ -1,15 +1,15 @@
 # maintainer: Michael Babker <michael.babker@joomla.org> (@mbabker)
 
-3.4.3-apache: git://github.com/joomla/docker-joomla@966275ada2148e343a68c8c03870f11cc7f5b89c apache
-3.4.3: git://github.com/joomla/docker-joomla@966275ada2148e343a68c8c03870f11cc7f5b89c apache
-3.4-apache: git://github.com/joomla/docker-joomla@966275ada2148e343a68c8c03870f11cc7f5b89c apache
-3.4: git://github.com/joomla/docker-joomla@966275ada2148e343a68c8c03870f11cc7f5b89c apache
-3-apache: git://github.com/joomla/docker-joomla@966275ada2148e343a68c8c03870f11cc7f5b89c apache
-apache: git://github.com/joomla/docker-joomla@966275ada2148e343a68c8c03870f11cc7f5b89c apache
-3: git://github.com/joomla/docker-joomla@966275ada2148e343a68c8c03870f11cc7f5b89c apache
-latest: git://github.com/joomla/docker-joomla@966275ada2148e343a68c8c03870f11cc7f5b89c apache
+3.4.3-apache: git://github.com/joomla/docker-joomla@ba5df1da80e0df1ec30ba93c6a369b9cacdc57fe apache
+3.4.3: git://github.com/joomla/docker-joomla@ba5df1da80e0df1ec30ba93c6a369b9cacdc57fe apache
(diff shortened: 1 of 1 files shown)


psftw:
The ``apache`` tag runs well, but I seem to be having trouble getting the ``fpm`` tag working.  I'm seeing connection reset errors when attempting to hit ``container-ip:9000`` in my browser.  Am I doing something wrong?

```
docker run --name mysql3 -e MYSQL_ROOT_PASSWORD=my-secret-pw -d mysql:5.6
docker run --name joomla3 --link mysql3:mysql -d joomla:fpm
```

mbabker:
Shouldn't be.  The configs are all based on the WordPress image since that
so closely matches Joomla's environment, I'll have to re-check it and see
if I missed something.

md5:
@psftw Browsers don't speak the FastCGI protocol, so you won't be able to test the image that way.

[See here](https://github.com/docker-library/official-images/blob/master/test/tests/php-fpm-hello-web/run.sh) for a test I wrote that's able to speak FastCGI directly to port 9000 for testing.

Another option would be to set up an appropriate Nginx config to front the FPM image. See this gist for an example: https://gist.github.com/md5/d9206eacb5a0ff5d6be0

yosifkit:
I only have one suggestion, since the htaccess file comes with in the zip, why not just a quick sed to just create the file with the single change?  That way you won't need to update or even track the file in the docker-joomla repo.

```bash
if [ ! -e .htaccess ]; then
    sed -r 's/^(Options -Indexes.*)$/#\1/' htaccess.txt > .htaccess
    chown www-data:www-data .htaccess
fi
```

And I have one question, is `JOOMLA_UPSTREAM_VERSION` needed? We have it in the wordpress image because we are slightly lazy; wordpress releases 4.0, but we treat it as 4.0.0 since 4.0.1 would also be 4.0.  And now @tianon is fixing it in [docker-library/wordpress#91](https://github.com/docker-library/wordpress/pull/91/files?w=1).

mbabker:
`sed` is something I'm not strong in so the way I did it works, but if that's quicker/easier I'll definitely push it in.

All of Joomla's version numbers use the major.minor.patch format always, so I can apply that change too.

yosifkit:
That was fast :metal:. LGTM.

Anything else @tianon?

yosifkit:
Build test of #929; 3cc5168da859df45eab761e0cc42a8997a356e4b (`joomla`):

```console
$ url="https://raw.githubusercontent.com/docker-library/official-images/3cc5168da859df45eab761e0cc42a8997a356e4b/library/joomla"
$ bashbrew build "$url"
Cloning joomla (git://github.com/joomla/docker-joomla) ...
Processing joomla:3.4.3-apache ...
Processing joomla:3.4.3 ...
Processing joomla:3.4-apache ...
Processing joomla:3.4 ...
Processing joomla:3-apache ...
Processing joomla:apache ...
Processing joomla:3 ...
Processing joomla:latest ...
Processing joomla:3.4.3-fpm ...
Processing joomla:3.4-fpm ...
Processing joomla:3-fpm ...
Processing joomla:fpm ...
testing joomla:3.4.3-apache
	'utc' [1/3]...passed
	'no-hard-coded-passwords' [3/3]...passed
testing joomla:3.4.3-fpm
	'utc' [1/3]...passed
	'no-hard-coded-passwords' [3/3]...passed
```

tianon:
Nice, definitely a +1 from me for using `sed` instead -- that'll make sure future updates to the `htaccess` upstream automatically make their way into the image. :+1:

LGTM

PR #968, opened 2015-08-17
Add Couchbase Server 3.1.0

ceejatec:
Adds Couchbase Server 3.1.0. Also incorporates some Dockerfile cleanups from myself and some more significant changes from @hookenz (switching to runit to launch PID 1, basing off Ubuntu official image rather than Centos).

### DIFF
--- a/library/couchbase
+++ b/library/couchbase
@@ -1,21 +1,21 @@
 # maintainer: Couchbase Docker Team <docker@couchbase.com> (@couchbase)
 
-latest: git://github.com/couchbase/docker@906b651a6f374c70832dd440375a821b1ec890fa enterprise/couchbase-server/3.1.0
-enterprise: git://github.com/couchbase/docker@906b651a6f374c70832dd440375a821b1ec890fa enterprise/couchbase-server/3.1.0
+latest: git://github.com/couchbase/docker@dcf715fcfdc88225700ee76e437d1d4522ae150c enterprise/couchbase-server/3.1.0
+enterprise: git://github.com/couchbase/docker@dcf715fcfdc88225700ee76e437d1d4522ae150c enterprise/couchbase-server/3.1.0
 
-3.1.0: git://github.com/couchbase/docker@906b651a6f374c70832dd440375a821b1ec890fa enterprise/couchbase-server/3.1.0
-enterprise-3.1.0: git://github.com/couchbase/docker@906b651a6f374c70832dd440375a821b1ec890fa enterprise/couchbase-server/3.1.0
+3.1.0: git://github.com/couchbase/docker@dcf715fcfdc88225700ee76e437d1d4522ae150c enterprise/couchbase-server/3.1.0
+enterprise-3.1.0: git://github.com/couchbase/docker@dcf715fcfdc88225700ee76e437d1d4522ae150c enterprise/couchbase-server/3.1.0
 
-3.0.3: git://github.com/couchbase/docker@906b651a6f374c70832dd440375a821b1ec890fa enterprise/couchbase-server/3.0.3
-enterprise-3.0.3: git://github.com/couchbase/docker@906b651a6f374c70832dd440375a821b1ec890fa enterprise/couchbase-server/3.0.3
+3.0.3: git://github.com/couchbase/docker@dcf715fcfdc88225700ee76e437d1d4522ae150c enterprise/couchbase-server/3.0.3
+enterprise-3.0.3: git://github.com/couchbase/docker@dcf715fcfdc88225700ee76e437d1d4522ae150c enterprise/couchbase-server/3.0.3
 
-3.0.2: git://github.com/couchbase/docker@906b651a6f374c70832dd440375a821b1ec890fa enterprise/couchbase-server/3.0.2
-enterprise-3.0.2: git://github.com/couchbase/docker@906b651a6f374c70832dd440375a821b1ec890fa enterprise/couchbase-server/3.0.2
+3.0.2: git://github.com/couchbase/docker@dcf715fcfdc88225700ee76e437d1d4522ae150c enterprise/couchbase-server/3.0.2
+enterprise-3.0.2: git://github.com/couchbase/docker@dcf715fcfdc88225700ee76e437d1d4522ae150c enterprise/couchbase-server/3.0.2
 
-community: git://github.com/couchbase/docker@906b651a6f374c70832dd440375a821b1ec890fa community/couchbase-server/3.0.1
-community-3.0.1: git://github.com/couchbase/docker@906b651a6f374c70832dd440375a821b1ec890fa community/couchbase-server/3.0.1
+community: git://github.com/couchbase/docker@dcf715fcfdc88225700ee76e437d1d4522ae150c community/couchbase-server/3.0.1
+community-3.0.1: git://github.com/couchbase/docker@dcf715fcfdc88225700ee76e437d1d4522ae150c community/couchbase-server/3.0.1
 
-2.5.2: git://github.com/couchbase/docker@906b651a6f374c70832dd440375a821b1ec890fa enterprise/couchbase-server/2.5.2
-enterprise-2.5.2: git://github.com/couchbase/docker@906b651a6f374c70832dd440375a821b1ec890fa enterprise/couchbase-server/2.5.2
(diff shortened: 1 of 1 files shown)


tianon:
Changeset: https://github.com/couchbase/docker/compare/695e9dbe4f8928ec4b66a83568d3d583c8215b1e...906b651a6f374c70832dd440375a821b1ec890fa#diff-8941c93c7af57eff48058be15eb5ecff

What's the rationale for the switch to `runit`?  Signal forwarding and zombie reaping are usually the only problems we see with software being run as PID1, and both of those are solved with something much slimmer and simpler than `runit` like [`tini`](https://github.com/krallin/tini), although it's worth noting that zombie reaping is even something being worked on in Docker itself (https://github.com/docker/docker/issues/11529).

ceejatec:
You can see @hookenz original report on the pull request here: https://github.com/couchbase/docker/pull/8

I'm not sure whether signal handling, zombie reaping, or some other issue was behind the problems he had with the original image, but his fairly extensive use of the container was greatly improved by this change. I didn't argue with the choice of tools to resolve the problem. Is there a particular reason that runit is "bad", other than some extra KB in the image?

hookenz:
@tianon - Because I'd never heard of tini before.  runit is not very big.  It's written in C and works well.  It also has the benefit that if you do nothing with application output, it automatically goes to stdout which is nice and docker friendly.  So if someone wants to switch on couchbase stdout/stderr logging you'll see it.

I know zombie reaping issues were being worked on in docker itself, but this is still not available and takes time to work it's way into CoreOS where I was using this.

yosifkit:
@hookenz I think the rationale behind using `tini` is the limited feature set. `tini` only does reaping and signal forwarding. While with runit you can also do multiple processes (which is tempting to do once the tool is there) and restart on failure, which could gloss over possible problems.

So is runit to solve the zombie problem, or is there an issue with couchbase needing multiple processes, or just something to restart it on failure?

hookenz:
@yosifkit - to be honest I added it on a hunch it was related to PID1.  I couldn't get couchbase server to run passed a few days and often even minutes on Amazon on CoreOS.  It usually failed immediately when adding a second server to the cluster during the rebalance phase.  I started up a lot of servers and did a lot of trial and error testing over several months and this problem occurred 100% of the time.  I was using CoreOS stable with Docker 1.6.  I never tried it with Docker on any other Host OS.

Without a full internal understanding of the Couchbase code and the meaning of all the debug I never quite fully understood what was happening.  I was seeing symptoms of services being started and then not functioning correctly and CPU spiking (due to beam.smp trying to reconnect in a tight loop via tcp to another Couchbase service on the same host and that service not responding for some reason). 

So as I said, on a hunch of what's different in the container environment and a full machine (PID1) I added runit.  The problems completely disappeared and we've been running this for several months now in a production environment.

ceejatec:
Does anything else need to be done to get this merged?

yosifkit:
Putting couchbase into `runit` feels like it glosses over the problem without finding out what the problem is. It also obscures how the container is run and could encourage users to break the "one concern per container". \</rant>

Ok, so a few questions and suggestions:

- drop the `RUN chmod 755` lines and just commit the chmod in git
 - some docker storage drivers have issues with the chmod on a separate layer
- download the deb via https, but also recommend using a gpg signature or at least a shasum ([security section of official-images readme](https://github.com/docker-library/official-images/#security))
- How does logging work now?  There used to be stuff on stdout/stderr that could be seen with `docker logs`.
 - ie, does it need `svlog` in the container too and can it log to stdout/stderr of the container to keep consistent with the rest of the official images?
- if I `docker run -it couchbase`, it no longer responds to `ctrl-c` and has to be stopped with `docker stop`
 - not sure if it is necessary, but just an "fyi"

hookenz:
@yosifkit - Thanks for the feedback.

If you or someone at Couchbase labs are able to spare the time to fix Couchbase so that it can run correctly as PID 1 and run for a month or two on CoreOS with docker 1.6 and then I think that we should probably switch back.  But I think that's a wasted effort because it would appear this relatively common issue may end up being fixed in when docker 1.9 comes out.

What we have right now with runit has been totally stable in production for me.

Regarding the one concern per container comment.  It still is really the case.  Yes people might be tempted to add other stuff, but I kind of doubt that.  The most important thing is that we have a container for Couchbase that is stable.  What it does on the inside I think most won't care.  I sure would not have had it worked for me before I made the switch to runit.

As for logging in runit, nothing needs to be done there.  runit will pass sub-process stdout on  by default so it should appear as normal.  The svlog stuff is only if you want to change that behaviour or create log files.

As for the other things you've suggested, they are all good suggestions.  I'll let @ceejatec decide whether he wants to act on those.

ceejatec:
Thanks for the thoughts, @hookenz.

@yosifkit - I have pushed a change which addresses most of your other suggestions:

1. Dropped chmod, checked in the files with correct permissions.
2. I can't seem to download with https: as our files are on S3 (not sure if we have a Cloudfront URL which would be the only way), but I did upload sha256sums for all installers and embed the sha256 checksum into all Dockerfiles as suggested.
3. Previously the only logging was a brief message pointing the user to localhost:8091 for the Couchbase console. I have restored that log messsage.
4. I'm not personally concerned about Ctrl-C not working as foregrounding the container is not the normal use case.

Hope you like it better now!

hookenz:
:+1

yosifkit:
Ok, newest changes look fine to me: https://github.com/couchbase/docker/compare/906b651a6f374c70832...dcf715fcfdc8822

Full changeset: https://github.com/couchbase/docker/compare/695e9dbe4f8928ec4b6...dcf715fcfdc8822

LGTM, Build test of #968; 9d0473453fa4bb5412677df76789e791006ec26c (`couchbase`):

```console
$ url="https://raw.githubusercontent.com/docker-library/official-images/9d0473453fa4bb5412677df76789e791006ec26c/library/couchbase"
$ bashbrew build "$url"
Fetching couchbase (git://github.com/couchbase/docker) ...
Processing couchbase:latest ...
Processing couchbase:enterprise ...
Processing couchbase:3.1.0 ...
Processing couchbase:enterprise-3.1.0 ...
Processing couchbase:3.0.3 ...
Processing couchbase:enterprise-3.0.3 ...
Processing couchbase:3.0.2 ...
Processing couchbase:enterprise-3.0.2 ...
Processing couchbase:community ...
Processing couchbase:community-3.0.1 ...
Processing couchbase:2.5.2 ...
Processing couchbase:enterprise-2.5.2 ...
Processing couchbase:community-2.2.0 ...
testing couchbase:latest
	'utc' [1/4]...passed
	'override-cmd' [4/4]...passed
testing couchbase:3.0.3
	'utc' [1/4]...passed
	'override-cmd' [4/4]...passed
testing couchbase:3.0.2
	'utc' [1/4]...passed
	'override-cmd' [4/4]...passed
testing couchbase:community
	'utc' [1/4]...passed
	'override-cmd' [4/4]...passed
testing couchbase:2.5.2
	'utc' [1/4]...passed
	'override-cmd' [4/4]...passed
testing couchbase:community-2.2.0
	'utc' [1/4]...passed
	'override-cmd' [4/4]...passed
```

tianon:
Ok, fair enough.  :+1: LGTM

ceejatec:
Thanks everyone!

hookenz:
Excellent.  Thanks guys, this is working great! I've switched my production clusters to this updated version now.

Here is what my systemd service files for CoreOS.

```
[Unit]
Description=Couchbase Server
After=docker.service
Requires=docker.service
RequiresMountsFor=/var/lib/couchbase

[Service]
TimeoutStartSec=0
Restart=on-failure
RestartSec=60
ExecStartPre=-/usr/bin/docker kill couchbase
ExecStartPre=-/usr/bin/docker rm couchbase
ExecStartPre=/usr/bin/docker pull couchbase/server:community-3.0.1
ExecStart=/usr/bin/docker run --name couchbase \
  --net host \
  --ulimit nofile=40960:40960 \
  --ulimit core=100000000:100000000 \
  --ulimit memlock=100000000:100000000 \
  -v /var/lib/couchbase:/opt/couchbase/var/lib/couchbase \
  couchbase/server:community-3.0.1
ExecStop=/usr/bin/docker stop couchbase

[Install]
WantedBy=multi-user.target
```

PR #993, opened 2015-08-24
Add basic docker-dind test

tianon:


test/tests/docker-dind/run.sh
```diff
@@ -0,0 +1,45 @@
...
+		false
+	fi
+	sleep 2
+done
+
+docker_ pull busybox
+
+docker_ run --rm busybox true
+
+docker_ create -i --name test busybox cat
```
md5: The name `test` seems a bit collision prone.

tianon: This is inside our container, so collision here should actually be
impossible.

md5: :+1:

yosifkit:
LGTM

PR #1001, opened 2015-08-27
Add override-cmd test

md5:
Ensures that the CMD for an image can be overridden

### DIFF
--- a/test/tests/override-cmd/run.sh
+++ b/test/tests/override-cmd/run.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+set -eo pipefail
+
+image="$1"
+
+# Test that we can override the CMD with echo
+hello="world-$RANDOM-$RANDOM"
+output="$(docker run --rm "$image" echo "Hello $hello")"
+[ "$output" = "Hello $hello" ]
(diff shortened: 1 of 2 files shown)


md5:
A couple points:

1. I made this a global test with exclusions for the single-binary images since I believe it's a requirement (or at least a strong suggestion) that users be able to run other binaries that are in the image as their `CMD`. If there are other exclusions, I can add them, but I don't have all the image pulled locally to test.
2. I'm assuming that `echo` is available in every "normal" image. That should be a safe assumption, I believe.

yosifkit:
LGTM

test/tests/override-cmd/run.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -eo pipefail
+
+image="$1"
+
+# Test that we can override the CMD with echo
+hello="world-$RANDOM-$RANDOM"
+output="$(docker run --rm "$image" echo "Hello $hello")"
```
md5: Now that I look at this again, we may be leaving dangling volumes for some images since `docker run --rm` won't remove those. I guess we need a separate `docker rm -v`.

I don't have time to update this right now, but I can do so later today.

md5: Actually, I'm wrong. It does remove volumes: https://github.com/docker/docker/pull/15718/files

tianon:
Nice, LGTM

md5:
Now let's wait for the complaining to start when some image can't actually run `echo` :metal:

PR #1002, opened 2015-08-27
Add jetty-hello-web test

md5:


md5:
This test is using the [`appropriate/curl`](https://hub.docker.com/r/appropriate/curl/) image to run `curl` (~8MB, based on `alpine`). Let me know if you're rather I build a `curl` client image as part of the test.

test/tests/jetty-hello-web/run.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+
+set -eo pipefail
+
+dir="$(dirname "$(readlink -f "$BASH_SOURCE")")"
+
+image="$1"
+
+# Pull a client image with curl for testing
+clientImage='appropriate/curl'
```
yosifkit: Why not just `buildpack-deps:jessie-curl`, it is what `java:jre` images are based on, so it will already exist :wink:.

md5: Good point. I'll update to that.

md5:
Updated to use `buildpack-deps:jessie-curl` instead of `appropriate/curl`.

Thanks @yosifkit!

md5:
I suppose I should remove the `docker pull` as well. Let me do that.

md5:
Updated again to only pull `buildpack-deps:jessie-curl` when it isn't there.

Despite the fact that `jetty` is based on the `java:jre` images and those are based on `buildpack-deps:jessie-curl`, it could still happen that the tag isn't pulled while the image is actually there.

In fact, I could probably just use the `jetty` image itself since it already has `curl`...

md5:
Updated again to just use the image under test to run `curl`.

md5:
Let me know if you'd like me to squash the commits.

yosifkit:
LGTM

tianon:
LGTM

md5:
:+1:

PR #1004, opened 2015-08-27
Specify "curl" command explicitly in jetty-hello-web

md5:
I realized after #1002 was merged that using `jetty` as the client test image will only work if I specify `curl` as the `CMD` :confounded:


tianon:
lol oops, LGTM

yosifkit:
LGTM

PR #1011, opened 2015-08-29
update clojure commit to Lein 2.5.2

Quantisan:


tianon:
https://github.com/Quantisan/docker-clojure/compare/bb413b92f25c48ac48db93ec236459765be8a618...868a7ba998a826eddadc62dbcb0942885cb520fa

Everything there looks pretty good except the switch from a full fingerprint to just a short ID for `gpg --recv-key` (line 26; for more on why this isn't great, see http://www.asheesh.org/note/debian/short-key-ids-are-bad-news.html).  For reference, the full fingerprint of that short key appears to be `2E708FB2FCECA07FF8184E275A92E04305696D78` (`Jean Niklas L'orange (hyPiRion) <jeannikl@hypirion.com>`).

Quantisan:
@tianon thanks for pointing that out. The commit now points to a full GPG key.

tianon:
LGTM :+1:

PR #1015, opened 2015-09-01
Support user-specified test config scripts

md5:
This makes it easier to use tests in this repository for images other than official ones.

### DIFF
--- a/test/run.sh
+++ b/test/run.sh
@@ -105,9 +105,8 @@ for dockerImage in "$@"; do
 		continue
 	fi
 	
-	noNamespace="${dockerImage##*/}"
-	repo="${noNamespace%:*}"
-	tagVar="${noNamespace#*:}"
+	repo="${dockerImage%:*}"
+	tagVar="${dockerImage#*:}"
 	#version="${tagVar%-*}"
 	variant="${tagVar##*-}"
 	
--- a/test/run.sh
+++ b/test/run.sh
@@ -32,7 +32,7 @@ while true; do
 		--dry-run) dryRun=1 ;;
 		--help|-h|'-?') usage && exit 0 ;;
 		--test|-t) argTests["$1"]=1 && shift ;;
-		--config|-c) additionalConfigs+=("$1") && shift ;;
+		--config|-c) additionalConfigs+=("$(readlink -f "$1")") && shift ;;
 		--) break ;;
 		*)
 			{
@@ -55,7 +55,7 @@ declare -A testAlias=()
 declare -A imageTests=()
 declare -A globalExcludeTests=()
 
-# load default config
+# load the default config
 . "$dir/config.sh"
 
 # load additional user-specified configs
--- a/test/run.sh
+++ b/test/run.sh
@@ -23,7 +23,7 @@ opts="$(getopt -o 'ht:c:?' --long 'dry-run,help,test:,config:' -- "$@" || { usag
 eval set -- "$opts"
 
 declare -A argTests=()
-declare -a additionalConfigs=()
+declare -a configs=()
 dryRun=
 while true; do
 	flag=$1
@@ -32,7 +32,7 @@ while true; do
 		--dry-run) dryRun=1 ;;
 		--help|-h|'-?') usage && exit 0 ;;
 		--test|-t) argTests["$1"]=1 && shift ;;
-		--config|-c) additionalConfigs+=("$(readlink -f "$1")") && shift ;;
+		--config|-c) configs+=("$(readlink -f "$1")") && shift ;;
 		--) break ;;
 		*)
 			{
@@ -55,12 +55,44 @@ declare -A testAlias=()
 declare -A imageTests=()
 declare -A globalExcludeTests=()
 
-# load the default config
-. "$dir/config.sh"
+# if there are no user-specified configs, use the default config
+if [ ${#configs} -eq 0 ]; then
+	configs+=("$dir/config.sh")
+fi
 
-# load additional user-specified configs
-for conf in "${additionalConfigs[@]}"; do
+# load the configs
+declare -A testPaths=()
+for conf in "${configs[@]}"; do
 	. "$conf"
+
+	# Determine the full path to any newly-declared tests
+	confDir="$(dirname "$conf")"
+
+	for testName in ${globalTests[@]}; do
+		[ "${testPaths[$testName]}" ] && continue
+
+		if [ -d "$confDir/tests/$testName" ]; then
+			# Test directory found relative to the conf file
+			testPaths[$testName]="$confDir/tests/$testName"
+		elif [ -d "$dir/tests/$testName" ]; then
+			# Test directory found in the main tests/ directory
+			testPaths[$testName]="$dir/tests/$testName"
+		fi
+	done
+
+	for image in ${!imageTests[@]}; do
+		for testName in ${imageTests[$image]}; do
+			[ "${testPaths[$testName]}" ] && continue
+
+			if [ -d "$confDir/tests/$testName" ]; then
+				# Test directory found relative to the conf file
+				testPaths[$testName]="$confDir/tests/$testName"
+			elif [ -d "$dir/tests/$testName" ]; then
+				# Test directory found in the main tests/ directory
+				testPaths[$testName]="$dir/tests/$testName"
+			fi
+		done
+	done
 done
 
 didFail=
@@ -110,7 +142,7 @@ for dockerImage in "$@"; do
 		
 		# run test against dockerImage here
 		# find the script for the test
-		scriptDir="$dir/tests/$t"
+		scriptDir="${testPaths[$t]}"
 		if [ -d "$scriptDir" ]; then
 			script="$scriptDir/run.sh"
 			if [ -x "$script" -a ! -d "$script" ]; then
--- a/test/run.sh
+++ b/test/run.sh
@@ -68,7 +68,7 @@ for conf in "${configs[@]}"; do
 	# Determine the full path to any newly-declared tests
 	confDir="$(dirname "$conf")"
 
-	for testName in ${globalTests[@]}; do
+	for testName in ${globalTests[@]} ${imageTests[@]}; do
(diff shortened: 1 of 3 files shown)


md5:
I just whipped this up in a few minutes out of a desire to use the `postgres` tests for my PostGIS image.

If y'all are on board with this change, I can update the `usage` function and add docs to the `test/README.md`.

md5:
The way I envisioned using this was to have an additional config file like this (named, say `postgis-config.sh` in the `test` directory of my repo):

```bash
testAlias[appropriate/postgis]=postgres

imageTests[appropriate/postgis]='
	postgis-some-test
	postgis-other-test
'
```

And running the tests like this:

```console
$ ~/official-images/test/run.sh -c ./test/postgis-config.sh appropriate/postgis:9.4-2.1
```

One sticking point is that the script is stripping the image namespace from the image name here: https://github.com/docker-library/official-images/blob/master/test/run.sh#L68

I could build my image as just `postgis` in my test setup, but that seems wrong. I'd rather build it as `appropriate/postgis` (or `mdillon/postgis` as it's called on Docker Hub since I can't rename/redirect the image without breaking things for existing users).

md5:
Here's an example of a successful test run with these changes: https://travis-ci.org/appropriate/docker-postgis/jobs/78168336

md5:
The `imageTests` usage is going to take some more thought to get it working right, but `testAlias` is working fine.

test/run.sh
```diff
@@ -47,14 +49,20 @@ if [ $# -eq 0 ]; then
...
-#   imageTests
-#   globalExcludeTests
+# declare configuration variables
+declare -a globalTests=()
+declare -A testAlias=()
+declare -A imageTests=()
+declare -A globalExcludeTests=()
+
+# load the default config
 . "$dir/config.sh"
```
tianon: What do you think about instead having the `--config` flag fill up the array, and if it's empty here, we put `"$dir/config.sh"` into it so that you're not forced to use our default config, but get to add it yourself if you want it? (and we don't have to add another flag to allow folks to ignore the default config if they don't want the behavior it provides)

md5: I'm fine with it either way. In my particular case, the initial goal was to use the stuff from the default config for a derived image, so I coded it this way.

I think the bigger issue is how/when to determine the relative path to individual test cases. I think that maintaining the simple naming convention for test cases currently used in `$imageTests` makes some sense, but it gets complicated if I've replaced the default `config.sh` with my own config. Let's say I wanted to do that and have it run `postgres-basics`, `postgres-initdb`, and `postgis-basics`. One might imagine a custom config that looks like this:

```bash
imageTests[appropriate/postgis] = '
	postgres-basics
	postgres-initdb
	postgis-basics
'
```

Given that configuration, how would the test runner know to look for the first two in `tests/postgres-*` relative to `~/official-images/test` while looking for my added test relative to the directory that the custom config is defined in?

My first thought about this was to have `run.sh` reinitialize `imageTests` to an empty associative array before loading each config, then copy it into a data structure that tracks each iteration of `imageTests` in another associative array keyed by the config file's location, but that seems kind of brittle. It seems like `~/official-images/test/tests/$NAME/` should always be checked if the test directory is not found in a `tests/$NAME` directory that's a sibling of the custom config itself.

tianon: Hmm yeah, interesting thoughts.  I think you're currently the "state of the
art" for tests outside our repo, so as long as the logic doesn't get _too_
hairy I'm +1 for whatever makes your life easier using our framework! :+1:

md5:
I made some updates to this PR to allow custom tests to be defined by user config files. The logic I used to determine the path of tests in the `globalTests` and `imageTests` data structures is basically this:

1. Each time a config is sourced, all of the entries in `$globalTests` and `$imageTests` are checked to determine their paths.
2. If a directory named `tests/$testName` exists as a sibling of the config file, that is used as the path
3. Otherwise, if a directory named `tests/$testName` exists as a sibling of the top-level `run.sh`, that is used as the path

I also made it so that the `$imageTests` associative array is expected to use namespaced image names for non-official images. The official images will use their normal, non-namespaced names.

In addition, I followed @tianon's suggestion of having `-c` override the default config. If the user wants the default config in addition to their own, they need to specify it (see [here](https://github.com/appropriate/docker-postgis/blob/4a20c3cc8cb28f4b5159836892a0f5019498b3c4/.travis.yml#L23) for an example).

Here's the diff from the version we discussed last: https://github.com/appropriate/docker-library-official-images/compare/c8665b5191f3eac28173a073aca3520ecee7d4fc...additional-test-config

md5:
I just pushed some docs to describe the use of an alternate config file: https://github.com/appropriate/docker-library-official-images/commit/5450bc4f820066dde6ff763bbb65425a9b52c62d

This stuff seems to work pretty well, in my opinion. One thing I noticed is that specifying `testAlias` will make `imageTests` be ignored for that image (see https://github.com/docker-library/official-images/blob/master/test/run.sh#L74-L75). The docs seem to imply that it should be additive, though. I didn't change this since it was existing behavior, but I could see it being useful to do something like the following (mentioned in an earlier comment):

```bash
testAlias[appropriate/postgis]=postgres

imageTests[appropriate/postgis]='
	postgis-basics
'
```

For my actual testing, I instead did [this](https://github.com/appropriate/docker-postgis/blob/4a20c3cc8cb28f4b5159836892a0f5019498b3c4/test/postgis-config.sh):

```bash
imageTests[appropriate/postgis]='
	postgres-basics
	postgres-initdb
	postgis-basics
'
```

test/run.sh
```diff
@@ -47,13 +49,51 @@ if [ $# -eq 0 ]; then
...
+	for testName in ${globalTests[@]}; do
+		[ "${testPaths[$testName]}" ] && continue
+
+		if [ -d "$confDir/tests/$testName" ]; then
+			# Test directory found relative to the conf file
+			testPaths[$testName]="$confDir/tests/$testName"
+		elif [ -d "$dir/tests/$testName" ]; then
+			# Test directory found in the main tests/ directory
+			testPaths[$testName]="$dir/tests/$testName"
+		fi
```
yosifkit: Maybe an `else` block to warn the user that a test was not found. Or maybe it just comes out when the tests are run?

test/run.sh
```diff
@@ -47,13 +49,51 @@ if [ $# -eq 0 ]; then
...
+	done
+
+	for image in ${!imageTests[@]}; do
+		for testName in ${imageTests[$image]}; do
```
yosifkit: Could we combine this loop with the one above? :pray:  I dislike the duplicated inner block.

yosifkit:
The tests should be additive. We add back the imageTests for the default name: (https://github.com/docker-library/official-images/blob/b2a8717401e0506ee19325d7c6b01ca42fedfbeb/test/run.sh#L78-L80).

test/run.sh
yosifkit: ie, `for testName in ${globalTests[@]} ${imageTests[@]; do`
magic

md5: Thanks. I also thought it would be nice to combine those, but I didn't realize that `${imageTests[@]}` would enumerate the values of an associated array.

md5:
> The tests should be additive. We add back the imageTests for the default name: (https://github.com/docker-library/official-images/blob/b2a8717401e0506ee19325d7c6b01ca42fedfbeb/test/run.sh#L78-L80).

Thanks! Not sure how I missed that...

test/run.sh
md5: I think that's problematic because it would limit configs to only referring to tests that are in their own sibling `tests` directory. Consider the case where only an alternate config file is used, but it wants to refer to tests in the standard library of official image tests. If `$dir/tests/$testName` is not checked, then the official image tests would not be found.

test/run.sh
md5: Combined. I reran the tests for my PostGIS repo and it worked fine.

test/run.sh
md5: Oh, wait, you meant adding an `else` block, not replacing the `elif` block.

At one point I had an `else` block here, but I figured it would be caught as an error when the tests are run, as you say.

md5: Specifically here: https://github.com/appropriate/docker-library-official-images/blob/additional-test-config/test/run.sh#L154

Although now that I look closer, it seems like `$scriptDir` will be blank in that case and result in a nonsensical error message.

Where would prefer it be caught? I'm thinking having an `elif [ -z "$scriptDir" ]` check after the `if [ -d "$scriptDir" ]` check might be good. Either that or just change the error message in the `else` block from `echo >&2 "error: $scriptDir is not a directory"` to `echo >&2 "error: unable to locate test '$t'"`

md5: I've updated the error message: https://github.com/docker-library/official-images/commit/83bffb89594b99acc780ceb98e5ccf3ba5518974

Hopefully that addresses any remaining concerns about these changes.

test/run.sh
```diff
@@ -65,9 +91,8 @@ for dockerImage in "$@"; do
 		continue
 	fi
 	
-	noNamespace="${dockerImage##*/}"
```
tianon: So, the theory here was that you could run tests on something like `myuser/mysql` and have it run the `mysql` tests appropriately -- I'm not sure that I've ever actually used that functionality myself, but I figure it's worth pointing out that this would be a change in case it matters to anyone (I certainly don't mind, since I've never actually used it).

tianon:
This seems fair now; LGTM.

test/run.sh
md5: I suppose it would be possible to check for test configs for both the namespaced and the non-namespaced verson of the test, but this thing is pretty complicated already.

Also, since it's not documented, I wonder if anyone's ever taken advantage of this functionality.

tianon: Yeah, +1 for axing it. :metal:

md5:
@yosifkit Can I get a "LGTM"? :smile: :pray:

yosifkit:
LGTM

md5:
Thanks! :+1:

PR #1016, opened 2015-09-01
Postgres test enhancements

md5:
This PR updates the `postgres-*` tests to allow the number of connection attempts and sleep duration to be configured. This is helpful when using the tests against derived images that take longer to initialize due to long-running `docker-entrypoint-initdb.d` scripts.

I also made the tests output a "." for each connection attempt to show that the test is still doing something. I think this would be a nice addition to the other tests that have a connection loop. Let me know if you want me to split that changes from this PR and include it in a separate PR that adds that functionality to all such tests.

### DIFF
--- a/test/tests/postgres-basics/run.sh
+++ b/test/tests/postgres-basics/run.sh
@@ -24,7 +24,10 @@ psql() {
 		"$@"
 }
 
-tries=10
+: ${POSTGRES_TEST_TRIES:=10}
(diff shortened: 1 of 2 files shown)


yosifkit:
LGTM, and that sounds great for the other tests too.

tianon:
Seems sane, yeah! :+1:

PR #1025, opened 2015-09-03
Improve test retry output

md5:
This PR updates all the tests that have a retry connection loop to output a period for each connection attempt. It also has them dump the output of `docker logs` to aid in debugging.

It's probably worth pulling this code into a function that can be reused across tests instead of copying and pasting the boilerplate.

### DIFF
--- a/test/tests/docker-dind/run.sh
+++ b/test/tests/docker-dind/run.sh
@@ -29,6 +29,7 @@ while ! docker_ version &> /dev/null; do
 		docker_ version # to hopefully get a useful error message
 		false
 	fi
+	echo >&2 -n .
 	sleep 2
 done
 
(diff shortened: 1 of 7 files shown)


md5:
Now I'm thinking that moving the `docker logs "$cid"` call into the ` trap "docker rm -vf $cid > /dev/null" EXIT` of each test might be a better pattern.

This will make it so that the container logs get dumped for any test failure, not just for the retry failure.

yosifkit:
LGTM

yosifkit:
Making a helper to reduce boilerplate would be nice too :+1:.

md5:
@yosifkit I started playing around with doing that. It would need to do an `eval` since the commands used for testing the connection sometimes involve a pipe.

md5:
I started working on helper function, but I'd like to do it as a separate PR. See here https://github.com/appropriate/docker-library-official-images/commit/0954f94d1ca631dae983fc37fb3193aa0c31b512

yosifkit:
Ok, then this looks fine to me.

tianon:
Nice, LGTM :+1:

PR #1040, opened 2015-09-10
Remove deprecated stackbrew files

yosifkit:


tianon:
LGTM

For anyone who still needs this older code, I've created a `stackbrew` branch (https://github.com/docker-library/official-images/tree/stackbrew) at the latest commit that had modified the `stackbrew` folder so that it's easily accessible. :+1:

PR #1044, opened 2015-09-11
Add test for php-cli image

### LINKED ISSUE #606: [test] php-hello-world
Create a simple hello-world application using the php cli to ensure that the language can compile (if necessary) and run (see ./test/tests/hylang-sh for an example of using expected-std-out.txt).

MBonell:
I added a hello world script to test the php-cli image.
Closes #606

### DIFF
--- a/test/tests/php-hello-world/run.sh
+++ b/test/tests/php-hello-world/run.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-IMAGE="$1"
-DIR_TEST="$(dirname "$(readlink -f "$BASH_SOURCE")")"
-DIR_CONTAINER="/var/www/html/hello-world"
+image="$1"
+dirTest="$(dirname "$(readlink -f "$BASH_SOURCE")")"
+dirContainer='/var/www/html/hello-world'
 
-docker run -it --rm -v $DIR_TEST:$DIR_CONTAINER -w $DIR_CONTAINER $IMAGE php index.php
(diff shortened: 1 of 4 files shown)


test/tests/php-hello-world/run.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+IMAGE="$1"
+DIR_TEST="$(dirname "$(readlink -f "$BASH_SOURCE")")"
+DIR_CONTAINER="/var/www/html/hello-world"
+
+docker run -it --rm -v $DIR_TEST:$DIR_CONTAINER -w $DIR_CONTAINER $IMAGE php index.php
```
tianon: Can you please adapt the style of this file to match other examples in the repo? (ie, lowercase/camelCase local variables, quoted variables in the `docker run` line, prefer single quotes where possible like `DIR_CONTAINER`'s value)

Also, can you add an `expected-std-out.txt` file like we have in [`hylang-sh`](https://github.com/docker-library/official-images/tree/master/test/tests/hylang-sh) for example so that we verify that the output is exactly what we expect as well?

tianon:
Neat, thanks for working on this!

MBonell:
Thank @tianon I updated the code :+1:

tianon:
Nice, LGTM :+1:

yosifkit:
LGTM

PR #1045, opened 2015-09-12
Retry script for tests

md5:
As discussed in #1025, I've refactored the retry functionality into a helper script.

I've converted all of the tests with retry loops to use this function, except for `mongo-basics`. The reason I didn't convert that one yet is that it uses a different command for the retry loop and for the final attempt to get error output.

I also don't much like the use of `. "$dir/../../retry.sh"`, but I couldn't figure out a better way to do it. Besides the hokey relative path thing, the reason I source `retry.sh` instead of just executing it is to ensure that functions declared in the test are still in scope without having to use `export -f` to make them visible.

### DIFF
--- a/test/tests/jetty-hello-web/run.sh
+++ b/test/tests/jetty-hello-web/run.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+[ "$DEBUG" ] && set -x
+
 set -eo pipefail
 
 dir="$(dirname "$(readlink -f "$BASH_SOURCE")")"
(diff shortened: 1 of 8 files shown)


test/tests/jetty-hello-web/run.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+[ "$DEBUG" ] && set -x
```
yosifkit: This looks unrelated :wink:, but I like it :smile:.

md5: Ha! I usually use `git commit --verbose` to see what I'm committing, but it doesn't do any good if you don't actually look at the diff. Let me know if you want me to `--amend` it away.

yosifkit: I'm ok with it.

yosifkit:
LGTM

md5:
ping @tianon

tianon:
Nice, LGTM :+1:

(Sorry I'm so slow sometimes. :innocent:)

md5:
:+1:

PR #1054, opened 2015-09-16
Added spagobi official library definition file

ghost:
Hi,
  I'm in charge to dockerize SpagoBI: it's a opensource suite for Business Intelligence. I would add SpagoBI official docker files to the official library repository of Docker. You can retrieve information of SpagoBI project from these sites:
web site: http://www.spagobi.org
github project: https://github.com/EngineeringSPA/SpagoBI
forge project: http://forge.ow2.org/projects/spagobi
Fiware catalogue : http://catalogue.fiware.org/enablers/data-visualization-spagobi

thanks

### DIFF
--- a/library/spagobi
+++ b/library/spagobi
@@ -1,0 +1,6 @@
+# maintainer: Fabrizio Lovato <fabrizio.lovato@eng.it> (@fabrizyo)
+
+5.1-all-in-one: git://github.com/EngineeringSPA/SpagoBI-Docker@5336ddd1a400b42d86cc42427736d022ce66ae08 5.1-all-in-one
+all-in-one: git://github.com/EngineeringSPA/SpagoBI-Docker@5336ddd1a400b42d86cc42427736d022ce66ae08 5.1-all-in-one
+5.1: git://github.com/EngineeringSPA/SpagoBI-Docker@5336ddd1a400b42d86cc42427736d022ce66ae08 5.1-all-in-one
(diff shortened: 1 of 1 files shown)


psftw:
Thanks for submitting a PR.  Sorry it has taken two weeks to get you some feedback.  I see that development has moved around into a few different repository names.  My comments are against EngineeringSPA/SpagoBI at fc771c5, but this just disappeared from GitHub as well, so you'll want to update the PR here to make it easier for others to review.  There are a few concerns that pop in to my mind while looking it over:

* The image should support port pass-through to the host, and not require one of the workarounds suggested in the doc PR.
* This looks like a bundling of a number of servlets and tomcat 7.0.47.  While I'm not a java expert, I question the inclusion of a two year old Tomcat vs. using "FROM tomcat", and wonder if it is possible to split up this application into small pieces (i.e. one for each servlet?).
* I'm not sure what the difference is between the all-in-one and the fiware versions, but this should be justified/spelled out.
* The main product zip is verified, but the mysql script zip is not?
* Image layers could be optimized a bit.  This diff saves about 250MB of space, some extra layers in the image, and appears to be functional: https://gist.githubusercontent.com/psftw/92f3a0fee7aa68c54643/raw/d210e008abe6584d3ad08235ef79558a534775b7/gistfile1.txt
* I assume there are a number of configuration options that would be required to setup a real-world usable production instance.  It wasn't straight-foward to find this kind of information on the website, so I think it would be useful to include a link or two in the doc PR to relevant references.

psftw:
Closing due to inactivity.

PR #1057, opened 2015-09-16
fix node:0 alias tag and add node:4 alias tag

chorrell:
This fixes an issue with the node:0 tag was pointing to 4.0.0 when is should point to 0.12.7. This also introduces node:4 which points to node:4.0.0


chorrell:
See https://github.com/nodejs/docker-node/issues/43

tianon:
Nice, LGTM :+1:

tianon:
(No `Dockerfile` content changes here -- merging.)

PR #1081, opened 2015-09-26
Bump Rakudo Star to 2015.09

hoelzro:


yosifkit:
Changeset: https://github.com/perl6/docker/compare/2d0512062a...b09f3df7a0f

Build test of #1081; 4d80fa1e4130e7f8dec8d05b261817b742a157bc (`rakudo-star`):

```console
$ url="https://raw.githubusercontent.com/docker-library/official-images/4d80fa1e4130e7f8dec8d05b261817b742a157bc/library/rakudo-star"
$ bashbrew build "$url"
Fetching rakudo-star (git://github.com/perl6/docker) ...
Processing rakudo-star:2015.09 ...
Processing rakudo-star:latest ...
testing rakudo-star:2015.09
	'utc' [1/4]...passed
	'override-cmd' [4/4]...passed
```

tianon:
I'm really sad about https://github.com/perl6/docker/commit/b09f3df7a0f570b55617286f2e721b79c0cff91a, and wonder if maybe it's less related to the shell and more related to the `TERM` environment variable?  Does it work if `-e TERM` is supplied?  How about `-e TERM=xterm` (which IMO would be a somewhat sane default for the `Dockerfile` to specify)?

hoelzro:
@tianon I'm not too happy about it either; I tried with `TERM=xterm`, but to no avail.  I can dig further into the issue, but the `bash` workaround at least works for the moment.

tianon:
Since it's "just" for readline goodness, do you think it'd make sense to instead remove it from here (so default command is just the raw interpreter), but then we document that for an interactive session, you probably want to use `rlwrap` explicitly?

hoelzro:
@tianon Since this image is largely a "try before you buy" version of Rakudo Star, I would really like to offer a positive experience to new users.

tianon:
Fair enough. :+1:

tianon:
I just tested with the freshly built version without the bash wrapper
though ("docker run -it --rm rakudo-star rlwrap perl6") and it appears
to run and work fine here, so I'm a little confused. :confused:

hoelzro:
Hmm...that's odd.  Could it be that Docker on my machine has something wrong with it?

tianon:
It's possible, but not very likely -- which distro, which version of
Docker, and what flags did you supply on run?

hoelzro:
@tianon I'm on Arch Linux, 64-bit, using Docker 1.8.2.  The command invocation I used was:

    $ docker run -it $IMAGE

I just built the image (without the bash hack) on another machine, and it works fine. (???) The only difference between the two that readily comes to mind is that the machine on which things work is using an LTS kernel (4.1.8), whereas the "bad" machine is using a bleeding edge Arch kernel (4.1.6).  Assumingly enough, the LTS kernel is ahead of the non-LTS one today.

hoelzro:
@tianon Ok, so wiping my `/var/lib/docker` and rebuilding on the "bad" machine makes everything work.  /me shrugs

yosifkit:
Revert and force push new commit here :pray:?

hoelzro:
@yosifkit That's the plan; I'll push in an hour or two.

yosifkit:
:+1: Thanks! Ping me when you are ready.

hoelzro:
@yosifkit Should be ready now!

yosifkit:
Changeset: https://github.com/perl6/docker/compare/2d0512062a...a04fb8ba173f10

LGTM

yosifkit:
Merging :tada:, since the diff is basically the same as the previous build.

PR #1087, opened 2015-10-01
Add bashbrew/bashbrew.sh support any whitespace delimiter

c0b:
in some case a tab looks better than space.

### DIFF
--- a/bashbrew/bashbrew.sh
+++ b/bashbrew/bashbrew.sh
@@ -212,7 +212,7 @@ for repoTag in "${repos[@]}"; do
 	
 	tags=()
 	for line in "${repoTagLines[@]}"; do
-		tag="$(echo "$line" | awk -F ': +' '{ print $1 }')"
+		tag="$(echo "$line" | awk -F ':\\s+' '{ print $1 }')"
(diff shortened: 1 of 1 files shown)


c0b:
this one has to bypass ./bashbrew because it changes ./bashbrew/bashbrew.sh

tianon:
From Travis:
```
touch: cannot touch `/home/travis/build/docker-library/official-images/bashbrew/logs/list-2015-10-01--02-37-24/list-aerospike:3.6.1: git://github.com/aerospike/aerospike-server.docker@23ec5eedece7d2714fbd49563a654fd97f376add.log': No such file or directory
```

This PR is breaking the script, probably because `awk` doesn't support `\s` -- I think it has to be `[[:space:]]`.  However, I think we've got other scripts that might need to be updated for this change too (probably over in the docs repo), so I'm hesitant to change the format for something so minor/cosmetic.

c0b:
true, this is not necessary,

when I was editing `library/erlang` it looks beautiful to have tabs support, I changed `bashbrew/bashbrew.sh` locally and it runs perfect for my erlang files.

PR #1096, opened 2015-10-05
Add official image for Neo4j

benbc:
Hello. I'm one of the Neo4j developers at Neo Technology.

We'd love to get an official Neo4j image accepted into the official images library. I've been working with @psftw on the image and we think it's ready for review now.

Please let me know if you think any changes are needed.

### DIFF
--- a/library/neo4j
+++ b/library/neo4j
@@ -1,0 +1,7 @@
+# maintainer: Ben Butler-Cole <ben@neotechnology.com> (@benbc)
+
+2.3.0: git://github.com/neo4j/docker-neo4j@f424f2e0bbe0efca3608ffdc031e27c746da275d 2.3.0
+latest: git://github.com/neo4j/docker-neo4j@f424f2e0bbe0efca3608ffdc031e27c746da275d 2.3.0
+
+2.3.0-enterprise: git://github.com/neo4j/docker-neo4j@f424f2e0bbe0efca3608ffdc031e27c746da275d 2.3.0-enterprise
+enterprise: git://github.com/neo4j/docker-neo4j@f424f2e0bbe0efca3608ffdc031e27c746da275d 2.3.0-enterprise


psftw:
LGTM! :tada:  Thank you for putting up with me.

Note: the issue with "docker stop" has already been fixed upstream, just not in milestone 3.  I tested against this change directly and it works great: https://github.com/neo4j/neo4j/commit/53d5742880b1d99b3f9e342fbf44e20e4568087c

benbc:
Thanks @psftw. If possible we'd like to hold of merging this until the Neo4j 3.0 GA release goes out of the door.

tianon:
I've got a couple comments I noticed after just a quick glance over.

```Dockerfile
RUN mv data /data \
    && ln --symbolic /data

VOLUME /data
```

Why explicitly force neo4j to put data in `/data` instead of just putting the volume directly on `/var/lib/neo4j/data` where neo4j (and presumably users of it) expect it to be?

```bash
if [ -d /conf ]; then
    rm --recursive --force conf
    ln --symbolic /conf
else
# ...
```

Similar comments regarding the contents of `neo4j.sh`, which frankly look a little strange.  What's with the explicit `/conf` directory that overrides all the automatic behavior of the startup script?  Why not mount directly in `/var/lib/neo4j/conf`?  How much of that "custom configuration" stuff could be done in the `Dockerfile` instead and thus be baked into the image?  At the very least, changing `org.neo4j.server.webserver.address` appears to be a good candidate for baking directly into the image.

psftw:
The top-level folders are there for convenience only.  I remember having a discussion with another Neo employee about it as I had a similar reaction, but it didn't bother me enough to raise as a blocker with Ben.

Agree, the stuff in neo4j.sh should address runtime needs only, and outside of the webserver address which could be put in the Dockerfile, I think the others are reasonable since they are driven by environment variables and need to modify java properties-style configuration files before server startup.

benbc:
@tianon We've put the volume at `/data` to make the UI simpler (`--volume=/some/path:/data` vs `--volume=/some/path:/var/lib/neo4j/data`). To me that makes for a simpler, easier to remember contract and one that's less likely to be broken by changes to the Neo4j package structure. (The same argument applied to `/conf` as well, obviously.)

Most of the configuration stuff needs to go into `neo4j.sh` because it's variable at runtime. I have put `org.neo4j.server.webserver.address` in there just in order to keep all the configuration in one place, but of course I can move it to the `Dockerfile` if you prefer.

We have future plans to make Neo4j itself respond to configuration via environment variables in order to integrate better with tools like Docker. That will greatly simplify this file, but we aren't able to make those upstream changes immediately.

In the meantime we're providing the ability to override the entire config via a volume so that users can modify more obscure config options without having to provide for them explicitly in `neo4j.sh`.

benbc:
I have updated the image to include support for clustering.

benbc:
@tianon I've looked again at the setting of `org.neo4j.server.webserver.address` in `neo4j.sh` and I'd really like to keep it as it is. It allows us to start and stop the database (in order to carry out initial configuration that is only available over the API) without exposing a transiently-available database to the outside world. (See https://github.com/neo4j/docker-neo4j/commit/3ad7a933cca620e676a89383c344ab323ef2eba9.)

benbc:
@tianon This image is in a state where it's good to go from the Neo side. Is there anything that you would like to see changed in it?

benbc:
@tianon We are hoping to get Neo4j into the official library in time to announce at a conference on Wednesday next week. Do you think that will be possible?

library/neo4j
```diff
@@ -0,0 +1,7 @@
+# maintainer: Ben Butler-Cole <ben@neotechnology.com> (@benbc)
+
+2.3.0-M03: git://github.com/neo4j/docker-neo4j@660ace32a9851511874ce56731446b1424bfc5cf 2.3.0-M03
+latest: git://github.com/neo4j/docker-neo4j@660ace32a9851511874ce56731446b1424bfc5cf 2.3.0-M03
+
+2.3.0-M03_enterprise: git://github.com/neo4j/docker-neo4j@660ace32a9851511874ce56731446b1424bfc5cf 2.3.0-M03_enterprise
+latest_enterprise: git://github.com/neo4j/docker-neo4j@660ace32a9851511874ce56731446b1424bfc5cf 2.3.0-M03_enterprise
```
tianon: Mind swapping these tags to use `-` instead of `_` and changing `latest_enterprise` to just `enterprise`?

yosifkit:
I'm a little confused by the ["dump-config" section](https://github.com/neo4j/docker-neo4j/blob/master/2.3.0-M03/docker-entrypoint.sh#L56-L63) in your entrypoint.  I don't see an easy way that a user could get the entrypoint-modified config to dump to their mount.

yosifkit:
Should probably to move the [install of `lsof`](https://github.com/neo4j/docker-neo4j/blob/7b7d358006bea9f64b2a845b5bff2eebf3098be9/2.3.0-M03_enterprise/Dockerfile#L12-L14) to before the [`ENV` variables](https://github.com/neo4j/docker-neo4j/blob/7b7d358006bea9f64b2a845b5bff2eebf3098be9/2.3.0-M03_enterprise/Dockerfile#L3-L8) to improve layer sharing between versions of the image and to improve build caching.

library/neo4j
benbc: @tianon Done both. Thanks.

benbc:
@yosifkit 
> Should probably to move the install of lsof to before the ENV variables

Good idea. Thank you. Done.

benbc:
@yosifkit A clean set of unmodified config can be dumped to `~/neo4j-conf` with
```
docker run --rm --volume=$HOME/neo4j-conf:/conf neo4j dump-config
```
This can then be modified by hand and applied with
```
docker run --detach --volume=$HOME/neo4j-conf:/conf neo4j
```

benbc:
We have one last change that we would like to make to this, in order to line up with the GA of our 2.3.0 release. We need to switch from the RC1 that the tags currently point at to the GA version. There will be no changes to the image other than updating the version number in the Dockerfile.

Our current plan is to dark-launch the GA on Tuesday evening Pacific time. At that point we can update the Dockerfile and this PR. Are you happy to take that change and merge this PR late on Tuesday or first thing on Wednesday morning?

benbc:
I have persuaded our PM team to dark-release the GA slightly earlier in order to make this easier. It'll go out at noon Pacific on Tuesday. I'll immediately update the references here to point to the GA version. If you're able to merge on Tuesday afternoon we would be extremely grateful.

yosifkit:
Sounds good to me, pending your update later today. :tada:

Anything else @tianon?

benbc:
@yosifkit @tianon We have released Neo4j 2.3.0 and I have updated the Dockerfile and this PR to point to the new release. Please merge if you are happy. We would be extremely grateful if you are able to push the button today.

tianon:
LGTM

yosifkit:
LGTM, build test forthcoming.

yosifkit:
Build test of #1096; 0b3b60bfed7ef6ecbe3ca35ab14d61d1c5f0d1f0 (`neo4j`):

```console
$ url="https://raw.githubusercontent.com/docker-library/official-images/0b3b60bfed7ef6ecbe3ca35ab14d61d1c5f0d1f0/library/neo4j"
$ bashbrew build "$url"
Cloning neo4j (git://github.com/neo4j/docker-neo4j) ...
Processing neo4j:2.3.0 ...
Processing neo4j:latest ...
Processing neo4j:2.3.0-enterprise ...
Processing neo4j:enterprise ...
testing neo4j:2.3.0
	'utc' [1/4]...passed
	'override-cmd' [4/4]...passed
testing neo4j:2.3.0-enterprise
	'utc' [1/4]...passed
	'override-cmd' [4/4]...passed
```

tianon:
Welcome to the club! :+1:

benbc:
@tianon @yosifkit @psftw Thanks for all your help.

PR #1112, opened 2015-10-12
Closes #591: Test for GCC image

### LINKED ISSUE #591: [test] gcc-c-hello-world
Create a simple hello-world application in `c` to ensure that the language can compile (if necessary) and run (see ./test/tests/hylang-sh for an example of using expected-std-out.txt).

MBonell:
Hello world C test for the GCC image

### DIFF
--- a/test/tests/gcc-c-hello-world/run.sh
+++ b/test/tests/gcc-c-hello-world/run.sh
@@ -5,6 +5,4 @@ image="$1"
 dirTest="$(dirname "$(readlink -f "$BASH_SOURCE")")"
 dirContainer='/usr/src/c'
 
-docker run --rm -v "$dirTest":"$dirContainer":rw -w "$dirContainer" "$image" gcc -o hello-world  hello-world.c
-"$dirTest"/hello-world
-rm -rf "$dirTest"/hello-world
+docker run --rm -v "$dirTest":"$dirContainer":ro -w "$dirContainer" "$image" sh -c 'gcc -o /hello-world hello-world.c && /hello-world'
--- a/test/tests/gcc-c-hello-world/run.sh
+++ b/test/tests/gcc-c-hello-world/run.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+set -e
+
+image="$1"
+dirTest="$(dirname "$(readlink -f "$BASH_SOURCE")")"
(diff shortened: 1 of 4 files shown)


test/tests/gcc-c-hello-world/run.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -e
+
+image="$1"
+dirTest="$(dirname "$(readlink -f "$BASH_SOURCE")")"
+dirContainer='/usr/src/c'
+
+docker run --rm -v "$dirTest":"$dirContainer":rw -w "$dirContainer" "$image" gcc -o hello-world  hello-world.c
+"$dirTest"/hello-world
+rm -rf "$dirTest"/hello-world
```
yosifkit: I would rather have the compiled file stay inside the container, that way we don't need the `rm` or the `rw` mount.
```
docker run --rm -v "$dirTest":"$dirContainer":ro -w "$dirContainer" "$image" sh -c 'gcc -o /hello-world hello-world.c && /hello-world'
```

yosifkit:
Thanks for the PR!

test/tests/gcc-c-hello-world/run.sh
md5: That would also make it more feasible to run this test under Docker Machine on a non-Linux host (other incompatibilities notwithstanding).

MBonell:
Thanks for the feedback! I'll change it

test/tests/gcc-c-hello-world/run.sh
MBonell: I was testing your suggestion, but the `rw` is needed, because the executable have to be created. Error:  `cannot open output file hello-world: Read-only file system`

md5: @MBonell Did you use exactly the command that @yosifkit suggested? It should be writing `hello-world` to `/`, which should be writable. It seems like you may have done `sh -c 'gcc -o hello-world hello-world.c && hello-world'` instead.

MBonell: @md5 You're right! My mistake, I missed the `/`

yosifkit:
LGTM

tianon:
LGTM

PR #1118, opened 2015-10-13
Closes #594: Test for Golang image

### LINKED ISSUE #594: [test] golang-hello-world
Create a simple hello-world application to ensure that the language can compile (if necessary) and run (see ./test/tests/hylang-sh for an example of using expected-std-out.txt).

MBonell:
Hello world test for the Golang image

### DIFF
--- a/test/config.sh
+++ b/test/config.sh
@@ -45,6 +45,7 @@ imageTests+=(
         	gcc-c-hello-world
 	'
 	[golang]='
+	        golang-hello-world
 	'
 	[haskell]='
 		haskell-cabal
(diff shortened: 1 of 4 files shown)


yosifkit:
LGTM

tianon:
Do you think it's worth directly using `go run` instead of `go build` so the Go tool figures out a good place to store the intermediate objects on-disk for us?

MBonell:
@tianon Well, in this case is a single file, so let's use just `go run` is simpler

tianon:
:metal:

PR #1129, opened 2015-10-15
Add script for building and testing PRs

yosifkit:
This is the script we use to generate our nice build tests.  Adding it to git for transparency.  Hopefully we can work on getting a CI server to run it so that the process becomes a little better automated :smile:.


tianon:
This looks just like the one on my machine (except with some added usage text)! :+1:

LGTM

PR #1130, opened 2015-10-15
Add a simple initial "bashbrew" Dockerfile

tianon:
cc @md5 :wink:

It's got a ways to go before it's really, really useful, but it's at least good enough for a lot of stuff as-is. :+1:

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+.gitignore
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-*~
-*.pyc
-/bashbrew/logs
-/bashbrew/src
+.git
+bashbrew/logs
+bashbrew/src
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,12 @@
+FROM docker:git
+
+RUN apk add --update \
+		bash \
+	&& rm -rf /var/cache/apk/*
+
+ENV DIR /usr/src/official-images
+WORKDIR $DIR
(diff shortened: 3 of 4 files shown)


md5:
Shouldn't it `apk add --update bash`?

md5:
Nevermind my inattention!

md5:
Nice start!

md5:
While I think this is an interesting approach, it doesn't really solve the problem with running tests in the Docker Machine scenario encountered on OS X, which is the main hurdle I encounter when working with official images.

Since I need to make the Docker socket available to the container with `-v /var/run/docker.sock:/var/run/docker.sock` in order to get access to a Docker daemon, that daemon is still running outside the image. This works fine for `bashbrew build`, but as soon as I need to run tests that use mounted volumes themselves, everything breaks. The reason for this is that the paths inside the container don't match the paths that the Docker daemon attempts to mount for the (sibling) test containers.

The workaround I normally use for this is to do `-v "$(pwd):$(pwd)" -w "$(pwd)"`, which makes the test file paths inside the container the same as outside the container. Since I'm running these tests from my home directory under `/Users`, which is mounted as a shared folder into the Docker VM, this ends up working, but it's a bit of a pain in the ass.

tianon:
Right!  I have some crazy ideas for that problem too that give us other
nice features, but I need to play a little first to see if they'll actually
work. :smiling_imp:

Thanks for looking and thinking over this. :heart:

yosifkit:
LGTM

PR #1142, opened 2015-10-23
Closes #592: Test for gcc cpp image

### LINKED ISSUE #592: [test] gcc-cpp-hello-world
Create a simple hello-world application in `c++` to ensure that the language can compile (if necessary) and run (see ./test/tests/hylang-sh for an example of using expected-std-out.txt).

MBonell:
Hello world test for the gcc c++ image

### DIFF
--- a/test/config.sh
+++ b/test/config.sh
@@ -43,7 +43,7 @@ imageTests+=(
 	'
 	[gcc]='
 		gcc-c-hello-world
-	        gcc-cpp-hello-world
+		gcc-cpp-hello-world
 	'
 	[golang]='
 		golang-hello-world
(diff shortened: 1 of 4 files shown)


test/config.sh
```diff
@@ -43,6 +43,7 @@ imageTests+=(
 	'
 	[gcc]='
 		gcc-c-hello-world
+	        gcc-cpp-hello-world
```
tianon: This is mixed spaces and tabs. :wink:

tianon:
LGTM sans minor whitespace issue :+1:

yosifkit:
LGTM

PR #1182, opened 2015-11-13
Fix "docker-build.sh" wrapper to handle images with ONBUILD instructions more automatically

tianon:
The way we handle this semi-gracefully is to create a fake "docker save" tarball that contains just a single layer whose parent is the image we're wanting to test, and whose "config" is the same as the parent image, but with "OnBuild" nulled out, and adjust the build to use that new layer as the base for the build instead.

### DIFF
--- a/test/tests/docker-build.sh
+++ b/test/tests/docker-build.sh
@@ -3,11 +3,11 @@ set -e
 
 # wrapper around "docker build" that creates a temporary directory and copies files into it first so that arbitrary host directories can be copied into containers without bind mounts, but accepts a Dockerfile on stdin
 
-# usage: ./docker-build.sh some-host-directory -t some-new-image:some-tag <<EOD
+# usage: ./docker-build.sh some-host-directory some-new-image:some-tag <<EOD
 #        FROM ...
 #        COPY dir/... /.../
 #        EOD
-#    ie: ./docker-build.sh .../hylang-hello-world -t librarytest/hylang <<EOD
+#    ie: ./docker-build.sh .../hylang-hello-world librarytest/hylang <<EOD
 #        FROM hylang
 #        COPY dir/container.hy /dir/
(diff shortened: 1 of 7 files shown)


test/tests/docker-build.sh
```diff
@@ -16,9 +16,63 @@ set -e
...
+	jsonString+=',"id":"'"$newId"'"}'
+	mkdir -p "$tmp/onbuild-hacks/$newId"
+	echo "$jsonString" > "$tmp/onbuild-hacks/$newId/json"
+	echo -n '1.0' > "$tmp/onbuild-hacks/$newId/VERSION"
+	dd if=/dev/zero of="$tmp/onbuild-hacks/$newId/layer.tar" bs=1k count=1 &> /dev/null # empty tar file
+	cat > "$tmp/onbuild-hacks/repositories" <<EOF
+{"$image":{"$tag":"$newId"}}
+EOF
+	tar -cC "$tmp/onbuild-hacks" . | docker load
+	rm -rf "$tmp/onbuild-hacks"
```
tianon: I'm thinking that maybe this whole block deserves to be a whole new script that this script invokes directly instead...

yosifkit: I agree

tianon: Updated, and still works. :+1:

`$ ./test-pr.sh 0 rails:onbuild`

Build test of #0; FAKE (`rails:onbuild`):

```console
$ bashbrew build "rails:onbuild"
Cloning rails (git://github.com/docker-library/rails) ...
Processing rails:onbuild ...
testing rails:onbuild
	'utc' [1/4]...passed
	'override-cmd' [4/4]...passed
```

tianon:
Now there's a generic script here anyone can use if they also have the need of really hacky `ONBUILD`-removal. :innocent:

yosifkit:
LGTM

PR #1201, opened 2015-11-20
Adding F# language

haf:
Finally there shall be light to the docker universe!

This aims to replace [the fsharp docker account](https://hub.docker.com/r/fsharp/fsharp/).

# Checklist for Review

  - https://github.com/docker-library/official-images/pull/1201#issuecomment-161607873
  - https://github.com/docker-library/docs/issues/583
  - https://github.com/docker-library/docs/issues/641
  - fbad0cb41146501d1ab573f3fde9d4443ae4b066 (https://github.com/fsprojects/docker-fsharp/blob/d19e5c72b46ee5248b1d3411809a0e3cb7fb9e5d/4.0.0.4/Dockerfile)
  - https://github.com/docker-library/official-images/pull/1201#issuecomment-161106320
  - https://github.com/docker-library/official-images/issues/1948

### DIFF
--- a/library/fsharp
+++ b/library/fsharp
@@ -1,0 +1,5 @@
+# maintainer: Henrik Feldt <henrik@haf.se> (@haf)
+
+4.0.0.4: git://github.com/fsprojects/docker-fsharp@d19e5c72b46ee5248b1d3411809a0e3cb7fb9e5d 4.0.0.4
+latest: git://github.com/fsprojects/docker-fsharp@d19e5c72b46ee5248b1d3411809a0e3cb7fb9e5d 4.0.0.4
+
--- a/library/fsharp
+++ b/library/fsharp
@@ -1,5 +1,5 @@
 # maintainer: Henrik Feldt <henrik@haf.se> (@haf)
 
-4.0.0.4: git://github.com/fsprojects/docker-fsharp@d19e5c72b46ee5248b1d3411809a0e3cb7fb9e5d 4.0.0.4
-latest: git://github.com/fsprojects/docker-fsharp@d19e5c72b46ee5248b1d3411809a0e3cb7fb9e5d 4.0.0.4
-
+4.0.0.4: git://github.com/fsprojects/docker-fsharp@f063b22fc1e9f3d78e025f3a3842299721cd78bf 4.0.0.4
+4.0.1.1: git://github.com/fsprojects/docker-fsharp@1a01ecbe00ef0c1399893ed5cda40c197614f48d 4.0.1.1
+latest: git://github.com/fsprojects/docker-fsharp@f063b22fc1e9f3d78e025f3a3842299721cd78bf 4.0.0.4
--- a/library/fsharp
+++ b/library/fsharp
@@ -2,4 +2,4 @@
 
 4.0.0.4: git://github.com/fsprojects/docker-fsharp@f063b22fc1e9f3d78e025f3a3842299721cd78bf 4.0.0.4
 4.0.1.1: git://github.com/fsprojects/docker-fsharp@1a01ecbe00ef0c1399893ed5cda40c197614f48d 4.0.1.1
-latest: git://github.com/fsprojects/docker-fsharp@f063b22fc1e9f3d78e025f3a3842299721cd78bf 4.0.0.4
+latest: git://github.com/fsprojects/docker-fsharp@1a01ecbe00ef0c1399893ed5cda40c197614f48d 4.0.0.4
--- a/library/fsharp
+++ b/library/fsharp
@@ -2,4 +2,4 @@
 
 4.0.0.4: git://github.com/fsprojects/docker-fsharp@f063b22fc1e9f3d78e025f3a3842299721cd78bf 4.0.0.4
 4.0.1.1: git://github.com/fsprojects/docker-fsharp@1a01ecbe00ef0c1399893ed5cda40c197614f48d 4.0.1.1
-latest: git://github.com/fsprojects/docker-fsharp@1a01ecbe00ef0c1399893ed5cda40c197614f48d 4.0.0.4
+latest: git://github.com/fsprojects/docker-fsharp@1a01ecbe00ef0c1399893ed5cda40c197614f48d 4.0.1.1


forki:
:heart:

haf:
Hello, is anyone home?

yosifkit:
Thanks for the poke.

I would think it would be simpler to just do `FROM mono:4.2.1.102` so that layers can be shared if the user already has a mono project that isn't fsharp.  But now I am confused: the mono image [includes fsharp](https://github.com/mono/docker/blob/39c80bc024a4797c119c895fda70024fbc14d5b9/4.2.1.102/Dockerfile#L15).  Should that be removed so that fsharp can be a separate image that inherits from mono?

haf:
I would prefer it if we don't inherit mono, as we need to control their version very granularly. It's often the case that they release versions that are incompatible with F#.

yosifkit:
The `FROM` can be as specific as necessary and then bumped when you have tested it.  But the `fsharp` package should be dropped from the mono image, since there would be no need for this image if that exists.

haf:
It should not be dropped from the mono image since it's a part of a baseline mono installation, aiming to be identical to the CLR on Windows.

However, there's a need for a language-owned image, such as this one. By having a specific language-owned image we can guarantee a certain amount of stability and testing as well as an up to date language package.

forki:
@haf I agree. Mono should come as always, otherwise we would break lots and lots of code. 
New fsharp image is designed for new use cases like fsharp websites.

dsyme:
@haf  @forki @yosifkit As F# language designer I do agree with this based on the concerns above.  The "fsharp" container has needs different to a "mono" container.

In addition, we are eventually likely to include tags on this container which would allow choosing a .NET Core substrate instead of Mono (.NET Core is the Microsoft .NET implementation for Linux).  

So having one well-managed "fsharp" base container is important both for the QA reasons above, and because Mono won't always be the only substrate that F# depends on for Linux.

tianon:
I think the reasoning for not going `FROM mono` does make sense, and is fair enough -- I've sent a PR with some minor nits I had (https://github.com/fsprojects/docker-fsharp/issues/7), but otherwise this is looking pretty good to me.

Once everything's dialed in and we're ready to merge, we'll have an appropriate PR over in the docs repo to document what the image is and how to use it, right? (https://github.com/docker-library/docs/blob/master/README.md) :smile:

psftw:
Hmmm.. image is failing to build for me, but I recall it worked ~1 month ago.

```
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 mono-devel : Depends: libmono-cecil-private-cil (< 4.2.1.103) but 4.2.2.10-0xamarin1 is to be installed
              Depends: mono-mcs (= 4.2.1.102-0xamarin1) but 4.2.2.10-0xamarin1 is to be installed
              Depends: mono-gac (= 4.2.1.102-0xamarin1) but 4.2.2.10-0xamarin1 is to be installed
              Depends: mono-xbuild (= 4.2.1.102-0xamarin1) but 4.2.2.10-0xamarin1 is to be installed
              Depends: libmono-cil-dev (= 4.2.1.102-0xamarin1) but 4.2.2.10-0xamarin1 is to be installed
              Depends: libmono-2.0-dev (< 4.2.1.102-0xamarin1.1~) but 4.2.2.10-0xamarin1 is to be installed
^[[91mE: Unable to correct problems, you have held broken packages.
```

It also looks like there are new versions of fsharp to include, and fsharp/fsharp on DockerHub is also outdated.  Please ping me when this PR is updated and I'll take a stab at testing.

haf:
Thanks for notifying me. It's strange that pinned packages have changed /cc @directhex

directhex:
It's not pinned. See http://www.mono-project.com/docs/getting-started/install/linux/#accessing-older-releases and the contents of http://download.mono-project.com/repo/debian/dists/wheezy/snapshots/ (although it seems @akoeplinger may have forgotten to make 4.2.2 symlinks)

akoeplinger:
@directhex @haf 4.2.2.10 (and 4.2.2.29 which I just pushed) are alpha releases of the next Mono 4.2 service release. afaik we don't provide snapshots for the alpha channel.

You're adding the Mono alpha-channel [here](https://github.com/fsprojects/docker-fsharp/blob/master/4.0.0.4/Dockerfile#L6), I'd suggest you remove that since Mono 4.2.1.102 on which you rely is the stable release.

yosifkit:
Any movement here?  Looks like the Dockerfile just needs to be updated to use the specific  `snapshots` repo rather then `main` or `alpha`.

haf:
Ok, some movement!

I've pushed an update to the repo to use the snapshots repo from X.

haf:
Correction, having problems connecting to the internet in the docker container. Upgrading it.

haf:
Do I need to do anything else?

haf:
Also, should we try and add CoreCLR, @enricosada?

haf:
Ping

yosifkit:
Two questions and I think this image looks solid to me. :+1:
- can we get a docs PR to https://github.com/docker-library/docs?
- why `--force-yes` on `apt-get install`? It seems unnecessary.

haf:
@yosifkit Removed `--force-yes` and PRed the docs.

tianon:
I'm getting `WARNING: The following packages cannot be authenticated!` (with a bunch of `mono` packages) while trying to build this -- are you sure `3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF` is the right key for `deb http://download.mono-project.com/repo/debian wheezy/snapshots 4.2.3.4/main` ?  Does it need an additional key?  (`--force-yes` is definitely not the right solution here :cry:)

Other than that, this looks like it's really, really close, and I appreciate that you folks have stuck with it! :+1:

tianon:
Oh strange, if I run it by hand, it appears to work fine.  Something strange is afoot here.

tianon:
I can't find any clues; my _only_ guess is maybe something related to the key being `2048R`, but if that were the case (and it shouldn't be in `trusty`, only `xenial` and up was that changed IIRC) we should see a warning or error on `apt-get update`, not problems during install.

Going to see if the repo has the same issue on `wheezy` proper (to narrow down whether it's something specific to `trusty`).

tianon:
Gah, not specific to `trusty` :cry: (happens in `wheezy` base image as well)

tianon:
Figured it out! https://github.com/fsprojects/docker-fsharp/pull/13 :+1:

tianon:
As of fsprojects/docker-fsharp#13 merging, the `Dockerfile` LGTM :+1: (needs commit references updated here)

haf:
Ping @tianon 

If you have more input, perhaps it's about time that you merge this and then we can take 'extra thoughts' later?

haf:
W.R.T the tests; if possible, please add the infrastructure and core files and I can write a hello world F# program. In fact: `printfn "Hello World"` inside a `Hello.fsx` file executed with `fsharpi`.

yosifkit:
LGTM, though you might want to add a `WORKDIR` to `4.0.0.4` as well, but that shouldn't hold up this PR.  (now I am off to review the docs)

cc @tianon 
Build test of #1201; 5f5a9d7 (`fsharp`):

```console

$ bashbrew build fsharp:4.0.0.4
warning: insecure protocol git:// detected: git://github.com/fsprojects/docker-fsharp
Building bashbrew/cache:761464f0a075b70f03c85218e8dfcf8ff5711546cdcda0b2fb2c82f6a1bfe231 (fsharp:4.0.0.4)
Tagging fsharp:4.0.0.4

$ test/run.sh fsharp:4.0.0.4
testing fsharp:4.0.0.4
	'utc' [1/4]...passed
	'override-cmd' [4/4]...passed


$ bashbrew build fsharp:4.0.1.1
Building bashbrew/cache:0fadb2c72927c254ea388862292faca3704bef6d8bb29b5ef13289795e5ca73d (fsharp:4.0.1.1)
Tagging fsharp:4.0.1.1
Tagging fsharp:latest

$ test/run.sh fsharp:4.0.1.1
testing fsharp:4.0.1.1
	'utc' [1/4]...passed
	'override-cmd' [4/4]...passed

```

tianon:
LGTM

```diff
diff --git a/4.0.0.4/Dockerfile b/4.0.1.1/Dockerfile
similarity index 96%
copy from 4.0.0.4/Dockerfile
copy to 4.0.1.1/Dockerfile
index 55cc2ae..1b88e9c 100644
--- a/4.0.0.4/Dockerfile
+++ b/4.0.1.1/Dockerfile
@@ -12,7 +12,7 @@ RUN apt-get -y update && \
     apt-get -y --no-install-recommends install nuget mono-devel ca-certificates-mono && \
     rm -rf /var/lib/apt/lists/*
 
-ENV FSHARP_VERSION 4.0.0.4
+ENV FSHARP_VERSION 4.0.1.1
 ENV FSHARP_PREFIX=/usr \
     FSHARP_GACDIR=/usr/lib/mono/gac \
     FSHARP_BASENAME=fsharp-$FSHARP_VERSION \
@@ -30,4 +30,5 @@ RUN mkdir -p /tmp/src && \
     cd ~ && \
     rm -rf /tmp/src
 
+WORKDIR /root
 CMD ["fsharpi"]
```

tianon:
(we'll merge as soon as the docs are ready -- otherwise, we'll have a blank Hub page :+1:)

PR #1230, opened 2015-12-02
update to crate 0.52.3 stable

mikethebeer:


yosifkit:
Changeset: https://github.com/crate/docker-crate/compare/d8ef1163d...ffcbb88405

Overall seems good to me, though I would suggest that for better layer sharing to move the `ENV version` to after installing python. Also `apt-get clean` doesn't actually do anything since it is built into the [debian images](https://github.com/docker/docker/blob/fcccf2dae4770bdb7781d57e20ae94b9565b67ac/contrib/mkimage/debootstrap#L94-L101).

mikethebeer:
Hi @yosifkit 

thanks for the update. I have changed our dockerfile related to your comment. You can merge this PR or close it since we have a new stable release PR already here https://github.com/docker-library/official-images/pull/1246

ty

yosifkit:
Thanks! We can just stick with the new one.

PR #1236, opened 2015-12-03
Updates for django drupal golang mongo mysql php

yosifkit:
- `django` bump to 1.9
- `drupal` bump to 8.0.1
- `golang` bump 1.5.2
- `mongo` bump to 3.2.0-rc6
- `mysql` see https://github.com/docker-library/mysql/pull/119
- `php` PHP 7.0 GA

PR #1237, opened 2015-12-04
Update to PHP7

Lewiscowles1986:
Incorporates php7 stable release


md5:
This appears to be a duplicate of #1236, although it looks like you updated the file by hand instead of running `generate-stackbrew-library.sh` (since only the `latest` tag moved and not `cli`, `apache`, or `fpm`).

See the diff here: https://github.com/docker-library/official-images/pull/1236/files#diff-a91f6a9f720a0e0615d1d888fd98516b

Lewiscowles1986:
sorry I did not notice this other PR (which I think should be several PRs), I'll close this one, I searched for php7 before creating, and I did not know about `generate-stackbrew-library.sh`, is there a place I can read about it, or is it pretty much run it from cli and read what it says?

Lewiscowles1986:
Also just updated my patch-1 branch just-in-case you cannot merge that huge PR, I would really appreciate having PHP7 updated (regardless of what is going on with go, mysql, mongo, drupal & django) :+1:

md5:
@Lewiscowles1986 I'm not aware of any documentation of `generate-stackbrew-library.sh`. As for your concerns about having multiple updates in a single PR, I wouldn't worry about it too much. Since @yosifkit is part of the Official Images team and has already successfully tested the PR, I expect @tianon will test and merge it tomorrow. If there were to be a problem with one of the images, I expect they would just drop it from the PR.

PR #1248, opened 2015-12-08
Add nasven.js image on top level Docker Hub Registry

brunoborges:
[Nasven.js](https://github.com/nasven/) is a Nashorn tool to get Maven artifacts in scope for Nashorn script execution, and easier classpath configuration.

### DIFF
--- a/library/nasven
+++ b/library/nasven
@@ -1,0 +1,3 @@
+# maintainer: Bruno Borges <bruno.borges@gmail.com> (@brunoborges)
+
+latest: git://github.com/nasven/nasven@5b68fd3a191819aabeb623f153e0d970668c98e4 docker


psftw:
I appreciate the enthusiasm, but this looks like your pet project, and therefore would not be something I could get behind for an official repository.  It's a bit too specific IMHO, and I would suggest continuing to promote this via a non-official repository for now.

I saw a thread discussing issues with OpenJDK on top of Alpine the other day, but couldn't find it to link to here in case it had ramifications for your work.  Instead, I found this PR which you may find interesting: https://github.com/docker-library/java/pull/62

yosifkit:
Agree with @psftw; Closing old PR.

PR #1266, opened 2015-12-11
Rename tags: latest to lts

warden:


yosifkit:
I would rather not remove `latest` tag since that is the default when just doing `docker pull jenkins` and would be very confusing for beginning users. The problem is that docker chose to make `latest` the default, rather than something like `default`.  I would be :+1: to adding `lts` to make it more clear what `latest` means. @carlossg, this is regarding your images, what you do think?

warden:
I am getting your explanation, however I'm not sure whether by just adding *lts* tag wouldn't make users think that lts is lts and latest is latest from master... Up to you guys though.

yosifkit:
That is why the [Docker Hub description](https://hub.docker.com/_/jenkins/) contains all the active tags in a handy list at the top, grouped by image:  (It should be displaying a header similar to below, but the Docker Hub is awaiting a fix.)

## Supported tags and respective Dockerfile links
- latest, 1.625.3 (Dockerfile)

tianon:
I'd also note that the tag names are ordered in the description in the same
order as in the file, so I usually start with the most specific version (in
this case, 1.625.3, and leave "latest" last).  It might make sense to make
"lts" be first so that it stays clear, ie "lts, latest, 1.625.3" or
"lts, 1.625.3,
latest".

carlossg:
latest should not be removed. If you want to add LTS we can discuss it in https://github.com/jenkinsci/docker/issues/170

yosifkit:
Thanks @carlossg!

PR #1272, opened 2015-12-14
Fix apt-get Best Practices Link

JacobHayes:
The `apt-get` best practices link currently points to `run`, fixed to point to `apt-get`.


tianon:
LGTM, thanks! :+1:

PR #1273, opened 2015-12-15
aerospike 3.7.0.1

volmarl:


yosifkit:
Just thought I would chime a warning on the `VOLUME ["/etc/aerospike/"]`.  Is it something that should change on run?  While docker will copy what exists in a volume on first start to a new docker managed volume, it will not do so with a bind mount (`-v /my/path/:/etc/aerospike/`).

Changeset: https://github.com/aerospike/aerospike-server.docker/compare/a5fb93bc482...fe600d210a

LGTM, Build test of #1273; bfad3a5 (`aerospike`):

```console
$ bashbrew build "aerospike"
Cloning aerospike (git://github.com/aerospike/aerospike-server.docker) ...
Processing aerospike:3.7.0.1 ...
Processing aerospike:latest ...
testing aerospike:3.7.0.1
	'utc' [1/4]...passed
	'override-cmd' [4/4]...passed
```

volmarl:
Hi Yosikit,
Lets hold on 3.7.0.1 as I have a patch released today 3.7.0.2
Let me go into details into what we are  trying to do with Volume and you can chime in on best practice.
The idea is to allow a user to modify the default aerospike.conf config within the container located at /etc/aerospike/aerospike.conf the plan is to use -v option when issuing a docker run
-v /my/path/:/etc/aerospike/
an updated  aerospike.conf would be under /my/path/ and would be loaded.

Would that work has planned?

appreciate your help.

tianon:
That'll work fine without specifying a volume in the Dockerfile, and is
exactly what we'd propose you document -- if the volume is specified
explicitly in the Dockerfile, then it actually becomes impossible to modify
the configuration in a subsequent Dockerfile (ie, FROM aerospike; COPY
some-new.conf /etc/aerospike/) if a user wanted to pre-bake an image with
their updated config. :+1:

volmarl:
Understood. Thanks for clarifying the issue. We are making the change in 3.7.0.2 and closing pull request for 3.7.0.1

tianon:
Cool, sounds good!  Thanks! :+1:

PR #1294, opened 2015-12-21
Add Varnish

ericandrewlewis:
Hi,

we've open sourced our Varnish Dockerfile and would like to contribute it to the community :smile:
# Checklist for Review
- [ ] associated with or contacted upstream?
  - https://www.varnish-cache.org/lists/pipermail/varnish-misc/2015-December/024747.html :cry:
- [ ] does a [documentation](https://github.com/docker-library/docs/blob/master/README.md) PR exist? (should be reviewed and merged at roughly the same time so that we don't have an empty image page on the Hub for very long)
- [ ] dockerization review for best practices and cache gotchas/improvements (ala [the official review guidelines](https://github.com/docker-library/official-images/blob/master/README.md#review-guidelines))?
- [ ] existing official images have been considered as a base? (ie, if `foobar` needs Node.js, has `FROM node:...` instead of grabbing `node` via other means been considered?)
- [ ] passes current tests? any simple new tests that might be appropriate to add? (https://github.com/docker-library/official-images/tree/master/test)

### DIFF
--- a/library/varnish
+++ b/library/varnish
@@ -1,0 +1,7 @@
+# maintainer: Eric Lewis <eric.lewis@nytimes.com> (@ericandrewlewis)
+# maintainer: Eric Buth <eric.buth@nytimes.com> (@buth)
+
+4.1.0: git://github.com/newsdev/docker-varnish@2338063953894a2b812712b75c48e0e4fd31a5c9
+4.1: git://github.com/newsdev/docker-varnish@2338063953894a2b812712b75c48e0e4fd31a5c9
+4: git://github.com/newsdev/docker-varnish@2338063953894a2b812712b75c48e0e4fd31a5c9
+latest: git://github.com/newsdev/docker-varnish@2338063953894a2b812712b75c48e0e4fd31a5c9


tianon:
Nice to finally see someone take a stab at an official Varnish image! :smile:

I've got a couple high-level comments/questions to start off the discussion:

1. Has Varnish upstream been contacted to see if they're interested in participating or at least endorsing? (and if they aren't interested, whether they'll mind if we press on without their participation)

2. Looking at https://www.varnish-cache.org/releases, it would appear that both 4.1 and 4.0 are still officially supported, so it'd be great to have that reflected here.  Also, they publish official binaries/packages, so I think we'd be remiss not to take advantage of that (especially so that the `Dockerfile`s are simpler and easier to maintain, but also so that we're creating an image that deploys Varnish the way that upstream recommends).

ericandrewlewis:
> Has Varnish upstream been contacted to see if they're interested in participating or at least endorsing? (and if they aren't interested, whether they'll mind if we press on without their participation)

I [reached out](https://www.varnish-cache.org/lists/pipermail/varnish-misc/2015-December/024669.html) to the Varnish community a few weeks ago to solicit feedback, which there was some of but not a ton. I'll send another note out on that thread with an update.

> Looking at https://www.varnish-cache.org/releases, it would appear that both 4.1 and 4.0 are still officially supported, so it'd be great to have that reflected here. Also, they publish official binaries/packages, so I think we'd be remiss not to take advantage of that (especially so that the Dockerfiles are simpler and easier to maintain, but also so that we're creating an image that deploys Varnish the way that upstream recommends).

In Varnish 3.0, VMODs (Varnish modules) were compiled against Varnish source. Since Varnish 4.0 VMODs can be compiled against a Varnish installation (not requiring source). However, many VMODs have not been updated to use this mechanism, and will still require Varnish source. Given that, we've found that having source available is a good default for backwards compatibility until more VMODs are updated. ([see discussion](https://www.varnish-cache.org/lists/pipermail/varnish-misc/2015-December/024675.html))

tianon:
Oh!  That thread is fantastic and really answers both my questions
perfectly; thanks for sharing. :smile:

Your most recent post requesting support for the efforts here is excellent,
thanks for doing that.  Do you want us to continue review in the meantime,
or do you want to wait for a response to make sure we don't waste effort if
it turns out the project would rather we not continue?

ericandrewlewis:
> Do you want us to continue review in the meantime, or do you want to wait for a response to make sure we don't waste effort if it turns out the project would rather we not continue?

Yes, let's hold on a response.

kevin1024:
Thanks for doing this.  In lieu of an official image I'm going to go use yours :+1:

avoinea:
```
  cd varnish-$VARNISH_VERSION && \
  ./autogen.sh && \
  ./configure && \
  make install && \
  rm ../varnish-$VARNISH_VERSION.tar.gz
```
Shouldn't you also cleanup src folder `varnish-$VARNISH_VERSION`? It adds **37M** to the final image.

teohhanhui:
@avoinea Some Varnish modules need to be built with Varnish source. See https://github.com/docker-library/official-images/pull/1294#issuecomment-168208329

andrerom:
Would be great to see a official Varnish image, lowers the barrier for using Varnish locally, for testing, and on simpler server setups.

However the proposed image does not seem to be usable as is like other official images. IMO purpose of an official image should be that it can be used for out of the box, and extended if you need more modules/plugins/*, not just as a base image. So a propoer entrypoint is at least needed here.

Also, maybe add an Alpine variant *without* the source to give incentive to users of other VMODs to get them updated to support building from install?

teohhanhui:
@andrerom The original repo seems to be inactive. We've forked it here: https://github.com/tripviss/docker-varnish

Hoping to make it better...

tianon:
@ericandrewlewis do you think we've given upstream enough time to raise objections at this point?  Any value in continuing to press them, or do you think we should press forward?  It looks like you've got some interested co-conspirators now. :smile:

ericandrewlewis:
I think we've given upstream enough time, yeah. The Varnish core team seems rather neutral on the topic, and if they ever want to get more involved I'm sure they could.

I can't commit to being a maintainer at this point, so I leave it to y'all to decide what to do going forward here. Cheers!

andrerom:
👍  I also have the impression they won't handle this but don't mind others do it.

For official image, what about adding a few things:
- entrypoint _(could be added later to, but useful)_
- Varnish 5
- Afaik should not take to much space, is official and very useful: https://github.com/varnish/varnish-modules _(download link on the readme)_

tianon:
Thanks @ericandrewlewis :heart: :+1:

I hope you don't mind if I close this -- if another group takes up the torch, they should refer back to this discussion! :metal:

PR #1303, opened 2015-12-25
Update ruby to include 2.3

ianneub:
This will also replace latest/onbuild/slim with ruby 2.3

### DIFF
--- a/library/ruby
+++ b/library/ruby
@@ -23,15 +23,24 @@
 
 2.2.4: git://github.com/docker-library/ruby@90b6b506a4b495cd337115d9a2b4ba6f80c72004 2.2
 2.2: git://github.com/docker-library/ruby@90b6b506a4b495cd337115d9a2b4ba6f80c72004 2.2
-2: git://github.com/docker-library/ruby@90b6b506a4b495cd337115d9a2b4ba6f80c72004 2.2
-latest: git://github.com/docker-library/ruby@90b6b506a4b495cd337115d9a2b4ba6f80c72004 2.2
 
 2.2.4-onbuild: git://github.com/docker-library/ruby@5d04363db6f7ae316ef7056063f020557db828e1 2.2/onbuild
 2.2-onbuild: git://github.com/docker-library/ruby@5d04363db6f7ae316ef7056063f020557db828e1 2.2/onbuild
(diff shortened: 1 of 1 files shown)


mcoms:
Pinging https://github.com/docker-library/ruby/pull/59 so it's clear there's a downstream.

library/ruby
```diff
@@ -23,15 +23,24 @@
...
 
 2.2.4-slim: git://github.com/docker-library/ruby@90b6b506a4b495cd337115d9a2b4ba6f80c72004 2.2/slim
 2.2-slim: git://github.com/docker-library/ruby@90b6b506a4b495cd337115d9a2b4ba6f80c72004 2.2/slim
-2-slim: git://github.com/docker-library/ruby@90b6b506a4b495cd337115d9a2b4ba6f80c72004 2.2/slim
-slim: git://github.com/docker-library/ruby@90b6b506a4b495cd337115d9a2b4ba6f80c72004 2.2/slim
+
+2.3.0: git://github.com/docker-library/ruby@b8e8fe051b3c3c1973a8998679bbd47d6ab13f69 2.3
+2.3: git://github.com/docker-library/ruby@b8e8fe051b3c3c1973a8998679bbd47d6ab13f69 2.3
+2: git://github.com/docker-library/ruby@b8e8fe051b3c3c1973a8998679bbd47d6ab13f69 2.3
+latest: git://github.com/docker-library/ruby@b8e8fe051b3c3c1973a8998679bbd47d6ab13f69 2.3
```
tianon: This file's contents are generated via `./generate-stackbrew-library.sh` from within a clone of the `ruby` repo, which puts these lines at commit `1b08f346713a1293c2a9238e470e086126e2e28f` instead.  Do you want to update this PR, or would you rather we take over from here?

ianneub: You can take it over. Thanks for teaching me a bit about the process. Should I go ahead and close the PR?

tianon: Yeah, sure; thanks for your efforts!  Not everyone uses a script to
generate this content, but all of the /docker-library/ repos do (and
several others as well). :+1:

ianneub:
@tianon will take over using the correct procedure. Thanks all!

tianon:
https://github.com/docker-library/official-images/pull/1306 :+1:

PR #1308, opened 2015-12-28
Add the notion of "explicit variants" to our test suite

tianon:
For an "explicit variant", only the tests specified _explicitly_ for that variant will be executed -- the initial target of this is to not run quite so many tests on ":onbuild" variants, thus no longer requiring our hacky remove-onbuild logic (since the ":onbuild" variants are usually just extra metadata and sometimes a few extra packages on top of the standard variant).

### DIFF
--- a/test/run.sh
+++ b/test/run.sh
@@ -54,6 +54,7 @@ declare -a globalTests=()
 declare -A testAlias=()
 declare -A imageTests=()
 declare -A globalExcludeTests=()
+declare -A explicitTests=()
 
 # if there are no user-specified configs, use the default config
 if [ ${#configs} -eq 0 ]; then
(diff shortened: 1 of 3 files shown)


test/run.sh
```diff
@@ -99,9 +100,30 @@ for dockerImage in "$@"; do
 	testRepo=$repo
 	[ -z "${testAlias[$repo]}" ] || testRepo="${testAlias[$repo]}"
 	
-	testCandidates=( "${globalTests[@]}" ${imageTests[$testRepo]} ${imageTests[$testRepo:$variant]} )
+	testCandidates=()
+	if [ -z "${explicitVariants[$variant]}" ]; then
```
yosifkit: Just a suggestion that we may want `$repo:$variant` as a key also.

tianon: Yeah, good call; updating :+1:

tianon: Updated!

yosifkit:
LGTM

md5:
@tianon Where's an example of the `remove-onbuild` logic?

While this idea seems OK on its face, I do have some concern that this could allow derived images to stop executing tests that they should be passing if someone forgets to add them as "explicit" tests. I'd rather see a test exclusion mechanism of some sort for derived images to list the parent image tests that they've opted out of.

tianon:
The script is at
https://github.com/docker-library/official-images/blob/503d8a710c541a493f5f17316c434a87c495500e/test/tests/remove-onbuild.sh;
it's invoked via
https://github.com/docker-library/official-images/blob/503d8a710c541a493f5f17316c434a87c495500e/test/tests/docker-build.sh,
which is the foundation of pretty much all our tests.  The problem with it
is that it no longer works on Docker master (the in-progress 1.10), due to
changes in the way images are stored and referenced, but on top of that,
running all the non-onbuild tests on onbuild variants was always something
that bothered me since we do review them to make sure they don't ever do
much more than mutate metadata and perhaps install a new package or two
(which are things that are already harder to test for), and only bothered
me more when we had to do crazy hacks to remove the "onbuild" metadata from
them in order to run the non-onbuild tests.

tianon:
Basically, we were doing really bad hacks for "docker build" based
tests on images with onbuilds to work at all, and it came back to
haunt us pretty quickly in a way that motivated me to finally figure
out a mostly-clean way to skip tests that weren't designed to handle
ONBUILD metadata explicitly (which we need to create some of to
replace this functionality for sure).

PR #1311, opened 2015-12-30
Fix python-hy to not use bash

ncopa:
The python alpine variant does not ship bash so we use /bin/sh instead.

### DIFF
--- a/test/tests/python-hy/container.sh
+++ b/test/tests/python-hy/container.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 set -e
 
 pip install -q hy
--- a/test/tests/python-hy/run.sh
+++ b/test/tests/python-hy/run.sh
@@ -1,1 +1,1 @@
-../run-bash-in-container.sh
+../run-sh-in-container.sh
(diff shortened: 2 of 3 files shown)


test/tests/run-bash-in-container.sh
```diff
@@ -4,4 +4,4 @@ set -e
 testDir="$(readlink -f "$(dirname "$BASH_SOURCE")")"
 runDir="$(dirname "$(readlink -f "$BASH_SOURCE")")"
 
-source "$runDir/run-in-container.sh" "$testDir" "$1" bash ./container.sh
+source "$runDir/run-in-container.sh" "$testDir" "$1" sh ./container.sh
```
tianon: This should instead be a new `run-sh-in-container.sh` script which is then symlinked into `test/tests/python-hy/run.sh` (where `run-bash-in-container.sh` is symlinked now).

tianon:
Other than the one nit, this seems sane to me. :+1:

ncopa:
updated

tianon:
Perfect, thanks! :+1:

LGTM

PR #832, opened 2015-06-16
Add "import readline" to python-imports test

tianon:
https://github.com/docker-library/python/issues/53

(also alphabetized the imports)

PR #1315, opened 2016-01-04
Replace readline test with sqlite and ssl

ncopa:
The alpine image does not have readline, so we test for sqlite3 and ssl
instead.


ncopa:
Ok. this should be rejected for now due to #832. We wait til upstream supports libedit properly-

PR #1317, opened 2016-01-04
Fix ruby-bundler and ruby-gems tests to not use bash

ncopa:
Don't assume that bash is available. Alpine does not ship bash by
default, so use the ligher /bin/sh instead.

### DIFF
--- a/test/tests/ruby-gems/container.sh
+++ b/test/tests/ruby-gems/container.sh
@@ -1,20 +1,19 @@
-#!/bin/bash
+#!/bin/sh
 set -e
 
 # list taken from https://rubygems.org/stats
-gems=(
-	thor
-	rake
-	rails
-	rack
-	activesupport
-	activerecord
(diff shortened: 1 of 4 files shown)


test/tests/ruby-gems/container.sh
```diff
@@ -1,20 +1,20 @@
...
+	activesupport \
+	activerecord \
+	actionpack \
+	json \
+	actionmailer \
+	activeresource \
 
-for gem in "${gems[@]}"; do
+
+for gem; do
```
yosifkit: Looks like this was originally over-complicated; maybe simplify by just putting the list directly into the for loop?

yosifkit:
LGTM, ping @tianon

tianon:
LGTM

GitHub doesn't notify when a PR is updated, so feel free to comment explicitly next time. :+1:

PR #1327, opened 2016-01-09
Note where arm64 builds are going now

tianon:
Many thanks to @hqhq for donating the namespace to our cause. :smile: :metal: :+1:


hqhq:
Ahh~~ lgtm, thanks :smile:

PR #1334, opened 2016-01-12
Add GlassFish 4.1.1 image from Oracle

brunoborges:
The AWS team received a request to add GlassFish 4.1.1 on December 9th, 2015 but until now the image is not available. A PR exists:

- https://github.com/aws/aws-eb-glassfish-dockerfiles/issues/3
- https://github.com/aws/aws-eb-glassfish-dockerfiles/pull/2

Oracle folks would like to start maintaining the top-level GlassFish image based on OpenJDK and Oracle Linux. This PR adds 4.1.1 version and directs 'latest' tag to this new version.

### DIFF
--- a/library/glassfish
+++ b/library/glassfish
@@ -1,8 +1,12 @@
-# maintainer: Amazon Web Services <https://aws.amazon.com/contact-us/> (@aws)
+# Maintainer: Bruno Borges <bruno.borges@oracle.com> (@brunoborges)
 
+latest: git://github.com/oracle/docker-images@5ce9431c6ebe5777924c29acdfc79758c42ca9f3 GlassFish/4.1.1
+4.1.1: git://github.com/oracle/docker-images@5ce9431c6ebe5777924c29acdfc79758c42ca9f3 GlassFish/4.1.1
+4.1.1-web: git://github.com/oracle/docker-images@5ce9431c6ebe5777924c29acdfc79758c42ca9f3 GlassFish/4.1.1-web
+
+# Deprecated images
 4.0-jdk7: git://github.com/aws/aws-eb-glassfish@aafcfc5e812dfb9b998105d3ca9da1b7f10664e1 4.0-jdk7
 4.0: git://github.com/aws/aws-eb-glassfish@aafcfc5e812dfb9b998105d3ca9da1b7f10664e1 4.0-jdk7
 
 4.1-jdk8: git://github.com/aws/aws-eb-glassfish@aafcfc5e812dfb9b998105d3ca9da1b7f10664e1 4.1-jdk8
 4.1: git://github.com/aws/aws-eb-glassfish@aafcfc5e812dfb9b998105d3ca9da1b7f10664e1 4.1-jdk8
-latest: git://github.com/aws/aws-eb-glassfish@aafcfc5e812dfb9b998105d3ca9da1b7f10664e1 4.1-jdk8
--- a/library/glassfish
+++ b/library/glassfish
@@ -3,10 +3,3 @@
 latest: git://github.com/oracle/docker-images@5ce9431c6ebe5777924c29acdfc79758c42ca9f3 GlassFish/4.1.1
 4.1.1: git://github.com/oracle/docker-images@5ce9431c6ebe5777924c29acdfc79758c42ca9f3 GlassFish/4.1.1
 4.1.1-web: git://github.com/oracle/docker-images@5ce9431c6ebe5777924c29acdfc79758c42ca9f3 GlassFish/4.1.1-web
-
-# Deprecated images
-4.0-jdk7: git://github.com/aws/aws-eb-glassfish@aafcfc5e812dfb9b998105d3ca9da1b7f10664e1 4.0-jdk7
-4.0: git://github.com/aws/aws-eb-glassfish@aafcfc5e812dfb9b998105d3ca9da1b7f10664e1 4.0-jdk7
-
-4.1-jdk8: git://github.com/aws/aws-eb-glassfish@aafcfc5e812dfb9b998105d3ca9da1b7f10664e1 4.1-jdk8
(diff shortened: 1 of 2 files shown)


psftw:
``FROM oracle/openjdk`` won't fly because an official image can't depend on a non-official image.  I think this is mentioned in the documentation, but will double check later to make sure.  You'll also want to remove the deprecated tags from the library/glassfish file -- they'll remain on DockerHub but won't be advertised in the generated tag list at the top of the documentation there.

I don't have any major objections to switching the base image from Debian to Oracle Linux, so long as we are clear in the docs that the image is built with OpenJDK on Oracle Linux as of 4.1.1 rather than reusing the existing ``java`` official repository which consists of OpenJDK on Debian.  I'm curious of the reasoning behind the switch.  Recently, you were arguing that the ``java`` images should be based on Alpine due to the reduced size.

psftw:
The downloaded artifact should also be verified somehow, per the documentation.  I'm not sure why an admin password is set globally and encoded in an environment variable.  I'm not a glassfish expert, so some extra context on these changes and how they affect usage would be helpful in order to do some basic testing.

brunoborges:
Hi Peter @psftw, yes I argued that the image behind repository "java" should be based on Alpine to be as small as possible (in fact, I even went further and said that some tags should be dropped because of very low adoption, like JREs tags).

We chose Oracle Linux because if this is to be an "official" image managed by Oracle, it then must be based on a Linux image that is maintained by Oracle, with an OpenJDK package built by Oracle (the one that goes into Oracle Linux yum repository).

About the password, we set it so a developer can use this image OOTB in different ways (example: the current image created by AWS does not allow user to access the web console if port 4848 is binded to the host, because GlassFish denies access to it from a different address, host address, than the IP the server runs on: container address). If a user tries to access http://localhost:4848, it will get the following message:

        **Configuration Error**
        Secure Admin must be enabled to access the DAS remotely. 

That's the reason we set a default password. If a user wants to set its own password, it is possible by extending the image and calling **asadmin** to modify it. This is no different than running WildFly (which comes with empty password for the admin bits).

If there is anything I must change on our image, let me know.

psftw:
Thanks for the clarification.  As for things that would need to change, you'll definitely want to verify the content of the download in the Dockerfile and remove the deprecated tags from this PR.  Please ping me when ready so I can dig a little deeper on the image.

> If a user wants to set its own password, it is possible by extending the image and calling asadmin to modify it.
I wonder if it makes sense to enable users to pass in an environment variable to the image and have an entrypoint script handle updating the admin password?  Are there other configuration options that would be convenient to process in an entrypoint script for convenience?  Whatever we go with here, I just want there to be clear documentation on the common use cases (i.e. setting custom admin password).

brunoborges:
I'd rather leave to GlassFish users the responsibility of configuring the GlassFish domain with **asadmin** in their extended images than adding a bunch of options to be passed as environment variables. This image is intended to be usable as quick as possible, but for anything beyond quick deployments, users will configure the domain with asadmin anyway (deploying artifacts, setting up DB connections, JMS queues, Security Realms, Network settings, etc... including setting a proper admin password).

brunoborges:
I've also included an MD5 check:

 - https://github.com/oracle/docker-images/commit/ebf286d21d9000439b62448949f094179fab0267

brunoborges:
I wanted to add that the proposed image, although based on Oracle Linux, is in fact smaller (~141.8 MB) than the current one.

brunoborges:
@psftw hey peter, any updates?

brunoborges:
Ping ?

brunoborges:
@psftw ?

brunoborges:
@psftw ?

psftw:
Please refactor `FROM oracle/openjdk:8` so that it doesn't depend on a non-official image -- I think you were going to make it `FROM oraclelinux:7` and install a pinned java openjdk package from Oracle repositories?  Other than that, I don't have a good way to test this image but don't see any other major issues with the Dockerfile.  I'd be up for doing some sanity testing with a good sample application WAR file that flexes a number of APIs if available.  Looking forward to the docs PR as well.  Thanks

yosifkit:
Oh, yay! I just noticed that there were updates on your [`Dockerfile`](https://github.com/oracle/docker-images/commit/4348bf131e55b2d1b3ddb9b1f5aac493d3be6674).  Just a few items of feedback:
- `FROM` should be a specific version of `oraclelinux` like `7`, rather than `latest`
- `--no-check-certificate` is superfluous on the http connection
- what is the license for the oracle/docker-images repo?
- can we get a PR to https://github.com/docker-library/docs?
  - this will fill in https://hub.docker.com/_/glassfish/

ping @tianon and @psftw for their feedback as well

asziranyi:
@psftw @yosifkit @brunoborges @tianon Hi Gentlemen! Is there any update on this issue? Thanks.

brunoborges:
You can find it here: https://hub.docker.com/r/glassfish/server/

asziranyi:
@brunoborges Yes, I know that and we use it already, but the official repository update would be also cool. Thank you!

brunoborges:
I say the official repository is "glassfish/server".

This one here is not "official" to me, as no one associated with the
GlassFish project has control over. I'd call it at most  a "top-level"
repository.

Glad you are already using "glassfish/server". :-)

*Bruno Borges*
+1 (650) 454-9628
*www.brunoborges.com <http://www.brunoborges.com>*

asziranyi:
Ok, agreed.

The meaning "official" is different from Docker and Oracle perspective.

Thanks,

Andras

psftw:
Sorry for the confusion.  The "glassfish" official image was abandoned by @zhengyal and then @brunoborges decided to work on "glassfish/server" instead.  This issue is no longer needed, and we are overdue on deprecating/removing the currently unmaintained "glassfish" image.  I'll take responsibility for helping that along.

PR #1372, opened 2016-01-26
Adding DoneJS Docker files

Macrofig:
Hello! We have docker files for [DoneJS](http://donejs.com/) that I'd like to add to the main library.

### DIFF
--- a/library/centos
+++ b/library/centos
@@ -1,10 +1,10 @@
 # maintainer: The CentOS Project <cloud-ops@centos.org> (@CentOS)
-# Build date 20151215
+# Build date 20151223
 
 # CentOS 7 rolling builds
-latest: git://github.com/CentOS/sig-cloud-instance-images@93117754c2a3de3c91ab563022f592e7940d293a docker
-centos7: git://github.com/CentOS/sig-cloud-instance-images@93117754c2a3de3c91ab563022f592e7940d293a docker
-7: git://github.com/CentOS/sig-cloud-instance-images@93117754c2a3de3c91ab563022f592e7940d293a docker
+latest: git://github.com/CentOS/sig-cloud-instance-images@e72a549845ecce5563d15e600f1e735a71095dcf docker
+centos7: git://github.com/CentOS/sig-cloud-instance-images@e72a549845ecce5563d15e600f1e735a71095dcf docker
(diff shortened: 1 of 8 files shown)


Macrofig:
Hello, I'm sure the team is busy but just wanted to check in and ask if there's anyone available to review this PR? Thanks a bunch!

tianon:
Hey @Macrofig, sorry for the delay -- we've definitely been pretty heavily backlogged. :disappointed:

I've taken a look, and the Dockerization here looks sane, but I'm a little iffy on whether DoneJS makes sense as an official image, given that it's a JavaScript framework and not a stand-alone tool/service, right?  I'm already gunning to deprecate `rails` (https://github.com/docker-library/docs/pull/503, and I think `django` will be my next target), since it's usually specified in the `Gemfile` of a project that uses it (making `ruby` a more appropriate base there), and the same goes for `django` and `requirements.txt` (making `python` a more appropriate base there).  Would the same be true here regarding the way most users will consume DoneJS (ie, explicit version requirements added to `package.json`)?

To put it another way, what's the benefit provided to the user by having an explicit top-level `donejs` image?

psftw:
Agree, I think we need to avoid images which are trivially derived from existing base images.  Also agree about django and rails -- would rather contribute some guidance on working with Docker to the upstream documentation of those projects. Thank you for putting this together, but I'm recommending we close it.

Macrofig:
Thanks for the input, I know your team is super busy so it is much appreciated!

I completely understand and agree. In fact, I think there is a bit of confusion because I saw the trends of frameworks being added to the official images and thought this was the way to go.  DoneJS is easily installed via NPM. Although having an official image would make it easier to deploy a DoneJS project,  we could support a repo on Docker Hub and achieve the same goals. So, no biggie on our end. :)

Regarding deprecating framework type images, perhaps it is possible to include this info somewhere in the project? Would the Readme be a good place for that? I'd be happy to create an issue/PR. Let me know what you all think. And thanks for the feedback!

psftw:
@Macrofig on paper today we are a bit ambiguous about what constitutes a good fit for the program, and it's not an easy thing to define either.  The bullet points from https://docs.docker.com/docker-hub/official_repos/ are the closest we have, but it doesn't mention frameworks explicitly.  Let's have the discussion in this repository, and I can get the docs.docker.com page updated based on it (since it seems to have been moved to a closed source repo).

PR #1391, opened 2016-01-28
Update registry:latest to point to registry 2.x

aaronlehmann:
The registry:latest tag currently points to registry 0.9.1. It has been
pointing to this old version for awhile to avoid breaking users
following outdated legacy registry documentation. Registry 2.x is
well-established now, and the time has come to change the latest tag.
Change it so that registry:latest is synonymous with registry:2.

Remove old 0.8.1 and 0.9.1 tags from the file.

cc @docker/distribution-maintainers

### DIFF
--- a/library/registry
+++ b/library/registry
@@ -3,10 +3,7 @@
 # maintainer: Richard Scothern <richard.scothern@gmail.com> (@RichardScothern)
 # maintainer: Aaron Lehmann <aaron.lehmann@docker.com> (@aaronlehmann)
 
-latest: git://github.com/docker/docker-registry@0.9.1
-0.8.1: git://github.com/docker/docker-registry@0.8.1
-0.9.1: git://github.com/docker/docker-registry@0.9.1
-
 2: git://github.com/docker/distribution-library-image@3aadce6abde20374ff286d444b24a0ac8949ba65
 2.2: git://github.com/docker/distribution-library-image@3aadce6abde20374ff286d444b24a0ac8949ba65
 2.2.1: git://github.com/docker/distribution-library-image@3aadce6abde20374ff286d444b24a0ac8949ba65
+latest: git://github.com/docker/distribution-library-image@3aadce6abde20374ff286d444b24a0ac8949ba65


yosifkit:
If `0.8.1` and `0.9.1` are not getting any updates, they might as well dropped from the list.  This will not prevent them from being pulled as the images will still exist on the Docker Hub.  This will remove them from the "Supported tags and respective Dockerfile links" section and would prevent them from being rebuilt when `ubuntu` images are updated.

I think this will need to be addressed as well: https://github.com/docker-library/official-images/pull/643#issuecomment-93818039.

dmp42:
Thanks @yosifkit 

The 0.X won't get any update in the future, so, I guess we can drop them.

Also, we can continue our previous conversation here: https://github.com/docker/distribution/pull/504#issuecomment-130054060

aaronlehmann:
Updated to remove references to 0.8.1 and 0.9.1.

RichardScothern:
LGTM.  @yosifkit : can you merge so I can update for the new registry release please?

tianon:
As agreed upon in https://github.com/docker-library/official-images/pull/643#issuecomment-93818039, `latest` should not change until the consistency issue is addressed.  I'm happy to help address it, and am willing to do the work to maintain the script in the long run if issues with it or new requirements come up.  I'm also willing to debate [the standards for acceptance](https://github.com/docker-library/official-images/blob/master/README.md#consistency) if there's a good discussion to be had there, but that is a separate, orthogonal discussion that is IMO not appropriate to be having in this context -- we should open a new issue to discuss that.

The best course of action to get 2.3 built and released ASAP is going to be to put this PR on the back-burner for now and just update the `2.x` references, returning to this PR once the agreed-upon modifications are in place to make this acceptable.

RichardScothern:
@tianon : Let's leave that discussion for now.  I will revert the change to `latest` in #1418 for this release.  Is that OK?

tianon:
@RichardScothern that sounds perfect, thanks! :+1: (I think we cross-commented at roughtly the same time :smile:)

RichardScothern:
@yosifkit : we would like to base the registry image on Alpine linux.  As we ship a bare binary, there is little need to have a few hundred megabytes of extra stuff.

Given that Alpine does not come installed with bash this does not meet the ability to run with bash requirement.  Like swarm, cirros, busybox etc, can we drop this requirement to reclaim the latest tag for the registry?

yosifkit:
Alpine Linux would not remove that requirement.  There is still a shell and many other programs available.  More info in https://github.com/docker/distribution/pull/504#issuecomment-176982357. Let's continue the discussion there.

tianon:
Since this is still open, it's worth noting that https://github.com/docker-library/official-images/pull/1838 added `2.5.0-rc.1` which fixes this issue, so I'm guessing `latest` will return when `2.5.0` goes GA? :smile: :tada: :balloon:

RichardScothern:
Correct.  I didn't close this since we're still in rc. 👍

yosifkit:
This was fixed in #1991. :tada:

PR #1420, opened 2016-02-06
Fix suggested "gpg" usage to stop relying on deprecated and insecure behavior

tianon:
cc @aburch -- does this look at least _more_ correct? :innocent: (anything we could do to make this pattern even better?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ Following the Docker guidelines it is highly recommended that the resulting imag
 
 Here is a snippet of a Dockerfile to add in tini (be sure to use it in `CMD` or `ENTRYPOINT` as appropriate):
 
-```dockerfile
+```Dockerfile
 # grab tini for signal processing and zombie killing
 ENV TINI_VERSION v0.9.0
 RUN set -x \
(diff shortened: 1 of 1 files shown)


tianon:
(Wanna fix it properly here at the source before replicating outwards.)

README.md
```diff
@@ -122,7 +122,7 @@ RUN set -x \
 	&& curl -fSL "https://github.com/krallin/tini/releases/download/$TINI_VERSION/tini.asc" -o /usr/local/bin/tini.asc \
 	&& export GNUPGHOME="$(mktemp -d)" \
```
md5: If this `GNUPGHOME` export is a best practice, it seems like it should be in the `python.tar.xz.asc` example below as well.

tianon: Yeah, that's a good point -- I was going to update that separately, but might as well include it here too.

tianon: Updated. :smile:

md5: :+1:

yosifkit:
LGTM

tianon:
I just realized we probably ought to link to the source of _where_ this is pointed out to be "insecure"...  :innocent:

From `gpg(1)` (http://manpages.debian.org/cgi-bin/man.cgi?query=gpg&manpath=Debian+unstable+sid):

>        --verify
>
>     ...
>
>               Note: If the option --batch is not used, gpg may assume  that  a
>               single  argument is a file with a detached signature and it will
>               try to find a matching data file by stripping certain  suffixes.
>               Using  this historical feature to verify a detached signature is
>               strongly discouraged; always specify the data file too.

PR #1423, opened 2016-02-09
WIP: adding notary server and signer official images

endophage:
Working on docs over here https://github.com/docker-library/docs/pull/483

Signed-off-by: David Lawrence <david.lawrence@docker.com> (github: endophage)

### DIFF
--- a/library/notary-server
+++ b/library/notary-server
@@ -1,4 +1,0 @@
-# maintainer: David Lawrence <david.lawrence@docker.com> (@endophage)
-
-latest: git://github.com/docker/notary-official-images@7dc2105a67a3532f9f470f18d1c8343594341893 notary-server
-0.2: git://github.com/docker/notary-official-images@7dc2105a67a3532f9f470f18d1c8343594341893 notary-server
--- a/library/notary-server
+++ b/library/notary-server
@@ -1,0 +1,4 @@
+# maintainer: David Lawrence <david.lawrence@docker.com> (@endophage)
+
+latest: git://github.com/docker/notary-official-images@7dc2105a67a3532f9f470f18d1c8343594341893 notary-server
+0.2: git://github.com/docker/notary-official-images@7dc2105a67a3532f9f470f18d1c8343594341893 notary-server
--- a/library/notary-signer
+++ b/library/notary-signer
@@ -1,0 +1,4 @@
+# maintainer: David Lawrence <david.lawrence@docker.com> (@endophage)
+
+latest: git://github.com/docker/notary-official-images@7dc2105a67a3532f9f470f18d1c8343594341893 notary-signer
+0.2: git://github.com/docker/notary-official-images@7dc2105a67a3532f9f470f18d1c8343594341893 notary-signer
--- a/library/notary-signer
+++ b/library/notary-signer
@@ -1,4 +1,0 @@
-# maintainer: David Lawrence <david.lawrence@docker.com> (@endophage)
-
-latest: git://github.com/docker/notary-official-images@7dc2105a67a3532f9f470f18d1c8343594341893 notary-signer
-0.2: git://github.com/docker/notary-official-images@7dc2105a67a3532f9f470f18d1c8343594341893 notary-signer
--- a/library/notary
+++ b/library/notary
@@ -1,0 +1,6 @@
+# maintainer: David Lawrence <david.lawrence@docker.com> (@endophage)
+
+server: git://github.com/docker/notary-official-images@3ea5b74928929415f2faaa5df574c4745bf6d286 notary-server
+server-0.2.0: git://github.com/docker/notary-official-images@3ea5b74928929415f2faaa5df574c4745bf6d286 notary-server
+signer: git://github.com/docker/notary-official-images@3ea5b74928929415f2faaa5df574c4745bf6d286 notary-signer
+signer-0.2.0: git://github.com/docker/notary-official-images@3ea5b74928929415f2faaa5df574c4745bf6d286 notary-signer


library/notary-server
```diff
@@ -0,0 +1,4 @@
+# maintainer: David Lawrence <david.lawrence@docker.com> (@endophage)
+
+latest: git://github.com/docker/notary-official-images@7dc2105a67a3532f9f470f18d1c8343594341893 notary-server
+0.2: git://github.com/docker/notary-official-images@7dc2105a67a3532f9f470f18d1c8343594341893 notary-server
```
diogomonica: Are we tagging this as 0.2?

library/notary-signer
```diff
@@ -0,0 +1,4 @@
+# maintainer: David Lawrence <david.lawrence@docker.com> (@endophage)
+
+latest: git://github.com/docker/notary-official-images@7dc2105a67a3532f9f470f18d1c8343594341893 notary-signer
+0.2: git://github.com/docker/notary-official-images@7dc2105a67a3532f9f470f18d1c8343594341893 notary-signer
```
diogomonica: Should we wait to tag with the final hash? Or just update it later?

yosifkit:
Ok, a few notes on the images:
- the files overwrite each other on `COPY`, so the image fails to even run with `Cannot mkdir: /notary/server is not a directory`

    > If \<dest> does not end with a trailing slash, it will be considered a regular file and the contents of \<src> will be written at \<dest>. --[docs.docker](https://docs.docker.com/engine/reference/builder/#copy).

  - maybe add a trailing slash to the COPYs or do a mkdir first (putting `WORKDIR` before the `COPY` did not help)
- since the image contains a full distro, it will need to be consistent with the rest of the official images: https://github.com/docker-library/official-images#consistency
- I would rather have `notary:server`, `notary:0.2-server`, `notary:signer`, and `notary:0.2-signer`, since they are all the same project and does not pollute the image namespace with duplicates.  I am not sure which variant would be `notary:0.2` and `notary:latest`.

endophage:
@yosifkit thanks for the feedback and apologies for the problems. I should probably have marked these as WIP. I'll make the improvements you've outlined.

My one concern with putting server and signer in the same hub repo is that they can operate entirely independently if configured to do so and it prevents there being a "latest" which will get pulled by default should not tag be provided.

tianon:
You could forgo having a "latest" tag at all if that helps -- it's not
unprecedented and will help folks who run it without a tag realize they out
to probably check out the docs and see what it is they actually need to
run. :+1:

- Tianon

endophage:
@tianon if there's already precedent for that then we can do that.

tianon:
Indeed, this wouldn't be the first, so let's plan on that. :+1:

If you need any help or advice in implementing the things @yosifkit mentioned, don't hesitate to reach out; we're happy to help further. :heart:

endophage:
@tianon @yosifkit Now that we've cut the notary 0.2.0 release, the time has come for me to wrap this up. I think I've covered your fixes and suggestions and I've updated the separate docs PR to reflect the single "notary" repository.

Let me know what other changes you'd like me to make.

tianon:
Image is looking pretty good, just two minor comments I've got:

1. probably want to remove `-latest` from the tag names here
    - `docker run notary:server` vs `docker run notary:server-latest`, which mirrors `docker run image:version` vs `docker run image` (implied `latest`)
2. will want to give https://github.com/docker-library/official-images/blob/master/README.md#consistency another once-over -- you'll either need to add a wrapper script of some kind, or switch from `alpine` to `scratch` :+1: (happy to help write something that makes sense for the former if needed)

endophage:
Sounds good. We want to use alpine because it's useful to be able to `exec sh` into a server container to debug.

What's the best way to get your gosu command into a container? I've seen a couple of examples of people just using wget but that seems less than ideal.

tianon:
Yeah, I definitely agree that it helps for debuggability.

Unfortunately, that's probably the best way if you want `gosu`; something like the following works pretty well (and includes PGP verification):
```dockerfile
# grab gosu for easy step-down from root
RUN set -x \
	&& curl -o /usr/local/bin/gosu -fSL "https://github.com/tianon/gosu/releases/download/1.7/gosu-$(dpkg --print-architecture)" \
	&& curl -o /usr/local/bin/gosu.asc -fSL "https://github.com/tianon/gosu/releases/download/1.7/gosu-$(dpkg --print-architecture).asc" \
	&& gpg --keyserver ha.pool.sks-keyservers.net --recv-keys B42F6819007F00F88E364FD4036A9C25BF357DD4 \
	&& gpg --batch --verify /usr/local/bin/gosu.asc /usr/local/bin/gosu \
	&& rm /usr/local/bin/gosu.asc \
	&& chmod +x /usr/local/bin/gosu
```

If `notary` doesn't need to write data anywhere on disk, you might be able to get away with just using `USER` for that, but I'm guessing there's some kind of on-disk storage involved too that'll need `chown`?

You might also be interested in https://github.com/docker-library/redis/pull/48 (and the linked issues, docker-library/rabbitmq#60, docker-library/cassandra#48, and docker-library/mongo#81) so that users of the image can take permissions into their own hands and use aribtrary `--user` values too. :smile:

endophage:
as these are meant to be prod containers, there's no writing to disk so `USER` should work fine. Will give it a go.

endophage:
@tianon I think the consistency stuff is fixed now, I've tested running various commands and it seems to be working correctly for things like `docker run ... sh`.

I've also updated the tags to remove the `-latest` part.

yosifkit:
Looking over the [docs](https://github.com/docker-library/docs/pull/483) and this should be good to go.
LGTM, Build test of #1423; 7dec03d (`notary`):

```console
$ bashbrew build "notary"
Cloning notary (git://github.com/docker/notary-official-images) ...
Processing notary:server ...
Processing notary:server-0.2.0 ...
Processing notary:signer ...
Processing notary:signer-0.2.0 ...
testing notary:server
	'utc' [1/4]...passed
	'override-cmd' [4/4]...passed
testing notary:signer
	'utc' [1/4]...passed
	'override-cmd' [4/4]...passed
```

tianon:
LGTM

PR #1436, opened 2016-02-12
Adds an official image for Consul.

slackpad:
Hi - I'm an employee at HashiCorp and we'd like to propose an official image for [Consul](https://www.consul.io/). The repo referenced here has some details about the tools we used:

https://github.com/hashicorp/docker-consul/tree/master/0.6

We've also got a first cut at our docs PR content here:

https://github.com/slackpad/docs/blob/official-consul-image/consul/content.md

Thanks for your feedback!

/cc @psftw

### DIFF
--- a/library/consul
+++ b/library/consul
@@ -1,5 +1,4 @@
 # maintainer: James Phillips <james@hashicorp.com> (@slackpad)
 
-0.6.3: git://github.com/hashicorp/docker-consul@bdbf63901c464fa7798eda1d19301be39e6069bb 0.6
-0.6.4: git://github.com/hashicorp/docker-consul@b1097f34f5b244eacfc6e05d1daf2c6e046afc39 0.6
-latest: git://github.com/hashicorp/docker-consul@b1097f34f5b244eacfc6e05d1daf2c6e046afc39 0.6
+0.6.4: git://github.com/hashicorp/docker-consul@d442f03c754b6faf1fc85c73e2713411ad0a2788 0.6
+latest: git://github.com/hashicorp/docker-consul@d442f03c754b6faf1fc85c73e2713411ad0a2788 0.6
--- a/library/consul
+++ b/library/consul
@@ -1,0 +1,4 @@
+# maintainer: James Phillips <james@hashicorp.com> (@slackpad)
+
+0.6.3: git://github.com/hashicorp/docker-consul@a452edd59700226b2df33d43046226baf574b5e0 0.6
+latest: git://github.com/hashicorp/docker-consul@a452edd59700226b2df33d43046226baf574b5e0 0.6
--- a/library/consul
+++ b/library/consul
@@ -1,4 +1,4 @@
 # maintainer: James Phillips <james@hashicorp.com> (@slackpad)
 
-0.6.3: git://github.com/hashicorp/docker-consul@a452edd59700226b2df33d43046226baf574b5e0 0.6
-latest: git://github.com/hashicorp/docker-consul@a452edd59700226b2df33d43046226baf574b5e0 0.6
+0.6.3: git://github.com/hashicorp/docker-consul@bdbf63901c464fa7798eda1d19301be39e6069bb 0.6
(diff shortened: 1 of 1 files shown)


psftw:
The latest iteration looks good to me.  I've got mixed feelings about redistributing dumb-init and gosu via your docker-base artifact, but I can see this strategy making more sense as other shared utilities are added over time.  Thanks for putting up with me so far.  I'll have some more comments on the doc PR when it's created -- overall really great job explaining things in the link above.

slackpad:
@psftw thanks for the update. You are correct about our motivation for docker-base - it's a good way to get the same level of trust for these small utilities as we have for our core app by building and signing them ourselves, and this will be convenient for adding similar things down the road or in images for our other projects.

slackpad:
@psftw I went ahead and opened the docs side of this as well so we can iterate on that while this is being reviewed.

kikitux:
@tianon any further comments?

tianon:
Sorry for the delay folks, our backlog is a bit out of control and we're working through it.

I'm definitely still having some confusion about what `docker-base.zip` is, and why it's necessary -- can you please elaborate?  It looks like it's just a single zip file which contains a binary for both `gosu` and `dumb-init`; have I understood that correctly?  I'm having a hard time understanding what the value is of doing it that way versus just downloading and verifying the tools directly from their associated sources, even in the context of "adding more tools" -- if one image needs more tools, that doesn't necessarily mean all of them do, does it?

slackpad:
Hi @tianon thanks for the feedback - I definitely understand what a Github backlog feels like :-)

You are correct - `docker-base.zip` is basically just a zip with those two binaries, though it's also signed. During earlier iterations this had some additional content like certificates but since we moved to Alpine that isn't necessary anymore. The remaining value of this now is that it lets us build and sign a release of these binaries so everything in the container above the base layer has the same level of release hygiene as `consul` itself. You've got signed releases for `gosu` but `dumb-init` does not, so it seemed easier to put all these extras into one place and build and sign them ourselves rather than rely on any individual projects. Other images would be able to pick only what they need for this, so it's definitely not our intention to always take everything from here in every image. I'd be happy to make the `Dockerfile` here more explicit and just take `gosu` and `dumb-init` so it's super explicit and wouldn't pick up extra things that may get added to `docker-base.zip` in the future.

tianon:
I can definitely appreciate the desire for signed artifacts (and we highly encourage them, ie https://github.com/docker-library/official-images/blob/master/README.md#security), but I don't see any signatures being downloaded or verified here -- was that just an oversight?  I think I can probably live with `docker-base.zip` if we verify both the sum and the signature, and make sure we only extract the bits we need per-image. :+1:

slackpad:
Ah good catch I verify the sun but not the sig. I'll add sig verification to docker base and consul and make the extract of docker base more selective.

tianon:
:metal:

You know the drill -- just give us a prod here when that's updated and we'll take a look. :smile:

slackpad:
Hi @tianon - updated the `Dockerfile` to pull the `SHA256SUM` files for each release along with their signatures and verify with the Hashicorp key. This makes maintenance of the `Dockerfile` much nicer as all we need to bump now is the version, there's no longer a pasted SHA in the `Dockerfile`. We take only the parts we need from `docker-base` as well.

slackpad:
Hi @tianon pinging this one up to see if you can get a chance to look this week. Thanks!

yosifkit:
:tada: :cake: :tada:  LGTM (pending docs update :smiley:)

cc @tianon
Build test of #1436; f8abd83 (`consul`):

```console
$ bashbrew build "consul"
Cloning consul (git://github.com/hashicorp/docker-consul) ...
Processing consul:v0.6.4 ...
Processing consul:latest ...
testing consul:v0.6.4
	'utc' [1/4]...passed
	'override-cmd' [4/4]...passed
```

slackpad:
Awesome - thanks for all the review feedback!

tianon:
LGTM

PR #1445, opened 2016-02-15
Added kaazing-gateway 5.0.1.46 release and Removed old Kaazing versio…

dpwspoon:
…n in favor of new version.  Support for KWIC added via configuration.

### DIFF
--- a/library/kaazing-gateway
+++ b/library/kaazing-gateway
@@ -1,4 +1,4 @@
 # maintainer: Kaazing Docker Maintainers <https://github.com/kaazing/gateway.docker/issues>
 
-latest: git://github.com/kaazing/gateway.docker@7a90d1f01df141a2a56b4ce0e9d337165cb8c677
-5.0.1.46: git://github.com/kaazing/gateway.docker@7a90d1f01df141a2a56b4ce0e9d337165cb8c677
+latest: git://github.com/kaazing/gateway.docker@d9630f76c2c3d201c7990f50ca553346cdc64c81
+5.0.1.46: git://github.com/kaazing/gateway.docker@d9630f76c2c3d201c7990f50ca553346cdc64c81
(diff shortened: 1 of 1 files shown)


yosifkit:
Sorry for the delay, I've been traveling the last week.

Changeset: https://github.com/kaazing/gateway.docker/compare/ab90b204bb...d9630f76c2c3d
That looks like a lot more than a version bump.  I'll try to dive into it in the morning to try and understand the additions (or feel free to jump in and explain).

From my initial pass, I would highlight https://github.com/docker-library/official-images#consistency in the change from `CMD` to `ENTRYPOINT`.

dpwspoon:
Thanks @yosifkit, apologies for the misleading PR title, things inadvertently got added together.  This updated the version and added support for a different type of configuration at runtime.  The [updated scripts work](https://github.com/kaazing/gateway.docker/issues/8) is generally done to make configuration of the gateway easier when wanting to use it for [KWIC](http://kaazing.com/kwic/); that's a light weight VPN like solution that runs ontop of websockets.  

I believe originally having the `CMD` vs `ENTRYPOINT` may always have been an error as the gateway.start script can take arguments such as `--config`.   I'll make the change using and entrypoint with a CMD and fallout to bash if the command is not recognized

yosifkit:
Is there a reason for overwriting the `gateway.start` that comes with Kaazing?  Couldn't this new `gateway.start` just be a new script that does an `exec` to the original `gateway.start`?

There appears to be quite a few variables defined at the top, but many seem to be unused.  Are they supposed to be environment defaults that the server will use?  Some might be better defined in the Dockerfile as ENV.

Also, the case line for `ambassador-server` is missing a `shift`. :wink:

https://github.com/docker-library/docs/pull/491 is the docs PR to go with this?

dpwspoon:
Thanks for the feedback, closing this for now as the script might get picked up by the main project. I'll open another one shortly to update the version

Issue #1448, opened 2016-02-16
glibc: CVE-2015-7547


tianon:
# CVE-2015-7547: glibc getaddrinfo stack-based buffer overflow

https://googleonlinesecurity.blogspot.no/2016/02/cve-2015-7547-glibc-getaddrinfo-stack.html

---

- [x] `alpine`: no glibc package available
- [x] `busybox`: https://github.com/docker-library/official-images/pull/1451
- [x] `centos` (RHEL derivative): https://github.com/docker-library/official-images/pull/1455
  - https://access.redhat.com/security/cve/CVE-2015-7547
- [ ] `crux`:
  - https://crux.nu/gitweb/?p=ports/core.git;a=history;f=glibc;hb=HEAD
- [x] `debian`: https://github.com/docker-library/official-images/issues/1450
  - https://security-tracker.debian.org/tracker/CVE-2015-7547
- [x] `fedora`: https://github.com/docker-library/official-images/pull/1456, https://github.com/docker-library/official-images/pull/1461
  - https://bugzilla.redhat.com/show_bug.cgi?id=CVE-2015-7547
- [x] `mageia`: https://github.com/docker-library/official-images/pull/1468
  - https://advisories.mageia.org/MGASA-2016-0079.html (https://lwn.net/Articles/676456/)
  - https://bugs.mageia.org/show_bug.cgi?id=17394
  - https://bugs.mageia.org/buglist.cgi?quicksearch=ALL%20CVE-2015-7547
- [x] `opensuse`: https://github.com/docker-library/official-images/pull/1460, https://github.com/docker-library/official-images/pull/1465
  - https://bugzilla.novell.com/show_bug.cgi?id=CVE-2015-7547
- [x] `oraclelinux`: https://github.com/docker-library/official-images/pull/1453
  - http://linux.oracle.com/cve/CVE-2015-7547.html
- [ ] `photon`:
  - https://github.com/vmware/photon/commit/fdf30faebdb4d1c253e5abbe39f1839536f1e5c0
  - https://github.com/vmware/photon/tree/master/SPECS/glibc
- [ ] `sourcemage`:
  - http://scmweb.sourcemage.org/?p=smgl/grimoire.git;a=history;f=libs/glibc
  - http://www.sourcemage.org/projects/grimoire/search?q=CVE-2015-7547
- [x] `ubuntu`: https://github.com/docker-library/official-images/pull/1454
[...]

PR #1468, opened 2016-02-22
Update Mageia 5 (especially for CVE-2015-7547)

tianon:
See also https://github.com/docker-library/official-images/issues/1448.

This also removes Mageia 4 (which was EOL on September 19th, 2015; https://www.mageia.org/en/support/#lifecycle).

cc @juanluisbaptiste -- it looks good; I just rebased your "dist" branch on master and squashed the two tarball-touching commits into a single one :+1: (https://github.com/juanluisbaptiste/docker-brew-mageia/commit/4da7cfe586ac0ef3fe3856d48d0410bf83c0cbe2)

PR #1469, opened 2016-02-22
Updated images glibc: CVE-2015-7547 #1448

juanluisbaptiste:
Updated images for glibc: CVE-2015-7547 #1448 and removed mageia 4 as it is EOL.


tianon:
Aha, changes the file in exactly the same way as #1468, so I'm merging. :+1:

Thanks @juanluisbaptiste!

PR #1454, opened 2016-02-17
Update ubuntu (especially for CVE-2015-7547)

tianon:
- `ubuntu:precise`: 20160217
- `ubuntu:trusty`: 20160217
- `ubuntu:wily`: 20160217
- `ubuntu:xenial`: 20160217.2

See also https://github.com/docker-library/official-images/issues/1448

PR #1485, opened 2016-02-28
ZDOCK-12 - Update php-zendserver images to patch against CVE2015-7547

davidl-zend:
Update php-zendserver images to patch against CVE2015-7547


yosifkit:
This [changeset](https://github.com/zendtech/php-zendserver-docker/compare/6590c650b46b02...500c6e7136deb) is just trying to update the `libc` from the base image; this was already accomplished when we updated the base image (https://github.com/docker-library/official-images/issues/1448, https://github.com/docker-library/official-images/pull/1454).

If there are vulnerabilities in your images that are because of the base image, then it should be fixed in the base image.

yosifkit:
This was fixed by the rebuild. Thanks!

PR #1553, opened 2016-03-18
Switch "php-ext-install" test to use POSIX instead of Bash (Alpine support)

tianon:


tianon:
See also https://github.com/docker-library/php/pull/202

yosifkit:
LGTM

Issue #1528, opened 2016-03-10
Document --privileged, and why its use is strongly discouraged


tianon:
We currently frown very, very heavily on `--privileged`, but the reasons why aren't really documented (I think they used to be part of the "`Dockerfile` Best Practices", but I can't seem to find any reference anymore -- might be that my search fu is rusty, but it still could do with a more explicit callout).

For reference (in case someone else wants to take this on), `--privileged` is essentially defined as "remove all restrictions" (especially cgroups, but caps as well).  In almost all cases, a mix of `--device` and `--cap-add` is much more appropriate (because it only adds the explicit permissions necessary for a particular application).

tianon:
To put this another way, there's absolutely no barriers in place to stop `--privileged` containers from modifying `/dev/sda1` or even more nefarious things like `/dev/kmem`.  At that point, we're barely even talking about a container anymore -- it's simply a partially namespaced process.

PR #1557, opened 2016-03-21
Discuss runtime privileges.

psftw:
Fixes https://github.com/docker-library/official-images/issues/1528

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,8 @@ This is one place that experience ends up trumping documentation for the path to
 
 #### Security
 
+##### Image Build
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -195,6 +197,12 @@ The `Dockerfile` should be written to help mitigate man-in-the-middle attacks du
 	    # install
 	```
 
+##### Runtime Configuration
+
+By default, Docker containers are executed with reduced privileges: whitelisted Linux capabilities, Control Groups, and a default Seccomp profile (1.10+ w/ host support).  Software running in a container may require additional privileges in order to function correctly, and there are a number of command line options to customize container execution. See [`docker run` Reference](https://docs.docker.com/engine/reference/run/) and [Seccomp for Docker](https://docs.docker.com/engine/security/seccomp/) for reference.
+
+Official Repositories that require additional privileges should specify the minimal set of command line options for the software to function, and nmay still be rejected if this introduces significant portability or security issues.  In general, `--privileged` is not allowed, but a combination of `--cap-add` and `--device` options may be acceptable.  Additionally, `--volume` can be tricky as there are many host filesystem locations that introduce portability/security issues (i.e. X11 socket).
```
yosifkit: `s/nmay/may/` :wink:

psftw:
Amended the typo, thanks @yosifkit

tianon:
Nice, LGTM :+1:

PR #1567, opened 2016-03-26
added NEM

rb2nem:
This PR add docker config file for running [NEM](http://nem.io)'s server (nis) and client (ncc).
NEM is number 12 of http://coinmarketcap.com/ and growing.
This work is based on unofficial images maintained at https://hub.docker.com/u/rb2nem/ , but with 
changes to follow the oficial images guidelines.
NEM is developed in Java and this image reuses the official Java images.
The images run only one process and each expose one port: 8989 for client and 7890 for server.
The main focus of these images is to make it easy to run a NIS node, and lower the barrier of
entry for developers interested in NEM.

### DIFF
--- a/library/nem
+++ b/library/nem
@@ -1,0 +1,5 @@
+# maintainer: rb2nem@gmail.com (@rb2nem)
+nis: git://github.com/rb2nem/docker-hub-nem@a0398ea3b132ee245ee0cbfa0f4f79cbab21190f nis
+ncc: git://github.com/rb2nem/docker-hub-nem@a0398ea3b132ee245ee0cbfa0f4f79cbab21190f ncc
+nis-0.6.74: git://github.com/rb2nem/docker-hub-nem@a0398ea3b132ee245ee0cbfa0f4f79cbab21190f nis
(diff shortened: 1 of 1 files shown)


rb2nem:
any feedback ?

psftw:
Thanks for submitting.  There are a few issues with the Dockerfile, but more importantly, I don't think we should start adding crypto-currency images to the program.  We've already strayed a bit from what I would consider "good" candidate images (namely, "drop-in solutions for popular programming language runtimes, data stores, and other services, similar to what a Platform-as-a-Service (PAAS) would offer." to quote https://docs.docker.com/docker-hub/official_repos/), though the majority do fit this description.  Beyond the philosophical mismatch, NEM (or Bitcoin, Litecoin, etc) as an official image would falsely be viewed as secure given the "Official Repository" flag, and the last thing I want is for someone to suffer financial loss due to a vulnerability in the image itself, a misconfiguration of Docker, or any number of other problems.  The NEM project can still create a "nem" user/organization account on DockerHub to redistribute official releases.

If you look around at existing official repository Dockerfiles, you'll see a number of conventions for optimizing layers.  Many of your RUN commands can be combined into one, which will optimize the image size and layer count significantly.

yosifkit:
Closing based on previous comment (https://github.com/docker-library/official-images/pull/1567#issuecomment-209014108).

PR #1641, opened 2016-04-17
Add an official image for Apache Storm

31z4:
As version 1.0.0 is a [huge milestone](http://storm.apache.org/2016/04/12/storm100-released.html) for Apache Storm it would be great to finally have an official image for it. See corresponding [docs PR](https://github.com/docker-library/docs/pull/550).

# Checklist for Review

  - https://github.com/docker-library/official-images/pull/1641#issuecomment-219223295
  - https://github.com/docker-library/docs/pull/550

### DIFF
--- a/library/storm
+++ b/library/storm
@@ -2,13 +2,13 @@ Maintainers: Elisey Zanko <elisey.zanko@gmail.com> (@31z4)
 GitRepo: https://github.com/31z4/storm-docker.git
 
 Tags: 0.9.7, 0.9
-GitCommit: f38117c7116512d45f1af04b514e9971e7214de8
+GitCommit: 93746fa3936afb3751565860632d3e49d53e9b0e
 Directory: 0.9.7
 
 Tags: 0.10.2, 0.10
-GitCommit: f38117c7116512d45f1af04b514e9971e7214de8
+GitCommit: 93746fa3936afb3751565860632d3e49d53e9b0e
 Directory: 0.10.2
 
 Tags: 1.0.2, 1.0, latest
-GitCommit: f38117c7116512d45f1af04b514e9971e7214de8
+GitCommit: 93746fa3936afb3751565860632d3e49d53e9b0e
 Directory: 1.0.2
--- a/library/storm
+++ b/library/storm
@@ -1,0 +1,6 @@
+# maintainer: Elisey Zanko <elisey.zanko@gmail.com> (@31z4)
+
+0.9.6: git://github.com/31z4/storm-docker@ba9faacec10470a4d16a779cdc448279cbf9e4df 0.9.6
+0.10.0: git://github.com/31z4/storm-docker@ba9faacec10470a4d16a779cdc448279cbf9e4df 0.10.0
+1.0.0: git://github.com/31z4/storm-docker@ba9faacec10470a4d16a779cdc448279cbf9e4df 1.0.0
+latest: git://github.com/31z4/storm-docker@ba9faacec10470a4d16a779cdc448279cbf9e4df 1.0.0
--- a/library/storm
+++ b/library/storm
@@ -1,6 +1,6 @@
 # maintainer: Elisey Zanko <elisey.zanko@gmail.com> (@31z4)
 
-0.9.6: git://github.com/31z4/storm-docker@eb42867e8e4f0d09b31c72d51dc31fcdcfac414e 0.9.6
-0.10.1: git://github.com/31z4/storm-docker@eb42867e8e4f0d09b31c72d51dc31fcdcfac414e 0.10.1
-1.0.1: git://github.com/31z4/storm-docker@eb42867e8e4f0d09b31c72d51dc31fcdcfac414e 1.0.1
-latest: git://github.com/31z4/storm-docker@eb42867e8e4f0d09b31c72d51dc31fcdcfac414e 1.0.1
+0.9.6: git://github.com/31z4/storm-docker@0f2664748a33162d95825c5f9ecb8444f61682b4 0.9.6
+0.10.1: git://github.com/31z4/storm-docker@0f2664748a33162d95825c5f9ecb8444f61682b4 0.10.1
+1.0.1: git://github.com/31z4/storm-docker@0f2664748a33162d95825c5f9ecb8444f61682b4 1.0.1
+latest: git://github.com/31z4/storm-docker@0f2664748a33162d95825c5f9ecb8444f61682b4 1.0.1
--- a/library/storm
+++ b/library/storm
@@ -1,6 +1,6 @@
 # maintainer: Elisey Zanko <elisey.zanko@gmail.com> (@31z4)
 
-0.9.6: git://github.com/31z4/storm-docker@a138393f9ef35c94e652d7c49c17285fe158c99b 0.9.6
-0.10.0: git://github.com/31z4/storm-docker@a138393f9ef35c94e652d7c49c17285fe158c99b 0.10.0
-1.0.0: git://github.com/31z4/storm-docker@a138393f9ef35c94e652d7c49c17285fe158c99b 1.0.0
-latest: git://github.com/31z4/storm-docker@a138393f9ef35c94e652d7c49c17285fe158c99b 1.0.0
+0.9.6: git://github.com/31z4/storm-docker@d050a2bd2f4b0aecfa007968e1d54448b15fa94d 0.9.6
+0.10.0: git://github.com/31z4/storm-docker@d050a2bd2f4b0aecfa007968e1d54448b15fa94d 0.10.0
+1.0.0: git://github.com/31z4/storm-docker@d050a2bd2f4b0aecfa007968e1d54448b15fa94d 1.0.0
+latest: git://github.com/31z4/storm-docker@d050a2bd2f4b0aecfa007968e1d54448b15fa94d 1.0.0
--- a/library/storm
+++ b/library/storm
@@ -1,6 +1,6 @@
 # maintainer: Elisey Zanko <elisey.zanko@gmail.com> (@31z4)
 
-0.9.6: git://github.com/31z4/storm-docker@d71c8addfcc86a38934349c46b28b806ea4c1c42 0.9.6
-0.10.1: git://github.com/31z4/storm-docker@d71c8addfcc86a38934349c46b28b806ea4c1c42 0.10.1
-1.0.1: git://github.com/31z4/storm-docker@d71c8addfcc86a38934349c46b28b806ea4c1c42 1.0.1
-latest: git://github.com/31z4/storm-docker@d71c8addfcc86a38934349c46b28b806ea4c1c42 1.0.1
+0.9.6: git://github.com/31z4/storm-docker@eb42867e8e4f0d09b31c72d51dc31fcdcfac414e 0.9.6
+0.10.1: git://github.com/31z4/storm-docker@eb42867e8e4f0d09b31c72d51dc31fcdcfac414e 0.10.1
+1.0.1: git://github.com/31z4/storm-docker@eb42867e8e4f0d09b31c72d51dc31fcdcfac414e 1.0.1
+latest: git://github.com/31z4/storm-docker@eb42867e8e4f0d09b31c72d51dc31fcdcfac414e 1.0.1
--- a/library/storm
+++ b/library/storm
@@ -1,6 +1,6 @@
 # maintainer: Elisey Zanko <elisey.zanko@gmail.com> (@31z4)
 
-0.9.6: git://github.com/31z4/storm-docker@0f2664748a33162d95825c5f9ecb8444f61682b4 0.9.6
-0.10.1: git://github.com/31z4/storm-docker@0f2664748a33162d95825c5f9ecb8444f61682b4 0.10.1
-1.0.1: git://github.com/31z4/storm-docker@0f2664748a33162d95825c5f9ecb8444f61682b4 1.0.1
-latest: git://github.com/31z4/storm-docker@0f2664748a33162d95825c5f9ecb8444f61682b4 1.0.1
+0.9.6: git://github.com/31z4/storm-docker@5fb2661adf5891ac5ccf26f927c73a7bc4b89187 0.9.6
+0.10.1: git://github.com/31z4/storm-docker@5fb2661adf5891ac5ccf26f927c73a7bc4b89187 0.10.1
+1.0.2: git://github.com/31z4/storm-docker@5fb2661adf5891ac5ccf26f927c73a7bc4b89187 1.0.2
+latest: git://github.com/31z4/storm-docker@5fb2661adf5891ac5ccf26f927c73a7bc4b89187 1.0.2
--- a/library/storm
+++ b/library/storm
@@ -1,6 +1,6 @@
 # maintainer: Elisey Zanko <elisey.zanko@gmail.com> (@31z4)
 
-0.9.6: git://github.com/31z4/storm-docker@d57881285dd264f2e194dacc120004c0cb4476d2 0.9.6
-0.10.1: git://github.com/31z4/storm-docker@d57881285dd264f2e194dacc120004c0cb4476d2 0.10.1
-1.0.1: git://github.com/31z4/storm-docker@d57881285dd264f2e194dacc120004c0cb4476d2 1.0.1
-latest: git://github.com/31z4/storm-docker@d57881285dd264f2e194dacc120004c0cb4476d2 1.0.1
+0.9.6: git://github.com/31z4/storm-docker@d71c8addfcc86a38934349c46b28b806ea4c1c42 0.9.6
+0.10.1: git://github.com/31z4/storm-docker@d71c8addfcc86a38934349c46b28b806ea4c1c42 0.10.1
+1.0.1: git://github.com/31z4/storm-docker@d71c8addfcc86a38934349c46b28b806ea4c1c42 1.0.1
+latest: git://github.com/31z4/storm-docker@d71c8addfcc86a38934349c46b28b806ea4c1c42 1.0.1
--- a/library/storm
+++ b/library/storm
@@ -1,6 +1,14 @@
-# maintainer: Elisey Zanko <elisey.zanko@gmail.com> (@31z4)
+Maintainers: Elisey Zanko <elisey.zanko@gmail.com> (@31z4)
+GitRepo: https://github.com/31z4/storm-docker.git
 
-0.9.6: git://github.com/31z4/storm-docker@5fb2661adf5891ac5ccf26f927c73a7bc4b89187 0.9.6
-0.10.1: git://github.com/31z4/storm-docker@5fb2661adf5891ac5ccf26f927c73a7bc4b89187 0.10.1
-1.0.2: git://github.com/31z4/storm-docker@5fb2661adf5891ac5ccf26f927c73a7bc4b89187 1.0.2
-latest: git://github.com/31z4/storm-docker@5fb2661adf5891ac5ccf26f927c73a7bc4b89187 1.0.2
+Tags: 0.9.7, 0.9
+GitCommit: f38117c7116512d45f1af04b514e9971e7214de8
+Directory: 0.9.7
+
+Tags: 0.10.2, 0.10
+GitCommit: f38117c7116512d45f1af04b514e9971e7214de8
+Directory: 0.10.2
+
+Tags: 1.0.2, 1.0, latest
+GitCommit: f38117c7116512d45f1af04b514e9971e7214de8
+Directory: 1.0.2
--- a/library/storm
+++ b/library/storm
@@ -1,6 +1,6 @@
 # maintainer: Elisey Zanko <elisey.zanko@gmail.com> (@31z4)
 
-0.9.6: git://github.com/31z4/storm-docker@d050a2bd2f4b0aecfa007968e1d54448b15fa94d 0.9.6
-0.10.0: git://github.com/31z4/storm-docker@d050a2bd2f4b0aecfa007968e1d54448b15fa94d 0.10.0
-1.0.0: git://github.com/31z4/storm-docker@d050a2bd2f4b0aecfa007968e1d54448b15fa94d 1.0.0
-latest: git://github.com/31z4/storm-docker@d050a2bd2f4b0aecfa007968e1d54448b15fa94d 1.0.0
+0.9.6: git://github.com/31z4/storm-docker@d57881285dd264f2e194dacc120004c0cb4476d2 0.9.6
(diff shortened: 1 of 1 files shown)


31z4:
Hey guys @psftw @tianon @yosifkit I've recently made some improvements to the image and the documentation. Any chance someone look at it and provide some feedback? Thanks!

tianon:
Hey @31z4, sorry for the delay!

A few thoughts:

1. `GPG_KEY` should be set to the full fingerprint (ie, `ACEFE18DD2322E1E84587A148DE03962E80B8FFD`), so that `gpg` can verify the key after it's fetched
2. `gpg --verify` should be `gpg --batch --verify` (see https://github.com/docker-library/official-images/pull/1420)
3. it's worth considering adding `../storm/bin` to the path such that users can expect `docker run storm storm --...` to work appropriately
4. is `/data` going to be used for mutable state of the application?  would it be appropriate to put it into a `VOLUME`?  (should users expect to need to back it up / will the application performance suffer if the data stored there is getting a performance hit for being on the CoW filesystem Docker provides?)
5. regarding `ENTRYPOINT`, you probably want to give https://github.com/docker-library/official-images/blob/master/README.md#consistency a read-through :smile:

31z4:
@tianon many thanks for your feedback!

1. Done.
2. Done.
3. Done.
4. `/data` can be used as a value for a `storm.local.dir` config param. This is where the Nimbus and Supervisor daemons save [their small amounts of state (like jars, confs, and things like that)](http://storm.apache.org/releases/1.0.0/Setting-up-a-Storm-cluster.html). I don't think it makes much sense to put it into a `VOLUME` by default as well as backing up this dir. I have no idea how CoW filesystem will affect Storm performance. Didn't find any benchmarks somehow related to that.
5. Done.

HeartSaVioR:
Great work!
Btw, maybe off-topic or not, I guess you may want to also address `drpc` (for supervisor nodes or just master node), `ui` (for master node), `logviewer` (for supervisor nodes) to docker-compose.
Please refer to https://github.com/wurstmeister/storm-docker.

tianon:
Nice! (https://github.com/31z4/storm-docker/compare/d050a2bd2f4b0aecfa007968e1d54448b15fa94d...d71c8addfcc86a38934349c46b28b806ea4c1c42)

Regarding GPG, I did `gpg --keyserver ha.pool.sks-keyservers.net --recv-key 8DE03962E80B8FFD`, and then `gpg --fingerprint`, which shows the full fingerprint. :+1:

Only comment I'd have left is that you probably want to add `CMD ["storm"]` so that users who do just `docker run storm` without a command get a running container (where `CMD` is just the default; `docker run storm some-other-command` will still work appropriately)! :+1:

31z4:
@tianon Unfortunately, we can't get working Storm with just a single command. At minimum we have to start the Nimbus and Supervisor daemons. Of course I can start those two using [`supervisor`](http://supervisord.org/). But I'd like to keep an image as simple as possible and avoid additional dependencies. What do you think? Does it make sense to have `CMD ["supervisor"]` which starts both Nimbus and Supervisor daemons inside a single container?

tianon:
Ah, naw -- that makes sense! :+1:  (https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#run-only-one-process-per-container)

I'm looking over the docs now, and I'm a little worried about the Zookeeper dependency.  We don't have an official `zookeeper` image yet (which you're aware of), but "No official images can be derived from, or depend on, non-official images." (https://github.com/docker-library/official-images/blob/master/README.md#repeatability)  Is there a way to run Storm without Zookeeper? (or perhaps do you know any members of the Zookeeper project who might be interested in collaborating on/maintaining an image? :smiling_imp:)

31z4:
@tianon There is no way to run Storm in cluster mode without Zookeeper. Though, Zookeeper is not required if you run Storm in [local mode](http://storm.apache.org/releases/1.0.1/Local-mode.html). I was also concerned about not having an official Zookeeper image. I planned to manage this after the Storm image. It seems that plan has changed 😃

tianon:
Oh, I suppose before we get too far down that rabbit hole, has there been any contact with either upstream to see if they're interested in collaboration on the image?  We like to make sure they're at least aware of what we're doing, especially since we've burned bridges by being too hasty without consulting upstream on whether they even wanted to be part of the official images, so we like to be a little more proactive than we've been in the past. :sweat_smile: :blush:

31z4:
@tianon [@ApacheStorm](https://twitter.com/ApacheStorm) retwitted my [tweet](https://twitter.com/e1z4/status/729887889333600256). I've also just created a [JIRA issue](https://issues.apache.org/jira/browse/STORM-1836). So I believe they're aware 😊

31z4:
@HeartSaVioR Thank you for the feedback! There are several reasons why I didn't include `drpc`,`ui` and `logviewer` into `docker-compose.yml`

1. I'd like to keep it as simple as possible and include only necessary services. Everything besides Zookeeper, Nimbus and Supervisor is optional.
2. Regarding `ui` and `logviewer` there are known issues ([one](https://issues.apache.org/jira/browse/STORM-1758), [two](https://issues.apache.org/jira/browse/STORM-1759)) which I'd like to address first to get a fully working dockerized `ui`. Have you ever encountered those issues?

harshach:
@31z4 ui and log viewer would be great to have in the docker image.

31z4:
@harshach as you run the `ui` using the same `storm` binary - it's already in the image. Here is an example how to run it assuming `nimbus` is serving on `nimbus-host`
    
      $ docker run -d -p 8080:8080 31z4/storm:1.0.1 storm ui -c nimbus.host=storm-nimbus

31z4:
@tianon One step closer #1765 😊

yosifkit:
Now that we have the zookeeper image, we can hopefully get this in soon!

The only concern I have is related to the docs where all the supporting containers are run with `--net container:other-container`; which removes some of the isolation provided by containers and prevents multi-host usage by tying all the containers to one host.  Could we swap the inline examples to be more like the compose example, using hostnames or IP addresses?  Using hostnames would require a docker network to be created, but that is what `kapacitor` does in their [docs](https://github.com/docker-library/docs/tree/3fc49079ae0fdfe5e5ac624cd36c5b0dc569d741/kapacitor#subscriptions).

31z4:
@yosifkit thanks for your feedback! I'll look into `kapacitor` docs. Also I'd like to make some improvements including using `docker-entrypoint.sh`.

tianon:
Am I understanding correctly that Storm requires both "nimbus" and "zookeeper" to function?  In the examples in the documentation, you seem to be using `-c` to supply both `storm.zookeeper.servers` and `nimbus.host`, presumably because those values default to `localhost` upstream?  Would it make sense to change those defaults in the `Dockerfile` so that users can still overwrite them, but so that by default, the image is slightly easier to use if you name or `--network-alias` your containers as `zookeeper` and `nimbus`?

tianon:
(See also https://github.com/docker-library/kibana/blob/b49429d8cf95f58a2ed1246b5ff24bbe0dd81911/4.6/Dockerfile#L50-L52 for an example in `kibana` of what I mean)

31z4:
@yosifkit @tianon guys, thanks for suggestions! I think the image is ready for the final review now.

abipit:
Hello @31z4, great job, thank you.

Just 2 points :
1) do you have the part for docker-compose to add a storm UI ?

2) I have a topology that works in other environment (storm with ambari) but my topology don't work in your environment.
_Error: A JNI error has occurred, please check your installation and try again
Exception in thread "main" java.lang.NoClassDefFoundError: backtype/storm/topology/IRichSpout_
I test a lot of thinks but without success, do you have an idea ?

Thanks

HeartSaVioR:
@abipit 
Regarding 2 you need to check your topology is ready to Storm 1.x since package base name is changed from `backtype/storm` to `org.apache.storm`.

abipit:
thank you @HeartSaVioR, great news, I will test with the storm container 0.10.2 !

For my first point, I didn't see this anwser with the command line to launch the UI, my bad :)

abipit:
For storm ui, I launch this command : 

docker run -d -p 8081:8080 --name=storm_ui 31z4/storm:0.10.2 storm ui -c nimbus.host=nimbus

But without success, in the UI, under the "Internal Server Error" section, I have : 
_java.lang.RuntimeException: org.apache.thrift7.transport.TTransportException: java.net.UnknownHostException: nimbus_

I made several tests but the storm UI is not correct. 
Do you have an idea ?
Thanks a lot

abipit:
It's ok for storm_UI but with the docker-compose, not with a command-line.

Because in command_line, the network for storm_UI was 172.18.0.0 while the 3 others containers created by the docker-compose, the network was 172.17.0.0

I put this code in the docker-compose : 
    storm_ui:
        image: 31z4/storm:0.10.2
        container_name: storm_ui
        command: storm ui -c nimbus.host=nimbus
        depends_on:
            - nimbus
            - zookeeper
        links:
            - nimbus
            - zookeeper
        restart: always
        ports:
            - 8081:8080

and it's OK now because all containers are in the same network 172.17....

abipit:
I found the solution to launch the storm_UI directly in command_line, I indicate that the network is the storm network :

`docker run -d -p 8090:8080 --name=storm_ui --net="storm_default" 31z4/storm:0.10.2 storm ui -c nimbus.host=nimbus`

31z4:
@HeartSaVioR @abipit thank you guys for your comments. Just added the Storm UI example into the [docs](https://github.com/docker-library/docs/pull/550).

tianon:
https://github.com/31z4/storm-docker/compare/5fb2661adf5891ac5ccf26f927c73a7bc4b89187...f38117c7116512d45f1af04b514e9971e7214de8 seems pretty sane to me! :+1: :heart:

yosifkit:
Dockerization seems sane to me. I have just a few questions about logging. Does it log to `STORM_LOG_DIR` by default?  The common case in most of the docker official images is to log to stdout so that logs can be viewed with `docker logs`; is this possible?  I think even just an example in the docs would be enough.

31z4:
@yosifkit Thanks for pointing this out. Found out that actually logs are not going to the `/logs` 😊 Will push a fix right away.

According to logging everything to stdout by default I think it doesn't make sense. Storm has pretty complex and meaningful [default logging configuration](https://github.com/apache/storm/tree/v1.0.2/log4j2).  The idea is to separate different log sources into different files to facilitate browsing and searching the logs. Logging also depends on the service you're running. For example `nimbus` has the following files in its `/logs`:

```
/logs/:
access-nimbus.log      access-web-nimbus.log  nimbus.log
```

While working `supervisor` has a bit more:

```
/logs/:
access-supervisor.log      access-web-supervisor.log  supervisor.log             workers-artifacts

/logs/workers-artifacts:
topology-1-1476256953

/logs/workers-artifacts/topology-1-1476256953:
6700  6701  6702  6703

/logs/workers-artifacts/topology-1-1476256953/6700:
gc.log.0.current    worker.log          worker.log.err      worker.log.metrics  worker.log.out      worker.pid          worker.yaml

/logs/workers-artifacts/topology-1-1476256953/6701:
gc.log.0.current    worker.log          worker.log.err      worker.log.metrics  worker.log.out      worker.pid          worker.yaml

/logs/workers-artifacts/topology-1-1476256953/6702:
gc.log.0.current    worker.log          worker.log.err      worker.log.metrics  worker.log.out      worker.pid          worker.yaml

/logs/workers-artifacts/topology-1-1476256953/6703:
gc.log.0.current    worker.log          worker.log.err      worker.log.metrics  worker.log.out      worker.pid          worker.yaml
```

Hope this makes sense. Will add a short note about logging in the image documentation.

yosifkit:
```diff
diff --git a/storm_0.10/Dockerfile b/storm_0.10/Dockerfile
new file mode 100644
index 0000000..c851d57
--- /dev/null
+++ b/storm_0.10/Dockerfile
@@ -0,0 +1,43 @@
+FROM openjdk:8-jre-alpine
+MAINTAINER Elisey Zanko <elisey.zanko@gmail.com>
+
+# Install required packages
+RUN apk add --no-cache \
+    bash \
+    python \
+    su-exec
+
+ENV STORM_USER=storm
+ENV STORM_CONF_DIR=/conf
+ENV STORM_DATA_DIR=/data
+ENV STORM_LOG_DIR=/logs
+
+# Add a user and make dirs
+RUN set -x \
+    && adduser -D "$STORM_USER" \
+    && mkdir -p "$STORM_CONF_DIR" "$STORM_DATA_DIR" "$STORM_LOG_DIR" \
+    && chown -R "$STORM_USER:$STORM_USER" "$STORM_CONF_DIR" "$STORM_DATA_DIR" "$STORM_LOG_DIR"
+
+ARG GPG_KEY=ACEFE18DD2322E1E84587A148DE03962E80B8FFD
+ARG DISTRO_NAME=apache-storm-0.10.2
+
+# Download Apache Storm, verify its PGP signature, untar and clean up
+RUN set -x \
+    && apk add --no-cache --virtual .build-deps \
+        gnupg \
+    && wget -q "http://www.apache.org/dist/storm/$DISTRO_NAME/$DISTRO_NAME.tar.gz" \
+    && wget -q "http://www.apache.org/dist/storm/$DISTRO_NAME/$DISTRO_NAME.tar.gz.asc" \
+    && export GNUPGHOME="$(mktemp -d)" \
+    && gpg --keyserver ha.pool.sks-keyservers.net --recv-key "$GPG_KEY" \
+    && gpg --batch --verify "$DISTRO_NAME.tar.gz.asc" "$DISTRO_NAME.tar.gz" \
+    && tar -xzf "$DISTRO_NAME.tar.gz" \
+    && chown -R "$STORM_USER:$STORM_USER" "$DISTRO_NAME" \
+    && rm -r "$GNUPGHOME" "$DISTRO_NAME.tar.gz" "$DISTRO_NAME.tar.gz.asc" \
+    && apk del .build-deps
+
+WORKDIR $DISTRO_NAME
+
+ENV PATH $PATH:/$DISTRO_NAME/bin
+
+COPY docker-entrypoint.sh /
+ENTRYPOINT ["/docker-entrypoint.sh"]
diff --git a/storm_0.10/docker-entrypoint.sh b/storm_0.10/docker-entrypoint.sh
new file mode 100755
index 0000000..a8aea91
--- /dev/null
+++ b/storm_0.10/docker-entrypoint.sh
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+set -e
+
+# Allow the container to be started with `--user`
+if [ "$1" = 'storm' -a "$(id -u)" = '0' ]; then
+    chown -R "$STORM_USER" "$STORM_CONF_DIR" "$STORM_DATA_DIR" "$STORM_LOG_DIR"
+    exec su-exec "$STORM_USER" "$0" "$@"
+fi
+
+# Generate the config only if it doesn't exist
+CONFIG="$STORM_CONF_DIR/storm.yaml"
+if [ ! -f "$CONFIG" ]; then
+    cat << EOF > "$CONFIG"
+storm.zookeeper.servers: [zookeeper]
+nimbus.seeds: [nimbus]
+storm.log.dir: "$STORM_LOG_DIR"
+storm.local.dir: "$STORM_DATA_DIR"
+EOF
+fi
+
+exec "$@"
diff --git a/storm_0.9/Dockerfile b/storm_0.9/Dockerfile
new file mode 100644
index 0000000..5830aee
--- /dev/null
+++ b/storm_0.9/Dockerfile
@@ -0,0 +1,43 @@
+FROM openjdk:8-jre-alpine
+MAINTAINER Elisey Zanko <elisey.zanko@gmail.com>
+
+# Install required packages
+RUN apk add --no-cache \
+    bash \
+    python \
+    su-exec
+
+ENV STORM_USER=storm
+ENV STORM_CONF_DIR=/conf
+ENV STORM_DATA_DIR=/data
+ENV STORM_LOG_DIR=/logs
+
+# Add a user and make dirs
+RUN set -x \
+    && adduser -D "$STORM_USER" \
+    && mkdir -p "$STORM_CONF_DIR" "$STORM_DATA_DIR" "$STORM_LOG_DIR" \
+    && chown -R "$STORM_USER:$STORM_USER" "$STORM_CONF_DIR" "$STORM_DATA_DIR" "$STORM_LOG_DIR"
+
+ARG GPG_KEY=ACEFE18DD2322E1E84587A148DE03962E80B8FFD
+ARG DISTRO_NAME=apache-storm-0.9.7
+
+# Download Apache Storm, verify its PGP signature, untar and clean up
+RUN set -x \
+    && apk add --no-cache --virtual .build-deps \
+        gnupg \
+    && wget -q "http://www.apache.org/dist/storm/$DISTRO_NAME/$DISTRO_NAME.tar.gz" \
+    && wget -q "http://www.apache.org/dist/storm/$DISTRO_NAME/$DISTRO_NAME.tar.gz.asc" \
+    && export GNUPGHOME="$(mktemp -d)" \
+    && gpg --keyserver ha.pool.sks-keyservers.net --recv-key "$GPG_KEY" \
+    && gpg --batch --verify "$DISTRO_NAME.tar.gz.asc" "$DISTRO_NAME.tar.gz" \
+    && tar -xzf "$DISTRO_NAME.tar.gz" \
+    && chown -R "$STORM_USER:$STORM_USER" "$DISTRO_NAME" \
+    && rm -r "$GNUPGHOME" "$DISTRO_NAME.tar.gz" "$DISTRO_NAME.tar.gz.asc" \
+    && apk del .build-deps
+
+WORKDIR $DISTRO_NAME
+
+ENV PATH $PATH:/$DISTRO_NAME/bin
+
+COPY docker-entrypoint.sh /
+ENTRYPOINT ["/docker-entrypoint.sh"]
diff --git a/storm_0.9/docker-entrypoint.sh b/storm_0.9/docker-entrypoint.sh
new file mode 100755
index 0000000..a8aea91
--- /dev/null
+++ b/storm_0.9/docker-entrypoint.sh
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+set -e
+
+# Allow the container to be started with `--user`
+if [ "$1" = 'storm' -a "$(id -u)" = '0' ]; then
+    chown -R "$STORM_USER" "$STORM_CONF_DIR" "$STORM_DATA_DIR" "$STORM_LOG_DIR"
+    exec su-exec "$STORM_USER" "$0" "$@"
+fi
+
+# Generate the config only if it doesn't exist
+CONFIG="$STORM_CONF_DIR/storm.yaml"
+if [ ! -f "$CONFIG" ]; then
+    cat << EOF > "$CONFIG"
+storm.zookeeper.servers: [zookeeper]
+nimbus.seeds: [nimbus]
+storm.log.dir: "$STORM_LOG_DIR"
+storm.local.dir: "$STORM_DATA_DIR"
+EOF
+fi
+
+exec "$@"
diff --git a/storm_latest/Dockerfile b/storm_latest/Dockerfile
new file mode 100644
index 0000000..972e3b3
--- /dev/null
+++ b/storm_latest/Dockerfile
@@ -0,0 +1,43 @@
+FROM openjdk:8-jre-alpine
+MAINTAINER Elisey Zanko <elisey.zanko@gmail.com>
+
+# Install required packages
+RUN apk add --no-cache \
+    bash \
+    python \
+    su-exec
+
+ENV STORM_USER=storm
+ENV STORM_CONF_DIR=/conf
+ENV STORM_DATA_DIR=/data
+ENV STORM_LOG_DIR=/logs
+
+# Add a user and make dirs
+RUN set -x \
+    && adduser -D "$STORM_USER" \
+    && mkdir -p "$STORM_CONF_DIR" "$STORM_DATA_DIR" "$STORM_LOG_DIR" \
+    && chown -R "$STORM_USER:$STORM_USER" "$STORM_CONF_DIR" "$STORM_DATA_DIR" "$STORM_LOG_DIR"
+
+ARG GPG_KEY=ACEFE18DD2322E1E84587A148DE03962E80B8FFD
+ARG DISTRO_NAME=apache-storm-1.0.2
+
+# Download Apache Storm, verify its PGP signature, untar and clean up
+RUN set -x \
+    && apk add --no-cache --virtual .build-deps \
+        gnupg \
+    && wget -q "http://www.apache.org/dist/storm/$DISTRO_NAME/$DISTRO_NAME.tar.gz" \
+    && wget -q "http://www.apache.org/dist/storm/$DISTRO_NAME/$DISTRO_NAME.tar.gz.asc" \
+    && export GNUPGHOME="$(mktemp -d)" \
+    && gpg --keyserver ha.pool.sks-keyservers.net --recv-key "$GPG_KEY" \
+    && gpg --batch --verify "$DISTRO_NAME.tar.gz.asc" "$DISTRO_NAME.tar.gz" \
+    && tar -xzf "$DISTRO_NAME.tar.gz" \
+    && chown -R "$STORM_USER:$STORM_USER" "$DISTRO_NAME" \
+    && rm -r "$GNUPGHOME" "$DISTRO_NAME.tar.gz" "$DISTRO_NAME.tar.gz.asc" \
+    && apk del .build-deps
+
+WORKDIR $DISTRO_NAME
+
+ENV PATH $PATH:/$DISTRO_NAME/bin
+
+COPY docker-entrypoint.sh /
+ENTRYPOINT ["/docker-entrypoint.sh"]
diff --git a/storm_latest/docker-entrypoint.sh b/storm_latest/docker-entrypoint.sh
new file mode 100755
index 0000000..a8aea91
--- /dev/null
+++ b/storm_latest/docker-entrypoint.sh
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+set -e
+
+# Allow the container to be started with `--user`
+if [ "$1" = 'storm' -a "$(id -u)" = '0' ]; then
+    chown -R "$STORM_USER" "$STORM_CONF_DIR" "$STORM_DATA_DIR" "$STORM_LOG_DIR"
+    exec su-exec "$STORM_USER" "$0" "$@"
+fi
+
+# Generate the config only if it doesn't exist
+CONFIG="$STORM_CONF_DIR/storm.yaml"
+if [ ! -f "$CONFIG" ]; then
+    cat << EOF > "$CONFIG"
+storm.zookeeper.servers: [zookeeper]
+nimbus.seeds: [nimbus]
+storm.log.dir: "$STORM_LOG_DIR"
+storm.local.dir: "$STORM_DATA_DIR"
+EOF
+fi
+
+exec "$@"
```

tianon:
LGTM

abipit:
@31z4 Great job, your image is now "Official" on the docker hub ;-) 👍 

About logs, is it possible to have the logviewer of supervisor on port 8000 ? 8000 is normally the default port, I tested to export it but without success, no page. It's for my developper, he don't have access on the server but he needs to consult the topology's log (in /logs/workers-artifacts/...).

Do you have an idea ? Thank you

31z4:
@abipit There are some [known issues](https://issues.apache.org/jira/browse/STORM-1342) with the `logviewer` in dockerized environments.

I see two major approaches here. The first is starting the `logviewer` in a separate container and using container linking and `--volumes-from`. And the second is starting both `logviewer` and `supervisor` inside a single container.

I'm thinking of creating an image variant with the `logviever` included. [RabbitMQ](https://hub.docker.com/_/rabbitmq/) does something similar with their `-managed` variant.

PR #1650, opened 2016-04-20
rethinkdb: Remove obsolete (unsupported) releases

stuartpb:
Per https://github.com/docker-library/official-images#library-definition-files, following the style used by Redis

Following discussion from https://github.com/rethinkdb/rethinkdb-dockerfiles/pull/27#issuecomment-212650916

cc @dalanmiller @danielmewes (please don't merge without review from at least one of these)

### DIFF
--- a/library/rethinkdb
+++ b/library/rethinkdb
@@ -1,37 +1,21 @@
 # maintainer: Daniel Alan Miller <dalanmiller@rethinkdb.com> (@dalanmiller)
 
-1.15.1: git://github.com/rethinkdb/rethinkdb-dockerfiles@0b7134d0d80d92ca461119056b1c534d61a2e3a8 jessie/1.15.1
-1.15.2: git://github.com/rethinkdb/rethinkdb-dockerfiles@0b7134d0d80d92ca461119056b1c534d61a2e3a8 jessie/1.15.2
 1.15.3: git://github.com/rethinkdb/rethinkdb-dockerfiles@0b7134d0d80d92ca461119056b1c534d61a2e3a8 jessie/1.15.3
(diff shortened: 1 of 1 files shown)


stuartpb:
@dalanmiller @danielmewes The assumption I went on in writing this patch is that RethinkDB supports each minor version going back to (at least) 1.15, but doesn't support running unpatched versions of those point-releases. Does that accurately reflect RethinkDB's support policy?

tianon:
LGTM! :+1:

dalanmiller:
I'm not sure what our support policy is @danielmewes, I"ll let you make the call.

tianon:
Maybe it would help to reframe the question a little bit: if there were a bug discovered in 1.15.3, would there be a 1.15.4?

danielmewes:
> Maybe it would help to reframe the question a little bit: if there were a bug discovered in 1.15.3, would there be a 1.15.4?

No. We have occasionally released bug fix releases for one major release before the current one, if a bug was particularly severe.
With our latest release being 2.3.0, we might still release a 2.2.7 in the future. But we will not release a further 2.1.x, 2.0.x, 1.16.x etc.

That being said, major releases do sometimes slightly break compatibility, and some applications might only work with a particular older version of RethinkDB (more precisely: with a particular X and Y in RethinkDB X.Y.z. A change in the z never breaks compatibility, so it should always be ok to only support the highest available z for a given X.Y).
I think we don't need to provide images for versions older than 2.0.x though.

yosifkit:
Thanks for the clarification @danielmewes.  @stuartpb, could you adjust the PR to drop the older  `1.X` versions as well?

PR #1649, opened 2016-04-20
Update docker-library images

tianon:
- `drupal`: 8.1 GA
- `mariadb`: add xtrabackup for Galera (docker-library/mariadb#47)
- `pypy`: 5.1.0
- `python`: empty `~/.cache` (docker-library/python#103)
- `redis`: use `su-exec` in Alpine variants (docker-library/redis#54)
- `rocket.chat`: 0.27.0

PR #1652, opened 2016-04-21
Update PyPy to 5.1

GaretJax:


yosifkit:
Thanks! We just merged this in https://github.com/docker-library/official-images/pull/1649. They have even finished pushing to the Docker Hub, so you can pull them! :tada:

PR #1677, opened 2016-04-27
travis: Allow check build and test pr

yongjhih:
Because travis has docker already

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: bash
+services: docker
 
 # allow for use of Docker-based workers
 sudo: false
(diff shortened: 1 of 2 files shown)


yosifkit:
We have not wanted to overload the Travis infrastructure with these builds. Some can take upwards of [an hour](https://travis-ci.org/docker-library/gcc/builds/74645801) to [build](https://travis-ci.org/docker-library/busybox/jobs/116950511) just one version of one image, which Travis would just stop the whole build after [50 minutes](https://docs.travis-ci.com/user/customizing-the-build/#Build-Timeouts).

We also often benefit from docker build cache on our local machines as well as the ability to docker inspect failing builds.

yongjhih:
Ok, I see, thanks for your explaining.

PR #1688, opened 2016-05-02
Jenkins latest should point to LTS, not 2.0

carlossg:
I jumped the gun making latest pointing to 2.0 when LTS is still in 1.651.1


tianon:
Whoops! :+1:

LGTM

PR #1694, opened 2016-05-03
Add correct tag mapping for node:5

chorrell:
`git pull node:5` should default to node:5.11.0.

This fixes nodejs/docker-node#174 where `git pull node:5` would grab the Node.js v5.10.1 image instead of the latest v5.11.0 image.

We fixed our generate-stackbrew-library.sh script in nodejs/docker-node#175 to default to the correct version.


tianon:
Nice, LGTM :+1:

PR #1721, opened 2016-05-09
Add Clear Linux Project base image (v8010)

bryteise:
Initial pull request for the Clear Linux Project base image. This image provides the latest released version of Clear Linux's os-core and os-core-update bundles upon which additional image types should be built. See https://clearlinux.org/ for more details on Clear Linux Project.

# Checklist for Review

  - https://github.com/docker-library/docs/issues/621
- [x] ~~existing official images have been considered as a base? (ie, if `foobar` needs Node.js, has `FROM node:...` instead of grabbing `node` via other means been considered?)~~
- [x] if `FROM scratch`, tarballs only exist in a single commit within the associated history?

### DIFF
--- a/library/clearlinux
+++ b/library/clearlinux
@@ -1,0 +1,4 @@
+# maintainer: William Douglas <william.douglas@intel.com>
+
+latest: git://github.com/clearlinux/docker-brew-clearlinux@0ceb073491db6b8e15c370e35c42e93c5ba2951e
+base: git://github.com/clearlinux/docker-brew-clearlinux@0ceb073491db6b8e15c370e35c42e93c5ba2951e


bryteise:
@tianon @yosifkit @moghedrin @psftw gentle prodding to see if there is something else I need to add here for review.

bryteise:
@tianon @yosifkit @Moghedrin @psftw I've updated the pull request and check list items, seems to mostly be missing the dockerization review.

tianon:
Is there a particular version number associated with this build, or is it a purely rolling-release distro? (not as familiar with Clear Linux as I'd like to be :smile:)

Also want to make sure you've considered `clear-linux` versus `clearlinux` -- we have images with both patterns (`oraclelinux` and `mongo-express`, for example), but wanted to make sure you knew you could choose. :smile:  (my own preference is for hyphenated, but I try not to be a stick in the mud about it, especially since https://github.com/clearlinux already seems to set precedence for upstream preferring the singleword form :smile:)

(Also, sorry for the delay!)

bryteise:
@tianon No worries! We try to release twice a day and each release is versioned. I haven't automated how this will be submitted yet for creating pull requests and I'm doubtful you want to be dealing with that kind of frequency.

Users would still be able to use our updater tool though to get the latest release once they deploy but it seems better to not have too stale of content as the base image. In other words guidance on how to handle updates would be greatly appreciated =).

I think we will end up not using the hyphen because of the github and clearlinux.org address avoid it but I did go back and forth myself as I liked the hyphen more myself!

tianon:
Hah yeah, twice-daily updates would be a bit excessive! :smile:  How heavy are the updates in terms of functional change? (ie, how far out do you think we could reasonably go before users will be affected in adverse ways if the image isn't updated?)

Regarding tags, I'd recommend adding the version number as an alias (especially since we'll keep older tags available when it's updated) so that users can pin to a specific version if they need slightly stronger reproducibility, but it's up to you. 
Is "Clear Linux Base" something specific and separate from just "Clear Linux"? (ie, what's the reasoning for the `:base` alias?  future variations of the image?)

bryteise:
@tianon The updates aren't usually too critical for the base image type as it is mostly just the updater itself, libc, curl, openssl, and bash for functionality. We do have this process in place for making updates faster that uses version deltas so we don't want to keep a version around for multiple months though.

So the aliases are an open question I wanted to raise. Currently we use a concept of bundles instead of packages in Clear Linux that translate to a set of functionality to do something. So for instance we have a editors bundle which contains a number of different editors instead of seperate vim or emacs packages.

I was wondering if it would be reasonable to add multiple different aliases to allow users to grab say a machine-learning image instead of needing to get the default base and doing installing the bundle after? This translates nicely to layers but I'm not sure how to best go about providing this on dockerhub officially.

So if we have a number of these aliases for different content I want to avoid also multiplexing that over versions (I'm also not sure keeping stale content is the right choice for our distro as we don't really support older builds).

tianon:
Yeah, since it sounds like that's already a concept defined in Clear Linux upstream, I think it makes sense to mirror that here with sub-images that include more of the standard bundles, assuming that mixing and matching them isn't going to be common (or only providing the most common ones and letting users do their own `Dockerfile` that's `FROM clearlinux:base` themselves if they want complex combinations, which with repository links and automated builds is pretty easy for them to keep up to date).

If you've got control of the `clearlinux` organization on the Docker Hub, you could also use that instead (ie, `clearlinux/editor` or something for the "editor" bundle), if that's more comfortable (or so that you can iterate faster and figure out what works before committing to anything here).

Regarding update frequency, I think updating something like weekly (revising that depending on external forces, such as a serious security vulnerability, installer/image bugs, or whatever) is sane enough if that works for you.  As long as it's something somewhat saner than twice daily, I don't mind letting you decide what makes sense. :smile:

tianon:
All that being said, Dockerization LGTM

(checking out docs now)

cc @yosifkit

tianon:
Hah, the test failure below is amusing!  Does Clear Linux not use `/etc/passwd`?

Build test of #1721; c294b0c (`clearlinux`):

```console

$ bashbrew build clearlinux:latest
warning: insecure protocol git:// detected: git://github.com/clearlinux/docker-brew-clearlinux
Building bashbrew/cache:57a5864d4275074752acad5e888f56508cd92fccad701865d53cbee54b7063ac (clearlinux:latest)
Tagging clearlinux:latest
Tagging clearlinux:base

$ test/run.sh clearlinux:latest
testing clearlinux:latest
	'utc' [1/4]...passed
	'no-hard-coded-passwords' [3/4]...cut: /etc/passwd: No such file or directory
failed
	'override-cmd' [4/4]...passed

```

The following images failed at least one test: clearlinux:latest

tianon:
There's also `Error response from daemon: No command specified.` (no default command) -- can I convince you to add `CMD ["/bin/bash"]` so `docker run -it --rm clearlinux` drops us into a shell without further intervention? :pray: :smile:

bryteise:
Ah we may try using the organization then for testing out what set of images makes sense first, that's a good idea. I don't think there would be too much mixing and matching regardless though as that's more for a user to decide on.

I think we'll be fine with once a week or fewer releases in this case as there shouldn't be much changing in the base for sure to need that kind of frequency.

The distro is [stateless](https://clearlinux.org/features/stateless) so we don't really ship anything the distro needs in /etc including /etc/passwd =).

Sure I'll add the CMD line, do you want me to just force push over the current request?

tianon:
Yeah, a force push would be grand, thanks! :+1:

bryteise:
@tianon Okay, I think this is all set with the CMD update and fresh release even =).

yosifkit:
I'm going to say LGTM, with the caveat that I need to improve the `no-hard-coded-passwords` test to account for a non-existent `/etc/password`.

Build test of #1721; 7841a35 (`clearlinux`):

```console

$ bashbrew build clearlinux:latest
warning: insecure protocol git:// detected: git://github.com/clearlinux/docker-brew-clearlinux
Building bashbrew/cache:60692a01860978edcc6e22890996e6e396a19b8ef87734ee56a541e221d18f1e (clearlinux:latest)
Tagging clearlinux:latest
Tagging clearlinux:base

$ test/run.sh clearlinux:latest
testing clearlinux:latest
	'utc' [1/4]...passed
	'no-hard-coded-passwords' [3/4]...cut: /etc/passwd: No such file or directory
failed
	'override-cmd' [4/4]...passed

```

The following images failed at least one test: clearlinux:latest

yosifkit:
We may run into the slow signing process like https://github.com/docker-library/official-images/issues/1961; it might take a little bit to be available. I have informed the group at Docker to expect this image so that they will be watching for it.

bryteise:
@yosifkit Ah okay, thanks for letting me know. Signing can take awhile =).

PR #1733, opened 2016-05-13
2.x branch does not yet include a LTS release

ndeloof:
one should not expose this one as official image


yosifkit:
LGTM

This doesn't remove images from the hub, but will update the "supported versions" list at the top.

tianon:
Any particular reason it's unsuitable?  I was honestly excited to see 2.x getting updates here. :smile:

ndeloof:
@tianon 2.x is available (as all other weekly builds) in `jenkinsci/jenkins` repository
official repo is about curated, so LTS releases. A 2.x will land here at some time, but too early for this. Typically, 2.0 is available as `jenkins:2.0` but includes a major security issue :-\

tianon:
Ah ok, fair enough! :disappointed:

PR #1771, opened 2016-05-24
Drop psftw from MAINTAINERS

psftw:
I'm not going anywhere, but I won't have as much
time to help with reviewing incoming proposals.


tianon:
:cry:

Would you be open to commenting yourself out instead of total removal? (ie, `# Peter ...`)

tianon:
LGTM :heart:

cc @yosifkit

yosifkit:
ok, :cry: we will miss your beneficial input, but understand that your time is limited.

LGTM

PR #1780, opened 2016-05-26
Use a client image with curl for testing.

hypnoce:
The jetty image may not have curl since there is an alpine variant.


md5:
Thanks @hypnoce! 👍 

@tianon @yosifkit This is to support the addition of a `jetty:alpine` image in https://github.com/appropriate/docker-jetty/issues/32

yosifkit:
LGTM

md5:
Thanks @yosifkit

PR #1790, opened 2016-06-01
Replace insecure git:// protocol with https://

ecneladis:
git:// protocol does not provide authentication and encryption.

> It uses the same data-transfer mechanism as the SSH protocol but without the encryption and authentication overhead.

https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols#The-Git-Protocol


ecneladis:
@yosifkit Ok, I'll check whole `docker-library` organisation and will submit pull requests.

How do you think external repositories can be handled? By improving documentation and fixing templates?

yosifkit:
Many of them would be fine with a PR to update their similar script and some would just need a note letting them know about it.  If you want to hold off for just a bit we are planning a rewrite of the library files to an rfc2822 format while keeping backwards compatibility and the scripts could be updated to do both changes at once.  (https://github.com/infosiftr/stackbrew/tree/rfc2822)

yosifkit:
Docs here could also be updated to recommend using `https`.

md5:
What benefit is there to using `https://` instead of `git://`?

Since none of these URLs are using any credentials, there is no authentication being added by switching to `https://`.

As for encryption and the ability to verify that it's actually coming from the real `github.com` that's provided by the SSL certificate verification, I'm not sure that provides a ton of value beyond what's provided by the cryptographic security of using a commit SHA.

I suppose it doesn't hurt, but it also doesn't seem necessary.

ecneladis:
@md5 First of all, using unencrypted channel for any network communication means unnecessary increasing risk of attack by some 3rd party, e.g. using some unpatched vulnerability in git.

Second, using SHA-1 in git wasn't though as a security mechanism, only as a consistency check. [1] Nevertheless for some time SHA1 was considered as a fairly secure, but we can observe gradual transition to stronger hash functions. [2] Some estimates provide that attacking SHA-1 may be feasible for malicious actors by 2018. [3]

Given that, it's not urgent, but it's better to be one step ahead and start implementing more secure practises.

[1] http://stackoverflow.com/a/28792805
[2] https://security.googleblog.com/2014/09/gradually-sunsetting-sha-1.html
[3] http://arstechnica.com/security/2012/10/sha1-crypto-algorithm-could-fall-by-2018/

md5:
@ecneladis Thanks!

I'm aware of the sunsetting of SHA-1, but I don't consider it very critical in the context of Git commits due to the difficulty of creating a wellformed commit with a hash collision. Still, I shouldn't have said “cryptographically secure”.

I almost deleted my comment before posting it since I realized there's no reason not to add a layer of SSL encryption, but I figured it could serve as a strawman. You bring up a good point about the possibility of an unmatched vulnerability in Git.

PR #1810, opened 2016-06-06
Explicitly note that Docker 1.10+ is required for the new Bashbrew

tianon:
This is especially thanks to https://github.com/docker/docker/issues/15785 and thus https://github.com/docker/docker/commit/c8cc4fb8d9a8d0cd155b8d26a3108e7762b623e3.

cc @yosifkit

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,13 @@
-FROM docker:1.9-git
+FROM docker:1.10-git
 
 # add "edge" since Alpine 3.3 only has Go 1.5 and we need 1.6+
 RUN sed -ri -e 'p; s!^!@edge !; s!v[0-9.]+!edge!' /etc/apk/repositories
 
-RUN apk add --update \
+RUN apk add --no-cache \
+# bash for running scripts
 		bash \
-		go@edge \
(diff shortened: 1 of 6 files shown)


tianon:
For a tad bit more detail, `git archive` _always_ creates the tar it generates with `uid/gid` of `0/0`, hard-coded.  For essentially any other tool, this is fine and ignored, but Docker 1.9 and below insisted on extracting to a temporary directory and trying to preserve permissions, and doing so client-side, which causes major problems for non-root Docker users.

tianon:
(This also updates `test-pr.sh` to try very hard to bind-mount `BASHBREW_CACHE` so that builds are a lot faster, and also run as our current user instead of root.)

yosifkit:
LGTM

tianon:
For posterity's sake, I feel I should add that I _did_ attempt to write a shim on top of a tar that just rewrites all the `uid/gid` from `0/0` to `geteuid/getegid`, but in the end we determined it wasn't worth the hassle/overhead just to keep support for Docker 1.9, especially since the way Bashbrew caches makes our reliance on Docker's cache quite a bit less pronounced.

PR #1813, opened 2016-06-06
Convert library/hylang to new 2822-based format`

tianon:
cc @paultag

### DIFF
--- a/library/hylang
+++ b/library/hylang
@@ -1,6 +1,7 @@
-# maintainer: Paul Tagliamonte <paultag@hylang.org> (@paultag)
+Maintainers: Paul Tagliamonte <paultag@hylang.org> (@paultag)
 
-latest: git://github.com/hylang/hy@0.11.1
-0: git://github.com/hylang/hy@0.11.1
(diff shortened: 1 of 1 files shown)


library/hylang
```diff
@@ -1,6 +1,7 @@
...
+Maintainers: Paul Tagliamonte <paultag@hylang.org> (@paultag)
 
-latest: git://github.com/hylang/hy@0.11.1
-0: git://github.com/hylang/hy@0.11.1
-0.11: git://github.com/hylang/hy@0.11.1
-0.11.1: git://github.com/hylang/hy@0.11.1
+# https://github.com/hylang/hy/tree/0.11.1
+Tags: 0.11.1, 0.11, latest
+GitRepo: https://github.com/hylang/hy.git
+GitFetch: refs/tags/0.11.1
```
tianon: The default of `refs/heads/master` would've worked fine here, but I figured since we know the exact tag that `HEAD`s at the tag we're interested in, we might as well be explicit and narrow here.

PR #1817, opened 2016-06-07
Apply constraints during "tag" and "push" (since they're only useful after "build", at least currently)

tianon:
cc @yosifkit


yosifkit:
LGTM

PR #1858, opened 2016-06-20
Update Mono 3.8 and 3.10 with workaround for mozroots

akoeplinger:
/cc @tianon this should fix the build break for the older Mono versions


akoeplinger:
Closing for now, just learned Mozilla might enforce https on hg.mozilla.org.

PR #1866, opened 2016-06-22
Correct typo in bashbrew push --namespace message

stuart-c:
Correct a typo in the help text for the bashbrew --namespace option for the push command


tianon:
Nice catch, thanks! :+1:

PR #1868, opened 2016-06-22
Add new pull subcommand

stuart-c:
Signed-off-by: Stuart Clark <stuart.clark@Jahingo.com>

A new subcommand which I find useful for priming the cache

### DIFF
--- a/bashbrew/go/src/bashbrew/cmd-pull.go
+++ b/bashbrew/go/src/bashbrew/cmd-pull.go
@@ -1,0 +1,44 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/codegangsta/cli"
+)
+
+func cmdPull(c *cli.Context) error {
+	repos, err := repos(c.Bool("all"), c.Args()...)
+	if err != nil {
+		return cli.NewMultiError(fmt.Errorf(`failed gathering repo list`), err)
+	}
+
+	uniq := c.Bool("uniq")
+	namespace := c.String("namespace")
+
+	if namespace == "" {
+		return fmt.Errorf(`"--namespace" is a required flag for "pull"`)
+	}
+
+	for _, repo := range repos {
+		r, err := fetch(repo)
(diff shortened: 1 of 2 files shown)


tianon:
Not sure how I feel about this, especially given that `docker pull` doesn't cache-fill anymore (as of Docker 1.10, which is our new minimum requirement).  We usually simulate this ourselves via `bashbrew list <args> | xargs -n1 docker pull`, which is pretty effective.

stuart-c:
This is basically the same thing though isn't it?

tianon:
Right, yes, that's the point I was trying to make -- one reasonably simple
line of shell for this one-off use case versus the 50 or so lines of Go for
something our official build server will never use and we only use in rare
cases ourselves (and furthermore we'd want to patch out for our official
build system in order to ensure it never ever pulls, as our configuration
there specifies via "Pull: never").

stuart-c:
While it probably isn't of much use for the official build servers that isn't the use-case I'm going for.

Ignoring the official images totally bashbrew is a very useful tool for CI building of Docker images where repeatability and careful control of tagging are desired. We find it a very useful tool but have a few extra requirements (e.g. pulling of images so we don't have to build everything from scratch) which would be preferable to include upstream instead of having to maintain a fork.

Regarding not wanting to use it on the official servers I'm not sure I understand the need to patch the feature out, as like the "Pull: never" step for building it won't ever pull unless you specifically request it.

[As an aside I'm looking to send over a couple more PRs for other small features which are probably not useful for the official build servers, including injection of build args]

yosifkit:
Thanks for spending time on this patch. We view the `bashbrew` tool as a build tool and `pull` doesn't seem to fit in that use case (especially since `docker pull` on `1.10+` is unable to fill build cache), so it is not something we want to add right now.

stuart-c:
No worries. I'll use the equivalent shell @tianon gave above instead.

PR #1885, opened 2016-06-25
Fix spurious "bad file descriptor" errors in bashbrew by using "exec.Cmd" objects more appropriately while streaming

tianon:


tianon:
GREEN CI YAY

nathanleclaire:
You know the saying -- "If you're going to vendor a @paultag lib, might as well vendor it all"

yosifkit:
LGTM

paultag:
Bwahahahaha

PR #1888, opened 2016-06-26
Add bashbrew as the entrypoint to make using the image easier

stuart-c:
Makes it simpler to use bashbrew. Currently no specific entrypoint is being set, so it uses the Docker entrypoint from the FROM line.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,3 +39,6 @@ COPY . $DIR
 RUN cd bashbrew/go && gb build
 
 VOLUME $BASHBREW_CACHE
+
+RUN ln -s "$PWD/bashbrew/bashbrew-entrypoint.sh" /usr/local/bin/bashbrew-entrypoint.sh
+ENTRYPOINT ["bashbrew-entrypoint.sh"]
--- a/bashbrew/bashbrew-entrypoint.sh
+++ b/bashbrew/bashbrew-entrypoint.sh
@@ -1,0 +1,14 @@
+#!/bin/sh
+set -e
+
+if [ "${1:0:1}" = '-' ]; then
+	set -- bashbrew "$@"
+fi
+
+# if our command is a valid bashbrew subcommand, let's invoke it through bashbrew instead
+# (this allows for "docker run bashbrew build", etc)
+if bashbrew "$1" --help > /dev/null 2>&1; then
+	set -- bashbrew "$@"
+fi
+
+exec "$@"


yosifkit:
This image is currently only used in the `test-pr.sh` script.  If we did want to have an entrypoint we should follow the same guidelines that the official images follow (ie, https://github.com/docker-library/official-images/#consistency).

stuart-c:
Makes perfect sense. I've updated to add a docker-entrypoint.sh file.

yosifkit:
Thanks! LGTM

Dockerfile
```diff
@@ -39,3 +39,6 @@ COPY . $DIR
 RUN cd bashbrew/go && gb build
 
 VOLUME $BASHBREW_CACHE
+
+COPY docker-entrypoint.sh /docker-entrypoint.sh
```
tianon: This is already copied into `$DIR` above, so this can be a symlink instead, and I'd rather we put this into the repo as `bashbrew/entrypoint.sh` and into the `PATH` as `bashbrew-entrypoint.sh` (so that our base image's `entrypoint` is still available via `docker-entrypoint.sh`; https://github.com/docker-library/docker/blob/99287145029122cf49321fa2a055d14240001d1d/1.11/Dockerfile#L21-L24)

ie:
```dockerfile
RUN ln -s "$PWD/bashbrew/entrypoint.sh" /usr/local/bin/bashbrew-entrypoint.sh
ENTRYPOINT ["bashbrew-entrypoint.sh"]
```

stuart-c: Updated

Dockerfile
```diff
@@ -39,3 +39,6 @@ COPY . $DIR
 RUN cd bashbrew/go && gb build
 
 VOLUME $BASHBREW_CACHE
+
+RUN ln -s "$PWD/bashbrew/entrypoint.sh" /usr/local/bin/bashbrew-entrypoint.sh
+ENTRYPOINT ["docker-entrypoint.sh"]
```
yosifkit: `s/docker-entrypoint/bashbrew-entrypoint/`

Also the new `docker-entrypoint.sh` needs to be moved and renamed so the `ln` works. :wink:

stuart-c:
Ahh! Sorrt @yosifkit I was being a bit slow. Updated...

yosifkit:
Looks like we need a `mv bashbrew-entrypoint.sh bashbrew/bashbrew-entrypoint.sh` so that it will be in the right spot.

stuart-c:
There we go

yosifkit:
LGTM

cc @tianon

tianon:
LGTM, thanks!

PR #1910, opened 2016-07-01
Convert centos over to new 2822-based format for Constraints

tianon:
The official build server was runs into "Untar re-exec error: exit status 1: output: operation not supported" with the latest tarballs, so I've converted to the new format so we can add "Constraints" and build "centos" officially on non-aufs.

cc @jperrin

(If you have a script that was building the previous output for you, I'd be happy to help convert it to output this newer format, but I couldn't find one. :smile:)

### DIFF
--- a/library/centos
+++ b/library/centos
@@ -1,45 +1,44 @@
-# maintainer: The CentOS Project <cloud-ops@centos.org> (@CentOS)
-# Build date 20160701
+Maintainers: The CentOS Project <cloud-ops@centos.org> (@CentOS)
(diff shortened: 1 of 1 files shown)


jperrin:
This is very likely due to 'ping' having capabilities restored in the tarball, which should be resolved by this pull request. https://github.com/docker/docker/pull/21251

See https://github.com/docker/docker/issues/20331 for info.

jperrin:
I wasn't generating the library info with a script, but the new format is handy. Once this is merged in, I'll start using this format

tianon:
:+1: sounds good, thanks!

PR #1911, opened 2016-07-02
Add ability to set build args

stuart-c:
Signed-off-by: Stuart Clark stuart.clark@Jahingo.com

Add a new option "BuildArgs" to the RFC2822 form of the image manifest which can contain comma separated lists of build arguments (in name=value format). These arguments are then passed to docker (via --build-arg in docker build) during build, with $VAR and ${VAR} environment substitutions allowed.

While not particularly useful for the official image builds it can be very handy for other users of bashbrew to pass build-time arguments from the build system (e.g. credentials which shouldn't be stored inside the image/Dockerfile)

For example:

Tags: latest
GitRepo: https://...
GitCommit: 12345
BuildArgs: arg1=value1, arg2=$HOME, arg3=the${USER}

### DIFF
--- a/bashbrew/go/src/bashbrew/cmd-build.go
+++ b/bashbrew/go/src/bashbrew/cmd-build.go
@@ -86,7 +86,7 @@ func cmdBuild(c *cli.Context) error {
 				}
 				defer archive.Close()
 
-				err = dockerBuild(cacheTag, archive)
+				err = dockerBuild(cacheTag, entry.BuildArgs, archive)
 				if err != nil {
 					return cli.NewMultiError(fmt.Errorf(`failed building %q (tags %q)`, r.RepoName, entry.TagsString()), err)
 				}
--- a/bashbrew/go/src/bashbrew/docker.go
+++ b/bashbrew/go/src/bashbrew/docker.go
@@ -137,8 +137,11 @@ func (r Repo) dockerBuildUniqueBits(entry *manifest.Manifest2822Entry) ([]string
 	}, nil
 }
 
-func dockerBuild(tag string, context io.Reader) error {
+func dockerBuild(tag string, build_args []string, context io.Reader) error {
 	args := []string{"build", "-t", tag, "--rm", "--force-rm"}
+	for _, arg := range build_args {
+		args = append(args, "--build-arg", os.ExpandEnv(arg))
+	}
 	args = append(args, "-")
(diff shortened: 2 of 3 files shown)


yosifkit:
I like the idea, but why limit it to build args? We could hopefully support any flag to docker build and might as well just call it something like BuildFlags.

I don't think we can reasonably support the `$var` expansion; it will inevitably lead to more complicated "shell-like" syntax requiring parsing (plus I don't like how `foo=${a` would become `foo=a`).

We should really start the changes over in https://github.com/docker-library/go-dockerlibrary/ so that we can the add that to the vendored copy here.  Maybe the syntax will require multi-line (if they want a space in a value)?

```
BuildFlags:
 --build-arg=test=foo bar
 --pull

# or this, with both being valid syntax
BuildFlags: --build-arg=test=foo --pull
```

stuart-c:
The idea was to get it all agreed here before doing a PR over at go-dockerlibrary (as just doing something over there wouldn't show why it might be wanted).

I'm not sure how useful making it totally generic would be and you'd probably make it a lot more complex. For example would you really want to be able to add the option --pull given at the bashbrew level you can already pass in --pull always/never/missing so having something in the per-tag/image level just feels confusing (if I said --pull never why is it still pulling?).

Also some options wouldn't be allowed anyway (such as --tag) as they are already being set within bashbrew and others like --memory also make sense to be able to set globally not (just) per-tag.

Regarding the "${a" becoming "a" the code is just using os.ExpandEnv from the standard Go library. Looking at the code if a ${ is found with no } it treats it the same as $"" (i.e. $ + no string) which of course when substituted becomes an empty string also. I suppose you could add some specific code to check for that situation and then fail the build, but then you lose the benefit of using the Go standard library (i.e. not having to write your own parser).

Regarding not supporting $ substitutions at all, while I can see some possible use in being able to set fixed per-tag build args for me the main usefulness comes from being able to use more dynamic args - for example setting things which might vary between places you run bashbrew (e.g. http_proxy) or for credentials which should never be stored in source code/text files.

Finally, I think "BuildArg: something=some with spaces" should already work as should multi-line entries too (you'd need the comma at the end of each line as per Maintainers: to follow the RFC style).

yosifkit:
Closing old issue.
Here is my current stance on build args for official images: https://github.com/docker-library/official-images/issues/4224#issuecomment-379373303

PR #1946, opened 2016-07-14
Add 'shell' image to run a Unix shell

bboreham:
The idea here is that running a shell to demonstrate features of Docker should be as quick and simple as possible, especially for newcomers.  If this PR is accepted, users could do:

    docker run -ti shell

and get an interactive shell.

Previously we used `docker run -ti ubuntu` for this purpose, but many tools were removed from the official Ubuntu image recently, including `ping`, `nslookup` and `sudo`.  Also, this image is based on Alpine so is less than 5MB compared to `ubuntu:latest` at 125MB.

This image only adds two things to `alpine`:
 * `CMD [ "/bin/sh" ]`.  Alpine's maintainers [have rejected this](http://github.com/gliderlabs/docker-alpine/pull/60)
 * it is called "shell", which is easier to explain to a complete beginner than "alpine".

### DIFF
--- a/library/shell
+++ b/library/shell
@@ -1,0 +1,4 @@
+# maintainer: Weaveworks Inc <help@weave.works> (@weaveworks)
+
+3.4: git://github.com/weaveworks/shell-image@357be742ceb4d830b7b900e3b3fdaa804d3fad8e versions/3.4
+latest: git://github.com/weaveworks/shell-image@357be742ceb4d830b7b900e3b3fdaa804d3fad8e versions/3.4


philwinder:
Love it. Simple, effective. 👍

tianon:
I like the sentiment, but given that "shell" isn't actually an upstream, I don't think this is an appropriate image for the official images program (`hello-world` and the DockerCon variants are our only current exceptions to that as far as I'm aware).

That being said, I've opened https://github.com/gliderlabs/docker-alpine/pull/199 to see if we can get attention on the default command again from the Alpine maintainers, especially given that the original reasons for not including `CMD` there no longer apply. :+1:

tianon:
The following new image might be relevant to your interests: [`bash`](https://hub.docker.com/_/bash/) / https://github.com/docker-library/official-images/pull/2217 :innocent:

PR #1765, opened 2016-05-22
Add an official image for Apache Zookeeper

31z4:
[Apache Zookeeper](http://zookeeper.apache.org/) is required for many distributed applications and services. In particular it's needed for #1641. [docs](https://github.com/docker-library/docs) PR is about to come.

# Checklist for Review

- [x] associated with or contacted upstream?
  - [ZOOKEEPER-1940](https://issues.apache.org/jira/browse/ZOOKEEPER-1940?focusedCommentId=15295665&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15295665) :confused:
- [x] does it fit into one of the common categories? ("service", "language stack", "base distribution")
- [x] is it reasonably popular, or does it solve a particular use case well?
- [x] does a [documentation](https://github.com/docker-library/docs/blob/master/README.md) PR exist? (should be reviewed and merged at roughly the same time so that we don't have an empty image page on the Hub for very long)
  - https://github.com/docker-library/docs/pull/590
- [x] dockerization review for best practices and cache gotchas/improvements (ala [the official review guidelines](https://github.com/docker-library/official-images/blob/master/README.md#review-guidelines))?
- [ ] 2+ dockerization review?
- [x] existing official images have been considered as a base? (ie, if `foobar` needs Node.js, has `FROM node:...` instead of grabbing `node` via other means been considered?)
- [x] ~~if `FROM scratch`, tarballs only exist in a single commit within the associated history?~~
- [ ] passes current tests? any simple new tests that might be appropriate to add? (https://github.com/docker-library/official-images/tree/master/test)

PR #1960, opened 2016-07-19
Add a note of clarification about who is expected to interact with the "new image checklist"

tianon:
This is as was discussed in https://github.com/docker-library/official-images/pull/1765#issuecomment-228353753 (showing that there was some definite confusion about the intended purpose of the checklist). :heart:


yosifkit:
LGTM

PR #1972, opened 2016-07-21
Remove "vivid" and "squeeze" from neurodebian

tianon:
See https://github.com/neurodebian/dockerfiles/pull/2


tianon:
cc @yarikoptic (just so you're kept in the loop :heart:)

PR #1979, opened 2016-07-22
Add a separate "EntryIdentifier" so that build output is more informative

tianon:
This changes something like:

    Using bashbrew/cache:277...af2 (buildpack-deps)

into:

    Using bashbrew/cache:277...af2 (buildpack-deps:wheezy)


tianon:
cc @yosifkit

yosifkit:
LGTM

PR #1987, opened 2016-07-27
Add hacky "diff-pr.sh" script for generating diffs of PRs

tianon:
This is what I've been slowly building up over time (as can be seen from the diffs I've been posting on PRs).  It still relies on internal knowledge of how Bashbrew works (and stores the Git data on-disk), but it's a start.  I figure it's as good as it's going to get for now until we can devote more time to working on the design aspects of Bashbrew to make this kind of thing officially supported instead of a hack.

### DIFF
--- a/diff-pr.sh
+++ b/diff-pr.sh
@@ -1,0 +1,159 @@
+#!/bin/bash
+set -eo pipefail
+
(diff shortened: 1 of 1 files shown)


yosifkit:
LGTM

PR #1990, opened 2016-07-28
Adjust bashbrew such that env overrides config appropriately

tianon:


tianon:
(this will fix `diff-pr.sh` in the face of `~/.config/bashbrew/flags` which specifies `Library:`)

yosifkit:
LGTM

PR #1998, opened 2016-07-29
SonarQube: add tag `alpine`

Godin:


tianon:
I'd prefer that this new tag be simply `:alpine` rather than `:latest-alpine`, and I've outlined my reasoning a bit over in #2009. :+1:

Godin:
@tianon okay, changed on `alpine`.

What about existing tags such as `lts-alpine` - should they be changed on `alpine-lts` ?

yosifkit:
The current ordering is fine and most of the images are fairly consistent with a `$version-$distro` (ie, most specific information to least specific) ordering.

LGTM

Godin:
@yosifkit cool, thanks for info

Godin:
Just wondering what prevents merge since PR was updated with requested change?

tianon:
Sorry about that @Godin!

LGTM :+1:

Godin:
@tianon no pb, thx!

PR #2009, opened 2016-08-01
Add the start of a "Tags and aliases" section to our documentation for creating/maintaining repositories

tianon:


yosifkit:
Should we add the recommendation of ordering your tag from most specific to least specific: ie `openjdk-8u92-jdk-alpine` vs `alpine-jdk-8u92-openjdk`.  The `openjdk` in this case feels out of place either way.

tianon:
Yeah, I think Java is a bad example in this case (https://github.com/docker-library/official-images/pull/2046).

I'm not sure whether that's worth talking about -- it's something general we do because we prefer it, but I'm not sure it's formal enough to be worth mentioning.  Maybe add that later?

yosifkit:
LGTM

PR #2017, opened 2016-08-03
Publishing old Neo4j point releases

benbc:
Also includes a change to the new library file format and the two latest point releases.

### DIFF
--- a/library/neo4j
+++ b/library/neo4j
@@ -1,3 +1,9 @@
+# Old point releases are included in this library file so that users
+# are not forced to upgrade to the latest point releases in order to
+# get security fixes to the base image. See
+# https://github.com/docker-library/official-images/issues/1864 for a
+# discussion of this point.
+
 Maintainers: Ben Butler-Cole <ben@neotechnology.com> (@benbc),
              Jonas Kalderstam <jonas.kalderstam@neotechnology.com> (@spacecowboy),
              Steven R Baker <steven.baker@neotechnology.com> (@srbaker)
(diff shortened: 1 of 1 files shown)


library/neo4j
```diff
@@ -1,17 +1,129 @@
...
+             Steven R Baker <steven.baker@neotechnology.com> (@srbaker)
 
-2.3.4: git://github.com/neo4j/docker-neo4j@3f379d33f5d08f2e483769427ab1e0b898a819ac 2.3.4
-2.3.5: git://github.com/neo4j/docker-neo4j@3f379d33f5d08f2e483769427ab1e0b898a819ac 2.3.5
-2.3: git://github.com/neo4j/docker-neo4j@3f379d33f5d08f2e483769427ab1e0b898a819ac 2.3.5
+Comment: Old point releases are included in this library file so that
+         users are not forced to upgrade to the latest point releases
+         in order to get security fixes to the base image. See
+         https://github.com/docker-library/official-images/issues/1864
+         for a discussion of this point.
```
yosifkit: so, this section by itself is causing parse issues. :cry: This is because there is a blank line so this first section is treated as a block that requires `Tags:`, etc.  I think the simplest solution is to just make this an actual comment by just starting each line with `#`.

benbc: @yosifkit Sure, I'll do that. I was trying to follow RFC 2822. If the format isn't exactly the same as that, maybe you could mention the differences in the README.

benbc:
We'd like to move our fork of this repo to different organization. I'll close the PR and open a new one.

PR #2037, opened 2016-08-09
Add a bit more explanation of how our format differs from RFC 2822

tianon:


tianon:
See also https://github.com/docker-library/official-images/pull/2017#discussion_r73469811, where it was pointed out that we aren't very clear about this. :+1:

> I was trying to follow RFC 2822. If the format isn't exactly the same as that, maybe you could mention the differences in the README.

README.md
```diff
@@ -227,11 +227,17 @@ The filename of a definition file will determine the name of the image repositor
 
 The manifest file format is officially based on [RFC 2822](https://www.ietf.org/rfc/rfc2822.txt), and as such should be familiar to folks who are already familiar with the "headers" of many popular internet protocols/formats such as HTTP or email.
 
+The primary additions are inspired by the way Debian commonly uses 2822 -- namely, lines starting with `#` are ignored and "paragraphs" (or "entries") are separated by a blank line.
+
+The first entry is the "global" metadata for the image. The only required field in the global entry is `Maintainers`, whose value is comma-separated in the format of `Name <email> (@github)` or `Name (@github)`. If fields such as `GitRepo` are specified in the global entry, they will be the default for the remainder of the file.
```
yosifkit: > If fields such as `GitRepo`

It feels like this might make people think that only `GitRepo` is valid in the global section as a default.  Maybe something like this? 
> Any field specified in the global entry will be the default for the rest of the entries/paragraphs and can be overridden in an individual paragraph

yosifkit:
Do we need to explain that a paragraph of just comments will ignore the extra new line?

tianon:
Done and done (although for the second one, I just included an example of doing precisely that -- not sure it's important enough to warrant spelling it out explicitly).

yosifkit:
LGTM

PR #2039, opened 2016-08-10
php builds ftp statically

patrickomeara:
`--enable-ftp` for php at build time

docker-library/php#236


yosifkit:
Looks like we got this in https://github.com/docker-library/official-images/pull/2041.  Thanks for the PR though!

PR #2020, opened 2016-08-04
Publishing old Neo4j point releases

benbc:
Also includes a change to the new library file format and the two latest point releases.

PR #2047, opened 2016-08-11
Allow globs in diff to get files beginning with `.`

yosifkit:
This was needed for https://github.com/docker-library/official-images/pull/2020 since they used `COPY ./local-package/*` for a file that starts with `.`.  Which is more similar to the glob matching that docker uses anyway.


tianon:
Indeed, Docker uses `filepath.Glob` which uses `filepath.Match` which this matches much closer. :+1:

PR #2051, opened 2016-08-12
Officially exempt "microsoft/windowsservercore" and "microsoft/nanoserver" from the "non-official images" exclusion

tianon:


yosifkit:
LGTM

PR #2072, opened 2016-08-19
Added the library definition file of monaserver (http://monaserver.ovh)

thomasjammet:
# Checklist for Review

-   [x] associated with or contacted upstream?
  - https://github.com/MonaSolutions is upstream for MonaServer
-   [ ] is it reasonably popular, or does it solve a particular use case well?
-   [ ] does a [documentation](https://github.com/docker-library/docs/blob/master/README.md) PR exist? (should be reviewed and merged at roughly the same time so that we don't have an empty image page on the Hub for very long)
-   [ ] dockerization review for best practices and cache gotchas/improvements (ala [the official review guidelines](https://github.com/docker-library/official-images/blob/master/README.md#review-guidelines))?
-   [ ] 2+ dockerization review?
-   [ ] existing official images have been considered as a base? (ie, if `foobar` needs Node.js, has `FROM node:...` instead of grabbing `node` via other means been considered?)
-   [ ] if `FROM scratch`, tarballs only exist in a single commit within the associated history?
-   [ ] passes current tests? any simple new tests that might be appropriate to add? (https://github.com/docker-library/official-images/tree/master/test)

### DIFF
--- a/library/monaserver
+++ b/library/monaserver
@@ -1,0 +1,5 @@
+Maintainers: Thomas Jammet <contact@monaserver.ovh> (@thomasjammet)
+GitRepo: https://github.com/MonaSolutions/docker-mona.git
+
+Tags: 1.2, 1, latest
+GitCommit: dbc236e235b62d8107d8e2c660f42486fa424aca


tianon:
Took a quick look at the Dockerization here, and I've got a few comments/questions:

1. `FROM alpine` has high potential for breaking when `:latest` changes from 3.4 to 3.5 -- that should be more explicit, like `FROM alpine:3.4`
2. does the `Makefile` for LuaJIT support parallelism? do the MonaServer `Makefile`s support parallelism? (if so, I recommend adding `-j "$(getconf _NPROCESSORS_ONLN)"` to those `make` invocations so they builds slightly faster)
3. since `/usr/local/bin` is in `PATH` by default, couldn't we skip `WORKDIR` and change `CMD` to be `CMD ["MonaServer", "--log=7"]` so that it's simpler for users to use/overwrite? (is there any particular reason why `MonaServer` ought to be started with CWD set to `/usr/local/bin`?)

Thanks!

yosifkit:
Closing, no response.

PR #2079, opened 2016-08-22
Add library file for powerdns

weldpua2008:
[PowerDNS ](https://www.powerdns.com/)is a DNS server, written in C++ and licensed under the GPL. It runs on most Unix derivatives. PowerDNS features a large number of different backends ranging from simple BIND style zonefiles to relational databases and load balancing/failover algorithms. A DNS recursor is provided as a separate program.
# Checklist for Review

-   [ ] associated with or contacted upstream?
-   [ ] is it reasonably popular, or does it solve a particular use case well?
-   [x] does a [documentation](https://github.com/docker-library/docs/blob/master/README.md) PR exist? (should be reviewed and merged at roughly the same time so that we don't have an empty image page on the Hub for very long)
  - docker-library/docs#668
-   [ ] dockerization review for best practices and cache gotchas/improvements (ala [the official review guidelines](https://github.com/docker-library/official-images/blob/master/README.md#review-guidelines))?
-   [ ] 2+ dockerization review?
-   [ ] existing official images have been considered as a base? (ie, if `foobar` needs Node.js, has `FROM node:...` instead of grabbing `node` via other means been considered?)
-   [ ] if `FROM scratch`, tarballs only exist in a single commit within the associated history?
-   [ ] passes current tests? any simple new tests that might be appropriate to add? (https://github.com/docker-library/official-images/tree/master/test)

### DIFF
--- a/library/powerdns
+++ b/library/powerdns
@@ -14,27 +14,3 @@ GitRepo: https://github.com/weldpua2008/docker-powerdns.git
 GitCommit: 1edaeceef4768047259f2834f1dbf341f98c8b55
 Directory: ubuntu/14.04/4/4.0.1/Dockerfile
 
-
-Tags: 4.0.0-ubuntu-16.04
-GitRepo: https://github.com/weldpua2008/docker-powerdns.git
-GitCommit: 1edaeceef4768047259f2834f1dbf341f98c8b55
-Directory: ubuntu/16.04/4/4.0.0/Dockerfile
-
-
-Tags: 4.0.1-ubuntu-16.04, 4.0-ubuntu-16.04
-GitRepo: https://github.com/weldpua2008/docker-powerdns.git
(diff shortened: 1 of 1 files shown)


weldpua2008:
Added a doc file: https://github.com/docker-library/docs/pull/668

tianon:
Apologies for the delay! :disappointed:

My first question would be whether you've been in contact with PowerDNS upstream to see if they're either interested in being involved, maintaining the image, providing their blessing for your ownership/maintainership of it, or would even rather there weren't an official image for PowerDNS at all (which has happened before)? :pray: :sweat_smile:

Moghedrin:
Closing because this pull request is stale and there hasn't been a response here. If there's more word, especially regarding PowerDNS upstream, we can reopen.

PR #2136, opened 2016-09-09
Remove outdated note about zombie solution in docker

yosifkit:
After reading the issue that this points to (https://github.com/docker/docker/issues/11529), it seems our note is not really true.  Digging deeper led me to this https://github.com/docker/docker/issues/23214#issuecomment-223627273, suggesting that this note be removed.


tianon:
LGTM

PR #2235, opened 2016-10-10
Add VMware Lightwave as official image

aizurov-vmw:


aizurov-vmw:
Hmm. Is there anything I'm missing? I've followed the format recommended in the readme but it seems to be trying to parse it as the old one and failing.

library/lightwave
```diff
@@ -0,0 +1,9 @@
+# VMware Lightwave
+Maintainers: Sriram Nambakam <snambakam@vmware.com> (@snambakam), Lightwave Checkins <lightwave-checkins@vmware.com> (@gerrit-lightwave)
+GitRepo: https://github.com/vmware/lightwave.git
+
```
tianon: There's a single "space" (` `) on this line which is turning the value of `GitRepo` into `https://github.com/vmware/lightwave.git\n` instead of separating the two entries.

aizurov-vmw: Thanks. Let me try a fresh pull request.

tianon:
You might want to go re-read through the README of this repository before
you do so -- I took a quick glance at your current Dockerfile and it's
very, very heavy-weight (running a full instance of systemd, which is very
unlikely to be accepted in any form), and not built "FROM" an official
image besides (I think you meant to base off the "photon" image instead of
the "vmware/photon" image).

PR #2236, opened 2016-10-10
Add VMware Lightwave as official image

aizurov-vmw:


tianon:
Quoting https://github.com/docker-library/official-images/pull/2235#issuecomment-252763031 here for posterity:

> You might want to go re-read through the README of this repository before
you do so -- I took a quick glance at your current Dockerfile and it's
very, very heavy-weight (running a full instance of systemd, which is very
unlikely to be accepted in any form), and not built "FROM" an official
image besides (I think you meant to base off the "photon" image instead of
the "vmware/photon" image).

tianon:
If there are to be future iterations of this PR, please force push to this branch (which will update this existing PR) rather than opening further new PRs (each of which creates a group of emails in the inboxes of not only the maintainers, but also every user who is watching this repository).  Thanks!

yosifkit:
Closing, no response

PR #2291, opened 2016-10-27
Bump version to force a rebuild

jefferai:
We had an issue with purging our CDN to remove an old version of the 0.6.2 package and it seems when this got built it was between when we originally created the package and when the purging was fixed.

This contains no changes, but merely updates the hash to force a rebuild in order to pull the correct/final version of 0.6.2 into the container.

### DIFF
--- a/library/vault
+++ b/library/vault
@@ -1,7 +1,7 @@
 # maintainer: Jeff Mitchell <jeff@hashicorp.com> (@jefferai)
 
 0.6.0: git://github.com/hashicorp/docker-vault@3d12aa78cdbdce22b3d3e30f1093843f21b0a8fa 0.6
-0.6: git://github.com/hashicorp/docker-vault@3d3957180d689ecddb537aa799a878171280e8a3 0.6.2
+0.6: git://github.com/hashicorp/docker-vault@7f2bb1991218a66ff483f9611acb885b322547f8 0.6.2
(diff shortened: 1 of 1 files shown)


yosifkit:
Without a change to the Dockerfile, this will not break the docker build cache so the rebuild won't have any effect.

jefferai:
OK. I'll push a comment change to the Dockerfile and update the PR.

jefferai:
Change pushed. It just explicitly codifies the normal stop signal.

tianon:
A comment change still won't modify the instructions, so will still get
eaten by build cache -- we can try purging our build cache to force a
rebuild, or you could look into embedding a SHA256 or similar into the
Dockerfile itself. :+1:

jefferai:
@tianon it's not a comment change, I added STOPSIGNAL.

tianon:
Right, but the addition is low in the `Dockerfile`, so cache isn't busted on the higher layers where the application is actually downloaded (and it's setting the `STOPSIGNAL` to the default value, which is a bit strange).

I think we should probably instead just purge all the cache on our build server and rebuild as-is (forcing it to redownload the latest).

jefferai:
@tianon I don't mind moving the STOPSIGNAL up to the top if that will fix it rather than you having to purge your entire cache. Let me know what is easiest for you.

tianon:
I think force-busting the cache on the server will be the cleanest so we don't clutter the `Dockerfile` (since it's a one-off issue). :+1:

yosifkit:
Vault was rebuilt and pushed, let us know if there is still an issue.

PR #2361, opened 2016-11-17
Update bashbrew Constraints logic with several fixes

tianon:
- only print a Constraints-related "skipping" warning for a given "repo:tag" once per session
- apply constraints while sorting too (so that `bashbrew build --all` on Windows doesn't try to `git fetch` Linux rootfs tarballs only to skip them during build, for example)
- add `--apply-constraints` flag to `from`, `children`, and `parents` subcommands


yosifkit:
Travis failure is unrelated to the PR.

PR #2466, opened 2016-12-20
Adding pydio to docker official images

ghecquet:
Pydio is a secure, self-hosted file sharing open-source software for businesses deployed on-premise or in a private cloud.

# Checklist for Review

    - https://github.com/pydio
    - https://github.com/docker-library/docs/pull/779
    - `FROM php:...`

### DIFF
--- a/library/pydio
+++ b/library/pydio
@@ -1,0 +1,15 @@
+Maintainers: Pydio Team <team@pydio.com> (@pydio),
+             Greg Hecquet <gregory@pydio.com> (@ghecquet)
+GitRepo: https://github.com/pydio/docker-library.git
+
+Tags: 7.0-apache, 7.0, latest
+GitCommit: 4bc599ce3f8400e233e5e92666a36f66d2b8ef08
+Directory: 7.0/apache
+
+Tags: 7.0-fpm
+GitCommit: 4bc599ce3f8400e233e5e92666a36f66d2b8ef08
+Directory: 7.0/fpm
+
+Tags: 7.0-cli
+GitCommit: 4bc599ce3f8400e233e5e92666a36f66d2b8ef08
+Directory: 7.0/cli
--- a/library/pydio
+++ b/library/pydio
@@ -3,13 +3,13 @@ Maintainers: Pydio Team <team@pydio.com> (@pydio),
 GitRepo: https://github.com/pydio/docker-library.git
 
 Tags: 7.0-apache, 7.0, latest
-GitCommit: 4bc599ce3f8400e233e5e92666a36f66d2b8ef08
+GitCommit: 8051e9086d80a43d427b94051eebd08ff5729658
 Directory: 7.0/apache
 
 Tags: 7.0-fpm
-GitCommit: 4bc599ce3f8400e233e5e92666a36f66d2b8ef08
+GitCommit: 8051e9086d80a43d427b94051eebd08ff5729658
 Directory: 7.0/fpm
 
 Tags: 7.0-cli
-GitCommit: 4bc599ce3f8400e233e5e92666a36f66d2b8ef08
+GitCommit: 8051e9086d80a43d427b94051eebd08ff5729658
 Directory: 7.0/cli


tianon:
I've been taking a look, and I've got a few comments so far:

1. `DEBIAN_FRONTEND=noninteractive` is very rarely necessary
2. `RUN pecl install apcu` is _definitely_ going to break at some point (we've had issues with bare no-version `pecl install` so regularly that I can no longer say it's "likely" to break :disappointed:); something like `RUN pecl install APCu-5.1.8` (tagging the version explicitly both so that PHP 7.0 compatibility is "for sure" and so that updates to APCu, which might not work properly with pydio, become an explicit choice to apply)
3. any particular reason to set _all_ of `LANG`, `LANGUAGE`, and `LC_ALL`?  (usually only `LANG` is necessary, but depends on the software)  Also, was `C.UTF-8` tested (it's included in the base image directly, and thus wouldn't require `locale-gen`)?
4. `pid` and `listen` are being set in `php.ini`, but are not PHP configuration options -- they're specifically for PHP-FPM, and thus would need to be over in `php-fpm.conf`, but I'm curious why those need to be customized rather than simply using the defaults from the `php:fpm` image?
5. `EXPOSE 443` in `pydio:apache` seems wrong -- the `httpd.conf` only references port 80, and `EXPOSE 80` is already in the `php:apache` upstream `Dockerfile` too, so I don't think any `EXPOSE` are really necessary here, right?
6. is there runtime data being stored in `/var/run/pydio` or `/var/lib/pydio`?  If so, wouldn't it make sense to make one or both of those a `VOLUME` directly in the image `Dockerfile`?
7. `pydio:cli` is using `ENTRYPOINT`, but doesn't appear to support the ["consistent interface" necessary for being an official image](https://github.com/docker-library/official-images#consistency)
8. while it's an interesting idea for `pydio:fpm` to pre-include configuration files for caddy, httpd, and nginx, I think that's probably something that's better suited to being part of the documentation (or at least explain to users how to _extract_ the example configuration from the image rather than using `volumes-from` and limiting the two services to be running on the same host in a swarm, for example)
9. `php:fpm` already includes `EXPOSE 9000` :+1:

ghecquet:
Thanks very much for your feedback.

I've removed or changed most of them [here.](https://github.com/pydio/docker-library/commit/3fef6f396e65d993eadce9e01e447c448d28c359) I just have a question regarding point 8 (below). Would you mind having a look and letting me know ?

1. Removed
2. Changed
3. Removed
4. Removed
5. Removed all `EXPOSE`
6. I've made a volume of both of them
7. Done
8. If I add a section to the documentation explaining how to docker cp the files so users can just use it as a template or elsewhere, would that be ok ? Often people are struggling with that configuration so I feel that would be a very nice addition to have that pre-included configuration. If you think that it isn't necessary I'll remove. 
9. Removed

ghecquet:
Sorry, I've only realised I needed to redo a pull request for my comment to be looked at.

Please let me know if I need to take action for point number 8.

Thanks in advance,
Greg

yosifkit:
:disappointed: sorry for the long reply :bowing_man:

My main comment is on the following section. `docker-php-ext-configure` on a module needs to happen before the `docker-php-ext-install` that installs it. So, the configure gd needs to be before the install gd.  Also, `ldap` is only configured but never installed.
```dockerfile
RUN docker-php-ext-install exif intl gd json mysqli pgsql mcrypt opcache xml \
    && docker-php-ext-configure gd --with-png-dir=/usr --with-jpeg-dir=/usr \
    && docker-php-ext-configure ldap --with-libdir=lib/x86_64-linux-gnu/ \
    && docker-php-ext-enable apcu
```

This should work a little better:
```dockerfile
RUN docker-php-ext-configure gd --with-png-dir=/usr --with-jpeg-dir=/usr \
    && docker-php-ext-configure ldap --with-libdir=lib/x86_64-linux-gnu/ \
    && docker-php-ext-install exif intl gd json ldap mysqli pgsql mcrypt opcache xml \
    && docker-php-ext-enable apcu
```

Secondary comment, I assume that since we took so long this will also need an `8.0` set of images?

yosifkit:

PR #2504, opened 2017-01-03
Make alpine 3.5 the latest

ncopa:


tianon:
LGTM

Merging since this only changes tags (no commit changes). :+1:

PR #2551, opened 2017-01-19
Update python-hy test

yosifkit:
New hylang `0.12.0` seems to work with python 3.6.

cc @tianon, @paultag

PR #2556, opened 2017-01-20
Revert "Update python-hy test"

tianon:
Reverts docker-library/official-images#2551

### DIFF
--- a/test/tests/python-hy/container.sh
+++ b/test/tests/python-hy/container.sh
@@ -2,7 +2,7 @@
 set -e
 
 # Hy doesn't work on 3.6+ :(
-if ! python -c 'import sys; exit((sys.version_info[0] == 3 and sys.version_info[1] >= 7) or sys.version_info[0] > 3)'; then
+if ! python -c 'import sys; exit((sys.version_info[0] == 3 and sys.version_info[1] >= 6) or sys.version_info[0] > 3)'; then
 	# TypeError: required field "is_async" missing from comprehension
 	echo >&2 'skipping Hy test -- no workie on Python 3.6+'
 	cat expected-std-out.txt # cheaters gunna cheat


tianon:
Heh, looks like the failures are actually caused by the new version of Hy, not by the test change. :innocent: :disappointed:

cc @paultag (any ideas for what we might be doing wrong?)

```console
$ docker run -it --rm python:3.6-slim bash
root@fb6e7113d667:/# pip install hy
Collecting hy
  Downloading hy-0.12.0.tar.gz (63kB)
    100% |████████████████████████████████| 71kB 1.6MB/s 
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-3xk7ny_l/hy/setup.py", line 35, in <module>
        (["git", "describe", "--tags", "--dirty"])
      File "/usr/local/lib/python3.6/subprocess.py", line 336, in check_output
        **kwargs).stdout
      File "/usr/local/lib/python3.6/subprocess.py", line 403, in run
        with Popen(*popenargs, **kwargs) as process:
      File "/usr/local/lib/python3.6/subprocess.py", line 707, in __init__
        restore_signals, start_new_session)
      File "/usr/local/lib/python3.6/subprocess.py", line 1326, in _execute_child
        raise child_exception_type(errno_num, err_msg)
    FileNotFoundError: [Errno 2] No such file or directory: 'git'
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-3xk7ny_l/hy/
```

Why should Hy require `git` just to install? :confused:

tianon:
Looks to be due to https://github.com/hylang/hy/pull/1135, which appears to have accounted for `git` failing, but not for `git` not even existing.

tianon:
Filed https://github.com/hylang/hy/issues/1205. :+1:

yosifkit:
Should we go with this revert or wait for the fix in https://github.com/hylang/hy/pull/1206?

tianon:
This revert is wrong too (tests still fail both with and without this revert) -- that's what led me to dig in and find the exact issue I filed.  I only left it open as a tracker to remind us that we _do_ need to figure out something (and pinning the test to Hy 0.11.x for now and including this revert might be the best way to do that until 0.12.1).

tianon:
ccac248 :+1:

PR #2567, opened 2017-01-23
Add a lein tag to clojure image

cap10morgan:
...for completness with new boot tag.

Now that we have both boot and lein variants, it makes sense to have a tag that means "just give me the latest version w/ lein."


tianon:
Simple tag change :+1:

LGTM

PR #2437, opened 2016-12-12
SonarQube 5.6.4 (LTS) and 6.2 (LATEST)

Godin:

PR #2575, opened 2017-01-24
added Groovy

keeganwitt:
# Checklist for Review

    - https://github.com/groovy
    - language stack
    - https://github.com/docker-library/docs/pull/807
    - `FROM openjdk`

### DIFF
--- a/library/groovy
+++ b/library/groovy
@@ -1,0 +1,42 @@
+Maintainers: Keegan Witt <keeganwitt@gmail.com> (@keeganwitt)
+GitRepo: https://github.com/groovy/docker-groovy.git
+
+Tags: 2.4.8-jdk7, 2.4-jdk7, jdk7
+GitCommit: 0402f0037b9bdf1d69ce844a381b81304877df3a
+Directory: jdk7
+
+Tags: 2.4.8-jre7, 2.4-jre7, jre7
+GitCommit: 0402f0037b9bdf1d69ce844a381b81304877df3a
+Directory: jre7
+
+Tags: 2.4.8-jdk7-alpine, 2.4-jdk7-alpine, jdk7-alpine
+GitCommit: 0402f0037b9bdf1d69ce844a381b81304877df3a
+Directory: jdk7-alpine
+
+Tags: 2.4.8-jre7-alpine, 2.4-jre7-alpine, jre7-alpine
+GitCommit: 0402f0037b9bdf1d69ce844a381b81304877df3a
+Directory: jre7-alpine
+
+Tags: 2.4.8-jdk8, 2.4-jdk8, jdk8, 2.4.8-jdk, 2.4-jdk, jdk
+GitCommit: 0402f0037b9bdf1d69ce844a381b81304877df3a
+Directory: jdk8
+
+Tags: 2.4.8-jre8, 2.4-jre8, jre8, 2.4.8-jre, 2.4-jre, jre, 2.4.8, 2.4, latest
+GitCommit: 0402f0037b9bdf1d69ce844a381b81304877df3a
+Directory: jre8
+
+Tags: 2.4.8-jdk8-alpine, 2.4-jdk8-alpine, jdk8-alpine, 2.4.8-jdk-alpine, 2.4-jdk-alpine, jdk-alpine
+GitCommit: 0402f0037b9bdf1d69ce844a381b81304877df3a
+Directory: jdk8-alpine
+
+Tags: 2.4.8-jre8-alpine, 2.4-jre8-alpine, jre8-alpine, 2.4.8-jre-alpine, 2.4-jre-alpine, jre-alpine, 2.4.8-alpine, 2.4-alpine, alpine
+GitCommit: 0402f0037b9bdf1d69ce844a381b81304877df3a
+Directory: jre8-alpine
+
+Tags: 2.4.8-jdk9, 2.4-jdk9, jdk9
+GitCommit: 0402f0037b9bdf1d69ce844a381b81304877df3a
+Directory: jdk9
+
+Tags: 2.4.8-jre9, 2.4-jre9, jre9
+GitCommit: 0402f0037b9bdf1d69ce844a381b81304877df3a
+Directory: jre9


tianon:
Heh, this is one I'm definitely excited to see (https://github.com/tianon/dockerfiles/blob/6be4867fba6ded46515a7fa2b8c8a0a55fa5d8f0/groovy/Dockerfile) :innocent: :smile:

A few initial thoughts/questions:

1. `FROM openjdk:8u111-jre` ought to be less specific, ie `openjdk:8-jre` (same for Alpine) so we can auto-rebuild the image when the base changes
2. please move the `ENV GROOVY_VERSION` bits down as far as they can safely go (probably just before the largest `RUN` where Groovy itself is downloaded), this will help maximize cache
3. any particular reason for the full-version specificity in `bash=4.3.42-r4`?
4. is this SHA256 something published with the upstream project, or something scraped specifically for this `Dockerfile`? (since you are upstream, I'm not too worried either way, but it's useful to note whether that's calculated locally or is an explicit release artifact)
5. since upstream provides detached signatures, it'd be a good idea to verify those as well (see the `Dockerfile` I linked above for an example of how to do something like that)
6. `apk del unzip` in a separate layer from the `apk add` won't actually accomplish anything except making the image larger with very little benefit -- I'd recommend using `apk add --virtual .build-deps unzip` to your main "install" `RUN` line instead followed by `apk del .build-deps` afterwards (and then adding `gnupg` as well for verifying the upstream signatures will be easier)
7. it's probably a good idea to link directly to https://issues.apache.org/jira/browse/GROOVY-7906 in a comment or something (it took me a bit to find the link to figure out what those `sed` lines were working around)
8. IMO it's worth considering using `-O` on your `wget` invocation to minimize the number of places `$GROOVY_VERSION` has to be used (and version-specific filenames on disk), but that's completely personal preference :innocent:

I'm happy to expound on any of these, or turn them into a concrete PR if that makes them easier to review/digest/implement. :+1:

keeganwitt:
Thanks for the feedback!  Sorry for what I'm sure are ignorant questions, but it'll help educate me if you could point me at some links.

**On the specific `FROM` and dependency versions**
I was attempting to make the Dockerfile as close to being a definitive definition as possible, and fully reproducible.  That is, if I build it again a month later, I get the same result. That's also why I picked specific versions of the dependencies I added in the Alpine versions. Is that considered not best practice?  If not, why?

**On moving `GROOVY_VERSION` down**
For my own education, is there any documentation that explains how the caching works?

**On signatures and hashes**
SHAs are not provided currently.  But if the signature matches, is there a reason to use the hash? I switched to the hash to avoid having to install gpg on Alpine, but now that I know about `--virtual` that may no longer be needed. I'm also working to find out if Gradle provides either a signature or hash.

keeganwitt:
On the tests, I have these running currently: https://travis-ci.org/groovy/docker-groovy not sure if we should add them to official images or what.

keeganwitt:
I implemented your suggestions [here](https://github.com/groovy/docker-groovy/tree/recommendations). I wondered if moving CMD to the top was a good idea for caching.  It it looks good, I'll merge in.

tianon:
> I take it the reason is so you can build new images to address JDK vulnerabilities without needing my intervention?

Yeah, definitely -- if the exact version of the dependency is important for the application in question, that's a little different, but otherwise we like to be able to handle security updates (and even just general simple version bumps) without explicit maintainer intervention where possible. :heart:

> Oh, I think I get it -- ...

Yep -- we like to both share the Docker build cache _across_ versions as aggressively as possible, but also want to make sure that simply re-building for a new Groovy version minimizes the amount of actual lines we have to execute. :smile:

> On signatures and hashes

Generally, we're -1 on hashes that you generate by hand while updating the `Dockerfile`, but I think it's fine in this case given that you are upstream (if you want to keep them, up to you).  In my mind, the SHA and the signature are separate (but related) concerns -- the signature verifies that we're getting bits that are blessed by the upstream release process (or at least the folks responsible for that release process), and the SHA verifies that they haven't unexpectedly changed.  Does that help?

> I implemented your suggestions here.

Changes look pretty good, although I'd still leave it up to you whether you'd like to do both `gpg` and SHA256. :+1:

(and explicit CC to @yosifkit for his thoughts/review)

> I wondered if moving CMD to the top was a good idea for caching.

Conceptually it's a little odd, since the `groovy` command doesn't exist until later, but functionally it's fine. :+1:

> On the tests, ...

Regarding tests, we should probably add a simple `groovy-hello-world` test similar to the other "Hello World" tests we've got so far -- I'm happy to write that though if you'd rather (since I'm already familiar with our testing setup here). :+1:

keeganwitt:
> In my mind, the SHA and the signature are separate (but related) concerns -- the signature verifies that we're getting bits that are blessed by the upstream release process (or at least the folks responsible for that release process), and the SHA verifies that they haven't unexpectedly changed.

True. But it's my understanding the detached signature is an encrypted hash, so verifying is a superset of both.  That is both the hash must match (which wouldn't be the case if it were improperly downloaded or tampered with), and encrypted by a trusted key. Unless I'm misunderstanding? I'm not opposed necessarily doing both, I just didn't see that it added any value.

>  I'm happy to write that though if you'd rather (since I'm already familiar with our testing setup here).

Sure, that'd be great! I hadn't looked much at how your testing was configured.

keeganwitt:
Two more questions (if you have any advice):
1. Should I add a symlink in _/usr/bin_?
1. Should I create a volume for _~/.groovy_?

yosifkit:
> Should I add a symlink in /usr/bin?

It looks like you are already adding `${GROOVY_HOME}/bin` to the path, what would the symlink be for?

> Should I create a volume for ~/.groovy?

Volumes are a contentious subject for some.  Tianon's comment sums up our current approach: https://github.com/docker-library/official-images/pull/2437#issuecomment-266578827.

keeganwitt:
> It looks like you are already adding ${GROOVY_HOME}/bin to the path, what would the symlink be for?

No functional reason.  I just thought I'd make sure it wasn't a best practice.

> Volumes are a contentious subject for some. Tianon's comment sums up our current approach: #2437 (comment).

It's a bit of a mixed case. _~/.groovy_ can have configuration in it (for example _~/.groovy/grapeConfig.xml_), but it can also have caches and logs, like _/.groovy/grapes_ and _~/.groovy/groovysh.history_, which might be appropriate for a volume. That's why I'm a little torn.

tianon:
Hmm, yeah -- if it has configuration files, I'd personally lean more towards _not_ having a `VOLUME` so that it's easier for users to use it flexibly (and even more especially since the path is relative to `HOME`, which we can't always guarantee will be consistent), but we'll leave that choice up to your discretion since it's so iffy. :+1: :heart:

keeganwitt:
I'm kind leaning that way too I think after pondering it some more.  The biggest cached stuff is Grapes, which not everyone necessarily uses.  I think we should be good to go on both PRs.  I'll take a look at the stackbrew for the next update.

yosifkit:
```diff
diff --git a/groovy_alpine/Dockerfile b/groovy_alpine/Dockerfile
new file mode 100644
index 0000000..ce9b10d
--- /dev/null
+++ b/groovy_alpine/Dockerfile
@@ -0,0 +1,51 @@
+FROM openjdk:8-jre-alpine
+
+CMD ["groovysh"]
+
+WORKDIR /opt
+ENV GROOVY_HOME /opt/groovy
+ENV PATH "${GROOVY_HOME}/bin:${PATH}"
+
+ENV GROOVY_VERSION 2.4.8
+RUN set -eu \
+	&& echo "Installing dependencies" \
+	&& apk add --no-cache \
+		bash \
+	\
+	&& echo "Installing build dependencies" \
+	&& apk add --no-cache --virtual .build-deps \
+		ca-certificates \
+		gnupg \
+		openssl \
+		unzip \
+	\
+	&& echo "Downloading Groovy" \
+	&& wget -O groovy.zip "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip" \
+	\
+	&& echo "Checking download signature" \
+	&& wget -O groovy.zip.asc "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip.asc" \
+	&& export GNUPGHOME="$(mktemp -d)" \
+	&& wget "http://www.apache.org/dist/groovy/KEYS" \
+	&& gpg --import KEYS \
+	&& gpg --batch --verify groovy.zip.asc groovy.zip \
+	&& rm -r "${GNUPGHOME}" \
+	&& rm groovy.zip.asc KEYS \
+	\
+	&& echo "Installing Groovy" \
+	&& unzip groovy.zip \
+	&& rm groovy.zip \
+	&& mv "groovy-${GROOVY_VERSION}" "${GROOVY_HOME}" \
+	\
+	&& echo "Applying workaround for https://issues.apache.org/jira/browse/GROOVY-7906" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/grape" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovy" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovyc" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovyConsole" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovydoc" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovysh" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/java2groovy" \
+	\
+	&& echo "Cleaning up build dependencies" \
+	&& apk del .build-deps \
+	\
+	&& groovy --version
diff --git a/groovy_jdk-alpine/Dockerfile b/groovy_jdk-alpine/Dockerfile
new file mode 100644
index 0000000..9928979
--- /dev/null
+++ b/groovy_jdk-alpine/Dockerfile
@@ -0,0 +1,51 @@
+FROM openjdk:8-jdk-alpine
+
+CMD ["groovysh"]
+
+WORKDIR /opt
+ENV GROOVY_HOME /opt/groovy
+ENV PATH "${GROOVY_HOME}/bin:${PATH}"
+
+ENV GROOVY_VERSION 2.4.8
+RUN set -eu \
+	&& echo "Installing dependencies" \
+	&& apk add --no-cache \
+		bash \
+	\
+	&& echo "Installing build dependencies" \
+	&& apk add --no-cache --virtual .build-deps \
+		ca-certificates \
+		gnupg \
+		openssl \
+		unzip \
+	\
+	&& echo "Downloading Groovy" \
+	&& wget -O groovy.zip "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip" \
+	\
+	&& echo "Checking download signature" \
+	&& wget -O groovy.zip.asc "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip.asc" \
+	&& export GNUPGHOME="$(mktemp -d)" \
+	&& wget "http://www.apache.org/dist/groovy/KEYS" \
+	&& gpg --import KEYS \
+	&& gpg --batch --verify groovy.zip.asc groovy.zip \
+	&& rm -r "${GNUPGHOME}" \
+	&& rm groovy.zip.asc KEYS \
+	\
+	&& echo "Installing Groovy" \
+	&& unzip groovy.zip \
+	&& rm groovy.zip \
+	&& mv "groovy-${GROOVY_VERSION}" "${GROOVY_HOME}" \
+	\
+	&& echo "Applying workaround for https://issues.apache.org/jira/browse/GROOVY-7906" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/grape" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovy" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovyc" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovyConsole" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovydoc" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovysh" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/java2groovy" \
+	\
+	&& echo "Cleaning up build dependencies" \
+	&& apk del .build-deps \
+	\
+	&& groovy --version
diff --git a/groovy_jdk/Dockerfile b/groovy_jdk/Dockerfile
new file mode 100644
index 0000000..a7afdf5
--- /dev/null
+++ b/groovy_jdk/Dockerfile
@@ -0,0 +1,28 @@
+FROM openjdk:8-jdk
+
+CMD ["groovysh"]
+
+WORKDIR /opt
+ENV GROOVY_HOME /opt/groovy
+ENV PATH "${GROOVY_HOME}/bin:${PATH}"
+
+ENV GROOVY_VERSION 2.4.8
+RUN set -eu \
+	&& echo "Downloading Groovy" \
+	&& wget -nv -O groovy.zip "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip" \
+	\
+	&& echo "Checking download signature" \
+	&& wget -nv -O groovy.zip.asc "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip.asc" \
+	&& export GNUPGHOME="$(mktemp -d)" \
+	&& wget -nv "http://www.apache.org/dist/groovy/KEYS" \
+	&& gpg --import KEYS \
+	&& gpg --batch --verify groovy.zip.asc groovy.zip \
+	&& rm -r "${GNUPGHOME}" \
+	&& rm groovy.zip.asc KEYS \
+	\
+	&& echo "Installing Groovy" \
+	&& unzip groovy.zip \
+	&& rm groovy.zip \
+	&& mv "groovy-${GROOVY_VERSION}" "${GROOVY_HOME}" \
+	\
+	&& groovy --version
diff --git a/groovy_jdk7-alpine/Dockerfile b/groovy_jdk7-alpine/Dockerfile
new file mode 100644
index 0000000..13ccc97
--- /dev/null
+++ b/groovy_jdk7-alpine/Dockerfile
@@ -0,0 +1,51 @@
+FROM openjdk:7-jdk-alpine
+
+CMD ["groovysh"]
+
+WORKDIR /opt
+ENV GROOVY_HOME /opt/groovy
+ENV PATH "${GROOVY_HOME}/bin:${PATH}"
+
+ENV GROOVY_VERSION 2.4.8
+RUN set -eu \
+	&& echo "Installing dependencies" \
+	&& apk add --no-cache \
+		bash \
+	\
+	&& echo "Installing build dependencies" \
+	&& apk add --no-cache --virtual .build-deps \
+		ca-certificates \
+		gnupg \
+		openssl \
+		unzip \
+	\
+	&& echo "Downloading Groovy" \
+	&& wget -O groovy.zip "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip" \
+	\
+	&& echo "Checking download signature" \
+	&& wget -O groovy.zip.asc "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip.asc" \
+	&& export GNUPGHOME="$(mktemp -d)" \
+	&& wget "http://www.apache.org/dist/groovy/KEYS" \
+	&& gpg --import KEYS \
+	&& gpg --batch --verify groovy.zip.asc groovy.zip \
+	&& rm -r "${GNUPGHOME}" \
+	&& rm groovy.zip.asc KEYS \
+	\
+	&& echo "Installing Groovy" \
+	&& unzip groovy.zip \
+	&& rm groovy.zip \
+	&& mv "groovy-${GROOVY_VERSION}" "${GROOVY_HOME}" \
+	\
+	&& echo "Applying workaround for https://issues.apache.org/jira/browse/GROOVY-7906" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/grape" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovy" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovyc" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovyConsole" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovydoc" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovysh" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/java2groovy" \
+	\
+	&& echo "Cleaning up build dependencies" \
+	&& apk del .build-deps \
+	\
+	&& groovy --version
diff --git a/groovy_jdk7/Dockerfile b/groovy_jdk7/Dockerfile
new file mode 100644
index 0000000..228fd29
--- /dev/null
+++ b/groovy_jdk7/Dockerfile
@@ -0,0 +1,28 @@
+FROM openjdk:7-jdk
+
+CMD ["groovysh"]
+
+WORKDIR /opt
+ENV GROOVY_HOME /opt/groovy
+ENV PATH "${GROOVY_HOME}/bin:${PATH}"
+
+ENV GROOVY_VERSION 2.4.8
+RUN set -eu \
+	&& echo "Downloading Groovy" \
+	&& wget -nv -O groovy.zip "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip" \
+	\
+	&& echo "Checking download signature" \
+	&& wget -nv -O groovy.zip.asc "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip.asc" \
+	&& export GNUPGHOME="$(mktemp -d)" \
+	&& wget -nv "http://www.apache.org/dist/groovy/KEYS" \
+	&& gpg --import KEYS \
+	&& gpg --batch --verify groovy.zip.asc groovy.zip \
+	&& rm -r "${GNUPGHOME}" \
+	&& rm groovy.zip.asc KEYS \
+	\
+	&& echo "Installing Groovy" \
+	&& unzip groovy.zip \
+	&& rm groovy.zip \
+	&& mv "groovy-${GROOVY_VERSION}" "${GROOVY_HOME}" \
+	\
+	&& groovy --version
diff --git a/groovy_jdk9/Dockerfile b/groovy_jdk9/Dockerfile
new file mode 100644
index 0000000..894660d
--- /dev/null
+++ b/groovy_jdk9/Dockerfile
@@ -0,0 +1,40 @@
+FROM openjdk:9-jdk
+
+CMD ["groovysh"]
+
+WORKDIR /opt
+ENV GROOVY_HOME /opt/groovy
+ENV PATH "${GROOVY_HOME}/bin:${PATH}"
+
+ENV GROOVY_VERSION 2.4.8
+RUN set -eu \
+	&& echo "Downloading Groovy" \
+	&& wget -nv -O groovy.zip "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip" \
+	\
+	&& echo "Installing build dependencies" \
+	&& apt-get update \
+	&& apt-get update && apt-get install -y --no-install-recommends \
+		gnupg \
+	&& rm -rf /var/lib/apt/lists/* \
+	\
+	&& echo "Checking download signature" \
+	&& wget -nv -O groovy.zip.asc "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip.asc" \
+	&& export GNUPGHOME="$(mktemp -d)" \
+	&& wget -nv "http://www.apache.org/dist/groovy/KEYS" \
+	&& gpg --import KEYS \
+	&& gpg --batch --verify groovy.zip.asc groovy.zip \
+	&& rm -rf "${GNUPGHOME}" \
+	&& rm groovy.zip.asc KEYS \
+	\
+	&& echo "Installing Groovy" \
+	&& unzip groovy.zip \
+	&& rm groovy.zip \
+	&& mv "groovy-${GROOVY_VERSION}" "${GROOVY_HOME}" \
+	\
+	&& echo "Cleaning up build dependencies" \
+	&& echo $(apt-mark showauto) \
+	&& apt-get remove -y --purge \
+		gnupg \
+	&& apt-get autoremove -y --purge \
+	\
+	&& groovy --version
diff --git a/groovy_jre7-alpine/Dockerfile b/groovy_jre7-alpine/Dockerfile
new file mode 100644
index 0000000..c76941f
--- /dev/null
+++ b/groovy_jre7-alpine/Dockerfile
@@ -0,0 +1,51 @@
+FROM openjdk:7-jre-alpine
+
+CMD ["groovysh"]
+
+WORKDIR /opt
+ENV GROOVY_HOME /opt/groovy
+ENV PATH "${GROOVY_HOME}/bin:${PATH}"
+
+ENV GROOVY_VERSION 2.4.8
+RUN set -eu \
+	&& echo "Installing dependencies" \
+	&& apk add --no-cache \
+		bash \
+	\
+	&& echo "Installing build dependencies" \
+	&& apk add --no-cache --virtual .build-deps \
+		ca-certificates \
+		gnupg \
+		openssl \
+		unzip \
+	\
+	&& echo "Downloading Groovy" \
+	&& wget -O groovy.zip "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip" \
+	\
+	&& echo "Checking download signature" \
+	&& wget -O groovy.zip.asc "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip.asc" \
+	&& export GNUPGHOME="$(mktemp -d)" \
+	&& wget "http://www.apache.org/dist/groovy/KEYS" \
+	&& gpg --import KEYS \
+	&& gpg --batch --verify groovy.zip.asc groovy.zip \
+	&& rm -r "${GNUPGHOME}" \
+	&& rm groovy.zip.asc KEYS \
+	\
+	&& echo "Installing Groovy" \
+	&& unzip groovy.zip \
+	&& rm groovy.zip \
+	&& mv "groovy-${GROOVY_VERSION}" "${GROOVY_HOME}" \
+	\
+	&& echo "Applying workaround for https://issues.apache.org/jira/browse/GROOVY-7906" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/grape" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovy" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovyc" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovyConsole" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovydoc" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovysh" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/java2groovy" \
+	\
+	&& echo "Cleaning up build dependencies" \
+	&& apk del .build-deps \
+	\
+	&& groovy --version
diff --git a/groovy_jre7/Dockerfile b/groovy_jre7/Dockerfile
new file mode 100644
index 0000000..24ea3be
--- /dev/null
+++ b/groovy_jre7/Dockerfile
@@ -0,0 +1,28 @@
+FROM openjdk:7-jre
+
+CMD ["groovysh"]
+
+WORKDIR /opt
+ENV GROOVY_HOME /opt/groovy
+ENV PATH "${GROOVY_HOME}/bin:${PATH}"
+
+ENV GROOVY_VERSION 2.4.8
+RUN set -eu \
+	&& echo "Downloading Groovy" \
+	&& wget -nv -O groovy.zip "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip" \
+	\
+	&& echo "Checking download signature" \
+	&& wget -nv -O groovy.zip.asc "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip.asc" \
+	&& export GNUPGHOME="$(mktemp -d)" \
+	&& wget -nv "http://www.apache.org/dist/groovy/KEYS" \
+	&& gpg --import KEYS \
+	&& gpg --batch --verify groovy.zip.asc groovy.zip \
+	&& rm -r "${GNUPGHOME}" \
+	&& rm groovy.zip.asc KEYS \
+	\
+	&& echo "Installing Groovy" \
+	&& unzip groovy.zip \
+	&& rm groovy.zip \
+	&& mv "groovy-${GROOVY_VERSION}" "${GROOVY_HOME}" \
+	\
+	&& groovy --version
diff --git a/groovy_jre9/Dockerfile b/groovy_jre9/Dockerfile
new file mode 100644
index 0000000..65adeac
--- /dev/null
+++ b/groovy_jre9/Dockerfile
@@ -0,0 +1,40 @@
+FROM openjdk:9-jre
+
+CMD ["groovysh"]
+
+WORKDIR /opt
+ENV GROOVY_HOME /opt/groovy
+ENV PATH "${GROOVY_HOME}/bin:${PATH}"
+
+ENV GROOVY_VERSION 2.4.8
+RUN set -eu \
+	&& echo "Downloading Groovy" \
+	&& wget -nv -O groovy.zip "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip" \
+	\
+	&& echo "Installing build dependencies" \
+	&& apt-get update \
+	&& apt-get update && apt-get install -y --no-install-recommends \
+		gnupg \
+	&& rm -rf /var/lib/apt/lists/* \
+	\
+	&& echo "Checking download signature" \
+	&& wget -nv -O groovy.zip.asc "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip.asc" \
+	&& export GNUPGHOME="$(mktemp -d)" \
+	&& wget -nv "http://www.apache.org/dist/groovy/KEYS" \
+	&& gpg --import KEYS \
+	&& gpg --batch --verify groovy.zip.asc groovy.zip \
+	&& rm -rf "${GNUPGHOME}" \
+	&& rm groovy.zip.asc KEYS \
+	\
+	&& echo "Installing Groovy" \
+	&& unzip groovy.zip \
+	&& rm groovy.zip \
+	&& mv "groovy-${GROOVY_VERSION}" "${GROOVY_HOME}" \
+	\
+	&& echo "Cleaning up build dependencies" \
+	&& echo $(apt-mark showauto) \
+	&& apt-get remove -y --purge \
+		gnupg \
+	&& apt-get autoremove -y --purge \
+	\
+	&& groovy --version
diff --git a/groovy_latest/Dockerfile b/groovy_latest/Dockerfile
new file mode 100644
index 0000000..b788fbb
--- /dev/null
+++ b/groovy_latest/Dockerfile
@@ -0,0 +1,28 @@
+FROM openjdk:8-jre
+
+CMD ["groovysh"]
+
+WORKDIR /opt
+ENV GROOVY_HOME /opt/groovy
+ENV PATH "${GROOVY_HOME}/bin:${PATH}"
+
+ENV GROOVY_VERSION 2.4.8
+RUN set -eu \
+	&& echo "Downloading Groovy" \
+	&& wget -nv -O groovy.zip "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip" \
+	\
+	&& echo "Checking download signature" \
+	&& wget -nv -O groovy.zip.asc "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip.asc" \
+	&& export GNUPGHOME="$(mktemp -d)" \
+	&& wget -nv "http://www.apache.org/dist/groovy/KEYS" \
+	&& gpg --import KEYS \
+	&& gpg --batch --verify groovy.zip.asc groovy.zip \
+	&& rm -r "${GNUPGHOME}" \
+	&& rm groovy.zip.asc KEYS \
+	\
+	&& echo "Installing Groovy" \
+	&& unzip groovy.zip \
+	&& rm groovy.zip \
+	&& mv "groovy-${GROOVY_VERSION}" "${GROOVY_HOME}" \
+	\
+	&& groovy --version
```

yosifkit:
Rather than downloading the keys during the build, we recommend embedding the full fingerprints in the Dockerfile and pull them from a keyserver like `ha.pool.sks-keyservers.net`.  This gives a second layer of verification so that the keys and build artifacts cannot be easily intercepted during the build process.

keeganwitt:
OK. I've made that change.

yosifkit:
```diff
diff --git a/groovy_alpine/Dockerfile b/groovy_alpine/Dockerfile
new file mode 100644
index 0000000..e8983e5
--- /dev/null
+++ b/groovy_alpine/Dockerfile
@@ -0,0 +1,51 @@
+FROM openjdk:8-jre-alpine
+
+CMD ["groovysh"]
+
+WORKDIR /opt
+ENV GROOVY_HOME /opt/groovy
+ENV PATH "${GROOVY_HOME}/bin:${PATH}"
+
+ENV GROOVY_VERSION 2.4.8
+RUN set -eu \
+	&& echo "Installing dependencies" \
+	&& apk add --no-cache \
+		bash \
+	\
+	&& echo "Installing build dependencies" \
+	&& apk add --no-cache --virtual .build-deps \
+		ca-certificates \
+		gnupg \
+		openssl \
+		unzip \
+	\
+	&& echo "Downloading Groovy" \
+	&& wget -O groovy.zip "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip" \
+	\
+	&& echo "Checking download signature" \
+	&& wget -O groovy.zip.asc "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip.asc" \
+	&& export GNUPGHOME="$(mktemp -d)" \
+	&& echo "Importing keys listed in http://www.apache.org/dist/groovy/KEYS from key server" \
+	&& gpg --keyserver ha.pool.sks-keyservers.net --recv-keys "0x41321490758AAD6F" "0x825C06C827AF6B66" "0x6A65176A0FB1CD0B" \
+	&& gpg --batch --verify groovy.zip.asc groovy.zip \
+	&& rm -r "${GNUPGHOME}" \
+	&& rm groovy.zip.asc \
+	\
+	&& echo "Installing Groovy" \
+	&& unzip groovy.zip \
+	&& rm groovy.zip \
+	&& mv "groovy-${GROOVY_VERSION}" "${GROOVY_HOME}" \
+	\
+	&& echo "Applying workaround for https://issues.apache.org/jira/browse/GROOVY-7906" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/grape" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovy" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovyc" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovyConsole" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovydoc" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovysh" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/java2groovy" \
+	\
+	&& echo "Cleaning up build dependencies" \
+	&& apk del .build-deps \
+	\
+	&& groovy --version
diff --git a/groovy_jdk-alpine/Dockerfile b/groovy_jdk-alpine/Dockerfile
new file mode 100644
index 0000000..15f9143
--- /dev/null
+++ b/groovy_jdk-alpine/Dockerfile
@@ -0,0 +1,51 @@
+FROM openjdk:8-jdk-alpine
+
+CMD ["groovysh"]
+
+WORKDIR /opt
+ENV GROOVY_HOME /opt/groovy
+ENV PATH "${GROOVY_HOME}/bin:${PATH}"
+
+ENV GROOVY_VERSION 2.4.8
+RUN set -eu \
+	&& echo "Installing dependencies" \
+	&& apk add --no-cache \
+		bash \
+	\
+	&& echo "Installing build dependencies" \
+	&& apk add --no-cache --virtual .build-deps \
+		ca-certificates \
+		gnupg \
+		openssl \
+		unzip \
+	\
+	&& echo "Downloading Groovy" \
+	&& wget -O groovy.zip "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip" \
+	\
+	&& echo "Checking download signature" \
+	&& wget -O groovy.zip.asc "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip.asc" \
+	&& export GNUPGHOME="$(mktemp -d)" \
+	&& echo "Importing keys listed in http://www.apache.org/dist/groovy/KEYS from key server" \
+	&& gpg --keyserver ha.pool.sks-keyservers.net --recv-keys "0x41321490758AAD6F" "0x825C06C827AF6B66" "0x6A65176A0FB1CD0B" \
+	&& gpg --batch --verify groovy.zip.asc groovy.zip \
+	&& rm -r "${GNUPGHOME}" \
+	&& rm groovy.zip.asc \
+	\
+	&& echo "Installing Groovy" \
+	&& unzip groovy.zip \
+	&& rm groovy.zip \
+	&& mv "groovy-${GROOVY_VERSION}" "${GROOVY_HOME}" \
+	\
+	&& echo "Applying workaround for https://issues.apache.org/jira/browse/GROOVY-7906" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/grape" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovy" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovyc" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovyConsole" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovydoc" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovysh" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/java2groovy" \
+	\
+	&& echo "Cleaning up build dependencies" \
+	&& apk del .build-deps \
+	\
+	&& groovy --version
diff --git a/groovy_jdk/Dockerfile b/groovy_jdk/Dockerfile
new file mode 100644
index 0000000..f1a092d
--- /dev/null
+++ b/groovy_jdk/Dockerfile
@@ -0,0 +1,28 @@
+FROM openjdk:8-jdk
+
+CMD ["groovysh"]
+
+WORKDIR /opt
+ENV GROOVY_HOME /opt/groovy
+ENV PATH "${GROOVY_HOME}/bin:${PATH}"
+
+ENV GROOVY_VERSION 2.4.8
+RUN set -eu \
+	&& echo "Downloading Groovy" \
+	&& wget -nv -O groovy.zip "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip" \
+	\
+	&& echo "Checking download signature" \
+	&& wget -nv -O groovy.zip.asc "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip.asc" \
+	&& export GNUPGHOME="$(mktemp -d)" \
+	&& echo "Importing keys listed in http://www.apache.org/dist/groovy/KEYS from key server" \
+	&& gpg --keyserver ha.pool.sks-keyservers.net --recv-keys "0x41321490758AAD6F" "0x825C06C827AF6B66" "0x6A65176A0FB1CD0B" \
+	&& gpg --batch --verify groovy.zip.asc groovy.zip \
+	&& rm -r "${GNUPGHOME}" \
+	&& rm groovy.zip.asc \
+	\
+	&& echo "Installing Groovy" \
+	&& unzip groovy.zip \
+	&& rm groovy.zip \
+	&& mv "groovy-${GROOVY_VERSION}" "${GROOVY_HOME}" \
+	\
+	&& groovy --version
diff --git a/groovy_jdk7-alpine/Dockerfile b/groovy_jdk7-alpine/Dockerfile
new file mode 100644
index 0000000..c1ade4b
--- /dev/null
+++ b/groovy_jdk7-alpine/Dockerfile
@@ -0,0 +1,51 @@
+FROM openjdk:7-jdk-alpine
+
+CMD ["groovysh"]
+
+WORKDIR /opt
+ENV GROOVY_HOME /opt/groovy
+ENV PATH "${GROOVY_HOME}/bin:${PATH}"
+
+ENV GROOVY_VERSION 2.4.8
+RUN set -eu \
+	&& echo "Installing dependencies" \
+	&& apk add --no-cache \
+		bash \
+	\
+	&& echo "Installing build dependencies" \
+	&& apk add --no-cache --virtual .build-deps \
+		ca-certificates \
+		gnupg \
+		openssl \
+		unzip \
+	\
+	&& echo "Downloading Groovy" \
+	&& wget -O groovy.zip "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip" \
+	\
+	&& echo "Checking download signature" \
+	&& wget -O groovy.zip.asc "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip.asc" \
+	&& export GNUPGHOME="$(mktemp -d)" \
+	&& echo "Importing keys listed in http://www.apache.org/dist/groovy/KEYS from key server" \
+	&& gpg --keyserver ha.pool.sks-keyservers.net --recv-keys "0x41321490758AAD6F" "0x825C06C827AF6B66" "0x6A65176A0FB1CD0B" \
+	&& gpg --batch --verify groovy.zip.asc groovy.zip \
+	&& rm -r "${GNUPGHOME}" \
+	&& rm groovy.zip.asc \
+	\
+	&& echo "Installing Groovy" \
+	&& unzip groovy.zip \
+	&& rm groovy.zip \
+	&& mv "groovy-${GROOVY_VERSION}" "${GROOVY_HOME}" \
+	\
+	&& echo "Applying workaround for https://issues.apache.org/jira/browse/GROOVY-7906" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/grape" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovy" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovyc" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovyConsole" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovydoc" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovysh" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/java2groovy" \
+	\
+	&& echo "Cleaning up build dependencies" \
+	&& apk del .build-deps \
+	\
+	&& groovy --version
diff --git a/groovy_jdk7/Dockerfile b/groovy_jdk7/Dockerfile
new file mode 100644
index 0000000..be78aaa
--- /dev/null
+++ b/groovy_jdk7/Dockerfile
@@ -0,0 +1,28 @@
+FROM openjdk:7-jdk
+
+CMD ["groovysh"]
+
+WORKDIR /opt
+ENV GROOVY_HOME /opt/groovy
+ENV PATH "${GROOVY_HOME}/bin:${PATH}"
+
+ENV GROOVY_VERSION 2.4.8
+RUN set -eu \
+	&& echo "Downloading Groovy" \
+	&& wget -nv -O groovy.zip "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip" \
+	\
+	&& echo "Checking download signature" \
+	&& wget -nv -O groovy.zip.asc "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip.asc" \
+	&& export GNUPGHOME="$(mktemp -d)" \
+	&& echo "Importing keys listed in http://www.apache.org/dist/groovy/KEYS from key server" \
+	&& gpg --keyserver ha.pool.sks-keyservers.net --recv-keys "0x41321490758AAD6F" "0x825C06C827AF6B66" "0x6A65176A0FB1CD0B" \
+	&& gpg --batch --verify groovy.zip.asc groovy.zip \
+	&& rm -r "${GNUPGHOME}" \
+	&& rm groovy.zip.asc \
+	\
+	&& echo "Installing Groovy" \
+	&& unzip groovy.zip \
+	&& rm groovy.zip \
+	&& mv "groovy-${GROOVY_VERSION}" "${GROOVY_HOME}" \
+	\
+	&& groovy --version
diff --git a/groovy_jdk9/Dockerfile b/groovy_jdk9/Dockerfile
new file mode 100644
index 0000000..011c0c8
--- /dev/null
+++ b/groovy_jdk9/Dockerfile
@@ -0,0 +1,42 @@
+FROM openjdk:9-jdk
+
+CMD ["groovysh"]
+
+WORKDIR /opt
+ENV GROOVY_HOME /opt/groovy
+ENV PATH "${GROOVY_HOME}/bin:${PATH}"
+
+ENV GROOVY_VERSION 2.4.8
+RUN set -eu \
+	&& echo "Downloading Groovy" \
+	&& wget -nv -O groovy.zip "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip" \
+	\
+	&& echo "Installing build dependencies" \
+	&& apt-get update \
+	&& apt-get update && apt-get install -y --no-install-recommends \
+		dirmngr \
+		gnupg \
+	&& rm -r /var/lib/apt/lists/* \
+	\
+	&& echo "Checking download signature" \
+	&& wget -nv -O groovy.zip.asc "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip.asc" \
+	&& export GNUPGHOME="$(mktemp -d)" \
+	&& echo "Importing keys listed in http://www.apache.org/dist/groovy/KEYS from key server" \
+	&& gpg --keyserver ha.pool.sks-keyservers.net --recv-keys "0x41321490758AAD6F" "0x825C06C827AF6B66" "0x6A65176A0FB1CD0B" \
+	&& gpg --batch --verify groovy.zip.asc groovy.zip \
+	&& rm -rf "${GNUPGHOME}" \
+	&& rm groovy.zip.asc \
+	\
+	&& echo "Installing Groovy" \
+	&& unzip groovy.zip \
+	&& rm groovy.zip \
+	&& mv "groovy-${GROOVY_VERSION}" "${GROOVY_HOME}" \
+	\
+	&& echo "Cleaning up build dependencies" \
+	&& echo $(apt-mark showauto) \
+	&& apt-get remove -y --purge \
+		dirmngr \
+		gnupg \
+	&& apt-get autoremove -y --purge \
+	\
+	&& groovy --version
diff --git a/groovy_jre7-alpine/Dockerfile b/groovy_jre7-alpine/Dockerfile
new file mode 100644
index 0000000..5c4b105
--- /dev/null
+++ b/groovy_jre7-alpine/Dockerfile
@@ -0,0 +1,51 @@
+FROM openjdk:7-jre-alpine
+
+CMD ["groovysh"]
+
+WORKDIR /opt
+ENV GROOVY_HOME /opt/groovy
+ENV PATH "${GROOVY_HOME}/bin:${PATH}"
+
+ENV GROOVY_VERSION 2.4.8
+RUN set -eu \
+	&& echo "Installing dependencies" \
+	&& apk add --no-cache \
+		bash \
+	\
+	&& echo "Installing build dependencies" \
+	&& apk add --no-cache --virtual .build-deps \
+		ca-certificates \
+		gnupg \
+		openssl \
+		unzip \
+	\
+	&& echo "Downloading Groovy" \
+	&& wget -O groovy.zip "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip" \
+	\
+	&& echo "Checking download signature" \
+	&& wget -O groovy.zip.asc "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip.asc" \
+	&& export GNUPGHOME="$(mktemp -d)" \
+	&& echo "Importing keys listed in http://www.apache.org/dist/groovy/KEYS from key server" \
+	&& gpg --keyserver ha.pool.sks-keyservers.net --recv-keys "0x41321490758AAD6F" "0x825C06C827AF6B66" "0x6A65176A0FB1CD0B" \
+	&& gpg --batch --verify groovy.zip.asc groovy.zip \
+	&& rm -r "${GNUPGHOME}" \
+	&& rm groovy.zip.asc \
+	\
+	&& echo "Installing Groovy" \
+	&& unzip groovy.zip \
+	&& rm groovy.zip \
+	&& mv "groovy-${GROOVY_VERSION}" "${GROOVY_HOME}" \
+	\
+	&& echo "Applying workaround for https://issues.apache.org/jira/browse/GROOVY-7906" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/grape" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovy" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovyc" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovyConsole" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovydoc" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/groovysh" \
+	&& sed -i "s|#!/bin/sh|#!/bin/bash|" "${GROOVY_HOME}/bin/java2groovy" \
+	\
+	&& echo "Cleaning up build dependencies" \
+	&& apk del .build-deps \
+	\
+	&& groovy --version
diff --git a/groovy_jre7/Dockerfile b/groovy_jre7/Dockerfile
new file mode 100644
index 0000000..07a1da7
--- /dev/null
+++ b/groovy_jre7/Dockerfile
@@ -0,0 +1,28 @@
+FROM openjdk:7-jre
+
+CMD ["groovysh"]
+
+WORKDIR /opt
+ENV GROOVY_HOME /opt/groovy
+ENV PATH "${GROOVY_HOME}/bin:${PATH}"
+
+ENV GROOVY_VERSION 2.4.8
+RUN set -eu \
+	&& echo "Downloading Groovy" \
+	&& wget -nv -O groovy.zip "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip" \
+	\
+	&& echo "Checking download signature" \
+	&& wget -nv -O groovy.zip.asc "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip.asc" \
+	&& export GNUPGHOME="$(mktemp -d)" \
+	&& echo "Importing keys listed in http://www.apache.org/dist/groovy/KEYS from key server" \
+	&& gpg --keyserver ha.pool.sks-keyservers.net --recv-keys "0x41321490758AAD6F" "0x825C06C827AF6B66" "0x6A65176A0FB1CD0B" \
+	&& gpg --batch --verify groovy.zip.asc groovy.zip \
+	&& rm -r "${GNUPGHOME}" \
+	&& rm groovy.zip.asc \
+	\
+	&& echo "Installing Groovy" \
+	&& unzip groovy.zip \
+	&& rm groovy.zip \
+	&& mv "groovy-${GROOVY_VERSION}" "${GROOVY_HOME}" \
+	\
+	&& groovy --version
diff --git a/groovy_jre9/Dockerfile b/groovy_jre9/Dockerfile
new file mode 100644
index 0000000..ecfc7b3
--- /dev/null
+++ b/groovy_jre9/Dockerfile
@@ -0,0 +1,42 @@
+FROM openjdk:9-jre
+
+CMD ["groovysh"]
+
+WORKDIR /opt
+ENV GROOVY_HOME /opt/groovy
+ENV PATH "${GROOVY_HOME}/bin:${PATH}"
+
+ENV GROOVY_VERSION 2.4.8
+RUN set -eu \
+	&& echo "Downloading Groovy" \
+	&& wget -nv -O groovy.zip "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip" \
+	\
+	&& echo "Installing build dependencies" \
+	&& apt-get update \
+	&& apt-get update && apt-get install -y --no-install-recommends \
+		dirmngr \
+		gnupg \
+	&& rm -r /var/lib/apt/lists/* \
+	\
+	&& echo "Checking download signature" \
+	&& wget -nv -O groovy.zip.asc "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip.asc" \
+	&& export GNUPGHOME="$(mktemp -d)" \
+	&& echo "Importing keys listed in http://www.apache.org/dist/groovy/KEYS from key server" \
+	&& gpg --keyserver ha.pool.sks-keyservers.net --recv-keys "0x41321490758AAD6F" "0x825C06C827AF6B66" "0x6A65176A0FB1CD0B" \
+	&& gpg --batch --verify groovy.zip.asc groovy.zip \
+	&& rm -rf "${GNUPGHOME}" \
+	&& rm groovy.zip.asc \
+	\
+	&& echo "Installing Groovy" \
+	&& unzip groovy.zip \
+	&& rm groovy.zip \
+	&& mv "groovy-${GROOVY_VERSION}" "${GROOVY_HOME}" \
+	\
+	&& echo "Cleaning up build dependencies" \
+	&& echo $(apt-mark showauto) \
+	&& apt-get remove -y --purge \
+		 dirmngr \
+		 gnupg \
+	&& apt-get autoremove -y --purge \
+	\
+	&& groovy --version
diff --git a/groovy_latest/Dockerfile b/groovy_latest/Dockerfile
new file mode 100644
index 0000000..0ff2ecf
--- /dev/null
+++ b/groovy_latest/Dockerfile
@@ -0,0 +1,28 @@
+FROM openjdk:8-jre
+
+CMD ["groovysh"]
+
+WORKDIR /opt
+ENV GROOVY_HOME /opt/groovy
+ENV PATH "${GROOVY_HOME}/bin:${PATH}"
+
+ENV GROOVY_VERSION 2.4.8
+RUN set -eu \
+	&& echo "Downloading Groovy" \
+	&& wget -nv -O groovy.zip "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip" \
+	\
+	&& echo "Checking download signature" \
+	&& wget -nv -O groovy.zip.asc "https://dist.apache.org/repos/dist/release/groovy/${GROOVY_VERSION}/distribution/apache-groovy-binary-${GROOVY_VERSION}.zip.asc" \
+	&& export GNUPGHOME="$(mktemp -d)" \
+	&& echo "Importing keys listed in http://www.apache.org/dist/groovy/KEYS from key server" \
+	&& gpg --keyserver ha.pool.sks-keyservers.net --recv-keys "0x41321490758AAD6F" "0x825C06C827AF6B66" "0x6A65176A0FB1CD0B" \
+	&& gpg --batch --verify groovy.zip.asc groovy.zip \
+	&& rm -r "${GNUPGHOME}" \
+	&& rm groovy.zip.asc \
+	\
+	&& echo "Installing Groovy" \
+	&& unzip groovy.zip \
+	&& rm groovy.zip \
+	&& mv "groovy-${GROOVY_VERSION}" "${GROOVY_HOME}" \
+	\
+	&& groovy --version
```

Starefossen:
Very much looking forward to this release 😄

keeganwitt:
One more question: do you think I should create a non-root user to run the Groovy processes as?

tianon:
For language base images I'm usually on the fence -- personally, I think that's the responsibility of the users of the image in most cases, but it really is up to you.  If you do choose to do so, I'd recommend you do so soon though (`node` is still dealing with a little bit of fallout from their adding a `node` user a bit after the fact, since it breaks some `Dockerfile`s that are `FROM node`).

keeganwitt:
If I do add, is it a best practice to also set the gid and uid manually?

tianon:
Definitely, even if you just set them explicitly to what's auto chosen so
that they stay consistent (both for users taking advantage of them and for
users who are doing their own thing around them).

PR #2633, opened 2017-02-07
Update ilianaw's Maintainers entry

tianon:
cc @ilianaw :heart:


ilianaw:
SGTM but I'm also planning on updating this image sometime today too :)

tianon:
We'll look forward to it! :+1:

tianon:
(Going to merge this just so we have this change in place regardless)

PR #2749, opened 2017-03-15
update centos monthly images as of 20170315

jperrin:


jperrin:
closing. missed a fix I needed to push in this update. will submit a corrected PR momentarily.

PR #2802, opened 2017-03-30
Update mongo-basics test to use a seccomp profile to get numactl access

yosifkit:
Uses Docker's default seccomp profile and adds `set_mempolicy` to the always allowed section.

Doesn't guarantee that the mongo test will always use numa, but it should now be available if the host is numa enabled.

### DIFF
--- a/test/tests/mongo-basics/run.sh
+++ b/test/tests/mongo-basics/run.sh
@@ -18,8 +18,8 @@ EOF
 
 # need set_mempolicy syscall to be able to do numactl for mongodb
 # if "set_mempolicy" is not in the always allowed list, add it
-extraSeccomp="$(echo "$seccomp" | docker run -i --rm $jqImage --tab '
-.syscalls[] |= if (
+extraSeccomp="$(echo "$seccomp" | docker run -i --rm "$jqImage" --tab '
+	.syscalls[] |= if (
 		.action == "SCMP_ACT_ALLOW"
 		and .args == []
 		and .comment == ""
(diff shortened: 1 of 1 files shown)


test/tests/mongo-basics/run.sh
```diff
@@ -3,12 +3,46 @@ set -e
...
+FROM alpine:3.5
+
+RUN apk add --no-cache jq
+
+ENTRYPOINT ["jq"]
+EOF
+
+# need set_mempolicy syscall to be able to do numactl for mongodb
+# if "set_mempolicy" is not in the always allowed list, add it
+extraSeccomp="$(echo "$seccomp" | docker run -i --rm $jqImage --tab '
```
tianon: `$jqImage` here ought to be quoted (ala, `"$jqImage"`)

test/tests/mongo-basics/run.sh
```diff
@@ -3,12 +3,46 @@ set -e
...
+.syscalls[] |= if (
```
tianon: Can you indent this line once? :pray:

test/tests/mongo-basics/run.sh
```diff
@@ -3,12 +3,46 @@ set -e
...
+')"
+
 cname="mongo-container-$RANDOM-$RANDOM"
-cid="$(docker run -d --name "$cname" "$image")"
+cid="$(docker run -d --security-opt seccomp=<(echo "$extraSeccomp") --name "$cname" "$image")"
 trap "docker rm -vf $cid > /dev/null" EXIT
 
 mongo() {
-	docker run --rm -i --link "$cname":mongo --entrypoint mongo "$image" --host mongo "$@"
+	docker run --rm -i --security-opt seccomp=<(echo "$extraSeccomp") --link "$cname":mongo --entrypoint mongo "$image" --host mongo "$@"
```
tianon: Oh, we should probably not use `--entrypoint` here so that `numactl` might actually be used. :sweat_smile:

PR #2825, opened 2017-04-06
Update piwik

infiniteproject:
bump version to 3.0.3


yosifkit:
Closing since #2826 is the right one

PR #2830, opened 2017-04-06
Bump docker version in Dockerfile to use supported release

yosifkit:
https://github.com/docker-library/docker/pull/48 reminded me that `1.13` was going away and then I saw this was on `1.10`.  This should help us see any new docker-related issues when building images.


tianon:
Hah, nutty -- I like it. :+1:

PR #2841, opened 2017-04-10
Update docker-library images

arpitjindal97:
moved 'latest' tag to 9.0 non-alpine version

### DIFF
--- a/library/tomcat
+++ b/library/tomcat
@@ -28,7 +28,7 @@ Tags: 7.0.77-jre8-alpine, 7.0-jre8-alpine, 7-jre8-alpine
 GitCommit: 7a16673434d629938f064b5aed86df44ee49b53a
 Directory: 7/jre8-alpine
 
-Tags: 8.0.43-jre7, 8.0-jre7, 8-jre7, jre7, 8.0.43, 8.0, 8, latest
+Tags: 8.0.43-jre7, 8.0-jre7, 8-jre7, jre7, 8.0.43, 8.0, 8
 GitCommit: 6be7d97a528019fd7cedb0f3ef0dca674713512b
 Directory: 8.0/jre7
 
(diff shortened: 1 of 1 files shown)


tianon:
This needs to be updated over in https://github.com/docker-library/tomcat/blob/78b5ae4bf558d626859cbb63b2837fd6980ca1f1/generate-stackbrew-library.sh first :+1: (which is what this file is generated from)

However, the reason this is still pointing at `8.0` rather than `8.5` or `9.0` is that `9.0.0.M19` (the latest `9.0` release) is still an alpha, and last I'd checked, `8.5` was as well, but it would appear that `8.5` is considered "stable" now, and we should update the `[8.0]='8 latest'` line to instead be `[8.5]='8 latest'`! :+1: :tada:

(See https://tomcat.apache.org/, especially "Tomcat 9.0.0.M19 (alpha) Released" and "Tomcat 8.5.13 Released")

tianon:
Looking at https://tomcat.apache.org/oldnews-2016.html, it would appear that `8.5.2` was the last release to be considered of pre-release status ("Tomcat 8.5.2 (beta) Released" followed by "Tomcat 8.5.3 Released") :+1: :metal:

tianon:
I'll give you some time if you'd like to make the PR over there, otherwise I'll likely make it soon-ishy. :+1:

(If you know that you'd either like to, would prefer not to, or don't care, please do let me know so we can get this switched over! :+1: :heart:)

arpitjindal97:
Can we change `[8.0]='jre7'` line to `[8.0]='jre8'` in [this](https://github.com/docker-library/tomcat/blob/78b5ae4bf558d626859cbb63b2837fd6980ca1f1/generate-stackbrew-library.sh) file ?

arpitjindal97:
8.5 is still not stable but we making 8.0 work with jre8 will make no problem i guess 🤘

tianon:
Not sure I understand what you mean -- `tomcat:8.0-jre8` is already available (and has been for quite some time).  What that line dictates is what folks receive when they only specify `tomcat:8.0`, and that's still `jre7` (and likely should stay that way).  If folks are touchy about the particular version of the JRE they're using, they should be using a more specific tag. :+1:

tianon:
I've filed https://github.com/docker-library/tomcat/pull/67 to update the `8` and `latest` aliases to point at `8.5` now. :+1:

arpitjindal97:
Thanks mate

arpitjindal97:
`docker` is still fetching 8.0.43. Is it because of [this](https://github.com/docker-library/repo-info/blob/master/repos/tomcat/remote/latest.md) file?

tianon:
Yep, see https://github.com/docker-library/official-images/pull/2847 :+1:

PR #2865, opened 2017-04-17
remove centos5 container references, reorder c6 containers

jperrin:
My understanding is that this will leave the c5 containers available, but won't advertise them. is that correct?


tianon:
Yes, that is correct -- we can remove them if you want, but we'd recommend leaving them.  This will simply remove them from the "Supported tags" list and stop us from rebuilding them. :+1:

tianon:
Merging since there's no real change here besides removals and reordering.

PR #2880, opened 2017-04-21
Use python command rather than python3 to allow cross distribution test

dmetzler:
When creating image on variant distrib, `python3` is not always recognized as a command (centos in that case), so the test fails. Using `python` is more generic.

### DIFF
--- a/test/tests/nuxeo-basics/run.sh
+++ b/test/tests/nuxeo-basics/run.sh
@@ -23,9 +23,9 @@ get() {
 			"http://nuxeo:8080/nuxeo/api/v1/$1"
 }
 
-python3() {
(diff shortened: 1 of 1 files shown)


tianon:
This might sometimes pull in Python 2 instead, right?  I'm not savvy enough with the differences to say whether it matters for the short snippet we're actually running, but maybe it's fine? :innocent: :+1:

dmetzler:
yes, it's only used for test to parse JSON payload. Syntax is the same in python 2 or 3 : I've tested it on both versions (centos bundles python2)

tianon:
Excellent, thanks for confirming 👍

PR #2897, opened 2017-04-26
Add LogicalDOC

car031:
We wish to add our Official image for LogicalDOC
Our git repository is logicaldoc/docker-logicaldoc

# Checklist for Review

    - https://github.com/logicaldoc
    - https://github.com/docker-library/docs/pull/896

### DIFF
--- a/library/logicaldoc
+++ b/library/logicaldoc
@@ -1,6 +1,6 @@
 Maintainers: Marco Meschieri <m.meschieri@logicaldoc.com> (@car031),
              Alessandro Gasparini <a.gasparini@logicaldoc.com> (@gasparez15)
-GitRepo: https://github.com/logicaldo/docker-logicaldoc.git
+GitRepo: https://github.com/logicaldoc/docker-logicaldoc.git
 
 Tags: latest, 7.6.4 
 GitCommit: 4a5e6ab215c45ddbde88c296b5093af6333a5364 
--- a/library/logicaldoc
+++ b/library/logicaldoc
@@ -1,0 +1,6 @@
+Maintainers: Marco Meschieri <m.meschieri@logicaldoc.com> (@car031),
+             Alessandro Gasparini <a.gasparini@logicaldoc.com> (@gasparez15)
+GitRepo: https://github.com/logicaldo/docker-logicaldoc.git
+
+Tags: latest, 7.6.4 
+GitCommit: 4a5e6ab215c45ddbde88c296b5093af6333a5364 


Moghedrin:
For reviewer reference: documentation PR at docker-library/docs#896.

car031:
Hi all, why is this image not already included? Has it some problems? Please leave feedback.

tianon:
Sorry @car031, we've usually got a pretty bad backlog of PRs across the docker-library repos, so things get a little lost sometimes. :disappointed:

A few comments after taking a brief look (I'm out-of-office today, but wanted to get you some initial feedback):

- `FROM phusion/baseimage` is definitely a problem
  - > No official images can be derived from, or depend on, non-official images
  - https://github.com/docker-library/official-images/tree/5dfff20bd03ffedf060ddf362be427ecf3a25638#repeatability
  - is there some specific functionality being provided by `phusion/baseimage` that is necessary for making LogicalDOC work in a container?
- the `MAINTAINER` instruction is deprecated, and I'd recommend simply removing it
  - https://docs.docker.com/engine/reference/builder/#maintainer-deprecated
  - https://docs.docker.com/engine/deprecated/#maintainer-in-dockerfile
- `apt-get upgrade` is often going to cause problems -- we prefer to instead update the base image and rebuild
  - > You should avoid RUN apt-get upgrade or dist-upgrade, as many of the “essential” packages from the base images won’t upgrade inside an unprivileged container. If a package contained in the base image is out-of-date, you should contact its maintainers.
  - https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#apt-get
- `RUN apt-get install ...` is going to break sometimes due to Docker cache
  - > Always combine RUN apt-get update with apt-get install in the same RUN statement
  - https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#apt-get
- installing Oracle Java is something that's not going to be acceptable -- Oracle does not currently provide an official Oracle Java image due to licensing restrictions, and it would not be appropriate for us to do so either; does LogicalDOC work with OpenJDK?
- `sudo`: does LogicalDOC itself use/require `sudo`?

car031:
Thanks for the feedback.
Please check, we removed the MAINTAINER, we removed the apt-get upgrade, we fixed the apt-get install.
LogicalDOC itself does not require sudo

We can use OpenJDK, can you give me the instruction to install it ?

We extend phusion because we need  a mechanism to execute daemons(like init.d).
What official image should we use to have the capability to have daemons ?

car031:
Ok, we COMPLETELY rewrote the image to extend from centos:7 that is official. We also respected all the other points you mentioned. Is it now ready for in clusion in the official images library ?

car031:
Hi tianon, had you time to look at the new version of our image? We used OpenJDK and we extended from centos:7 official image. Do you think we are ready for inclusion now ?

car031:
Hi all, we really wish to have our image included in the official images, and are putting efforts to reach the goal. Can some one give use feedback ? Thanks in advance.

car031:
Hi @tianon, we really wish to have our image included in the official images, and are putting efforts to reach the goal. Can some one give use feedback ? Thanks in advance.

tianon:
> We can use OpenJDK, can you give me the instruction to install it ?

I'd recommend checking out the official image for OpenJDK: https://hub.docker.com/_/openjdk/

> We extend phusion because we need a mechanism to execute daemons(like init.d).
> What official image should we use to have the capability to have daemons ?

This is contrary to the goals of Docker and the official images.  It works, but it ends up creating an artificial barrier to scaling and deployment flexibility.  See https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#each-container-should-have-only-one-concern for more details on this point (that entire document has a lot of good tips for creating concise highly-repeatable `Dockerfile`s).

Can the daemons that need to run for LogicalDOC run in separate containers and talk to one another over the network?

For example, see the section in https://hub.docker.com/_/sentry/ about running Sentry's server and a Celery beat and workers (all from the same image source, only modifying the command).

car031:
Hi @tianon,, sure, we extended from official image **centos:7** and we installed **OpenJDK.**  MOreover we dropped all the daemons so this is no more an issue,
So i think we respected all your concerns.
Is now our image ready to be included ?

car031:
Hi @tianon. soprry to bother but we really tried to literally follow all your indications, we the image now appears to respect all things. Can you check again ?

car031:
Hi @tianon. sorry to bother but we really tried to literally follow all your indications, we the image now appears to respect all things. Can you check again ?

car031:
hello, can someone give us a feedback to proceed with this ?

kinghuang:
Not speaking with any official capacity, but here's a few of my observations.

* Is there something that prevents you from using the official `openjdk` or `tomcat` images? It is preferable to extend an existing official image rather than installing those components locally in this image.
* The Dockerfile has `/docker-entrypoint.sh` set as both the `ENTRYPOINT` and `CMD`. That means the container is going to execute `/docker-entrypoint.sh /docker-entrypoint.sh` by default when it runs. You should probably set one or the other, not both.
* The entrypoint script doesn't handle any run arguments. Perhaps you should make `$LDOC_HOME/bin/logicaldoc.sh` the entrypoint and `run` the command, or even the entire line the command and not override the entrypoint.
* The volume declared for `$LDOC_HOME/conf` is unnecessary. Users can mount a volume at that path whether or not you declare a `VOLUME` in the Dockerfile.
* There are a lot of packages installed by your Dockerfile. Are packages like `zip`, `unzip`, `wget`, `ftp`, `gcc`, etc. really needed to run LogicalDOC? If there are packages needed only during image build, they should be installed and removed within the `RUN` instruction that uses them to minimize the image size and potential vulnerabilities.
  * Multi-stage builds are a related option for steps like making Tesseract OCR
* Is it possible to run through LogicalDOC's installation process during image build rather than when the image is run?

car031:
Thanks, yes we expanded from openjdk.

The packages installed are really needed. After our first pull submission we drastically changed the Docker file and now we think to be completely compatible and of course we extended an official image: the openjdk

car031:
Hi All, can someone validate our last version of the proposed official image ?

car031:
Hi All, can someone validate our last version of the proposed official image ? We extended another official image and we think to have respectes all the compliancy rules

yosifkit:
:disappointed: sorry for the slow reply

I've been working to build and test this locally, but it seems that it is required to register for a trial key on logicaldoc.com to even get past the login screen.  I think it would be most useful to have an identical Dockerfile for the Community Edition which I assume does not have the same requirement. We have a few other images that provide commercial versions of their open-source products via official-images but they also have a community image to get users an easy trial/demo (like neo4j and couchbase).

The compose file in your repo was helpful but cannot be started with `docker-compose up -d` since it requires that mysql be up and ready for connections before logicaldoc is started (just a warning that users might report this as an issue).  In [wordpress](https://github.com/docker-library/wordpress/blob/93bc135b2fad658fe041447209c1a379a16dd302/docker-entrypoint.sh#L190-L209) we solve it with a simple timeout loop that waits until a mysql connection is ready and then starts the real process.

I have some of the following in some changes locally and can make it a PR if you like.

- The Dockerfile has `/docker-entrypoint.sh` set as both the `ENTRYPOINT` and `CMD`. That means the container is going to execute `/docker-entrypoint.sh /docker-entrypoint.sh` by default when it runs. You should set it as just the entrypoint or command, not both.
- if it is set as entrypoint, please keep in mind the [consistency section](https://github.com/docker-library/official-images/tree/b8d835e8961ad474247780513bcb71a82a25eb88#consistency); in summary `docker run -it --rm logcialdoc bash` or any other executable in the image should work

- don't run `yum upgrade` in a container
  > You should avoid RUN apt-get upgrade or dist-upgrade, as many of the “essential” packages from the base images won’t upgrade inside an unprivileged container. If a package contained in the base image is out-of-date, you should contact its maintainers.

- maximize docker build cache by moving environment variables that change often as late in the Dockerfile as possible
  - for example, `ENV LDOC_VERSION="7.7.1"` should be after the `RUN` that installs tesseract
  - `LIBREOFFICE_VERSION` could be moved to just before the line it is needed

- any files downloaded from the net should be verified ([security section](https://github.com/docker-library/official-images/tree/b8d835e8961ad474247780513bcb71a82a25eb88#security))
  - ftp.foolabs.com/pub/xpdf has `sig` files so it can be verified with gpg
  - libreoffice produces pgp signatures as well. [link for 5.4.0](http://download.documentfoundation.org/libreoffice/stable/5.4.0/rpm/x86_64/LibreOffice_5.4.0_Linux_x86-64_rpm_langpack_en-GB.tar.gz.mirrorlist)
  - is there any way to verify the `leptonica` or `tesseract` files or at least get them over https? (to prevent possible MITM attacks to the build servers)  Are there releases that are packaged for easier consumption?

- does logicalDoc use `gcc gcc-c++ make` and the `-devel` packages beyond the `pip install j2cli` and compiling of leptonica and tesseract? Removing them in the same RUN that they are used saved my image ~200MB.

- ensure that there are not extra processes in the container that would interfere with signal handling
  - make sure any shell scripts that start a different process end with `exec my-other-executable`
  - fixing this causes the classic pid 1 reaping zombie problem (http://blog.dscpl.com.au/2015/12/issues-with-running-as-pid-1-in-docker.html), but easily fixed by adding `dumb-init` or `tini` or using `--init` on `docker run`

yosifkit:

PR #2949, opened 2017-05-13
Refactor the "multiarch" section of the README to make it clear these things are now "officially supported" (in many cases)

tianon:


tianon:
See also https://github.com/opencontainers/image-spec/pull/650 for where these ARM names/buckets come from.

README.md
```diff
@@ -8,16 +8,19 @@ See [Docker's documentation](https://docs.docker.com/docker-hub/official_repos/)
...
--	ARMv5 (`armel`): https://hub.docker.com/u/armel/
--	ARMv7 (`armhf`): https://hub.docker.com/u/armhf/
--	ARMv8 (`arm64`): https://hub.docker.com/u/aarch64/
+-	IBM z Systems (`s390x`): https://hub.docker.com/u/s390x/
 -	POWER8 (`ppc64le`): https://hub.docker.com/u/ppc64le/
--	System z (`s390x`): https://hub.docker.com/u/s390x/
 -	x86/i686 (`i386`): https://hub.docker.com/u/i386/
+-	ARMv8 (`arm64v8`): https://hub.docker.com/u/arm64v8/
+-	ARMv7 (`arm32v7`): https://hub.docker.com/u/arm32v7/
+-	ARMv6 (`arm32v6`): https://hub.docker.com/u/arm32v6/ (Raspberry Pi 1, Raspberry Pi Zero)
```
stevvooe: Should we have a blip in OCI about concatenating these?

tianon: If you think there'd be value in such a thing, sure -- we needed something to differentiate for pushing images into buckets, so this is what we settled on based on that PR's granularity. :sweat_smile:

stevvooe: Ok, we'll leave it as is. The problem I am worried about here is that you have `arm32` whereas OCI has `arm`.

I don't think this is a big deal.

PR #2985, opened 2017-05-25
Add "Table of Contents" of README.md

PeterDaveHello:
Let's improve the README.md readability by adding "Table of Contents" so that we don't need to scroll down and up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,32 @@
 
 [![Build Status](https://travis-ci.org/docker-library/official-images.svg?branch=master)](https://travis-ci.org/docker-library/official-images)
 
+## Table of Contents
+
+- [What are "Official Images"?](#what-are-official-images)
+- [Architectures other than amd64?](#architectures-other-than-amd64)
+- [Contributing to the standard library](#contributing-to-the-standard-library)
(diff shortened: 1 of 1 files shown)


yosifkit:
This is a great idea, but I'd like to ensure it doesn't get out of sync with any changes in the future.  We'll need to add this to an automated task. Did you use a specific tool to generate this?

PeterDaveHello:
Can use https://npmjs.com/package/doctoc but will need nodejs, acceptable?

PeterDaveHello:
@yosifkit would you like me to integrate with `doctoc`? Thanks.

PeterDaveHello:
@yosifkit ?

yosifkit:
Sorry for the long tail, I don't often take time for non-image related issues.  I'd rather not add a dependency on a new language (even though it could just be dropped into another docker image).  I did look at https://github.com/ekalinin/github-markdown-toc, but it could quickly exhaust github api limits.

Given that we always use `#` style headers we don't need a full markdown parser; I think we could whip up something that would generate it with just bash and awk.

PeterDaveHello:
Oops, was doing it, never mind.

PR #3000, opened 2017-05-31
fix a typo in README.md

PureWhiteWu:
I found a typo.
'it it' should be 'it is'


tianon:
Nice catch, thanks! :+1:

PR #3007, opened 2017-06-01
Adding SWI-Prolog official image.

ninjarobot:
This adds an official image for SWI-Prolog.

Supporting docs are submitted in a PR: https://github.com/docker-library/docs/pull/920

Here is a very quick "hello world" test for the image:
```console
# echo 'hello:-write("Hello, world!"),nl,halt.' >> hello.pl; swipl -q -f hello.pl -t hello; rm hello.pl
Hello, world!
```

# Checklist for Review

    - https://github.com/SWI-Prolog
    - language stack
    - https://github.com/docker-library/docs/pull/920

### DIFF
--- a/library/swipl
+++ b/library/swipl
@@ -2,5 +2,5 @@ Maintainers: Dave Curylo <dave@curylo.org> (@ninjarobot)
 GitRepo: https://github.com/SWI-Prolog/docker-swipl.git
 
 Tags: latest, 7.5.11
-GitCommit: 6adb96cb5bcd670c593452add9de417f841617aa
+GitCommit: e8d1f0bf3db5e9cbfa331aba06bb740a76a81f83
 Directory: 7.5.11/stretch
--- a/library/swipl
+++ b/library/swipl
@@ -1,6 +1,6 @@
 Maintainers: Dave Curylo <dave@curylo.org> (@ninjarobot)
 GitRepo: https://github.com/SWI-Prolog/docker-swipl.git
 
-Tags: latest, 7.5.7
-GitCommit: 05c70112197e4c59da8af37743a493ca9a3266fb
-Directory: 7.5.7/jessie
+Tags: latest, 7.5.11
+GitCommit: 379993742ae105d4db2395bfeb0aa0441df54b56
+Directory: 7.5.11/stretch
--- a/library/swipl
+++ b/library/swipl
@@ -2,5 +2,5 @@ Maintainers: Dave Curylo <dave@curylo.org> (@ninjarobot)
 GitRepo: https://github.com/SWI-Prolog/docker-swipl.git
 
 Tags: latest, 7.5.11
-GitCommit: 379993742ae105d4db2395bfeb0aa0441df54b56
+GitCommit: 6adb96cb5bcd670c593452add9de417f841617aa
 Directory: 7.5.11/stretch
--- a/library/swipl
+++ b/library/swipl
@@ -1,0 +1,6 @@
+Maintainers: Dave Curylo <dave@curylo.org> (@ninjarobot)
+GitRepo: https://github.com/SWI-Prolog/docker-swipl.git
+
+Tags: latest, 7.5.7
+GitCommit: 05c70112197e4c59da8af37743a493ca9a3266fb
+Directory: 7.5.7/jessie


tianon:
@ninjarobot sorry for the delay (and thanks so much for being patient with us) :heart:

Here's a few thoughts as I look through the Dockerization:

- `alpine`
  1. `FROM alpine` (unqualified) is potentially likely to break at a newer version of Alpine, so I'd recommend pinning that (`FROM alpine:3.6`, for example)
  2. `apk --update add --no-cache ...` is redundant, and actually will leave cache files around -- you probably wanted to simply use `apk add --no-cache` (without `--update`) in all three of those lines
  3. pulling packages from Edge into a stable Alpine release can sometimes cause serious issues -- does Alpine 3.6 have the updated packages that necessitated that addition of Edge?
- `jessie`
  1. due to our delay, `stretch` is now released and probably worth considering rebasing on top of :innocent:
  2. `apt-get dist-upgrade` often causes issues inside containers, so we prefer to update the base image directly instead (which will trigger rebuilds of `swipl` automatically)
  3. instead of `build-essential`, you likely want to just add the specific packages you need (from https://packages.debian.org/stretch/build-essential: "If you do not plan to build Debian packages, you don't need this package. ")
  4. `curl -O https://raw.githubusercontent.com/SWI-Prolog/docker-swipl/master/build` is definitely a non-starter -- that should be pulling from either a specific explicit commit, or better yet, simply embed the steps to build directly in the `Dockerfile` (are they really all that different from the Alpine build? `./configure && make && make install`)
  5. `apt-get purge` has a `--auto-remove` flag that can be used to avoid the explicit `apt-get autoremove` invocation
  6. it might be worth considering consolidating the "build dependencies" into a variable so that they can be used twice rather than repeated (which will help avoid missing any on accident)
  7. `apt-get clean` is applied automatically by the base image, and thus will do very little/nothing (see https://github.com/debuerreotype/debuerreotype/blob/ce5cb9f5a0eb2aa171587339d50126139750a6d8/scripts/debuerreotype-minimizing-config#L43-L66)
  8. instead of using `apt-get install` after purging the very packages you're reinstalling, you might consider using `apt-mark manual` to mark those packages as manually installed before purging (and thus save them from being auto-removed), and then they won't be installed twice in the same `RUN` line

tianon:
So, is that `build` file the same one that http://www.swi-prolog.org/build/unix.html talks about? (the one included with the SWI-Prolog source tarballs?)

Could any necessary modifications/adjustments to it happen in the `Dockerfile` to the copy downloaded with the source rather than checking a copy into the `docker-swipl` repo?

yosifkit:
Just a couple comments to prevent extra work in your image:

- The line to install run time dependencies should be above the apt-get purge. Many of these are dependencies of the build deps and would be needlessly uninstalled and then re-installed. Related would be to remove `libarchive13` from build deps so it won't get removed.
  ```Dockerfile
  apt-get install -y libarchive13 libgmp10 libossp-uuid16 libssl1.1 libdb5.3 libpcre3 libedit2 libgeos++ libspatialindex4v5 unixodbc odbc-postgresql tdsodbc libmariadbclient18 && \
  ```

- `libgeos++` is not a package name, and instead gets treated like a regex which installs more packages than you probably intended.  I'm not sure if you needed the C or C++ version of the library (C: `libgeos-c1v5`, C++: `libgeos-3.5.1`).  If you only need the C version (assumed since you install `gcc` and not `g++`), then you can also swap the build dep from `libgeos++-dev` to `libgeos-dev`.
  ```console
  Note, selecting 'libgeos-c1' for regex 'libgeos+'
  Note, selecting 'libgeos-c1v5' for regex 'libgeos+'
  Note, selecting 'libgeos-dbg' for regex 'libgeos+'
  Note, selecting 'libgeos-dev' for regex 'libgeos+'
  Note, selecting 'libgeos-doc' for regex 'libgeos+'
  Note, selecting 'libgeos-ruby1.8' for regex 'libgeos+'
  Note, selecting 'libgeos++-dev' for regex 'libgeos+'
  Note, selecting 'libgeos-3.5.1' for regex 'libgeos+'
  Note, selecting 'ruby-geos' instead of 'libgeos-ruby1.8'
  ```

- On the run dependencies, you could even put them in their own `RUN` line. This would allow separate versions of swipl to share the layer which would save users a little bandwidth between version bumps that have the same Debian image base.
  ```Dockerfile
  RUN apt-get update && \
  		libarchive13 \
  		libdb5.3 \
  		libedit2 \
  		libgeos-c1v5 \
  		libgmp10 \
  		libmariadbclient18 \ \
  		libossp-uuid16 \
  		libpcre3 \
  		libspatialindex4v5 \
  		libssl1.1 \
  		odbc-postgresql \
  		tdsodbc \
  		unixodbc
  ```

yosifkit:
Thanks! [`libarchive13`](https://github.com/SWI-Prolog/docker-swipl/blob/6adb96cb5bcd670c593452add9de417f841617aa/7.5.11/stretch/Dockerfile#L21) is still in the `BUILD_DEPS` so it gets removed (even though it is also in the run deps layer), this causes something like `use_module(library(prolog_pack)).` to fail.  Other than that, LGTM.

yosifkit:
<details>
<summary>Diff:</summary>

```diff
error: failed fetching repo "swipl"
unable to find a manifest named "swipl" (in "/tmp/tmp.o8uAAmgmAV/oi/library" or as a remote URL)
error: failed fetching repo "swipl"
unable to find a manifest named "swipl" (in "/tmp/tmp.o8uAAmgmAV/oi/library" or as a remote URL)
error: failed fetching repo "swipl"
unable to find a manifest named "swipl" (in "/tmp/tmp.o8uAAmgmAV/oi/library" or as a remote URL)
error: failed fetching repo "swipl"
unable to find a manifest named "swipl" (in "/tmp/tmp.o8uAAmgmAV/oi/library" or as a remote URL)
diff --git a/_bashbrew-arches b/_bashbrew-arches
index e69de29..20ba69b 100644
--- a/_bashbrew-arches
+++ b/_bashbrew-arches
@@ -0,0 +1 @@
+swipl:7.5.11 @ amd64
diff --git a/_bashbrew-list b/_bashbrew-list
index e69de29..75a14a5 100644
--- a/_bashbrew-list
+++ b/_bashbrew-list
@@ -0,0 +1,2 @@
+swipl:7.5.11
+swipl:latest
diff --git a/swipl_7.5.11/Dockerfile b/swipl_7.5.11/Dockerfile
new file mode 100644
index 0000000..06e3d87
--- /dev/null
+++ b/swipl_7.5.11/Dockerfile
@@ -0,0 +1,37 @@
+FROM debian:stretch-slim
+LABEL maintainer "Dave Curylo <dave@curylo.org>, Michael Hendricks <michael@ndrix.org>"
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+    libarchive13 \
+    libgmp10 \
+    libossp-uuid16 \
+    libssl1.1 \
+    libdb5.3 \
+    libpcre3 \
+    libedit2 \
+    libgeos-c1v5 \
+    libspatialindex4v5 \
+    unixodbc \
+    odbc-postgresql \
+    tdsodbc \
+    libmariadbclient18 && \
+    rm -rf /var/lib/apt/lists/*
+RUN SWIPL_VER=7.5.11 && \
+    SWIPL_CHECKSUM=ab8a21ef88e410fc8dfb421bcba7884687fdf08ed85c191e51d806aa831d95e9 && \
+    BUILD_DEPS='make gcc wget libarchive-dev libgmp-dev libossp-uuid-dev libpcre3-dev libreadline-dev libedit-dev libssl-dev zlib1g-dev libdb-dev libgeos-dev libspatialindex-dev unixodbc-dev' && \
+    apt-get update && apt-get install -y --no-install-recommends $BUILD_DEPS && \
+    mkdir /tmp/src && \
+    cd /tmp/src && \
+    wget http://www.swi-prolog.org/download/devel/src/swipl-$SWIPL_VER.tar.gz && \
+    echo "$SWIPL_CHECKSUM  swipl-$SWIPL_VER.tar.gz" >> swipl-$SWIPL_VER.tar.gz-CHECKSUM && \
+    sha256sum -c swipl-$SWIPL_VER.tar.gz-CHECKSUM && \
+    tar -xzf swipl-$SWIPL_VER.tar.gz && \
+    cd swipl-$SWIPL_VER && \
+    cp build.templ build && \
+    sed -i '/PREFIX=$HOME/c\PREFIX=/swipl' build && \
+    sed -i '/# export DISABLE_PKGS/c\export DISABLE_PKGS="jpl xpce"' build && \
+    sed -i '/# export EXTRA_PKGS/c\export EXTRA_PKGS="db space"' build && \
+    chmod u+x build && ./build && \
+    apt-get purge -y --auto-remove $BUILD_DEPS && \
+    cd /usr/bin && rm -rf /tmp/src && ln -s /swipl/bin/swipl swipl && rm -rf /var/lib/apt/lists/*
+CMD ["swipl"]
```

</details>

yosifkit:
Build test of #3007; c05098a4 (`swipl`):

```console

$ bashbrew build swipl:latest
Building bashbrew/cache:7656810c20976d2978541bdfed195b1dbc85df4b497799ec5354c9091ca43dbb (swipl:latest)
Tagging swipl:latest
Tagging swipl:7.5.11

$ test/run.sh swipl:latest
testing swipl:latest
	'utc' [1/4]...passed
	'override-cmd' [4/4]...passed

```

PR #3058, opened 2017-06-13
Add an initial "manifest-tool" implementation

tianon:
This is the next major milestone towards https://github.com/docker-library/official-images/issues/2289.  This adds manifest list support for our new `Architectures` fields, and technically also for `SharedTags` fields, but I haven't tested that support yet (priority ATM is `Architectures` -- we'll hit back on `SharedTags` after we get `Architectures` in good shape).

See https://hub.docker.com/u/trollin/ for a few repos I've pushed using this. :+1:

### DIFF
--- a/bashbrew/go/src/bashbrew/cmd-put-shared.go
+++ b/bashbrew/go/src/bashbrew/cmd-put-shared.go
@@ -4,10 +4,49 @@ import (
 	"fmt"
 	"os"
 	"path"
-	"strings"
 
 	"github.com/codegangsta/cli"
+
+	"github.com/docker-library/go-dockerlibrary/architecture"
+	"github.com/docker-library/go-dockerlibrary/manifest"
 )
+
+func entriesToManifestToolYaml(r Repo, entries ...*manifest.Manifest2822Entry) (string, error) {
+	yaml := ""
+	entryIdentifiers := []string{}
+	for _, entry := range entries {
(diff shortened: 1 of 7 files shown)


tianon:
cc @estesp -- if you've got time to even glance over this, that'd be awesome :heart: :pray: :+1:

bashbrew/go/src/bashbrew/cmd-put-shared.go
```diff
@@ -21,24 +60,55 @@ func cmdPutShared(c *cli.Context) error {
...
+				return err
+			}
+
+			for _, tag := range r.Tags(namespace, false, entry) {
+				tagYaml := fmt.Sprintf("image: %s\n%s", tag, yaml)
+				fmt.Printf("Putting %s\n", tag)
+				if err := manifestToolPushFromSpec(tagYaml); err != nil {
+					return fmt.Errorf("failed pushing %q (%q)", tag, entry.TagsString())
+				}
+			}
```
tianon: See https://github.com/estesp/manifest-tool/issues/23 for some discussion of making `manifest-tool` more efficient for our use case (and thus removing the need for so many nested loops :innocent:).

tianon:
I've now tested the `SharedTags` portion of this code, and it works too! :smile: :innocent: :tada:

estesp:
Looks good! I'll work on the tags issue as soon as I can get around to it later this week!

tianon:
You're a hero; thanks @estesp :+1:

PR #3115, opened 2017-06-24
Add arm32v5 support for Debian

tianon:


tianon:
Merging, since this doesn't change the contents (simply adds a new arch). :+1:

PR #3121, opened 2017-06-26
Remove Oracle Linux 5 images as it has reached EOL.

Djelibeybi:
I'm hoping this will remove the 5 and 5.11 tags from the Docker Hub official repo.


yosifkit:
Users will still be able to `docker pull` them, but the "supported tags" list on the README will no longer include them.

PR #3172, opened 2017-07-11
Add a new maintainer for the websphere-liberty image

jamiecoleman92:


tianon:
FYI @dibbles -- you're on the hook now! :smile:

PR #3189, opened 2017-07-12
Update go-dockerlibrary to pull in the new "getenv" function

tianon:
See https://github.com/docker-library/go-dockerlibrary/pull/15.


yosifkit:
Build failure unrelated.

PR #3193, opened 2017-07-13
Use the first arch of each entry in "diff-pr.sh"

tianon:
This allows us to diff `windows-amd64`-only images, but doesn't solve the more general problem of how to cleanly show a diff across all architectures.


tianon:
Classic case of https://github.com/docker-library/official-images/pull/3193/files?w=1 being easier to review. :innocent:

PR #3221, opened 2017-07-20
Remove Talon (@moghedrin) from maintainers list.

daghack:
Sad days, but I will no longer be a maintainer, moving forward.


MAINTAINERS
```diff
@@ -1,4 +1,3 @@
 Tianon Gravi <admwiggin@gmail.com> (@tianon)
 Joseph Ferguson <yosifkit@gmail.com> (@yosifkit)
-Talon Bowler <nolat301@gmail.com> (@moghedrin)
 # Emeritus: Peter Salvatore <peter@psftw.com> (@psftw)
```
tianon: I think you should instead swap to `# Emeritus:` like @psftw did -- makes it clear he did contribute a lot, and this is about the only way we can point that out to folks. :smile:

PR #3254, opened 2017-07-29
Support for new Ghost version 1.0

sharkoz:
In the new 1.0 release of Ghost, the setup process is not initiated through a http redirect but via javascript.

So the previous test checking that `/ghost/` would send a redirect to `setup` fails.

The js version now checks an API, `/ghost/api/v0.1/authentication/setup/`
If setup hasn't occured yet, this API returns `{"setup":[{"status":false}]}`
Otherwise it returns `{"setup":[{"status":true}]}`

This PR updates the test done to check if image is unconfigured for ghost images `ghost:1*` (version 1)

It fixes the failed test in PR https://github.com/docker-library/ghost/pull/67

### DIFF
--- a/test/tests/ghost-basics/run.sh
+++ b/test/tests/ghost-basics/run.sh
@@ -22,7 +22,7 @@ _request() {
 	docker run --rm --link "$cid":ghost "$clientImage" \
 		curl -fs -X"$method" "$@" "http://ghost:2368/$url"
 }
-
+sleep 60
 # Make sure that Ghost is listening and ready
 . "$dir/../../retry.sh" '-t' '30' '_request GET / --output /dev/null'
 
--- a/test/tests/ghost-basics/run.sh
+++ b/test/tests/ghost-basics/run.sh
@@ -27,10 +27,9 @@ _request() {
 . "$dir/../../retry.sh" '_request GET / --output /dev/null'
 
 # Check that /ghost/ redirects to setup (the image is unconfigured by default)
-if [[ $serverImage == ghost:1* ]]
-then
-	_request GET '/ghost/api/v0.1/authentication/setup/' | grep -q 'status":false'
-else
-	_request GET '/ghost/#/' -I | grep -q '^Location: .*setup'
-fi
+ghostVersion="$(docker inspect --format '{{range .Config.Env}}{{ . }}{{"\n"}}{{end}}' "$serverImage" | awk -F= '$1 == "GHOST_VERSION" { print $2 }')"
+case "$ghostVersion" in
+	0.*) _request GET '/ghost/' -I | grep -q '^Location: .*setup' ;;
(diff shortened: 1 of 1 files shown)


test/tests/ghost-basics/run.sh
```diff
@@ -27,4 +27,10 @@ _request() {
 . "$dir/../../retry.sh" '_request GET / --output /dev/null'
 
 # Check that /ghost/ redirects to setup (the image is unconfigured by default)
-_request GET '/ghost/' -I | grep -q '^Location: .*setup'
+if [[ $serverImage == ghost:1* ]]
+then
+	_request GET '/ghost/api/v0.1/authentication/setup/' | grep -q 'status":false'
```
yosifkit: How often will the api version change?  It'll be a bit annoying if failing tests just mean the test needs a bump (like we have now with 1.0).  Maybe we just need to do a simpler test to see that a `GET` request succeeds?

 ```bash
_request GET '/ghost/' -L --output /dev/null
```

File: test/tests/ghost-basics/run.sh [hunk 2]
```diff
@@ -27,4 +27,10 @@ _request() {
 . "$dir/../../retry.sh" '_request GET / --output /dev/null'
 
 # Check that /ghost/ redirects to setup (the image is unconfigured by default)
-_request GET '/ghost/' -I | grep -q '^Location: .*setup'
+if [[ $serverImage == ghost:1* ]]
```
yosifkit: Maybe instead pull the full ghost version from the image since the test could run on `ghost:latest`
```bash
ghostVersion="$(docker inspect --format '{{range .Config.Env}}{{ . }}{{"\n"}}{{end}}' "$serverImage" | awk -F= '$1 == "GHOST_VERSION" { print $2 }')"
case "$ghostVersion" in
	0.*) _request GET '/ghost/' -I | grep -q '^Location: .*setup' ;;
	*)   _request GET '/ghost/api/v0.1/authentication/setup/' | grep -q 'status":false' ;;
esac
```

sharkoz: It looks great thanks, I'm adding it right now.

test/tests/ghost-basics/run.sh
sharkoz: I can't speak for the ghost dev team on this, but this API is pretty straightforward, it does one single thing and I saw no announcements about changes on this. But maybe @acburdine has some insight ? And for the version, it's true `v0.1` doesn't seam very stable but for now there is no `latest` tag :smile: 

Anyway, for the simple `GET` request, it wouldn't be enough as this URL still exists after setup it's the admin URL.

yosifkit: Yeah, it does seem that this API endpoint will be unlikely to change; we'll just deal with it when it comes.

PR #3320, opened 2017-08-13
Add Conan-server

ovidiub13:
# Checklist for Review

-	[x] 1. associated with or contacted upstream?
    - https://github.com/conan-io
-	[x] 2. does it fit into one of the common categories? ("service", "language stack", "base distribution")
    - it's a dependency manager for C, likely similar to `maven` or `gradle`, so requires a bit more evaluation
-	[x] 3. is it reasonably popular, or does it solve a particular use case well?
-	[ ] 4. does a [documentation](https://github.com/docker-library/docs/blob/master/README.md) PR exist? (should be reviewed and merged at roughly the same time so that we don't have an empty image page on the Hub for very long)
-	[ ] 5. dockerization review for best practices and cache gotchas/improvements (ala [the official review guidelines](https://github.com/docker-library/official-images/blob/master/README.md#review-guidelines))?
    - `ADD https://...`
    - `CMD /conan...`
-	[ ] 6. 2+ dockerization review?
-	[ ] 7. existing official images have been considered as a base? (ie, if `foobar` needs Node.js, has `FROM node:...` instead of grabbing `node` via other means been considered?)
    - `FROM python`, but currently a tad over-specific (`python:3.6.1-alpine3.6`, ought to be `python:3.6-alpine3.6` instead)
-	[x] 8. ~~if `FROM scratch`, tarballs only exist in a single commit within the associated history?~~
-	[ ] 9. passes current tests? any simple new tests that might be appropriate to add? (https://github.com/docker-library/official-images/tree/master/test)

### DIFF
--- a/library/conan-server
+++ b/library/conan-server
@@ -1,4 +1,5 @@
-Maintainers: Ovidiu-Florin Bogdan <ovidiu.b13@gmail.com> (@ovidiub13)
+Maintainers: Ovidiu-Florin Bogdan <ovidiu.b13@gmail.com> (@ovidiub13),
+             Luis Martinez de Bartolome Izquierdo <Luis Martinez de Bartolome Izquierdo> (@lasote)
 GitRepo: https://github.com/conan-io/conan-docker-tools.git
 
 Tags: 0.25.1, latest
--- a/library/conan-server
+++ b/library/conan-server
@@ -1,5 +1,5 @@
 Maintainers: Ovidiu-Florin Bogdan <ovidiu.b13@gmail.com> (@ovidiub13),
-             Luis Martinez de Bartolome Izquierdo <Luis Martinez de Bartolome Izquierdo> (@lasote)
+             Luis Martinez de Bartolome Izquierdo <lasote@gmail.com> (@lasote)
 GitRepo: https://github.com/conan-io/conan-docker-tools.git
 
 Tags: 0.25.1, latest
--- a/library/conan-server
+++ b/library/conan-server
@@ -1,0 +1,6 @@
+Maintainers: Ovidiu-Florin Bogdan <ovidiu.b13@gmail.com> (@ovidiub13)
+GitRepo: https://github.com/conan-io/conan-docker-tools.git
+
+Tags: 0.25.1, latest
+GitCommit: 1e40380b3b9565caceddcb946c774ced31a74424
+Directory: conan_server


ovidiub13:
@tianon
1. Yes. The Dockerfile is stored on a conan-io repository
2. What evaluation are you referring to?
3. It is a JFrog product, with support in Artifactory, used by many open source projects and companies.
4. IIRC the documentation PR was supposed to be made after this one got an OK.
5. I'm guessing you're saying it's a problem that I'm downloading this to `/` and not to a specific directory?
6. Can I ping someone for this?
7. Good point. I'll change that.
8. --
9. I'm looking into those.

tianon:
I'm looking at the Dockerization here, and trying to understand why this `Dockerfile` is what seems to be so overly complex? (especially the addition of `gunicorn` and `entrypoint.sh`)

Isn't something like the following sufficient? (and arguably better, since it provides both `conan` and `conan_server` as available commands / wrapper scripts)
```dockerfile
FROM python:3.6-alpine3.6

ENV CONAN_VERSION=0.27.0

RUN pip install "conan==${CONAN_VERSION}"

EXPOSE 9300

CMD ["conan_server"]
```

ovidiub13:
@tianon 

* The version installed by `pip` is created with `pyinstaller`, which, the Conan developers tell me, only embeds a limited python interpreter, which brings some drawbacks.
* `gunicorn` is needed because it's a better wsgi server than what the `conan_server` has.
* having `conan` (client) installed in the container brings no benefit, as it is a client used to communicate with the server, and manage local sources (build, package, upload, etc.). It should be installed on the developers machine.
* can add the command from the `entrypoint.sh` directly to the `Dockerfile` if that's necessary.

tianon:

Sorry for the delay!

Fair enough, in that case I've got a few concerns about the `Dockerfile` as-is; the chief among them is `ADD` with a remote URL -- that creates a layer which whose space can then never be reclaimed, not to mention re-downloads the full remote asset on every `docker build` to verify whether it has changed.

The `entrypoint.sh` script is fine, but it ought to include `exec` so that `/bin/sh` doesn't stay resident (on top of the `/bin/sh` that's going to stay resident from `CMD` not using the JSON string array syntax); here's a slightly modified `Dockerfile` that ends up running _just_ `gunicorn` (so that signal handling, etc works properly), and which uses `wget` for downloading the Conan source so that it can be cached properly (and updates to 1.5.2 and Alpine 3.8):

```dockerfile
FROM python:3.6-alpine3.8

ENV CONAN_VERSION=1.5.2

RUN set -eux; \
	wget -O 'conan.zip' "https://github.com/conan-io/conan/archive/${CONAN_VERSION}.zip"; \
	unzip conan.zip; \
	rm conan.zip; \
	mv "conan-${CONAN_VERSION}" /conan; \
	cd /conan; \
	pip3 install -r conans/requirements.txt; \
	pip3 install -r conans/requirements_server.txt; \
	pip3 install gunicorn

EXPOSE 9300

WORKDIR /conan

RUN { echo '#!/bin/sh' && echo 'exec gunicorn -b 0.0.0.0:9300 -w 4 -t 300 conans.server.server_launcher:app'; } > /usr/local/bin/conan-server && chmod +x /usr/local/bin/conan-server

CMD ["conan-server"]
```

Thoughts?

ovidiub13:
Looks good. I need to find some time and update everything, so we can hopefully merge it.

ovidiub13:
The community has developed the Official Conan.io `conan-server` Docker image in https://hub.docker.com/r/conanio/conan_server/.

This PR is no longer necessary. Thank you for all your support in developing this PR and for sharing your experience in writing better `Dockerfile`s.

PR #3359, opened 2017-08-24
Add Architectures to geonetwork

longquan7:


tianon:
>  this file is generated via https://github.com/geonetwork/docker-geonetwork/blob/f80d7f7745202678805f28a1616f1b44386b1572/generate-stackbrew-library.sh

Please get upstream/maintainer buy-in before making this sort of change to their image, especially since they will be expected to support the result.  Thanks!

longquan7:
@tianon Ok,thanks

PR #3408, opened 2017-09-05
Add clojure Architectures based on openjdk base image Architectures

tianon:
I've personally tested this on the listed arches except `arm32v5` (since it's slow, I've only got it tested partway, but there's no reason it shouldn't work).

cc @Quantisan @cap10morgan -- do you have any issues with adding this? :smile:


tianon:
Also, as noted in https://github.com/Quantisan/docker-clojure/issues/37, Alpine has to wait for https://github.com/gliderlabs/docker-alpine/issues/304. :+1:

tianon:
friendly ping @Quantisan @cap10morgan :pray:

Quantisan:
@tianon LGTM. Very cool! Thanks for getting on top of this for us!

PR #3421, opened 2017-09-07
Update to jenkins 2.60.3.

francois2metz:
https://github.com/jenkinsci/docker/commit/dec928fd4eadbbbbcd6a35a3ad2baa38b537e846

### DIFF
--- a/library/jenkins
+++ b/library/jenkins
@@ -2,8 +2,8 @@
 # maintainer: Michael Neale <mneale@cloudbees.com> (@michaelneale)
 # maintainer: Carlos Sanchez <csanchez@cloudbees.com> (@carlossg)
 
-latest: git://github.com/jenkinsci/docker@15dc59d7dbd47da5259a50a9ebfa8895d594444f
-2.60.2: git://github.com/jenkinsci/docker@15dc59d7dbd47da5259a50a9ebfa8895d594444f
+latest: git://github.com/jenkinsci/docker@dec928fd4eadbbbbcd6a35a3ad2baa38b537e846
+2.60.3: git://github.com/jenkinsci/docker@dec928fd4eadbbbbcd6a35a3ad2baa38b537e846
 
-alpine: git://github.com/jenkinsci/docker@c24687d11870e6a408678a394c1298008773bf03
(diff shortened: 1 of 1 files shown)


library/jenkins
```diff
@@ -2,8 +2,8 @@
...
 
-latest: git://github.com/jenkinsci/docker@15dc59d7dbd47da5259a50a9ebfa8895d594444f
-2.60.2: git://github.com/jenkinsci/docker@15dc59d7dbd47da5259a50a9ebfa8895d594444f
+latest: git://github.com/jenkinsci/docker@dec928fd4eadbbbbcd6a35a3ad2baa38b537e846
+2.60.3: git://github.com/jenkinsci/docker@dec928fd4eadbbbbcd6a35a3ad2baa38b537e846
 
-alpine: git://github.com/jenkinsci/docker@c24687d11870e6a408678a394c1298008773bf03
-2.60.2-alpine: git://github.com/jenkinsci/docker@c24687d11870e6a408678a394c1298008773bf03
+alpine: git://github.com/jenkinsci/docker@dec928fd4eadbbbbcd6a35a3ad2baa38b537e846
+2.60.3-alpine: git://github.com/jenkinsci/docker@dec928fd4eadbbbbcd6a35a3ad2baa38b537e846
```
yosifkit: This changes the Alpine version to use the same Dockerfile as the Debian version. I see that you have a `Dockerfile-alpine`, but the official images do not have a way to specify a different Dockerfile; it needs to be in another folder or a different branch like it was previously.

francois2metz:
Thanks for the review. Closing as the official jenkins image is now deprecated.

yosifkit:
cc @ndeloof, @carlossg I thought the jenkins image would still receive updates for `2.60.x` LTS?

> The images found here will receive no further updates after LTS 2.60.x
>
> \- https://github.com/docker-library/docs/pull/948

ndeloof:
This PR indeed is valid for 2.60.3 (not sure why we didn't published it earlier)

carlossg:
@ndeloof is not because the alpine branch is still 2.60.2

carlossg:
see #3431

PR #3422, opened 2017-09-07
Add a "--dry-run" flag to "build", "tag", "push", and "put-shared"

tianon:
The intention is to use this for doing `put-shared` on our "heavy hitters" (repositories which we've been asked to leave at least 15 minutes in between pushes of any of them, so as to be kind to the automated build queues), allowing us to determine whether we need to do a push before we invoke it and then force a 15 minute wait before pushing another repository.


tianon:
So, rough pseudo-code would be something like:

1. for each heavy hitter:
    1. `put-shared --dry-run REPO` has stdout?
        1. `put-shared REPO`
        2. sleep 15 minutes

tianon:
See https://github.com/docker-library/official-images/pull/3423 for which repos are heavy hitters. :+1:

PR #3442, opened 2017-09-12
Add a "--single-arch" flag to "put-shared" for pushing SharedTags to architecture-specific namespaces

tianon:
Without this, `amd64/hello-world:latest` (etc.) will no longer exist. :sweat_smile:

### DIFF
--- a/bashbrew/go/src/bashbrew/cmd-put-shared.go
+++ b/bashbrew/go/src/bashbrew/cmd-put-shared.go
@@ -12,25 +12,29 @@ import (
 	"github.com/docker-library/go-dockerlibrary/manifest"
 )
 
-func entriesToManifestToolYaml(r Repo, entries ...*manifest.Manifest2822Entry) (string, time.Time, error) {
+func entriesToManifestToolYaml(singleArch bool, r Repo, entries ...*manifest.Manifest2822Entry) (string, time.Time, error) {
 	yaml := ""
 	mru := time.Time{}
 	entryIdentifiers := []string{}
 	for _, entry := range entries {
 		entryIdentifiers = append(entryIdentifiers, r.EntryIdentifier(*entry))
 
-		for _, arch := range entry.Architectures {
+		for _, entryArch := range entry.Architectures {
+			if singleArch && entryArch != arch {
+				continue
+			}
+
 			var ok bool
 
 			var ociArch architecture.OCIPlatform
-			if ociArch, ok = architecture.SupportedArches[arch]; !ok {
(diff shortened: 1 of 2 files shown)


bashbrew/go/src/bashbrew/cmd-put-shared.go
```diff
@@ -48,9 +52,6 @@ func entriesToManifestToolYaml(r Repo, entries ...*manifest.Manifest2822Entry) (
 			}
 		}
 	}
-	if yaml == "" {
-		return "", time.Time{}, fmt.Errorf("failed gathering images for creating %q", entryIdentifiers)
-	}
```
tianon: I did verify -- `manifest-tool` and the registry are both happy to accept an empty manifest list. :+1:

tianon: (This is necessary for cases like `arm32v6/hello-world:latest` which doesn't exist but is still a `SharedTag` we supposedly need to try pushing.)

tianon:
Intended usage:
```diff
diff --git a/multiarch/vars.groovy b/multiarch/vars.groovy
index 8021659..5266592 100644
--- a/multiarch/vars.groovy
+++ b/multiarch/vars.groovy
@@ -329,6 +329,9 @@ def bashbrewBuildAndPush(context) {
 
        def dryRun = context.sh(returnStdout: true, script: '''
                bashbrew push --dry-run --namespace "$TARGET_NAMESPACE" "$ACT_ON_IMAGE"
+               if [ -n "$BASHBREW_ARCH" ]; then
+                       bashbrew --arch-namespace "$ACT_ON_ARCH = $TARGET_NAMESPACE" put-shared --dry-run --single-arch --namespace "$TARGET_NAMESPACE" "$ACT_ON_IMAGE"
+               fi
        ''').trim()
        if (dryRun == '') {
                // if we don't need to push anything, let's not (and let's tell whoever invoked us that we didn't, so they scan skip other things too)
@@ -341,6 +344,13 @@ def bashbrewBuildAndPush(context) {
                                bashbrew push --namespace "$TARGET_NAMESPACE" "$ACT_ON_IMAGE"
                        '''
                }
+               if (env.BASHBREW_ARCH) {
+                       retry(3) {
+                               sh '''
+                                       bashbrew --arch-namespace "$ACT_ON_ARCH = $TARGET_NAMESPACE" put-shared --single-arch --namespace "$TARGET_NAMESPACE" "$ACT_ON_IMAGE"
+                               '''
+                       }
+               }
        }
 
        return 'push'
```

PR #3480, opened 2017-09-21
OrientDB add ppc64le

vaibhavsood:
Updates to include the ppc64le dockerfiles from orientdb upstream

### DIFF
--- a/library/orientdb
+++ b/library/orientdb
@@ -11,7 +11,7 @@ Architectures: amd64
 GitCommit: 8a9633c19fa9c53a5446b9b62997ff389813e785
 Directory: 2.1
 
-Tags: 2.2.27, latest
+Tags: 2.2.27
 Architectures: amd64, ppc64le
 amd64-GitCommit: 7de6c9c056ac364a1bd93fffae01b2f81a4b5029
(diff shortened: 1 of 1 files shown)


tianon:
While I'm a huge fan of more supported architectures, as-is this is definitely not going to be something we can accept in the official images.  The Dockerization itself has a number of minor issues, but I think the most important is that I don't see a strong reason why `ppc64le` can't use the same `Dockerfile` as `amd64` (pending https://github.com/gliderlabs/docker-alpine/issues/304, in this case, due to being Alpine-only), as requested in https://github.com/docker-library/official-images#multiple-architectures.

As it stands, these `ppc64le` images are from an entirely different base image (`ibmjava:latest` instead of `openjdk:8-jdk-alpine`), and as such the user experience is going to be potentially quite different (and the debugging experience certainly will be very different).

vaibhavsood:
Okay sure, i guess will have to wait for docker-alpine to be enabled for ppc64le, will then work with upstream to use the amd64 Dockerfile for ppc64le as well, will get back here once that happens, thanks!

tianon:
I'm going to close this in the meantime, especially since it's already outdated.  Let's continue the discussion upstream. :+1:

PR #3539, opened 2017-10-05
Match the constraint used by Fedora and CentOS

Djelibeybi:
This constraint does apply to Oracle Linux too and should've been added years ago. Mea culpa.


tianon:
Heh, this is mostly moot now that all our builders use `overlay2` consistently, but :+1: (might as well add it for consistency/correctness)

LGTM

PR #3606, opened 2017-10-22
Fix std libs test for Ruby 2.5

mtsmfm:
mathn.rb is removed from stdlib since 2.5.

https://github.com/ruby/ruby/blob/v2_5_0_preview1/NEWS#stdlib-compatibility-issues-excluding-feature-bug-fixes

----------

I split my PR just in case.

https://github.com/docker-library/official-images/pull/3605

Please feel free to close #3605 or ask me rebasing this PR after merging #3605


yosifkit:
LGTM! Thanks

PR #3845, opened 2017-12-26
Finally remove pypy:*-onbuild

tianon:
See docker-library/pypy#19.


tianon:
Removals only, merging!

PR #3888, opened 2018-01-10
Add official image for PM2 runtime

vmarchaud:
# Summary

This PR add the `pm2` file as requested in the README, the PR for the documentation is [here](https://github.com/docker-library/docs/pull/1118).

Github repo: https://github.com/Unitech/pm2

# Checklist for Review

           - @keymetrics / @Unitech / @vmarchaud
           - not sure but i would say `service`
           - see https://github.com/Unitech/pm2
          - https://github.com/docker-library/docs/pull/1118
         - We have our tests [here](https://github.com/keymetrics/docker-pm2/tree/master/example-app) and the build are done [here](https://travis-ci.org/keymetrics/docker-pm2)

### DIFF
--- a/library/pm2
+++ b/library/pm2
@@ -1,0 +1,86 @@
+# This file is generated via https://github.com/keymetrics/docker-pm2/blob/3b715c519b6b5453ceef63c578805a2ea879b87a/docker-tags.sh
+# 2018-01-10 14:36:46
+
+Maintainers: Keymetrics.io <https://github.com/keymetrics/docker-pm2> (@keymetrics)
+GitRepo: https://github.com/keymetrics/docker-pm2.git
+
+Tags: latest-alpine
+GitCommit: 2ff618938de9d75396f709da58f894bfdc525dc6
+Directory: tags/latest/alpine
+
+Tags: latest-stretch
+GitCommit: ff45312e04cbebb51193c07e9798a8c8ae914e68
+Directory: tags/latest/stretch
+
+Tags: latest-jessie
+GitCommit: ff45312e04cbebb51193c07e9798a8c8ae914e68
+Directory: tags/latest/jessie
+
+Tags: latest-slim
+GitCommit: ff45312e04cbebb51193c07e9798a8c8ae914e68
+Directory: tags/latest/slim
+
+Tags: latest-wheezy
+GitCommit: ff45312e04cbebb51193c07e9798a8c8ae914e68
+Directory: tags/latest/wheezy
+
+Tags: 4-alpine
+GitCommit: 2ff618938de9d75396f709da58f894bfdc525dc6
+Directory: tags/4/alpine
+
+Tags: 4-stretch
+GitCommit: 23d531ea377bff96e2cf23d3438f813078d019e5
+Directory: tags/4/stretch
+
+Tags: 4-jessie
+GitCommit: 23d531ea377bff96e2cf23d3438f813078d019e5
+Directory: tags/4/jessie
+
+Tags: 4-slim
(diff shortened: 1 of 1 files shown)


vmarchaud:
We know that currently we use the latest version each time we build the image, and we are looking for solution but since we need the have different image for each `node` version, we will need to build 20 images (5 for each node version) and our versioning would be weird : 
- `pm2:2.9.2-node8-alpine` 
- `pm2:2.8.0-node7-wheezy` 

Do someone has an idea on how to handle this without these problems ?

TimWolla:
As a Docker user that previously used pm2 outside of containers [1] I feel like this encourages users to not use Docker in a way Docker is optimized for: Immutable, reproducible containers that run a single atomic service.

The main issue of the PID 1 problem inside containers is easily solved by adding the `--init` flag, clustering should be done using Docker swarm and the last major issue “Graceful application Start and Shutdown” I don't understand (though I'm a bit out of the loop with pm2's features).

All in all I don't really see the benefit in an official pm2 image that after all only saves the user a single `RUN npm install pm2 -g` in their Dockerfile and causes maintenance work for both you and the official images team.

Am I missing something?

[1] In fact I still do on a machine for a few legacy services.

Unitech:
Hi @TimWolla,

Using PM2 to run Node.js in production is beneficial for these reasons:
- Faster restart; instead of having the whole containers being restarted in case of exception / memory overflow etcetera, having pm2 handling this directly inside containers reduce the downtime of the instance (from seconds to milliseconds)
- Process fallback; having two stateless processes load balanced with pm2 (you might be familiar with the cluster mode) avoid sending http queries to a a dying process, increasing again the SLA of the application
- Process flow control; pm2 offers different methods for executing code before and after the process get online/shutdown
- Automatic source map discovery; when running transpiled source code with pm2, in case of exception, pm2 will automatically resolve the exception and will print out the original line

Just to cite some of the key features when using pm2 inside containers (w/o speaking about process files, automatic json logs transposition and so on) and we are moving toward offering more and more features for Node.js users that use containers.

I don't think the "single atomic service" is still a credo at Docker; I let you compare how PHP, Python, Java applications are handled inside containers.

After all it's a question of technical choice. We have a growing user base using Docker & PM2 and we are now doing efforts to answer to the community by offering a good integration with Docker (now there is even a pm2-docker binary that is a pre-configured PM2 for Docker + we built an Docker official image and now there is this PR).

Best,
Alexandre

tianon:
@vmarchaud I think there's been a little misunderstanding here -- @TimWolla is definitely not confused about the benefits of PM2, but on the contrary is a fan (and is actively using it in at least one deployment, from the sounds of it)

The thing @TimWolla is trying to point out is that having a dedicated `pm2` image doesn't make much sense, and that it'd make a lot more sense for folks to simply include `RUN npm install ...` and `CMD` in their own images (since that's all that's required).  That would cut out the maintenance burden on you folks for maintaining every `FROM node` + supported `pm2` version combination, us the maintenance burden of reviewing those, and our servers the queue length to build/push them all.

There have been other images somewhat similar to what's proposed here, at least in their intended usage, that have since been deprecated.

The most similar that comes to mind is `celery`: https://github.com/docker-library/docs/pull/880 (https://github.com/docker-library/celery/issues/12, https://github.com/docker-library/celery/issues/1#issuecomment-287655769).  It's only useful in the context of adding one's own application code, and is not meaningful to run as a standalone service (similar to `pm2`).

Other examples I can think of are `django` (https://github.com/docker-library/docs/pull/630) and `rails` (https://github.com/docker-library/docs/pull/503).

vmarchaud:
@tianon Forgot to answer but thanks for the explanation, i understand that the cost is too high here !

PR #3415, opened 2017-09-06
Update erlang with multiple architectures

tianon:
See https://github.com/c0b/docker-erlang-otp/pull/70. :metal:

cc @c0b

PR #3956, opened 2018-01-31
Build as a multi-arch image

c0b:
this add multi-arch support as similar to #3415 to Erlang; I did manually test amd64 and arm64v8, and as Elixir designed in pure Elixir/Erlang code only, no C code, should be able to run anywhere that Erlang VM runs, so I added other architectures from each architecture supported by base Erlang image

| Elixir version | FROM Erlang version | Arch |
|---|--|--|
| 1.6 1.5 | Erlang 20 | amd64, arm32v7, arm64v8, i386, s390x, ppc64le (the alpine didn't verify on arm32v7) |
| 1.4 1.3 | Erlang 19 | amd64, arm32v7, arm64v8, i386, s390x |
| 1.2 | Erlang 18 | amd64, arm32v7, arm64v8, i386, s390x |

resolves c0b/docker-elixir#47

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ services: docker
 env:
   - DIR=master
   - DIR=1.5
+  - DIR=1.5 VARIANT=slim
   - DIR=1.4
   - DIR=1.4 VARIANT=slim
   - DIR=1.3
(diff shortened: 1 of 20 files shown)


c0b:
sorry did something wrong; close this one and open 3957 instead

PR #3958, opened 2018-01-31
Upgrade XWiki to 10.0

gdelhumeau:


vmassol:
It was closed since it's included in https://github.com/docker-library/official-images/pull/3973

Issue #3383, opened 2017-08-29
Add multi-stage build support


emilevauge:
Hi folks,
I'm opening this issue to track [multi-stage builds](https://docs.docker.com/engine/userguide/eng-image/multistage-build/) support in official-images.
Any ETA on this ?

StefanScherer:
Same demand to build optimal Node.js image in nodejs/docker-node#362 as Windows isn't as good in cleaning up temp files/packages/cached MSI.
* Windows builds still use Docker 1.12.2-cs2-ws-beta (17.06.1-ee-2 is available)
* ARM build are already use Docker 17.06.1-ce
* AMD builds use Docker 17.03.2-ce
Is there anything the community can help with?

yosifkit:
Here is a list we came up with so far, feel free to comment with more.

Positive points:
 - can generate nanoserver images where the installer exe would not normally run
 - upcoming nanoserver ([version 1709](https://blogs.technet.microsoft.com/windowsserver/2017/09/13/sneak-peek-3-windows-server-version-1709-for-developers/)) based images are mostly useless without multi-stage since they have no powershell
 - can download or generate a full rootfs (no more huge tarballs in git with force pushed branches)
 - large build container with small end result (hello-world, traefik)

Counter points:
 - can mostly be accomplished via long `RUN` lines
   - this will be made easier with https://github.com/moby/moby/issues/34423
 - increases space required on build servers
 - can greatly increase build time (busybox building binaries: 37 min for just amd64, vs a Dockerfile with `ADD busybox.tar.xz /`)
 - updates to either base image will require a rebuild and push
 - is a single binary image useful?
   - hard to debug, no shell or other utilities
   - saves about 4 MB to be from scratch vs alpine

This is more complex than just the Docker version of the systems building the images.
 - how do we clean dangling images without invalidating cache of every mutli-stage image?
 - hardcoded assumptions of a single `FROM` in many places
   - bashbrew itself
     - one example of many: in order to provide more robust build caching, bashbrew does a docker tag on an image with a hash of unique bits that includes the parent image ID
   - many jenkins jobs, shell scripts, tooling, etc used in building, tagging, pushing
     - finding instances of this assumption is difficult, but is still only half the battle; for example, we use the `Architectures` of the parent image of a given `Dockerfile` to determine which `Architectures` this image supports -- in a multi-stage world, that needs to instead calculate the set intersection of _all_ the parent images' `Architectures`
[...]

PR #3980, opened 2018-02-06
Add Express-Gateway to the official images

XVincentX:
Hello guys,

I'd love Express-Gateway, the project we're building, be part of the Docker Official Image program.

I'm quite sure we'll have a lot to work on, but this is a start.

Thanks a lot!

# Checklist for Review

    - https://github.com/docker-library/docs/pull/1144

### DIFF
--- a/library/express-gateway
+++ b/library/express-gateway
@@ -2,6 +2,6 @@ Maintainers: The Express-Gateway Team <info@express-gateway.io> (@express_gatewa
 GitRepo: https://github.com/ExpressGateway/docker-express-gateway.git
 
 Tags: 1.x, 1.13.x, 1.13.0
-Architectures: arm32v7, arm64v8, amd64, i386, ppc64le, s390x
-GitCommit: 52ffe46a4fe4a21700a9c66225d19425cb7a104e
+Architectures: arm64v8, amd64, i386, ppc64le, s390x
+GitCommit: 0c92fd48415e952be0aa037a1c7355df2da64e0e
 Directory: alpine
--- a/library/express-gateway
+++ b/library/express-gateway
@@ -1,7 +1,7 @@
 Maintainers: The Express-Gateway Team <info@express-gateway.io> (@express_gateway)
 GitRepo: https://github.com/ExpressGateway/docker-express-gateway.git
 
-Tags: 1.x, 1.10.x, 1.10.2
+Tags: 1.x, 1.11.x, 1.11.0
 Architectures: arm32v7, arm64v8, amd64, i386, ppc64le, s390x
-GitCommit: 266104b97f83e825293bcee6bffa644139628e71
+GitCommit: dde8e25dc3a698482b3fdf086c2a67a4fd471207
 Directory: alpine
--- a/library/express-gateway
+++ b/library/express-gateway
@@ -1,5 +1,8 @@
 Maintainers: The Express-Gateway Team <info@express-gateway.io> (@express_gateway)
 GitRepo: https://github.com/ExpressGateway/express-gateway.git
 
+Tags: 1.x, 1.8.x, 1.8.2
+GitCommit: df515d429bd06c596839d512290cc3fd3a3ba82b
+
 Tags: 1.x, 1.7.x, 1.7.3
 GitCommit: b715a408bd5c39636c0925b1739ffbc71072cc70
--- a/library/express-gateway
+++ b/library/express-gateway
@@ -2,5 +2,6 @@ Maintainers: The Express-Gateway Team <info@express-gateway.io> (@express_gatewa
 GitRepo: https://github.com/ExpressGateway/docker-express-gateway.git
 
 Tags: 1.x, 1.10.x, 1.10.2
+Architectures: arm32v7, arm64v8, amd64, i386, ppc64le, s390x
 GitCommit: 266104b97f83e825293bcee6bffa644139628e71
 Directory: alpine
--- a/library/express-gateway
+++ b/library/express-gateway
@@ -1,7 +1,11 @@
 Maintainers: The Express-Gateway Team <info@express-gateway.io> (@express_gateway)
 GitRepo: https://github.com/ExpressGateway/express-gateway.git
 
-Tags: 1.x, 1.8.x, 1.8.2
+Tags: 1.x, 1.9.x, 1.9.1
+GitCommit: 182691960d62d9398f7483e9122174124faa02bc
+
+
+Tags: 1.8.x, 1.8.2
 GitCommit: df515d429bd06c596839d512290cc3fd3a3ba82b
 
 Tags: 1.7.x, 1.7.3
--- a/library/express-gateway
+++ b/library/express-gateway
@@ -1,7 +1,7 @@
 Maintainers: The Express-Gateway Team <info@express-gateway.io> (@express_gateway)
 GitRepo: https://github.com/ExpressGateway/docker-express-gateway.git
 
-Tags: 1.x, 1.11.x, 1.11.0
+Tags: 1.x, 1.12.x, 1.12.0
 Architectures: arm32v7, arm64v8, amd64, i386, ppc64le, s390x
-GitCommit: dde8e25dc3a698482b3fdf086c2a67a4fd471207
+GitCommit: 7d329fb7c09a5c946bf3661e27abbffddd3b09ea
 Directory: alpine
--- a/library/express-gateway
+++ b/library/express-gateway
@@ -2,4 +2,4 @@ Maintainers: The Express-Gateway Team <info@express-gateway.io> (@express_gatewa
 GitRepo: https://github.com/ExpressGateway/express-gateway.git
 
 Tags: 1.x, 1.7.x, 1.7.3
-GitCommit: b1859d557491c0f6ddd07d8b8bb1c48bed5c57de
+GitCommit: b715a408bd5c39636c0925b1739ffbc71072cc70
--- a/library/express-gateway
+++ b/library/express-gateway
@@ -1,12 +1,6 @@
 Maintainers: The Express-Gateway Team <info@express-gateway.io> (@express_gateway)
-GitRepo: https://github.com/ExpressGateway/express-gateway.git
+GitRepo: https://github.com/ExpressGateway/docker-express-gateway.git
 
-Tags: 1.x, 1.9.x, 1.9.1
-GitCommit: 182691960d62d9398f7483e9122174124faa02bc
-
-
-Tags: 1.8.x, 1.8.2
-GitCommit: df515d429bd06c596839d512290cc3fd3a3ba82b
-
-Tags: 1.7.x, 1.7.3
-GitCommit: b715a408bd5c39636c0925b1739ffbc71072cc70
+Tags: 1.x, 1.10.x, 1.10.2
+GitCommit: 266104b97f83e825293bcee6bffa644139628e71
+Directory: alpine
--- a/library/express-gateway
+++ b/library/express-gateway
@@ -1,7 +1,7 @@
 Maintainers: The Express-Gateway Team <info@express-gateway.io> (@express_gateway)
 GitRepo: https://github.com/ExpressGateway/docker-express-gateway.git
 
-Tags: 1.x, 1.12.x, 1.12.0
+Tags: 1.x, 1.13.x, 1.13.0
 Architectures: arm32v7, arm64v8, amd64, i386, ppc64le, s390x
-GitCommit: 7d329fb7c09a5c946bf3661e27abbffddd3b09ea
+GitCommit: 52ffe46a4fe4a21700a9c66225d19425cb7a104e
 Directory: alpine
--- a/library/express-gateway
+++ b/library/express-gateway
@@ -1,0 +1,5 @@
+Maintainers: The Express-Gateway Team <info@express-gateway.io> (@express_gateway)
+GitRepo: https://github.com/ExpressGateway/express-gateway.git
+
+Tags: 1.x, 1.7.x, 1.7.3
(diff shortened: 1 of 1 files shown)


XVincentX:
@yosifkit Any chance you guys could have a look into this?

XVincentX:
@tianon Sorry to ping you personally — I realized you have been merging some pull requests lately and I was wondering whether you could help me with this.

Is there anything else I should be doing here? Thanks!

tianon:

Sorry for the delay!

I think my first concern here is around the location of the `Dockerfile` -- I don't think you really want to have the `Dockerfile` you submit to us to be tied into your main Express Gateway development repository, especially given that the `Dockerfile` as-is won't build properly via our tooling (there's no way to populate that `ARG EG_VERSION` value here), and we need some way to ensure the `Dockerfile` rebuilds properly (and always contains the same version when rebuilt).  In addition, your development repository will likely have a lot more churn in it, so it's harder to separate commits touching things the `Dockerfile` uses from other normal development commits.

There's quite a bit of content in https://github.com/docker-library/official-images#review-guidelines that I think will be helpful (especially "repeatability"), but there are also little things that'll need updating like swapping `CMD node -e "require('express-gateway')().run();"` over to `CMD ["node", "-e", "require('express-gateway')().run();"]` (which ensures there's no extra `/bin/sh` process running).

XVincentX:
@tianon 

Hey, thanks a lot for the reply!

Good point about the separate repository. This was something I wanted to do since the beginning, but given we have some automation on the repository (we automatically build Docker Images and other stuff when we publish a new version) I wanted to make sure that

1. The image makes sense as an official one
2. You guys would review the thing.

I'd be more than happy to make the change once we're all set with the review. Does that sound good to you?

> but there are also little things that'll need updating like swapping CMD node -e "require('express-gateway')().run();" over to CMD ["node", "-e", "require('express-gateway')().run();"] (which ensures there's no extra /bin/sh process running).

Oh cool, I didn't know such thing. Why is that happening when writing the `CMD` as a string?

XVincentX:
@tianon I've just merged a change that's changing the CMD statement. Is there anything else you want me to address before I start moving the Docker images in their own repository?

tianon:
> I'd be more than happy to make the change once we're all set with the review. Does that sound good to you?

Sure, fair enough -- will you still be maintaining the older versions, or will they be dropped?  (It's up to you -- the guideline we usually recommend is that you should only maintain whatever current releases are actively maintained upstream unless you've got a good reason to deviate.)

---

> Oh cool, I didn't know such thing. Why is that happening when writing the `CMD` as a string?

See https://docs.docker.com/engine/reference/builder/#cmd:

> If you use the *shell* form of the `CMD`, then the `<command>` will execute in
> `/bin/sh -c`:
> 
>     FROM ubuntu
>     CMD echo "This is a test." | wc -
> 
> If you want to **run your** `<command>` **without a shell** then you must
> express the command as a JSON array and give the full path to the executable.
> **This array form is the preferred format of `CMD`.** Any additional parameters
> must be individually expressed as strings in the array:
> 
>     FROM ubuntu
>     CMD ["/usr/bin/wc","--help"]

---

> Is there anything else you want me to address before I start moving the Docker images in their own repository?

You've got a couple other items being `COPY`'d into the image -- will those be copied over to the new repository?  Are those not things that get included automatically when you install from Yarn?  It would also be acceptable to fetch those via a tagged release (ala https://github.com/ExpressGateway/express-gateway/tree/v1.10.2/lib/config ...), but it is preferred for those to be fetched in some way that provides some sort of fetching verification (checksum, GPG signature, etc), if possible (which Yarn and friends usually provide, so it'd be really great if those bits came directly through Yarn, but understandable if they don't / that's not feasible / the defaults don't make sense in a Docker environment).

Just to be clear: is Express-Gateway something users would run as a standalone service?  Or is it more similar to something like Django (and thus more likely to be listed in a project's `requirements.txt` and installed alongside the project code)?  We used to have an image for Django and Celery, for example, and found that the maintenance burden of them far outweighed any possible benefit of the image (since one doesn't run Django directly, but rather uses it _with_ an application listed in `requirements.txt` and downloaded/installed like any/all other dependencies), so the image was somewhere between hard to use and entirely useless. :sweat_smile:

XVincentX:
>  Is Express-Gateway something users would run as a standalone service?

Definitely. Express Gateway is an API Gateway proxying request and manipulating then. You download the image and it's up and running listening on a port.

> Will those be copied over to the new repository? Are those not things that get included automatically when you install from Yarn?

I think these files are already included in the yarn package. That would just be matter of moving files around. I'll figure this out.

> we usually recommend is that you should only maintain whatever current releases are actively maintained upstream

That makes sense. Perfect!

XVincentX:
@tianon Are multi stage builds supported?

TimWolla:
> @tianon Are multi stage builds supported?

They are not. See #3383.

XVincentX:
Thanks @TimWolla. I can live with that for now!

XVincentX:
@tianon @TimWolla This should now be ready.

XVincentX:
Actually I have got a followup question — is there any way to provide a redirect from the old image `expressgateway/express-gateway` to the new one that'll be published? Or is the "migration" something I would have to communicate around to all the users?

XVincentX:
@tianon Any update on this?

XVincentX:
@tianon @TimWolla @yosifkit 

Guys — this has been here for a lot of time. Is there anything I can do to speed up the process?

TimWolla:
@XVincentX Please note: I'm not a maintainer of the Official Images library. I “only” maintain a few official images and thus are aware of the process. Thus this is entirely out of my control.

XVincentX:
I see, sorry for involving you in. I really have no idea what to do here :(

tianon:
What does it do if the provided VOLUME is empty, like it will be if a user
bind-mounts instead of making a new "Docker volume"?  If the user lets
Docker create a new volume, it'll get pre-filled with the content of the
image at the volume location, which will be this default configuration
you've provided, but in every other case it'll stay empty, which I imagine
will probably make Express-Gateway fail to start?

XVincentX:
Yes, it won't be able to start saying there are missing files @tianon 

Any suggestions on how to handle this better?

tianon:
If there's not a way to convince Express-Gateway itself to regenerate in the face of an empty directory (or for it to provide sane defaults that can then be modified and saved via the Admin API endpoints you referenced), then I guess the best/only option is probably something hacky like what the WordPress image does -- put the default configuration files elsewhere, during startup detect whether key files we expect exist, and if they don't, copy in the defaults.

It definitely seems weird that the API is allowed to mutate the configuration, but if that's how it's supposed to work, so be it!

XVincentX:
@tianon Thanks for the response! I'm so glad we're interacting so quickly after all this time waiting!

> If there's not a way to convince Express-Gateway itself to regenerate in the face of an empty directory (or for it to provide sane defaults that can then be modified and saved via the Admin API endpoints you referenced), then I guess the best/only option is probably something hacky like what the WordPress image does -- put the default configuration files elsewhere, during startup detect whether key files we expect exist, and if they don't, copy in the defaults.

I guess this could be done in a `docker-entrypoint.sh` file — I would copy the files somewhere, mount the volume — check the files existence and if not — copy the files in the volume to initialize it — is it correct?

Followup question on that — why is crashing not an option? Wouldn't be ok to say "yeah ok thanks for the volume but I need some files into?" Just out of curiosity.

> It definitely seems weird that the API is allowed to mutate the configuration, but if that's how it's supposed to work, so be it!

I can understand the concern — but I won't tell you the story how the project ended up being there. It's a change I think we'll introduce, but in the next major version though, so 2.0.0.

XVincentX:
@tianon 

I've added now a `docker-entrypoint.sh` with the required logic to check for file presence and put defaults in case there are not provided. Let me know if I've missed something else or not.

XVincentX:
@tianon Ping

tianon:
Ok, let's get back on this! :sweat_smile: :heart: (Your patience is very much appreciated.)

The `docker-entrypoint.sh` was added, but does appear to have been removed by your version bumping bot in https://github.com/ExpressGateway/docker-express-gateway/commit/b4886838dafedfefae74a63d0a5031904afb45b2 (FYI).

The entrypoint (https://github.com/ExpressGateway/docker-express-gateway/blob/d836632e50e5a26a9fdb33fa384e253a0cd11fb2/alpine/docker-entrypoint.sh) doesn't currently meet our https://github.com/docker-library/official-images#consistency requirements -- here's what I'd recommend:

1. Adjust `Dockerfile` something like so:

    ```dockerfile
    CMD ["node", "-e", "require('express-gateway')().run();"]
    ```

2. Adjust the end of `docker-entrypoint.sh` something like so:

    ```bash
         ...
    fi

    exec "$@"
    ```

There's a similar approach in https://github.com/mongo-express/mongo-express-docker which might be useful as an example (although it uses a dummy `mongo-express` command to indicate that `node app` will be run so that the particulars of the exact `node` command that gets invoked don't have to be part of the `Dockerfile`'s `CMD` instruction).

Also, you've currently got your default config content split between the files you `COPY` and the `entrypoint.sh` script (which will likely eventually get out of sync) -- it would probably be worthwhile to instead `COPY` them to somewhere outside the volume so that the script can simply `cp` them if they don't exist.  What do you think?

To clarify what I meant by "included in the yarn package", I meant some of these:

```console
$ find / -name gateway.config.yml -o -name system.config.yml
/usr/local/share/.config/yarn/global/node_modules/express-gateway/bin/generators/gateway/templates/getting-started/config/gateway.config.yml
/usr/local/share/.config/yarn/global/node_modules/express-gateway/bin/generators/gateway/templates/getting-started/config/system.config.yml
/usr/local/share/.config/yarn/global/node_modules/express-gateway/bin/generators/gateway/templates/basic/config/gateway.config.yml
/usr/local/share/.config/yarn/global/node_modules/express-gateway/bin/generators/gateway/templates/basic/config/system.config.yml
/usr/local/share/.config/yarn/global/node_modules/express-gateway/lib/config/gateway.config.yml
/usr/local/share/.config/yarn/global/node_modules/express-gateway/lib/config/system.config.yml
/usr/local/share/.cache/yarn/v2/npm-express-gateway-1.14.0-3948cb03ca51f0d629576af07fee2275eb4bcf8e/bin/generators/gateway/templates/getting-started/config/gateway.config.yml
/usr/local/share/.cache/yarn/v2/npm-express-gateway-1.14.0-3948cb03ca51f0d629576af07fee2275eb4bcf8e/bin/generators/gateway/templates/getting-started/config/system.config.yml
/usr/local/share/.cache/yarn/v2/npm-express-gateway-1.14.0-3948cb03ca51f0d629576af07fee2275eb4bcf8e/bin/generators/gateway/templates/basic/config/gateway.config.yml
/usr/local/share/.cache/yarn/v2/npm-express-gateway-1.14.0-3948cb03ca51f0d629576af07fee2275eb4bcf8e/bin/generators/gateway/templates/basic/config/system.config.yml
/usr/local/share/.cache/yarn/v2/npm-express-gateway-1.14.0-3948cb03ca51f0d629576af07fee2275eb4bcf8e/lib/config/gateway.config.yml
/usr/local/share/.cache/yarn/v2/npm-express-gateway-1.14.0-3948cb03ca51f0d629576af07fee2275eb4bcf8e/lib/config/system.config.yml
```

If any of these is suitable for the initial volume initialization, then you might be able to skip those `COPY` bits entirely and just `cp` these in the entrypoint instead, right?  That gives you less duplication to maintain (so is likely worth considering).

(Those files in `.cache` are probably ripe for pruning to make the image a little smaller too -- maybe there's a way to ask Yarn not to cache them, or perhaps a `yarn cache clean` command of some kind to remove them properly?)

Other than that, I think this looks OK and we should look to get you across the finish line here shortly. :+1:

XVincentX:
@tianon 

> The docker-entrypoint.sh was added, but does appear to have been removed by your version bumping bot in ExpressGateway/docker-express-gateway@b488683 (FYI).

Correct. I've restored it and updated the template.

> Adjust Dockerfile something like so

Done

> Adjust the end of docker-entrypoint.sh something like so

Done

> perhaps a yarn cache clean command…

That's a way. Or, in case the docker builds you guys do allow to set ´--shm-size´ option during the build phase, then I could use that as a cache directory so its content gets discarded automatically once the build is done.

Good point also about the gateway and system configuration file. I'll look into the alternatives to see if that makes sense or not.

XVincentX:
@tianon Done! I should have addressed all your comments!

yosifkit:
There are also now `config/models/*` files that are dropped into the folder that becomes a volume and not addressed by the entrypoint.

Most of the time, relying on Docker's autocopy when using a `docker volume` is not great; in order to work for all users, you have to have a second copy of the files and replicate the behavior in the entrypoint for those doing a bind-mount. You could either `cp` them from the examples included with the npm install, or `COPY` them elsewhere, like `/usr/local/etc/eg/`, and `cp` from there to the volume on start-up. If there are quite a few files, perhaps something like `rsync --ignore-existing` would do what you need?

XVincentX:
Yeah I forgot about the model files, I'll have a look into this later. Thanks for pointing it out @yosifkit

XVincentX:
@tianon @yosifkit 
I've added the latest tag. Shall I remove the `credentials.json file not found in / - copying now...` sentences here and there?

yosifkit:
The output could be useful to users, so I would just skip them if the command isn't `node` or something.
```bash
# maybe like this?
if [ "$1" = 'node' ]; then 
# do copy stuff
fi
exec "$@"
```

XVincentX:
Is it ok if I address this in a following PR?

yosifkit:
> Is it ok if I address this in a following PR?

Fine by me. It passes the tests we have :smile:

XVincentX:
This is so great! Finally it happened!

PR #4002, opened 2018-02-12
Add common image

arthurdm:
Adding a new tag to websphere-liberty, called `common`, which will be used by some of the other tags to reduce the number of layers and remove the need to run the `installUtility` command in other images.

### DIFF
--- a/library/websphere-liberty
+++ b/library/websphere-liberty
@@ -7,6 +7,9 @@ Architectures: amd64, i386, ppc64le, s390x
 Tags: kernel
 Directory: ga/developer/kernel
 
+Tags: common
+Directory: ga/developer/common
+
 Tags: microProfile
 Directory: ga/developer/microProfile
 


tianon:
I'm a bit confused -- isn't this already what `websphere-liberty:kernel` represents? (which is the base image of almost all other variants)

```console
$ bashbrew from --uniq websphere-liberty
websphere-liberty:kernel: ibmjava:8-jre
websphere-liberty:microProfile: websphere-liberty:kernel
websphere-liberty:webProfile6: websphere-liberty:kernel
websphere-liberty:webProfile7: websphere-liberty:kernel
websphere-liberty:javaee7: websphere-liberty:webProfile7
websphere-liberty:beta: ibmjava:8-jre
```

arthurdm:
hi @tianon - this is a layer above the Kernel.  So it's the kernel + features that will be needed by both the `webProfile7` tag and the `microProfile` tag.  The current Dockerfiles for webProfile/microprofile have a layer that run `installUtility` (for example https://github.com/WASdev/ci.docker/blob/master/ga/developer/webProfile7/Dockerfile#L21), so the new `common` layer will run that (expensive) command, so that the other images can start from that cached common layer.

arthurdm:
The goal is that after this image is created in docker hub, I will make the modifications to both `webProfile7` and `microProfile` to be simply:

```
FROM websphere-liberty:common
COPY server.xml /config/
```

yosifkit:
I thought microProfile and webProfile7 were quite different; currently their `installUtility` layers are 37MB and 75.1MB, respectively.  Wouldn't this add extra packages that would be considered bloat for a microProfile user?

tianon:
The proposed changes should be made together in a single PR -- we don't need to build and push `websphere-liberty:common` before we can build images based on it (the `test-pr` and image build scripts will build things in the right order automatically).

Also, this could be accomplished without an explicit tag by making the first few `RUN` lines of the two `Dockerfile`s identical and letting Docker's build cache do the right thing (since we make use of that quite heavily).

arthurdm:
Thanks for the feedback @yosifkit - we're taking another look at those layers and if it's worth the extra size.

@tianon - the PR to change the Dockerfiles for `webProfile7` and `microProfile` would be to another GitHub repo (https://github.com/WASdev/ci.docker), so I couldn't have a single PR that delivered here and there at the same time.   You bring up a very good point about the cached layer.  So if we flip the order of `RUN` and `COPY` in the 2 Dockerfiles in question, and make the `RUN` command exact the same, then the cache benefits will be the same as if those layers were in a standalone `common` image?

tianon:
Yes, if the textual contents of the two `RUN` commands are identical, they will get the cache benefits that the explicit `common` tag provides, but implicitly instead.

arthurdm:
Thanks @tianon - I will close this PR then, and will take advantage of this cached layer strategy.

PR #4032, opened 2018-02-21
adding arm32v7 to vault

chrishoffman:


library/vault
```diff
@@ -3,6 +3,6 @@ Maintainers: Chris Hoffman <choffan@hashicorp.com> (@hoffmancs),
 GitRepo: https://github.com/hashicorp/docker-vault.git
 
 Tags: 0.9.4, latest
-Architectures: amd64, arm64v8, arm32v6, i386
+Architectures: amd64, arm64v8, arm32v7, arm32v6, i386
```
yosifkit: `alpine:3.7` is not supported on arm32v7.

https://github.com/docker-library/official-images/blob/72971c38907657e4216e84161d8b785798b9c587/library/alpine#L29-L30

chrishoffman: I was trying to get this to work on a Rasberry Pi 3 but will likely need to create another image variant to support that.

tianon: This arm32v6 image should work as-is on Pi3 in 32bit mode, and arm64v8 in 64bit mode.

chrishoffman: I am seeing this:

```
pi@raspberrypi:~ $ docker run vault
Unable to find image 'vault:latest' locally
latest: Pulling from library/vault
docker: no matching manifest for linux/arm in the manifest list entries.
See 'docker run --help'.

pi@raspberrypi:~ $ lscpu
Architecture:          armv7l
Byte Order:            Little Endian
CPU(s):                4
On-line CPU(s) list:   0-3
Thread(s) per core:    1
Core(s) per socket:    4
Socket(s):             1
Model:                 4
Model name:            ARMv7 Processor rev 4 (v7l)
CPU max MHz:           1200.0000
CPU min MHz:           600.0000
BogoMIPS:              38.40
Flags:                 half thumb fastmult vfp edsp neon vfpv3 tls vfpv4 idiva idivt vfpd32 lpae evtstrm crc32```

tianon: Ah, looks like it had a build failure -- https://doi-janky.infosiftr.net/job/multiarch/view/images/view/vault/

(poking it now)

tianon: Also, looking at https://releases.hashicorp.com/vault/0.9.4/, it doesn't appear `arm64` actually exists? (hence the 403 forbidden error on those builds)

tianon: Give it a try on that same device now and it should work. :+1:

chrishoffman: Thanks, it works!  I will check on the arm64 release.

PR #4048, opened 2018-02-25
Replace bash with sh

MetalArend:
Alpine images implement sh instead of bash. Do they fail this guideline, or is it good enough to provide any shell available?


tianon:
Thanks @MetalArend!  The intent was for any shell (or command) the image contains -- I'd like to keep the default referencing `bash`, but mentioning `sh` explicitly is a great idea, thanks! (I've applied e3cde539 to that end). :heart:

PR #4148, opened 2018-03-20
Removed 'latest' tag from postgres variant

doublebyte1:


tianon:
LGTM (tag changes only, so merging)

:heart:

PR #4184, opened 2018-03-28
Add a flag to force push images in case a newer but wrong image was pushed in its place

yosifkit:
This is part of one possible solution to help us with cleaning up https://github.com/docker-library/openjdk/issues/180.

### DIFF
--- a/bashbrew/go/src/bashbrew/cmd-push.go
+++ b/bashbrew/go/src/bashbrew/cmd-push.go
@@ -18,6 +18,7 @@ func cmdPush(c *cli.Context) error {
 	uniq := c.Bool("uniq")
 	namespace := c.String("namespace")
 	dryRun := c.Bool("dry-run")
+	withForce := c.Bool("force")
 
 	if namespace == "" {
 		return fmt.Errorf(`"--namespace" is a required flag for "push"`)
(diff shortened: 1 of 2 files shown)


yosifkit:
@tianon has a different solution that will work better. (a script on the build server to push any image that is different than the one on the hub)

bashbrew/go/src/bashbrew/cmd-push.go
```diff
@@ -44,7 +45,7 @@ func cmdPush(c *cli.Context) error {
 
 				created := dockerCreated(tag)
 				lastUpdated := fetchDockerHubTagMeta(tag).lastUpdatedTime()
-				if created.After(lastUpdated) {
+				if created.After(lastUpdated) || withForce {
```
tianon: I think we should probably skip `fetchDockerHubTagMeta` in the case of `--force`, something like:
```go
if !withForce {
	created := dockerCreated(tag)
	lastUpdated := fetchDockerHubTagMeta(tag).lastUpdatedTime()
	if !created.After(lastUpdated) {
		fmt.Fprintf(os.Stderr, "skipping %s (created %s, last updated %s)\n", tag, created.Local().Format(time.RFC3339), lastUpdated.Local().Format(time.RFC3339))
		continue
	}
}
fmt.Printf("Pushing %s\n", tag)
...
```

PR #4200, opened 2018-04-02
Liberty 18001

NottyCode:
* Update to Liberty 18.0.0.1.
* Stop using a server.xml checked into git, instead use the server.xml from the downloaded zip from maven central.


yosifkit:
Can you rebase your branch and force push to resolve the conflict?

NottyCode:
screwed up my repository so I'll resubmit

PR #4206, opened 2018-04-03
Update 2.4-alpine to reference 2.4-alpine3.7

FreekingDean:
ruby:2.4-alpine still references ruby:2.4-alpine3.4. This is no longer the latest alpine version for ruby2.4. This PR updates the tag to reference alpine3.7 instead.


tianon:
```bash
# this file is generated via https://github.com/docker-library/ruby/blob/bf3b9e755c36c25dde89c832c293abe87292e482/generate-stackbrew-library.sh
```
:wink:

This should be discussed further over in https://github.com/docker-library/ruby/issues/199.

FreekingDean:
Oh man, I feel like a dope, thank you!

PR #4221, opened 2018-04-05
pull request from photon:1.0-20180226

gerrit-photon:
pull request from photon:1.0-20180226


gerrit-photon:
Also a test pull request. Please skip it

PR #4239, opened 2018-04-11
Fix CentOS branch capitalization

tianon:
FYI @jperrin


tianon:
(This manifests as `fatal: Couldn't find remote ref refs/heads/centos-7armhfp`.)

tianon:
No material changes here, merging. :+1:

PR #4196, opened 2018-03-31
Rename piwik to matomo

pierreozoux:
linked to https://github.com/matomo-org/docker/pull/88

PR #2698, opened 2017-02-27
Remove deprecated "java" manifest (updating rapidoid and tomee to account for the deprecation)

tianon:
cc @nmihajlovski @lordofthejars (FYI)

PR #2046, opened 2016-08-11
Add a new "openjdk" image (renamed from "java") so that "java" can be deprecated

tianon:
https://github.com/docker-library/docs/pull/660

PR #4243, opened 2018-04-12
Exclude removed files

muhlba91:
When **removing** official docker images (e.g. #4196, #2046, #2698), the build will fail since it detects the removed file as an image to check for.

In fact, we should **ignore** those removed files and mark the build as passed once all other changed/added images have been built successfully. Therefore, we have to filter the `git diff` results according to their status.

### DIFF
--- a/bashbrew/travis.sh
+++ b/bashbrew/travis.sh
@@ -26,7 +26,7 @@ elif [ "$(git diff --numstat "$UPSTREAM...$HEAD" -- . | wc -l)" -ne 0 ]; then
 	echo >&2 'Changes in bashbrew/ detected!'
 	extraCommands=1
 else
-	repos=( $(git diff --numstat "$UPSTREAM...$HEAD" -- ../library | awk -F '/' '{ print $2 }') )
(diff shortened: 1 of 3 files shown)


muhlba91:
Can you please investigate why this build fails now? Imo, the `rev-parse` part hasn't changed in my changes.

Log output:
```
bashbrew list --build-order --all
...
command: git ["rev-parse" "fd39ce9631c3b6a110b6c1e8e5756a20d0863a77^{commit}"]
fatal: ambiguous argument 'fd39ce9631c3b6a110b6c1e8e5756a20d0863a77^{commit}': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

muhlba91:
@yosifkit @tianon - does anyone of you has a lead on the error occurring in the build? :)

tianon:
Sorry for the large delay. :bow: :heart:

I appreciate the idea, but I don't think image deletion happens often enough that we need to handle this case explicitly -- we don't require passing checks for a merge, and instead rely on maintainer discretion, so it doesn't cause us any real harm.  IMO this also helps make image deletions (which we don't want to encourage :sweat_smile: :heart:) more obvious.

PR #4301, opened 2018-04-28
Update node to have correct `latest` tags

SimenB:
See https://github.com/nodejs/docker-node/pull/705

Reported here: https://github.com/docker-library/official-images/pull/4289#issuecomment-385134176


tianon:
Tag changes only; merging 👍

teppeis:
@SimenB @tianon 
`node:10` is now available, but `node:latest` is not updated (it's same as `node:9`).
Also `node:alpine` and `node:wheezy` are updated,  but `node:slim` and `node:stretch` are NOT updated.

tianon:
I think some of the code we've got to optimize pushes got confused here -- I've gone and poked things and tags appear to be correct now. :+1:

PR #4308, opened 2018-05-01
Adding a CentOS tag

arthurdm:
Adding a new WebSphere Liberty image that builds upon the CentOS base image.  This OS is often used in cloud environment that require the docker container to run as non-root, so the corresponding [Dockerfile](https://github.com/WASdev/ci.docker/blob/master/ga/developer/centos/Dockerfile) accomplishes this.  

The technique of externally referenced multi-stage is used to avoid code duplication for fetching IBM Java and IBM WebSphere Liberty binaries.


arthurdm:
will resubmit with an update

PR #4326, opened 2018-05-07
Adding centOS tag

arthurdm:
Adding a tag for CentOS-based image.  This image runs as non-root, which allows it to run in restricted environments such as OpenShift.

This image is also important for developers that want to use CentOS for dev and RHEL for production.

### DIFF
--- a/library/websphere-liberty
+++ b/library/websphere-liberty
@@ -3,7 +3,6 @@ Maintainers: Wendy Raschke <wraschke@us.ibm.com> (@wraschke),
              Arthur De Magalhaes <arthurdm@ca.ibm.com> (@arthurdm)
 GitRepo: https://github.com/WASdev/ci.docker.git
 GitCommit: 6c582d04a7201e23f1af9dc0d955a0f66cb53354
-
 Architectures: amd64, i386, ppc64le, s390x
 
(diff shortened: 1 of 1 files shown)


tianon:
This `Dockerfile` includes `COPY --from=websphere-liberty:kernel ...` which suffers from the same issues as multi-stage builds (enumerated/discussed over in https://github.com/docker-library/official-images/issues/3383#issuecomment-329259161).

arthurdm:
Thx @tianon - I added a comment in that issue.

tianon:
I'm not sure I fully understand the importance of providing a CentOS-based
image as well -- isn't part of the point of using Java and tools
like WebSphere Liberty that you don't care as much about the underlying
operating system?  What does this change for usage of the image from a
functional perspective?

arthurdm:
hi @tianon - any updates on this?

yosifkit:
This comment still applies: https://github.com/docker-library/official-images/pull/4326#issuecomment-387190856. Unfortunately, none of our tooling can handle multi-stage builds including just `COPY --from` another official image.

arthurdm:
ok.   thanks @yosifkit - I will close this PR then.

PR #4375, opened 2018-05-21
Add Friendica

nupplaphil:
Hi,

I do want to push some Friendica base images to the official-images.
This is the first PR to make things work. In case we get merged, we prepare the images for other architectures (e.g. `i386`) and maybe other PHP version (e.g. `5.6`).

Upstream support: @friendica @MrPetovan @tobiasd

Base distribution for Friendica (a decentralized social network)

See https://friendi.ca

See https://github.com/docker-library/docs/pull/1234

As good as we can. The images are influenced by the docker-images of https://github.com/nextcloud/docker

I'm afraid, I don't think so. @MrPetovan @tobiasd made a review, but we are not THAT familiar with docker. I'm creating personal docker-images/containers for about 2 years.

The base is `php:7.1` (apache, fpm, fpm-alpine)

See https://travis-ci.org/friendica/docker

### DIFF
--- a/library/friendica
+++ b/library/friendica
@@ -26,7 +26,7 @@ Architectures: amd64
 GitCommit: 9c954f4d6bc90c61f9b3552fbbf8a5f6e70580ad
 Directory: develop/fpm
 
-Tags: 208-05-dev-fpm-alpine, develop-fpm-alpine, dev-fpm-alpine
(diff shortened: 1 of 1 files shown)


nupplaphil:
I'm closing this PR because I have some "homework" to do to make a sufficient repo for good docker images.

Therefore I will go on using the "friendica/server" repo instead until we get to the point that we are stable enough to get merged to the official repository :-)

PR #4452, opened 2018-06-12
mark all openjdk architectures as supported

keeganwitt:
For the same reasons as #4451, we can mark all Java supported architectures as supported for this image.


tianon:
See https://github.com/docker-library/official-images/pull/4451#issuecomment-397169679:

> Unfortunately, the arches for each tag need to match the arches of the tag they're `FROM`.  We usually generate the appropriate list via the `bashbrew` command, ala https://github.com/docker-library/python/blob/bbbc37fff3411a34deef30dd9b34dc938fe7b134/generate-stackbrew-library.sh#L44-L58 / https://github.com/docker-library/python/blob/bbbc37fff3411a34deef30dd9b34dc938fe7b134/generate-stackbrew-library.sh#L130.

keeganwitt:
I see.  I had some mismatches (specifically with the Alpine images).  I've corrected that.

PR #4487, opened 2018-06-22
HarperDB for Official Docker Image

zaxharperdb:
HarperDB is a fully indexed SQL and NoSQL database written in NodeJS, visit harperdb.io for more information on the product offerings.  The docker image will be based on the free community edition with the ability to register, in the future, for enterprise license.

### DIFF
--- a/library/harperdb
+++ b/library/harperdb
@@ -1,0 +1,8 @@
+Maintainers: Zachary Fowler <zachary@harperdb.io> (@zaxharperdb)
+GitRepo: https://github.com/HarperDB/official-docker/
+
+Tags: 1.0.0, 1.1, 1.2 latest
+amd64-GitCommit: d3dcdf4b3453036baa5b3e78f735374bc0ecaab2
+amd64-Directory: amd64/
+arm32v7-GitCommit: d3dcdf4b3453036baa5b3e78f735374bc0ecaab2
+arm32v7-Directory: arm32v7/
--- a/library/harperdb
+++ b/library/harperdb
@@ -1,7 +1,7 @@
 Maintainers: Zachary Fowler <zachary@harperdb.io> (@zaxharperdb)
 GitRepo: https://github.com/HarperDB/official-docker/
 
-Tags: 1.0.0, 1.1, 1.2 latest
+Tags: 1.0.0, 1.1, 1.2, latest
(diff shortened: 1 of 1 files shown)


zaxharperdb:
Thank you for the update.
I have not been one to poke you about. ;-)
I figured you have many request, you are a popular framework.
I appreciate your work and diligence.
Continue to be righteous.
Cheers,
-Z-

tianon:
Sorry for the delay (and thanks for your patience)! :heart:

> ```dockerfile
> ... && apt-get clean
> ```

Running `apt-get clean` is unnecessary -- the `debian:stretch-slim` base image performs that same function automatically following every `apt-get install` (what you might have meant instead is `rm -rf /var/lib/apt/lists/*`); see `/etc/apt/apt.conf.d/docker-clean` inside the image for more details.

> ```dockerfile
> ADD ./amd64/HarperDB-ce.tar /usr/local/bin/
> ```

This gives me a massive amount of pause -- what's inside `HarperDB-ce.tar`?  Where does it come from?  How is it built?  Why is it different for each architecture?  Is it officially published somewhere on the internet it could be reliably pulled from, thus allowing this complex two-`Dockerfile` repo structure to be flattened to a single multi-arch `Dockerfile` (which would also resolve pathing issues around the context directory which simply won't work as-is)?

Is the community edition of HarperDB open source, or merely free-to-use?  There won't be a click-through "accept these terms" for the EULA here (as there appears to be even for the Docker options on https://products.harperdb.io/download), so I'm not sure if the added direct-usage exposure of being a top-level image on the Docker Hub matches what the website seems to imply about usage of CE here.

> ```dockerfile
> COPY ./entrypoint.sh /usr/local/bin
> ```

The Docker cache behavior around `COPY`/`ADD` is still somewhat unpredictable -- this line should definitely move down as far as possible (likely just before `ENTRYPOINT` or `USER`) in order to allow Docker's cache to be as efficient as possible.

> ```dockerfile
> RUN cd /usr/local/bin/HarperDB/bin && ...
> ```

This line appears to be "installing" HarperDB -- does that duplicate many of the files provided by `HarperDB-ce.tar`, thus creating a larger-than-necessary layer/image?

---

The `entrypoint.sh` script appears to be kind of long -- is there a reason it's so involved?  Shouldn't more of this behavior belong in the official bits HarperDB distributes upstream instead?

PR #4540, opened 2018-07-06
Release Neo4j 3.4.3

jennyowen:
I know we skipped the 3.4.2 release. When we released 3.4.2 to Maven we accidentally packaged the wrong jar dependencies and so it's in an unknown state. You can't un-release from maven so we had to jump straight to 3.4.3 on all platforms.
I don't know what the Docker rules are on whether we can skip a release number.


jennyowen:
closing PR. Neo4j 3.4.3 has problems too. Please do not merge.

PR #4599, opened 2018-07-17
Add Architectures tag to support multiple architectures

odidev:
Mediawiki working fine for both ```AMD64 and ARM64v8``` architectures, without any code change.

So a normal build run is required to validate both architectures.
I have added ```Architectures tag``` to extend official support for ```arm64v8``` as well.

Please have a look and let me know, if anyhting else is required to make ```mediawiki``` ```docker pull```able for ```arm64v8``` too.

Signed-off-by: Odidev <odidev@puresoftware.com>

### DIFF
--- a/library/mediawiki
+++ b/library/mediawiki
@@ -3,6 +3,7 @@ Maintainers: David Barratt <dbarratt@wikimedia.org> (@davidbarratt),
              addshore <addshorewiki@gmail.com> (@addshore)
 GitRepo: https://github.com/wikimedia/mediawiki-docker.git
 GitCommit: 2dea40647eb85d6e0fc607904e0779644cf869c9
+Architectures: amd64, arm64v8
 
 Tags: stable, latest, 1.31, 1.31.0
 Directory: stable
--- a/library/mediawiki
+++ b/library/mediawiki
@@ -3,7 +3,7 @@ Maintainers: David Barratt <dbarratt@wikimedia.org> (@davidbarratt),
              addshore <addshorewiki@gmail.com> (@addshore)
 GitRepo: https://github.com/wikimedia/mediawiki-docker.git
 GitCommit: 2dea40647eb85d6e0fc607904e0779644cf869c9
-Architectures: amd64, arm64v8
+Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, ppc64le
 
 Tags: stable, latest, 1.31, 1.31.0
 Directory: stable


yosifkit:
I can't find any discussion on the tracker about adding arm64 to the MediaWiki images (https://phabricator.wikimedia.org/project/view/3094/).  

cc maintainers: @davidbarratt, @legoktm, @addshore  Do you want to add more architectures (in this case `arm64v8`) to the MediaWiki image?  I would guess that you should be able to add any that the base php images support.  Let us know what you would like to do here and if we can help.

legoktm:
MediaWiki should be fully architecture independent, so +1 from me.

davidbarratt:
That's fine with me.

@odidev can you update your PR to list all of these?
https://github.com/docker-library/official-images/blob/master/library/php#L13
thanks!

davidbarratt:
@yosifkit somewhat unrelated, shouldn't Docker just use, by default, whatever the base image supports rather than needing to copy/maintain our own list?

odidev:
@davidbarratt : Do you mean, that i should raise PR to support these?
```Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, ppc64le```
instead of just ```amd64, arm64v8``` ?

And what about testing of other architectures?

**Comment by davidbarratt** [comment]
@odidev Yes. We should support whatever platforms PHP supports. (per @legoktm)

I have no idea how this could be tested, but I'm ok _allowing_ the option and allowing people to bring up issues on different platforms as they find them.

**Comment by odidev** [comment]
@davidbarratt : :laughing: Good point. I agree with your thoughts.
I will update this PR to support all the mentioned architectures.

**Comment by odidev** [comment]
@davidbarratt : Thanks for approving the changes. :smile: 
Hope to see ```docker pull mediawiki``` on ```arm64v8``` soon.

**Comment by davidbarratt** [comment]
@odidev just out of curiosity, what ARM device(s) are you using?

**Comment by odidev** [comment]
@davidbarratt : I have Softiron-OD1000 boards with arm64v8 architecture.

**Comment by yosifkit** [comment]
@davidbarratt, yeah most projects aren't as architecture agnostic so they can't just inherit arches of the parent image.

LGTM

<details>
<summary>Diff:</summary>

```diff
diff --git a/_bashbrew-arches b/_bashbrew-arches
index 36d1546..9e6572a 100644
--- a/_bashbrew-arches
+++ b/_bashbrew-arches
@@ -1,3 +1,18 @@
 mediawiki:1.27.4 @ amd64
+mediawiki:1.27.4 @ arm32v5
+mediawiki:1.27.4 @ arm32v7
+mediawiki:1.27.4 @ arm64v8
+mediawiki:1.27.4 @ i386
+mediawiki:1.27.4 @ ppc64le
 mediawiki:1.30.0 @ amd64
+mediawiki:1.30.0 @ arm32v5
+mediawiki:1.30.0 @ arm32v7
+mediawiki:1.30.0 @ arm64v8
+mediawiki:1.30.0 @ i386
+mediawiki:1.30.0 @ ppc64le
 mediawiki:1.31.0 @ amd64
+mediawiki:1.31.0 @ arm32v5
+mediawiki:1.31.0 @ arm32v7
+mediawiki:1.31.0 @ arm64v8
+mediawiki:1.31.0 @ i386
+mediawiki:1.31.0 @ ppc64le
```

</details>

PR #4637, opened 2018-07-26
bashbrew: add private registry and Jenkins compatibility

7AC:
* account for `File:` in manifest (see https://github.com/docker-library/go-dockerlibrary/pull/16)
* fallback on `$GERRIT_REFSPEC` for fetching, which allows to checkout/build pending Gerrit reviews via Jenkins
* pass `$DOCKER_BUILD_COMMIT` argument to docker build, which gives the image being built a hold on its own version (e.g., to use in `<binary> --version` inside the image)
* add `-registry` to drive push, allowing to push selectively against private registries and not only DockerHub


tianon:
This is a _lot_ more than simply adding support for the new `File:` added in https://github.com/docker-library/go-dockerlibrary/pull/16, and I'm sorry to say that most of it cannot be accepted as-is (especially `GERRIT_REFSPEC` and `DOCKER_BUILD_COMMIT` -- we don't and won't use Gerrit so cannot comfortably commit to maintaining that bit of code, and `--build-arg` will error out if the argument is not defined in the `Dockerfile` being built).  The behavior around `--registry` should already be possible via `--namespace` (the Docker Hub API code is merely an optimization -- it should error out and be ignored fairly quickly if pushing to a non-Hub registry via `--namespace`).

7AC:
@tianon to the points you brought up:

* "This is a lot more than simply adding support for the new File:": yes sorry about that, we've been using this fork for a while now and patches have been piling up
* "we don't and won't use Gerrit so cannot comfortably commit to maintaining that bit of code": I hear that, I was hoping that having it as a fallback would be acceptable but I understand. Unfortunately that has become a critical part of getting images tested before actually changing the manifest, so we can get the build validated where we actually change the Dockerfile or so. The alternative is merging the Dockerfile without building (or building images that are based on it), betting a build failure when updating the manifest, and having to go back and change the repo where we modified the Dockerfile again. This is not ideal because between the two edits to the Dockerfile the image in the source repo is broken
* "--build-arg will error out if the argument is not defined in the Dockerfile being built": I don't think that's true, we have plenty of Dockerfiles that don't use it and build just fine.
* "The behavior around --registry should already be possible via --namespace (the Docker Hub API code is merely an optimization": we want to leverage this optimization with our registry too, and then parse the log and figure out what images were actually pushed and trigger more automation based on that

tianon:
> Unfortunately that has become a critical part of getting images tested before actually changing the manifest, so we can get the build validated where we actually change the Dockerfile or so.

For our automated `update.sh` jobs (https://doi-janky.infosiftr.net/job/update.sh/), we just update the manifest file in the CI's working copy (faking commits into `~/.cache/bashbrew/git` if necessary) in order to achieve a similar result both without making real commits of `library/xxx` and without changes to `bashbrew`.  Perhaps a similar approach could work? (using `GERRIT_REFSPEC` to `sed` a copy of the library file with the desired change)

> I don't think that's true, we have plenty of Dockerfiles that don't use it and build just fine.

Ah, you're correct -- it used to be an error, but it appears to have been changed to a warning at some point:

```console
$ docker build --build-arg TEST=test -

FROM bash

Sending build context to Docker daemon  2.048kB
Step 1/1 : FROM bash
 ---> 59507b30b48a
[Warning] One or more build-args [TEST] were not consumed
Successfully built 59507b30b48a
```

> we want to leverage this optimization with our registry too, and then parse the log and figure out what images were actually pushed and trigger more automation based on that

So, the reason we implemented this in the first place was because Docker Hub is slow (not because we really needed to know when the push would be a no-op).  I still think this would be better implemented for non-slow registries outside `bashbrew` (especially since the alternative is to pull in `docker/distribution` code to properly parse the registry value out of a given image name, and that's a bit heavier than I'd like to add here -- even implementing the registry protocol is a bit heavier than I'd like `bashbrew` to be without strong reason, not to mention maintaining functionality we then won't be using or testing).

tianon:
Docker itself doesn't support pushing to a registry unless it's part of the
image name.

tianon:
Source of our Jenkins jobs:
https://github.com/docker-library/oi-janky-groovy/blob/5a4fab6c423b6790a253bf6a0beb32ad4946b983/update.sh/target-pipeline.groovy

7AC:
Are you guys open to a separate PR for the `DOCKER_BUILD_COMMIT` change alone? We build our binaries in the first stage of a Dockerfile, and we set the version at compile time. Without that variable we can't set the version while building inside Bashbrew.

7AC:
posted here https://github.com/docker-library/official-images/pull/5545

PR #4658, opened 2018-07-31
Add mqtt.cool

gfinocchiaro:
# Checklist for Review

    - See
      * https://github/mqttcool 
     - I would say service

### DIFF
--- a/library/mqtt.cool
+++ b/library/mqtt.cool
@@ -1,18 +1,10 @@
 Maintainers: Lightstreamer Server Development Team <support@lightreamer.com> (@lightstreamer)
 GitRepo: https://github.com/MQTTCool/Docker.git
 
-Tags: 1.0.3-jre8, 1.0-jre8, 1.0.3, 1.0
-GitCommit: 3cf73a0f2207ea6a8cdea575f646dd7192b27ab6
-Directory: 1.0/jre8
-
-Tags: 1.0.3-jre8-alpine, 1.0-jre8-alpine, 1.0.3-alpine
-GitCommit: 3cf73a0f2207ea6a8cdea575f646dd7192b27ab6
-Directory: 1.0/jre8-alpine
-
 Tags: 1.1.0-jre8, 1.1-jre8, 1.1.0, 1.1, 1, latest
-GitCommit: 3cf73a0f2207ea6a8cdea575f646dd7192b27ab6
+GitCommit: 03a22f403a632015b6c1952861120adf1bda18e3
 Directory: 1.1/jre8
 
 Tags: 1.1.0-jre8-alpine, 1.1-jre8-alpine, 1.1.0-alpine, 1.0-alpine
-GitCommit: 3cf73a0f2207ea6a8cdea575f646dd7192b27ab6
+GitCommit: 03a22f403a632015b6c1952861120adf1bda18e3
 Directory: 1.1/jre8-alpine


gfinocchiaro:
@tianon  any news?

tianon:
Your patience is appreciated!  A few initial thoughts:

Looking at https://lightstreamer.com/download/#mqtt11, this appears to officially request acceptance of a EULA before downloading -- how do you intend to capture that with the Docker image?  Is this the community edition, or the "enterprise" edition?  Is MQTT.Cool open-source? (I can't seem to find source for it, and https://mqtt.cool/ is pretty vague :smile:)

> ```dockerfile
> FROM openjdk:8-jre
> ```

Have you tried using `openjdk:8-jre-slim`?  It should be quite a bit smaller, and if MQTT.Cool works with it, it should result in a much slimmer image.

> ```dockerfile
> WORKDIR /mqtt.cool
> ```

This seems to only apply to the `RUN` line immediately following it -- why not simply use `cd` instead and avoid the extra image metadata change?

> ```dockerfile
> # Replace the fictitious jdk path with the JAVA_HOME environment variable in the launch script file
>         && sed -i -- 's/\/usr\/jdk1.8.0/$JAVA_HOME/' bin/unix-like/mc.sh \
> ```

This seems like something that ought to be pushed into the upstream tarball, right?  How are users expected to consume this script without Docker as-is?  Maybe even something like `${JAVA_HOME:-/usr/jdk1.8.0}` in the upstream version would make sense if that default value somehow has meaning somewhere that `JAVA_HOME` isn't set? (it seems like having that script upstream enforce users setting `JAVA_HOME` appropriately makes more sense, IMO)

> ```dockerfile
> # Addjust logging configurations in order to log only on standard output
>         && sed -i -e '123,$s/<appender-ref ref="LSDailyRolling" \/>/<appender-ref ref="LSConsole" \/>/' \
>                   -e '/<appender-ref ref="LSDailyRolling" \/>/ d' conf/lightstreamer_log_conf.xml \
>         && sed -i -e '67,$s/<appender-ref ref="MQTTCoolRolling" \/>/<appender-ref ref="LSConsole" \/>/' mqtt_connectors/mqtt_master_connector_log_conf.xml \
> ```

The intent of these lines is definitely :+1: :100:, but the implementation seems a bit fragile -- those are exact line numbers, aren't they?  Isn't this going to be a bit painful to maintain and ensure is correct in the future?

---

Given that this appears to be a service that users run, why do they care which JRE version it's using?  To put that another way, why is the JRE version relevant enough to include in the image tags here?  Wouldn't users interact with this container via the standard exposed APIs, which should be agnostic to JRE version?

gfinocchiaro:
> Your patience is appreciated! A few initial thoughts:
> 
😄 
> Looking at https://lightstreamer.com/download/#mqtt11, this appears to officially request acceptance of a EULA before downloading -- how do you intend to capture that with the Docker image? Is this the community edition, or the "enterprise" edition? Is MQTT.Cool open-source? (I can't seem to find source for it, and https://mqtt.cool/ is pretty vague 😄)
> 
It is the Community Edition, even if the Enterprise Edition could be enabled by purchasing a license.
No, MQTT.Cool is not open-source.
https://mqtt.cool will be live in few days 😄 
> > ```dockerfile
> > FROM openjdk:8-jre
> > ```
> 
> Have you tried using `openjdk:8-jre-slim`? It should be quite a bit smaller, and if MQTT.Cool works with it, it should result in a much slimmer image.

Actually there is `open-jdk:8-jre-alpine` too. Tried with `openjdk:8-jre-slim`, but unfortunately it looks like that `gpg` is not included. Any tip?
> 
> > ```dockerfile
> > WORKDIR /mqtt.cool
> > ```
> 
> This seems to only apply to the `RUN` line immediately following it -- why not simply use `cd` instead and avoid the extra image metadata change?

Yes, right! Updated.
> 
> > ```dockerfile
> > # Replace the fictitious jdk path with the JAVA_HOME environment variable in the launch script file
> >         && sed -i -- 's/\/usr\/jdk1.8.0/$JAVA_HOME/' bin/unix-like/mc.sh \
> > ```
> 
> This seems like something that ought to be pushed into the upstream tarball, right? How are users expected to consume this script without Docker as-is? Maybe even something like `${JAVA_HOME:-/usr/jdk1.8.0}` in the upstream version would make sense if that default value somehow has meaning somewhere that `JAVA_HOME` isn't set? (it seems like having that script upstream enforce users setting `JAVA_HOME` appropriately makes more sense, IMO)

Very good observations, even if at this stage it is very hard for us getting back to this 1.1 version, considered that a brand new 1.2 package will be released in the next few days. That said, I'm going to update the upstream script as per your suggestion.
Anyway, here [our tips](https://lightstreamer.com/mqtt-docs/baseparent/MQTT.Cool%20Getting%20Started%20Guide.html#_configure) regarding JAVA_HOME.
> 
> > ```dockerfile
> > # Addjust logging configurations in order to log only on standard output
> >         && sed -i -e '123,$s/<appender-ref ref="LSDailyRolling" \/>/<appender-ref ref="LSConsole" \/>/' \
> >                   -e '/<appender-ref ref="LSDailyRolling" \/>/ d' conf/lightstreamer_log_conf.xml \
> >         && sed -i -e '67,$s/<appender-ref ref="MQTTCoolRolling" \/>/<appender-ref ref="LSConsole" \/>/' mqtt_connectors/mqtt_master_connector_log_conf.xml \
> > ```
> 
> The intent of these lines is definitely 👍 💯, but the implementation seems a bit fragile -- those are exact line numbers, aren't they? Isn't this going to be a bit painful to maintain and ensure is correct in the future?

Yes, right: just deeply simplified.
> 
> Given that this appears to be a service that users run, why do they care which JRE version it's using? To put that another way, why is the JRE version relevant enough to include in the image tags here? Wouldn't users interact with this container via the standard exposed APIs, which should be agnostic to JRE version?

JRE version is important too because of server-side development feature as illustrated [here](https://lightstreamer.com/mqtt-docs/baseparent/MQTT.Cool%20Getting%20Started%20Guide.html#_the_mqtt_cool_hook).

Let me know other thoughts.
Thanks

gfinocchiaro:
Once again a lot of problems with pull/rebase stuff... Closing and opening a new one...

tianon:
Ok, I've managed to close it.  In the future, please don't try rebasing or merging back to master anymore -- if you need that, let me know and I'll do the dance for you (because it takes a fair amount of effort to fix this when it does happen, not to mention the noise, so I'd rather just help you more hands-on doing the rebase to avoid this problem entirely instead). :heart:

gfinocchiaro:
Thank you so much and apologize for bother you with that.

PR #4663, opened 2018-08-01
library: mariadb: Disable ARM64 builds for Xenial

lag-linaro:
MariaDB do not supply the appropriate packages.

Signed-off-by: Lee Jones <lee.jones@linaro.org>

### DIFF
--- a/library/mariadb
+++ b/library/mariadb
@@ -20,7 +20,7 @@ GitCommit: 195c8432d2f434a50416dbb5ab2dec4b1978d89f
 Directory: 10.1
 
 Tags: 10.0.35-xenial, 10.0-xenial, 10.0.35, 10.0
-Architectures: amd64, arm64v8, i386, ppc64le
+Architectures: amd64, i386, ppc64le
 GitCommit: f8437ada7c617d2a6bb1c30bb30d0367b93e0ed2
 Directory: 10.0
 


lag-linaro:
Without this patch, the Docker builders fail with an "unstable build".

tianon:
The error here is additionally the following bit from the build log:

```console
+ [[ percona-xtrabackup == percona* ]]
/bin/sh: 1: [[: not found
```

The problem here is that `[[` is a Bashism, so doesn't work in a `RUN` line (which uses `/bin/sh`), so we need to adjust that.

Also, this entire file is generated -- if we make this change manually in this way, the generator will just put it right back, so we need to fix the test in the repository's scripts (which auto-detect architecture support) in order to fix this properly; see https://github.com/docker-library/mariadb/pull/195.

lag-linaro:
So a project's Dockerfile can overwrite the architectures mentioned in the Official Images repo?

tianon:
Not exactly.  The relationship here is that `library/mariadb` is generated via https://github.com/docker-library/mariadb/blob/ae7aca28f456769efa7ead255e2bcc3475285c2f/generate-stackbrew-library.sh.  That script sources the architecture list from the relevant `Dockerfile`s, which are in turn generated via https://github.com/docker-library/mariadb/blob/ae7aca28f456769efa7ead255e2bcc3475285c2f/update.sh (which uses the actual upstream repository to determine what the exact list of supported architectures should be).

That's why https://github.com/docker-library/mariadb/pull/195 was the correct fix first (which led to roughly this same change, but in a way that will last instead of being a one-off fix).

lag-linaro:
Wow, okay.  This process is certainly more complicated than I initially gave it credit for.

Thanks for the explanation.

PR #4666, opened 2018-08-02
Add ARM64v8 support.

odidev:
This PR has been raised in link with [this](https://github.com/odoo/docker/issues/195).

Odoo code is successfully building and running for ```arm64v8``` using ```tag-11```.
I can provide build and run logs too if required for confirmation.

### DIFF
--- a/library/odoo
+++ b/library/odoo
@@ -1,11 +1,14 @@
 # maintainer: Christophe Monniez <moc@odoo.com> (@d-fence)
 
+Architectures: amd64
 9.0: git://github.com/odoo/docker@67e5ceb0fa3478a3e3c3428d8062a4dab45be29f 9.0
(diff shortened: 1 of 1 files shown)


tianon:
cc @d-fence @odony @sle-odoo @aab-odoo 

Can you confirm whether this is something Odoo is willing to support?

odony:
@tianon yes we're willing to support it, makes sense as we also build RPI-based devices with Odoo embedded .

However we'd like to bump up our releases, so I've created #4706 that includes the update revision number, and credits @odidev. Thanks!

odidev:
Thanks @odony @tianon for your kind support.
@odony - Really appreciate your consideration :)

tianon:
Replaced by https://github.com/docker-library/official-images/pull/4706 :+1:

PR #4680, opened 2018-08-03
Add Multiple architecture support for geonetwork

odidev:
This PR is created in link with https://github.com/geonetwork/docker-geonetwork/issues/5

The support for multiple architecture has been added in continuation to conversation that as ```Tomcat ``` supports multiple architectures then same can be supported by ```geonetwork``` as well.

 https://github.com/docker-library/official-images/blob/83445dba81738f34d37e13fa811b34841d1d6def/library/tomcat#L38

I can provide build and run logs as required for confirmation as well.

Regards,

### DIFF
--- a/library/geonetwork
+++ b/library/geonetwork
@@ -4,25 +4,31 @@ Maintainers: Joana Simoes <joana.simoes@geocat.net> (@doublebyte1)
 GitRepo: https://github.com/geonetwork/docker-geonetwork
 
 Tags: 3.0.5, 3.0
+Architectures: amd64, arm32v5, arm32v7, i386
 GitCommit: c2af30125f631f6357d3af3837ec4b80f04b5917
 Directory: 3.0.5
 
 Tags: 3.0.5-postgres, 3.0-postgres
+Architectures: amd64, arm32v5, arm32v7, i386
 GitCommit: e5cf5da76f557481c35b4a4d3e3cac77768a1302
 Directory: 3.0.5/postgres
 
 Tags: 3.2.2, 3.2
+Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, ppc64le, s390x
 GitCommit: a9d9f3f07c0f2b8badadb74206647d7f6dea4b24
 Directory: 3.2.2
 
 Tags: 3.2.2-postgres, 3.2-postgres
+Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, ppc64le, s390x
 GitCommit: 7ba5f533d9ee31cf1a22d4d6a1a84fbb74f86466
 Directory: 3.2.2/postgres
 
 Tags: 3.4.3, 3.4, latest
+Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, ppc64le, s390x
 GitCommit: 21efb004beb5ee0675dd9ee0c807ca756b49b9a3
 Directory: 3.4.3
 
 Tags: 3.4.3-postgres, 3.4-postgres, postgres
+Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, ppc64le, s390x
 GitCommit: 21efb004beb5ee0675dd9ee0c807ca756b49b9a3
 Directory: 3.4.3/postgres


doublebyte1:
I support this PR. @tianon: what do you think?

lag-linaro:
LGTM

tianon:
(Doing a couple cross-architecture build tests on this one. :+1:)

<details>
<summary>Diff:</summary>

```diff
diff --git a/_bashbrew-arches b/_bashbrew-arches
index 75b6afa..78633a4 100644
--- a/_bashbrew-arches
+++ b/_bashbrew-arches
@@ -1,6 +1,42 @@
 geonetwork:3.0 @ amd64
+geonetwork:3.0 @ arm32v5
+geonetwork:3.0 @ arm32v7
+geonetwork:3.0 @ arm64v8
+geonetwork:3.0 @ i386
+geonetwork:3.0 @ ppc64le
+geonetwork:3.0 @ s390x
 geonetwork:3.0-postgres @ amd64
+geonetwork:3.0-postgres @ arm32v5
+geonetwork:3.0-postgres @ arm32v7
+geonetwork:3.0-postgres @ arm64v8
+geonetwork:3.0-postgres @ i386
+geonetwork:3.0-postgres @ ppc64le
+geonetwork:3.0-postgres @ s390x
 geonetwork:3.2 @ amd64
+geonetwork:3.2 @ arm32v5
+geonetwork:3.2 @ arm32v7
+geonetwork:3.2 @ arm64v8
+geonetwork:3.2 @ i386
+geonetwork:3.2 @ ppc64le
+geonetwork:3.2 @ s390x
 geonetwork:3.2-postgres @ amd64
+geonetwork:3.2-postgres @ arm32v5
+geonetwork:3.2-postgres @ arm32v7
+geonetwork:3.2-postgres @ arm64v8
+geonetwork:3.2-postgres @ i386
+geonetwork:3.2-postgres @ ppc64le
+geonetwork:3.2-postgres @ s390x
 geonetwork:latest @ amd64
+geonetwork:latest @ arm32v5
+geonetwork:latest @ arm32v7
+geonetwork:latest @ arm64v8
+geonetwork:latest @ i386
+geonetwork:latest @ ppc64le
+geonetwork:latest @ s390x
 geonetwork:postgres @ amd64
+geonetwork:postgres @ arm32v5
+geonetwork:postgres @ arm32v7
+geonetwork:postgres @ arm64v8
+geonetwork:postgres @ i386
+geonetwork:postgres @ ppc64le
+geonetwork:postgres @ s390x
````

</details>

tianon:
So, this isn't fully correct.

The `geonetwork:3.0.5` image is `FROM tomcat:8.0-jre7`, which has `Architectures: amd64, arm32v5, arm32v7, i386` (given that it's `FROM openjdk:7-jre` which is `FROM buildpack-deps:jessie-curl` which is `FROM debian:jessie`, which recently transitioned to LTS status and thus lost most of the architectures it supports).

On the other hand, 3.4.2 and 3.2.2 appear to be OK (since they're `FROM tomcat:8.0-jre8`, which descends from Stretch currently).

Is 3.0.5 still actively supported by upstream? (If not, removing it could be one option.)

If 3.0.5 is still actively supported, it will need a separate `Architectures:` list which is a subset of `tomcat:8.0-jre7`'s.

odidev:
Hi @tianon , I also agree with your comments.
Should I modify the Pull request, or can it be handled at your end?

Regards,

odidev:
Modified the **PR**, as per guidance from @tianon 

Regards,

tianon:
Doesn't https://github.com/geonetwork/docker-geonetwork/blob/master/generate-stackbrew-library.sh need to be updated to account for these changes? (or is that not going to be used anymore?)

odidev:
@tianon - I think @doublebyte1 is in a better position to answer your query.

@doublebyte1 - Can you please have a look at this.
With your support, we can accelerate the inclusion of ```arm64v8``` in official support list.

Regards,

doublebyte1:
@tianon @odidev The 3.0.5 version is no longer supported by upstream. We can safely remove this tag from the docker image. I have updated the [generate-stackbrew-library.sh
](https://github.com/geonetwork/docker-geonetwork/commit/7161069d48e316a2723d1d24ec98521b2bc886ef) and created a [PR](https://github.com/docker-library/official-images/pull/4707) to remove this tag from the geonetwork library. To separate concerns a bit, maybe the architecture should go on a separate PR? What do you think?

doublebyte1:
@tianon I am actually wondering if we should remove 3.2. as well, as there will be no more releases for this tag

tianon:
> To separate concerns a bit, maybe the architecture should go on a separate PR?

I don't think it matters too much from our perspective -- if you want them in separate commits, that's fine (removals are trivial merges), but we don't mind having them lumped together either.

> I am actually wondering if we should remove 3.2. as well

If it won't get any further releases, that sounds like EOL to me. :smile:

Removing the tags here will remove them from the "Supported" section on the Hub readme (and will prevent us from spending cycles rebuilding them on the official build server), but the tags will still be available to users who want them.  (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)

doublebyte1:
@tianon I removed the 3.2.2 tag on this [PR](https://github.com/docker-library/official-images/pull/4711)
@odidev I guess after this being merged you can add the architecture tags, or would you like me to do it?

odidev:
New PR raised to add support:
https://github.com/docker-library/official-images/pull/4724

PR #4702, opened 2018-08-07
Update sl container for sl7 to current (build-date=20180807)

sl-team:


tianon:
Can you please combine this and #4701 into a single PR so the CI can pass? (it also helps our build test, because otherwise `sl:6` fails to fetch here and `sl:7` fails to fetch over there)

jcpunk:
I'll look into what we've got going on here.  Each image comes out of a separate build system so it may take a bit of doing on our end.

PR #4711, opened 2018-08-10
Update GeoNetwork library

doublebyte1:
- removed 3.2.2 tag (deprecated)


tianon:
LGTM (removal only) :+1:

PR #4732, opened 2018-08-15
Updating tag orders

arthurdm:
Simply a cosmetic change, but we would like to update the order in which the tags show up in Docker Hub so that we show the most popular used tags first.


yosifkit:
LGTM!

PR #4758, opened 2018-08-22
Add more clarification about what "Official" means

tianon:
This is in response to questions we get from time to time around what "official" means and why our program is called "official images" (as in, what does that even mean), so this is an attempt to explain that a little better.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,18 @@
 
 See [Docker's documentation](https://docs.docker.com/docker-hub/official_repos/) for a good high-level overview of the program.
 
+## What do you mean by "Official"?
+
+In many cases, the images in this program are not only supported but also *maintained* directly by the relevant upstream projects.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,6 +6,18 @@
 
 See [Docker's documentation](https://docs.docker.com/docker-hub/official_repos/) for a good high-level overview of the program.
 
+## What do you mean by "Official"?
+
+In many cases, the images in this program are not only supported but also *maintained* directly by the relevant upstream projects.
+
+For some, they're developed in collaboration with the upstream project (or with the explicit blessing of the upstream project).
+
+In all cases, we strive to create images that are true to the upstream project's vision for how their project is intended to be consumed, occationally adding additional behavior to make them more friendly to use within Docker / containerization in general.
```
yosifkit: occationally? :wink:

tianon: What the heck; I even did `:set spell` in Vim, what's wrong with my spellchecker....

PR #4768, opened 2018-08-24
Add Multi-arch support to OrientDB

odidev:
This PR is raised after an agreed discussion in the ticket https://github.com/orientechnologies/orientdb-docker/issues/48
and the merged PR at https://github.com/orientechnologies/orientdb-docker/pull/51

Please have a look and provide your guidance.

Regards,

### DIFF
--- a/library/orientdb
+++ b/library/orientdb
@@ -1,15 +1,34 @@
-# maintainer: Luigi Dell'Aquila <l.dellaquila@orientdb.com> (@luigidellaquila)
+Maintainers: Luigi Dell'Aquila <l.dellaquila@orientdb.com> (@luigidellaquila)
+GitRepo: https://github.com/orientechnologies/orientdb-docker.git
 
-2.0.18: git://github.com/orientechnologies/orientdb-docker@8a9633c19fa9c53a5446b9b62997ff389813e785 2.0
+Tags: 3.0.6, latest
+Architectures: amd64, arm64v8, ppc64le
+GitCommit: bc7e1c48bedbd259df885dd922a13f79dfef0107
+amd64-Directory: 3.0/x86_64/alpine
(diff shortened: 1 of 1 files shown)


tianon:
cc @luigidellaquila

I'm looking over these, and I'm really not seeing a very compelling reason for these to be completely separate `Dockerfile`s -- can you elaborate on the reasoning behind that choice?

See https://github.com/docker-library/official-images#multiple-architectures for more information on our multi-architecture stance.

odidev:
Hi @tianon 

The changes are done w.r.t, to work done in ticket https://github.com/orientechnologies/orientdb-docker/issues/48

However, it would be great if you can provide us with some solution to this problem, as I am aware of the fact that you are the expert in this domain :smiley: 

Regards,

lag-linaro:
@tianon although your points seem to be valid and reasonable, apart from the fact that it was this PR which bought the issue to your attention, I'm not sure I see the connection between them and @odidev's submission.

IMHO @odidev has kept loyal to the current project workflow.  I believe it would be more helpful for you to raise a new issue against the parent project in order to discuss them with the associated maintainers instead of discussing here.

This PR LGTM.

tianon:
It's related to this PR because this is the PR which introduced multi-arch to this official image, and the method by which it has been implemented is not something we're willing to accept, regardless of how long it has been accepted in that repository (until now, it has never been part of the bits we review).

As such, I'm closing this -- we're happy to discuss and brainstorm further to help implement this in an acceptable way, but it will require upstream cooperation (@luigidellaquila) and changes to that structure in the repository.

lag-linaro:
The patch looked as though multi-arch was present already.  I saw X86 and PPC.

However, I do understand your reasoning.

Thanks for your feedback.  I'll turn my hand to it tomorrow.

PR #4800, opened 2018-08-31
Adding Couchbase Autonomous Operator for Kubernetes

ceejatec:


tianon:
My initial question here would definitely be around why this is a separate image rather than simply tags of the existing `couchbase` image?  For example, WP-CLI is added to the `wordpress` image in tags of the form `wordpress:cli-CLIVERSION`, which works pretty well.  It might also be worth considering putting this in a dedicated "couchbase" organization on Docker Hub instead (since it appears to be an add-on for the core couchbase server?).

ceejatec:
I proposed it as a separate image since it's a completely separate product from Couchbase Server. There's no common code. Technically we could push it as separate tags to the existing "couchbase" image, but that seems like it'd be very confusing both for you and for our customers.

We do have a couchbase/ organization on Docker Hub where we do various releases and pre-releases, and in fact couchbase/operator already exists as an "unofficial" version of this product (just like we have couchbase/server in addition to the Official Docker Image "couchbase"). But we'd like to have an Official Docker image of this product, for the same reasons as we wanted to have the Official Docker image for Couchbase Server.

ceejatec:
Any other questions or concerns regarding this image proposal?

ceejatec:
Is there a step we may have missed? Anything else we can do to continue moving forward here?

yosifkit:
This doesn't seem to fit in the ["generally useful"](https://docs.docker.com/docker-hub/official_repos/#how-do-i-create-a-new-official-repository) image category.  Let me clarify this position.  As far as I can tell a user will never use `docker run` on this image, which is very different from the rest of the official-images.  In the regular use case, they will probably not even know the name of the image since it is embedded in a yaml file that is provided to them by [couchbase.com/downloads](https://www.couchbase.com/downloads), so the visibility in the official images doesn't seem to be important.  There is only amd64, so automatic manifest lists are not needed.

So, I don't see what benefit the official images add, aside from our review of Dockerfile changes, with the drawback of extra work for Couchbase to engage in our process for updates.

From my perspective, the full name of this image is irrelevant to users and is just as 'official' when from the `couchbase/` organization on Docker Hub.  So you could use [automated builds](https://docs.docker.com/docker-hub/builds/) and [repository links](https://docs.docker.com/docker-hub/builds/#repository-links) to have an up-to-date image built without having our bottleneck but still have software officially from Couchbase.org.

ceejatec:
Whoops, didn't intend my last commit to be for this PR.

I'm going to close this PR for now to avoid future confusion. If this Official Image becomes desired anew in future I'll re-propose.

PR #4816, opened 2018-09-06
Fedora Update 20180906

cverna:
Signed-off-by: Clement Verna <cverna@tutanota.com>

### DIFF
--- a/library/fedora
+++ b/library/fedora
@@ -25,16 +25,15 @@ ppc64le-Directory: ppc64le/
 Tags: rawhide
 Architectures: amd64, arm32v7, arm64v8, ppc64le
 GitFetch: refs/heads/rawhide
-GitCommit: a3db183bc886c67e9e326b39c59b7ceda5c64aa4
+GitCommit: 3638c329a8590f71b2922b24494f66b6c932d8d2
 amd64-Directory: x86_64/
-arm32v7-Directory: armhfp/
 arm64v8-Directory: aarch64/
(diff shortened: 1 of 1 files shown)


cverna:
It seems that the CI build got stuck, it is possible to run it again ?

yosifkit:
Diff: https://gist.github.com/docker-library-bot/811d6a9862823cdb4e9a5f2194c42ed3

library/fedora
```diff
@@ -25,16 +25,15 @@ ppc64le-Directory: ppc64le/
 Tags: rawhide
 Architectures: amd64, arm32v7, arm64v8, ppc64le
 GitFetch: refs/heads/rawhide
-GitCommit: a3db183bc886c67e9e326b39c59b7ceda5c64aa4
+GitCommit: 3638c329a8590f71b2922b24494f66b6c932d8d2
 amd64-Directory: x86_64/
-arm32v7-Directory: armhfp/
```
tianon: Doh, not sure how we missed this -- this removes the `arm32v7` directory, but doesn't remove `arm32v7` from the `Architectures:` list (so it tries to build periodically and fails every time).  I'll be committing a fixed `Architectures:` value shortly.

tianon: https://github.com/docker-library/official-images/commit/0c44196c8968dfac4569c3d12a7b3dd093c5985e :+1:

cverna: @tianon thanks :+1:

PR #4903, opened 2018-10-01
Remove deprecated "piwik"

tianon:
As noted in https://github.com/docker-library/docs/pull/1206/files#diff-d161a79ef9b88daccb45cd0f03f1c4bd, this will receive no further updates and users are encouraged to use `matomo` instead.

FYI @pierreozoux


tianon:
> This image is officially deprecated in favor of [the `matomo` image](https://hub.docker.com/_/matomo/), and will receive no further updates after 2018-09-30 (Sep 30, 2018). Please adjust your usage accordingly.

(Quoting the exact deprecation text here for posterity.)

PR #4938, opened 2018-10-11
Update Ubuntu (20181005)

tianon:
cc @mwhudson


mwhudson:
Cosmic is going to be released next week, is it worth waiting for GA? Or maybe we can just update cosmic then?

tianon:
IMO worth waiting, good call. :+1:

tianon:
(Especially given the large splash effect updating any `ubuntu` tag has :smile:)

PR #4980, opened 2018-10-22
Add support for "File:" in manifest files

tianon:
See https://github.com/docker-library/go-dockerlibrary/pull/16 for format implementation details.

This feature is intended to be used extremely sparingly, but allows for specifying the `-f` argument to `docker build` relative to `Directory:` (via a new `File:` field), especially so multiple builds can share build context.

### DIFF
--- a/bashbrew/go/src/bashbrew/cmd-build.go
+++ b/bashbrew/go/src/bashbrew/cmd-build.go
@@ -90,7 +90,7 @@ func cmdBuild(c *cli.Context) error {
 					}
 					defer archive.Close()
 
-					err = dockerBuild(cacheTag, archive)
+					err = dockerBuild(cacheTag, entry.ArchFile(arch), archive)
 					if err != nil {
 						return cli.NewMultiError(fmt.Errorf(`failed building %q (tags %q)`, r.RepoName, entry.TagsString()), err)
 					}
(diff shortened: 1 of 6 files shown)


tianon:
Here's the example manifest file I was testing with:

```
Maintainers: foo (@bar)
GitRepo: https://github.com/docker-library/healthcheck.git
GitCommit: fc5ddb227e406301b0891f66732c00592dbbe3c6
Directory: mysql

Tags: mysql

Tags: mariadb
File: Dockerfile.mariadb

Tags: percona
File: Dockerfile.percona
```

It's worth noting that if one parses that with existing `bashbrew`, it comes back out like this:
```
Maintainers: foo (@bar)
GitRepo: https://github.com/docker-library/healthcheck.git
GitCommit: fc5ddb227e406301b0891f66732c00592dbbe3c6
Directory: mysql

Tags: mysql, mariadb, percona
```

So, anywhere this is introduced, it's going to be prudent to rename the _existing_ `Dockerfile` too such that if `bashbrew` isn't updated, it will result in an error something like `Dockerfile not found`.

For example, if we wanted to update https://github.com/docker-library/httpd to flatten that 2.4 folder so that we didn't have a duplicated `docker-entrypoint.sh` file, we wouldn't simply move `2.4/alpine/Dockerfile` to `2.4/Dockerfile.alpine`, but would _also_ move `2.4/Dockerfile` to `2.4/Dockerfile.debian`.

PR #4986, opened 2018-10-24
Add nodejs v11, update v10, remove unsupported architectures & variations

PeterDaveHello:
v11:
- https://github.com/nodejs/docker-node/pull/881
- https://github.com/nodejs/node/releases/tag/v11.0.0
- https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V11.md
- https://github.com/nodejs/docker-node/commit/119e51741fdeb1d2b25a942a20070d4c8a3ccc3a

v10:
- https://github.com/nodejs/docker-node/pull/885
- https://github.com/nodejs/node/releases/tag/v10.3.0
- https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V10.md

Jessie removal:
- https://github.com/nodejs/docker-node/pull/863

i386 removal:
- https://github.com/nodejs/docker-node/pull/823

PR #5007, opened 2018-10-30
Add "naughty-from.sh" script inspired by the report of the same name

tianon:
This verifies that the "Architectures" field of each entry in a given pull request is a subset of its parent.

See https://github.com/docker-library/oi-janky-groovy/blob/39d0ad07eb2d2bbd9f0c956c74d14f2c0421dce2/reports/naughty-from-pipeline.groovy / https://doi-janky.infosiftr.net/job/reports/job/naughty-from/ for the original/current version of this report (which this implementation is heavily inspired by).


tianon:
See https://github.com/docker-library/official-images/commit/5760aaf2bf60e61acbe19ff28b3ff8fab35b6c3f and https://travis-ci.org/docker-library/official-images/builds/448170278 for an example of failure. :+1:

tianon:
(This will help catch issues like https://github.com/docker-library/official-images/pull/4986 in a more proactive way -- it does so via a simple script which maintainers can use to verify their own compliance trivially if so desired.)

PR #3930, opened 2018-01-23
Remove images based on old or not-yet-supported images

yosifkit:
- `neurodebian:zesty`
  - `ubuntu:zesty` is EOL as of January 13, 2018
    - see also https://github.com/docker-library/official-images/pull/3901 and https://github.com/neurodebian/dockerfiles/pull/9
  - recommend removing since the base image will no longer receive updates
  - cc @yarikoptic

- `open-liberty:kernel-java8-ibmsfj` (and children)
  - cannot build on any non-amd64 arch since there is no base image
  - `ibmjava:8-sfj-alpine` currently only supports amd64 architecture
  - can be added back once ibmjava adds more arches for their alpine based image https://github.com/docker-library/official-images/pull/3418#issuecomment-359383227
  - cc @NottyCode
  - YMMV, but you might be interested in the scripts we use to keep our image/arch combinations aligned: [example](https://github.com/docker-library/redis/blob/cf9b54aee062b2fabe2d29ac39f1f2574807b0e5/generate-stackbrew-library.sh) from redis whose output becomes [`library/redis`](https://github.com/docker-library/official-images/blob/b805d4a4470032c51f54825867c81c6d89050049/library/redis)

PR #5008, opened 2018-10-30
Ditch "neurodebian:artful" now that it's EOL

tianon:
It's that time again, @yarikoptic. :sweat_smile: :heart:

### DIFF
--- a/library/neurodebian
+++ b/library/neurodebian
@@ -1,6 +1,6 @@
 Maintainers: Yaroslav Halchenko <debian@onerussian.com> (@yarikoptic)
 GitRepo: https://github.com/neurodebian/dockerfiles.git
-GitCommit: 88af27888a5aee76924c3b6840fc47f9cd8f63cb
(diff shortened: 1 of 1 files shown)


tianon:
See https://github.com/docker-library/official-images/pull/3930 for the last place we did/discussed this, especially this bit:

@yarikoptic:

> Pardon my ignorant question, especially if I asked it before -- pruning those build scripts doesn't mean that already existing images would be removed and no longer available, correct? I just to make sure that we still have access to already existing images.

@yosifkit:

> Correct, the images will still be available to pull from the Docker Hub (and be seen on https://hub.docker.com/r/library/neurodebian/tags/), but they won't be under the "Supported tags and respective Dockerfile links" section.

tianon:
(Just deletions, no build test necessary.)

yarikoptic:
Thank you @tianon . R.I.P. artful!

PR #5012, opened 2018-10-30
Add mqtt.cool

gfinocchiaro:
# Checklist for Review

    - See
      * https://github/mqttcool 
     - I would say service


gfinocchiaro:
@tianon when you will be available, let's resume here this [thread](https://github.com/docker-library/official-images/pull/4658#issuecomment-434265412).

tianon:
I know you're not doing it on purpose, but this has become a pattern, and it's got me really worried -- have you identified what it is about your workflow which keeps causing this issue for you?  It's very disruptive for us (and takes a fair amount of manual effort to clean up each time), and frankly does not encourage us to want to include more Lightstreamer-maintained repositories in our program, so before we go any further I'd like to try and identify what it is that's causing this and whether and how we can address it so that we can move forward with some proper image review here.

gfinocchiaro:
Apologize for that, but it happens only for mqtt.cool. On the contrary, other Lightstreame related repositories are good.
It happens every time I try a rebase, which simply does not work for me and don't know how to fix it.
Could you please show me the right workflow or indicate some other resource? Also, consider that repository is there for several months, and different reviews on the image content have already been conducted from your side.
If you prefer, I can start once more with a fresh new repository. Let me know, thanks.

PR #5037, opened 2018-11-05
Gocd Server docker image

ChrisRo89:
Created a repository with GoCD based on CentOS for docker, contributers and feedback are welcome


tianon:
My first question (before this gets too much further) would be whether you're part of the GoCD community and/or whether they're involved or interested in having this move forward?

ChrisRo89:
Hi, thanks for the question both is not the case. But I working for a company which is using GoCD and the original docker image looks not that nice and so I created one on my self with the needs an was thinking that might be interesting as well for others. Am 05.11.2018 18:07 schrieb Tianon Gravi <notifications@github.com>:My first question (before this gets too much further) would be whether you're part of the GoCD community and/or whether they're involved or interested in having this move forward?

PR #5042, opened 2018-11-07
Update photon image

gerrit-photon:
Update photon image


gerrit-photon:
Please skip it since it remove 3.0 docker tag in photon docker file. I will update my script and submit a new pull request later.

Thanks,
Michelle

PR #5051, opened 2018-11-08
Update photon image

gerrit-photon:
Update photon image


gerrit-photon:
Skip it since 3.0 tag is not included in the photon docker file

PR #5052, opened 2018-11-08
Update photon image

gerrit-photon:
Update photon image


gerrit-photon:
This update include glib, libgd, rpm and elfutils.

PR #5113, opened 2018-11-27
Library definition files for onboarding Amazon Corretto as an official docker image.

EricEdens:
This PR adds a new official image containing [Amazon Corretto 8](https://aws.amazon.com/corretto/) installed into [Amazon Linux 2](https://hub.docker.com/_/amazonlinux/).

The image contains the full JDK 8, at the latest security update 8u192, with common commands such as `java` and `javac` symlinked into the user's path.

### DIFF
--- a/library/amazoncorretto
+++ b/library/amazoncorretto
@@ -1,0 +1,8 @@
+Maintainers: Amazon Corretto Team (@corretto),
+             Eric Edens (@ericedens),
+             iliana weller (@ilianaw)
+GitRepo: https://github.com/corretto/corretto-8-docker.git
+
+Tags: 8, 8u192, 8-al2-full, latest
+GitFetch: refs/heads/8-al2-full
+GitCommit: ca83f84faf06f5ca115ac922926e37fb183a2126
--- a/test/config.sh
+++ b/test/config.sh
@@ -17,6 +17,7 @@ imageTests[:onbuild]+='
 '
 
 testAlias+=(
+	[amazoncorretto]='openjdk'
 	[iojs]='node'
 	[jruby]='ruby'
 	[pypy]='python'


EricEdens:
PR for docs: https://github.com/docker-library/docs/pull/1369

library/amazoncorretto
```diff
@@ -0,0 +1,5 @@
+Maintainers: Eric Edens (@ericedens)
```
ilianaw: I'd strongly recommend listing a team email alias here (that can accept external mail) and/or a couple of other team members with GitHub accounts.

library/amazoncorretto
```diff
@@ -0,0 +1,5 @@
...
+GitRepo: https://github.com/corretto/corretto-8-docker.git
+
+Tags: 8, 8u192, 8-al2-full, latest
+GitCommit: d654b54d9d01d7867a4e2141eb5515f2ff54cab7
```
ilianaw: The Dockerfile in this commit doesn't lock the yum install to a specific version, so later rebuilds on this commit will build a newer version than implied here (8u192).

library/amazoncorretto
EricEdens: Good idea! I've added the Corretto organization and you.

library/amazoncorretto
EricEdens: Good call. I added a branch for the 8-al2-full series, and then added a commit (https://github.com/corretto/corretto-8-docker/commit/e83e992e4f6227de1562c3665cfd32d4639fba9a) specifically for 8u192 within that series.

shatil:
@EricEdens why call this image `amazoncorretto` and not just `corretto`, considering that's the name of your Github project? There's nothing else in the Docker Library using that name: https://hub.docker.com/_/corretto/

And... why publish it as a top-level image, rather than as a tag for OpenJDK? Given that Corretto _is_ an OpenJDK fork, it'd be awesome if it showed up where someone would go looking for an OpenJDK Docker Image to use: https://hub.docker.com/_/openjdk/

tianon:
I'll let the maintainers here respond to that first question (since it's really up to them and we don't have a strong preference one way or another), but on your second point, that'd be because Corretto is a _fork_ of OpenJDK, not "OpenJDK" (so I think Oracle would have some opinions on whether it's applicable as tags of the `openjdk` image, just as they likely would for IBM Java, for example, or something from Azul).

This was more of a gray area when we had the old `java` image, but that's since been deprecated (and even there, "Java" is a particular project from Oracle ("Oracle and Java are registered trademarks of Oracle and/or its affiliates.").

EricEdens:
**Why call this image amazoncorretto and not just corretto?**

The distribution's official name is [Amazon Corretto](https://aws.amazon.com/corretto/), which motivated the naming of the image. We also wanted to be consistent with the pattern used by the [_/amazonlinux](https://hub.docker.com/_/amazonlinux/) official image.

Good point regarding the name of the GitHub repository. There have been talks about re-aligning that, and if that happens we'll update the documentation in [docker-library/docs](https://github.com/docker-library/docs).

tianon:
Just curious, what's the significance of `al2-full`?  I get the "Amazon Linux 2" bit (I think), but what is the `full` referring to?  That it's the "full" JDK+JRE?

Another curiosity (no review bearing) -- this is starting with a fork of OpenJDK 8; do you plan to add other versions (11+, for example) later as well?

> `yum -y update`

This is not a great idea -- this will update core packages from the base distribution, which will likely cause this new layer to be unexpectedly larger and shadow files from the base image.  If there are updates that need to be applied in the base image, those should be added to an updated `amazonlinux:2` tag which will then automatically result in this image being rebuilt with the updated packages.

> `amazon-linux-extras enable corretto8`

In my own testing, this allowed to specify a version constraint as well -- is that necessary or useful, or is the constraint on `java-1.8.0-amazon-corretto-devel-1.8.0_192.b12` sufficient? (My `yum`/`dnf` experience isn't as strong as it should be, so I just want to be sure. :heart:)

We have a few tests (https://github.com/docker-library/official-images/tree/master/test) that we currently run on the `openjdk` image -- would any of those be something you'd be interested in enabling and/or having us help hook up for this image too? (Current Java-specific tests are `java-hello-world` to verify basic functionality, `java-uimanager-font` to verify some obscure font-related bug we had in Debian, and `java-ca-certificates` to verify that Java can correctly pick up the appropriate CA certificate store and make a basic TLS request -- they all appear to pass on your existing image as-is.)

Other than that, this looks pretty good (it's really simple :smile:).  The image currently comes it at a pretty hefty ~511MB, but I think that can likely be improved over time?

EricEdens:
> Just curious, what's the significance of al2-full? I get the "Amazon Linux 2" bit (I think), but what is the full referring to? That it's the "full" JDK+JRE?

'8-al2-full' is the fully-qualified tag, which we've designed to allow for future growth. In this case it means:
  - Lock to a specific major version of Java (8) and receive updates as they are available.
  - Lock to Amazon Linux 2, and receive updates as they are available.
  - Receive the 'full' variant that includes the JDK, JRE, and all headful components (Such as X11, ALSA, and CUPS).

In addition to the 'full' variant, we foresee supporting a 'minimal' variant that would only include the JRE and would not have pull in the headful dependencies.

So, although not available today, this tag system would support something like '11-al-minimal', which would mean:
  - Lock to Java 11 and receive updates as they are available.
  - Lock to Amazon Linux, receiving updates as they are available, which may include bumping to a new LTS release if the current LTS release becomes end-of-life.
  - Receive only runtime-related components, and do not include dependencies on headful components.

> yum -y update

All very good points here. I'll remove it in the next iteration.

> is the constraint on java-1.8.0-amazon-corretto-devel-1.8.0_192.b12 sufficient? 

Similar to the usage of `yum update`, my intention was to ensure that users receive the latest updates. In doing so, however, I created a Dockerfile that resulted in non-repeatable builds. The [latest revision](https://github.com/corretto/corretto-8-docker/commit/ca83f84faf06f5ca115ac922926e37fb183a2126) includes the version constraint.


> We have a few tests 

Good idea. I onboarded to the test runner and all pass locally.

shatil:
Thanks @tianon, that distinction between pure OpenJDK and a fork makes sense to me now.

> In addition to the 'full' variant, we foresee supporting a 'minimal' variant that would only include the JRE and would not have pull in the headful dependencies.

Awesome! @EricEdens really looking forward to the `8-jre-headless` tag :)

tianon:
<details>
<summary>Diff:</summary>

```diff
diff --git a/_bashbrew-arches b/_bashbrew-arches
index e69de29..f8dddd9 100644
--- a/_bashbrew-arches
+++ b/_bashbrew-arches
@@ -0,0 +1 @@
+amazoncorretto:latest @ amd64
diff --git a/_bashbrew-list b/_bashbrew-list
index e69de29..1d3b0d3 100644
--- a/_bashbrew-list
+++ b/_bashbrew-list
@@ -0,0 +1,4 @@
+amazoncorretto:8
+amazoncorretto:8u192
+amazoncorretto:8-al2-full
+amazoncorretto:latest
diff --git a/amazoncorretto_latest/Dockerfile b/amazoncorretto_latest/Dockerfile
new file mode 100644
index 0000000..5dd602a
--- /dev/null
+++ b/amazoncorretto_latest/Dockerfile
@@ -0,0 +1,4 @@
+FROM amazonlinux:2
+RUN amazon-linux-extras enable corretto8 \
+ && yum -y install java-1.8.0-amazon-corretto-devel-1.8.0_192.b12-1.amzn2 \
+ && yum clean all
````

</details>

**Comment by tianon** [comment]
LGTM (reviewing docs PR now)

Build test of #5113; fd782696; `amd64` (`amazoncorretto`):

```console

$ bashbrew build amazoncorretto:8
Building bashbrew/cache:9f2e15a09c81b5c73e106401b634a9fa2a7ac53cc509f3724cf74493fc8913bc (amazoncorretto:8)
Tagging amazoncorretto:8
Tagging amazoncorretto:8u192
Tagging amazoncorretto:8-al2-full
Tagging amazoncorretto:latest

$ test/run.sh amazoncorretto:8
testing amazoncorretto:8
	'utc' [1/7]...passed
	'cve-2014--shellshock' [2/7]...passed
	'no-hard-coded-passwords' [3/7]...warning: garbage password detected for 'root': '*LOCK*'
passed
	'override-cmd' [4/7]...passed
	'java-hello-world' [5/7]...passed
	'java-uimanager-font' [6/7]...passed
	'java-ca-certificates' [7/7]...passed

```

PR #5124, opened 2018-11-29
euleros: Update GitCommit and add AArch64 support

lag-linaro:
Signed-off-by: Lee Jones <lee.jones@linaro.org>

### DIFF
--- a/library/euleros
+++ b/library/euleros
@@ -1,7 +1,11 @@
 Maintainers: Li Ruilin <liruilin4@huawei.com> (@Kazero0)
 GitRepo: https://github.com/euleros/euleros-docker-images.git
-GitFetch: refs/heads/master
-GitCommit: e1e5020fe9fd5030c9e5d34d5e86a74b379eff36
+GitCommit: 7f25a5a4c70cf8d29167dbaeb36ff3a7c6a621a8
+
+Tags: 2.3
+Architectures: arm64v8
+arm64v8-Directory: 2.3/aarch64/
 
 Tags: 2.2, latest
+Architectures: amd64
 Directory: 2.2


lag-linaro:
Cc: @liangchenye

liangchenye:
LGTM and thanks @lag-linaro

lag-linaro:
Thanks.  One quick question for @tianon and @yosifkit:

If the GitCommits are the same, do I still need to provide architecture specific ones?

odidev:
Hi @lag-linaro 
I checked module ```cirros```, and there ```GitCommit``` is kept common for all architectures.
It seems if architectures specific commit is not provided then ```GitCommit``` is picked as the default one.

odidev:
LGTM :)

lag-linaro:
@tianon @yosifkit any idea why this PR has been missed?

I see that others have been merged in the mean time.

tianon:
I've definitely got some strong concerns about the maintenance of this image.  The last update to this image was Nov 14, 2017 (which is also when it was initially added...).  Looking at EulerOS, I find that one of the primary selling points is "enhanced security", but it's literally never seen an image update (until now, I'll get to that in a minute), which is kind of the opposite of "enhanced security".  Is this image still an active project and something that we should continue to list as supported within the official images program?

---

On the subject of the technical details of this PR, as-is it adds a new version which supports `arm64v8`-_only_, which I don't think was the intention.

lag-linaro:
> On the subject of the technical details of this PR, as-is it adds a new version which supports `arm64v8`-_only_, which I don't think was the intention.

Yes, that is the intention of this PR.

tianon:
Does the new version of upstream EulerOS not support amd64?

lag-linaro:
Most likely.  But someone would have to create/upload an appropriate rootfs to add support for it.

tianon:
Right, so I think that's the bigger issue.  This image has literally
never been updated since being initially proposed, and if the image is
not going to be maintained, we definitely shouldn't add more
architectures and should instead look to officially deprecate it.

lag-linaro:
> Right, so I think that's the bigger issue. This image has literally never been updated since being initially proposed, and if the image is not going to be maintained, we definitely shouldn't add more architectures and should instead look to officially deprecate it.

@liangchenye FYI

odidev:
Rootfs for the ```AMD64``` architecture can be generated in the same way as it has been done for ```aarch64``` architecture.
Right now, EulerOS 2.3 officially supports both amd64 & aarch64 architectures at their official platform.

But Docker-hub still has EulerOS-2.2 due to some reasons unknown resulting in just amd64 support.
Once EulerOS will be upgraded to 2.3 then it will offiailly support multi-arch resulting in a higher number of users who can use it and keep the repository involved with issues/PR's

But yes, we need to take a step further to allow support for new architectures like aarch64 to allow experimentations and keep the traffic ongoing.

Regards,

tianon:
@liangchenye what's the game plan here?  Do you plan to continue to support and maintain the EulerOS official image?  Will it be receiving updates in a timely manner in the future?  Will this bump be updated to include `amd64` updates as well?

Should we deprecate EulerOS or do a soft-deprecation "call for maintainers" like we did with Crux?

liangchenye:
Hi @tianon , 
EulerOS is keeping update: https://developer.huawei.com/ict/site-euleros/euleros/repo/yum/.
But the docker image is not updated timely.

I asked EulerOS release team member, he told me they will assign someone to keep track of this issue.
I think we will do a better job in the future.   In the future, both `x86` and `arm64` will be released at the same time.

lag-linaro:
@liangchenye that is good news.

@tianon what's holding this back from being merged currently?

tianon:
This isn't merged because this PR as-is is definitely unacceptable -- it adds `euleros:2.3`, but only supports ARM64, which is definitely not an accurate representation of upstream (and was the catalyst for this conversation about maintenance of the `euleros` image in the first place).

What this needs to be considered acceptable is architecture parity, even if only as a good faith gesture that this isn't simply a drive-by contribution of initial ARM64 support that will then sit and rot in the same way the existing `euleros` artifacts have.

> I asked EulerOS release team member, he told me they will assign someone to keep track of this issue.

Any movement on this point?

tianon:
@liangchenye friendly ping?

> I asked EulerOS release team member, he told me they will assign someone to keep track of this issue.

Any movement on this point?

tianon:
> @liangchenye friendly ping?
> 
> > I asked EulerOS release team member, he told me they will assign someone to keep track of this issue.
> 
> Any movement on this point?

tianon:
If we don't hear back soon, we'll likely be starting the image deprecation process. 😉

liusdu:
@tianon I am the guy @liangchenye  mentioned before -_-.  Sorry for this late.  For the passed several weeks, we were working on internal issues.  Generally we release a EulerOS version( supporting both arm64 and amd64) for every three months. 

We will update the release plan and send a pr(for arm64 and arm64) in two weeks.

yosifkit:
Thanks for the update @liusdu! Let us know if you have any questions.

I'll go ahead a close this PR since a replacement with both architectures will be coming soon. :heart:

liusdu:
@tianon @yosifkit  it is the first time to push a new requses https://github.com/docker-library/official-images/pull/5611 . if any question ,let me know!

PR #5200, opened 2018-12-18
Remove Groovy & Gradle one time pushes

keeganwitt:


keeganwitt:
Do you know of a good way for me to troubleshoot/test these?  I've just been letting Travis run the build (and a few other tests) and I don't know that Travis supports things like ARM, etc.

tianon:
You can use something like
https://github.com/kata-containers/proxy/pull/84/files for ppc64le, but it
is undocumented and AFAIK not fully finalized yet.

For arm, you should be able to use one of Amazon's new instance types if
you don't already have a cheap device.  There's also at least one CI
provider that offers free arm builds but I don't recall which one ATM. 😅

keeganwitt:
It looks like a dependency of Groovy (Jansi) doesn't work with those architectures.  Maybe I should just remove those variants.

PR #5220, opened 2018-12-21
Added static-file-server to official images.

halverneus:
From the best I can tell, this static-file-server is the most popular single-purpose static file service on Docker Hub (>216,000 downloads and averaging >2,000 downloads per day). It is a small, easily configurable image for serving content over HTTP or HTTPS.

# Checklist for Review

### DIFF
--- a/library/static-file-server
+++ b/library/static-file-server
@@ -1,5 +1,5 @@
 Maintainers: Jeromy Streets <halverneus@gmail.com> (@halverneus)
-GitRepo: https://github.com/halverneus/static-file-server.git
+GitRepo: https://github.com/halverneus/static-file-server-docker.git
 GitCommit: f4b66b41af8c0592b0f431181081e8c026182c78
 
 Tags: v1.5.1-linux, linux
--- a/library/static-file-server
+++ b/library/static-file-server
@@ -1,0 +1,24 @@
+Maintainers: Jeromy Streets <halverneus@gmail.com>
+GitRepo: https://github.com/halverneus/static-file-server.git
+GitCommit: f4b66b41af8c0592b0f431181081e8c026182c78
+
+Tags: v1.5.1-linux, linux
+SharedTags: v1.5.1, latest
+Architectures: amd64, arm32v6, arm32v7, arm64v8, i386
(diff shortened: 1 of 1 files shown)


tianon:
Just to clarify, what makes this different from something like, say, NGINX or Apache's HTTPd?

> It is a small, easily configurable image for serving content over HTTP or HTTPS.

To compare, `nginx:alpine` is ~8MB on-disk (and NGINX has a large community of users / configuration examples available online and a long track record for being solid, not to mention a lot of features like websocket proxying as well), where this image appears to be ~4MB on-disk, which by itself doesn't seem like a particularly compelling difference IMO.

halverneus:
I don't disagree that it doesn't provide any feature that Nginx or Apache can already provide. The appeal for most users is the simplicity. It does one thing and does it well. This is similar to Traefik, which also provides no features not covered by Nginx or Apache. There are many applications that only need a static file server, but the barrier of entry for those without adminstration experience is a bit high.
I wouldn't have considered adding this to the official library if not for its popularity, despite a complete lack of promotion. Because of the user base, the application is 100% unit and functionally tested before every release and makes every effort to follow all best practices.
Having said that, I have no disagreement with your assessment. I just think it exemplifies how simple it should be to use a Docker service and this is a niche that I believe would encourage the use of Docker for those just starting out.
If you disagree, I will withdraw my pull request without complaint.

tianon:
I want to be totally clear that I think this project is neat and I do not in any way want to discourage you from working on it (especially if it suits a need you have). :heart:

However, in reviewing the project, I'm afraid I don't think it fits the criteria of this program -- it seems to have a small (but growing!) popularity with a set of Docker users, but doesn't have name recognition (nor a name that _can_ really get name recognition, a Google search for "static file server" gives millions of results :smile:), and doesn't have a very broad base of contributors (of either code or issues) on GitHub yet either.

Given also that we have both NGINX and Apache's HTTPd which can both serve this use case out-of-the-box (and `nginx:alpine` is roughly the same image size), I'm inclined to suggest that perhaps we can revisit this much further in your project's future when it's gained a bit more maturity and traction/vocal userbase?

halverneus:
That is completely fair and I'm in no way discouraged. Thank you for your consideration!

PR #5256, opened 2019-01-04
Update Solr

makuk66:
This update contains improvements to the docker-solr build process and a gosu version update; There is no new version of Solr, or changes to user scripts.

Changes:
- update gosu to the latest version
- update the travis build environment for the latest docker
- convert to multi-stage builds, see e.g. https://github.com/docker-solr/docker-solr/blob/master/7.6/Dockerfile . Review/feedback appreciated.


tianon:
Regarding multi-stage builds, see https://github.com/docker-library/official-images/issues/3383.

makuk66:
I'm reading that as "don't do that yet". OK, I'll revert that.

PR #5260, opened 2019-01-06
Move TOC Out Of Its Misery

Dargmuesli:
Remove the suprising effect that the README's table of contents has when encountered in the middle of reading.

(or tell me why it is there, because then I didn't get it while creating this PR)


tianon:
It was down further to put more focus on those initial questions (which we get so often that we wanted them highlighted as soon as possible, especially since our `README` is so long).  Ideally, it will probably be mostly contributors/maintainers who even need the TOC, since the bulk of the content is targeted at them.

PR #5282, opened 2019-01-11
Add mqtt.cool

gfinocchiaro:
# Checklist for Review

    - See
      * https://github/mqttcool 
     - I would say service

### DIFF
--- a/library/mqtt.cool
+++ b/library/mqtt.cool
@@ -1,0 +1,10 @@
+Maintainers: Lightstreamer Server Development Team <support@lightreamer.com> (@lightstreamer)
+GitRepo: https://github.com/MQTTCool/Docker.git
+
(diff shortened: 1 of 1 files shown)


yosifkit:
Extreme apologies for the slow reply. :bowing_man: New images unfortunately get the least priority but often take the most time to review, so they often fall by the wayside.

Since every change of every image is reviewed by humans, one of the criteria is that new images are "reasonably popular" to be able to get the most out of the available time we have.

In reviewing the project, I'm afraid I don't think it currently fits the criteria of this program. The current [Docker Hub image](https://hub.docker.com/r/mqttcool/mqtt.cool) has a small number of pulls and the repos within https://github.com/mqttcool don't yet have much community activity.

We recommend continuing keeping the `mqttcool/mqtt.cool` image up-to-date for your users and we are open to a future discussion

PR #5298, opened 2019-01-15
[nats] Release v1.4.0

kozlovic:
Details can be found [here](https://github.com/nats-io/gnatsd/releases/tag/v1.4.0)


kozlovic:
@tianon Thanks! Do you automatically look for corresponding PR in doc or should I link the doc PR when submitting to official-images so that you don't have to look for it?
In case, here is the corresponding doc for this PR: https://github.com/docker-library/docs/pull/1401

yosifkit:
We don't often look for a docs PR at the same time as an update PR, since most of the time we don't have one. We just process the queue of new PRs in each repo.  (but your link is helpful to remind us that they go together)

PR #5306, opened 2019-01-17
Update docker-library images

tianon:
- `drupal`: 8.6.7, 8.5.10, 7.63
- `ghost`: 2.11.1
- `haproxy`: 1.9.2
- `joomla`: 3.9.2
- `openjdk`: 11.0.2, 12-ea+27, 13-ea+3, debian `8u181-b13-2~deb9u1`, Windows Server 1809 (docker-library/openjdk#269)
- `pypy`: weird `update.sh` flakiness again (no changes) -- we really need to look into that :sweat_smile:

PR #5311, opened 2019-01-17
Update Joomla! to 3.9.2

J0WI:


tianon:
This was included in https://github.com/docker-library/official-images/pull/5306 -- would you like me to stop including `joomla` in those bumps? (it would certainly make my life simpler not to have to worry about `joomla` bumps there :sweat_smile:)

tianon:
(cc @mbabker)

J0WI:
Oops, I've overseen this! Thanks for keeping Joomla! up to date!

tianon:
You should be able to run it in a Docker container (as long as you can get the `bashbrew` binary and the repo source in there, which isn't too difficult), if that helps. :smile: :heart:

PR #5327, opened 2019-01-22
Add EventStore image

riccardone:
Added EventStore for official docker image.
The image is Open Source.

# Checklist for Review

	-	https://github.com/docker-library/docs/pull/1405

### DIFF
--- a/library/eventstore
+++ b/library/eventstore
@@ -1,0 +1,7 @@
+Maintainers: EventStore Team <info@eventstore.org> (@eventstore)
+
+GitRepo: https://github.com/EventStore/eventstore-docker
+
+Tags: release-4.1.1-hotfix1
+GitCommit: cd0eafec71baffbcc0b9a4bb7c58e996cf290d34
+Directory: 4.1.1-hotfix1


library/eventstore
```diff
@@ -0,0 +1,7 @@
+Maintainers: EventStore Team <info@eventstore.org> (@eventstore)
+
+GitRepo: https://github.com/EventStore/eventstore-docker
+
+Tags: release-4.1.1-hotfix1
```
TimWolla: You probably want to have `latest`, `4` and `4.1` aliases.

library/eventstore
```diff
@@ -0,0 +1,7 @@
...
+GitCommit: cd0eafec71baffbcc0b9a4bb7c58e996cf290d34
```
TimWolla: I cannot find that commit in https://github.com/EventStore/eventstore-docker.

tianon: Yeah, it appears this is a commit on the main EventStore repo, not the eventstore-docker repo: https://github.com/EventStore/eventstore/commit/cd0eafec71baffbcc0b9a4bb7c58e996cf290d34

I'd recommend giving https://github.com/docker-library/official-images#contributing-to-the-standard-library a read-through (and @TimWolla's other comments were definitely spot-on :heart:).  There appear to be quite a few things noted in the docs there that aren't adhered to here.

I'd also add that `apt-get autoremove` isn't likely to do anything (since you don't have any packages being _removed_ here), and `apt-get clean` is done automatically by the base image itself.  Also, if your install process is leaving extra files in `/tmp` or `/var/tmp`, I'd highly recommend trying to figure out why (since installing a Debian package shouldn't typically be leaving trash in those places).

yosifkit:
Gif to brighten the day:


Apologies for the long delay. :bowing_man:

Is this PR still a desirable addition? I see that there was a PR to adjust the Dockerfile, but it seems to have stalled (https://github.com/EventStore/eventstore-docker/pull/61). How can we help?

thefringeninja:
We have moved our `Dockerfile` to the main repository: https://github.com/EventStore/EventStore/blob/master/Dockerfile

I think we have addressed most concerns (except `HEALTHCHECK`) by completely changing the way it is built, but I don't have permission to push to our fork. Once I get them I will update this PR.

yosifkit:
Again, apologies for the delays. :bowing_man: New images unfortunately get the least priority but often take the most time to review, so they often fall by the wayside.

Unfortunately moving the Dockerfile (while useful for local development) doesn't work well as the target for official-images. Since the Dockerfile uses `COPY ./src .`, it essentially means the entire EventStore source code is now part of the Docker build context, and thus becomes part of the reviewed artifacts. (see `diff-pr.sh` for how we generate our diff; it is basically taking all docker build contexts from the starting `library/EventStore` file and diffing them to the new contexts from the PR).

Regarding multi-stage builds for official-images, see our [FAQ entry on multi-stage builds](https://github.com/docker-library/faq/#multi-stage-builds) and the beginning of this [comment (docker-library/ghost#209)](https://github.com/docker-library/ghost/issues/209#issuecomment-582188371) for my current thoughts.

-------------------

I can continue to review the Dockerfile in `EventStore/EventStore`, but it doesn't seem to have the same goal as a Dockerfile for an official-image (being geared toward developer usage). Since you have releases available as `.deb`, `.tar.gz` and `Mono`, it would make sense to just use those (https://github.com/docker-library/faq/#why-do-so-many-official-images-build-from-source).

As for review on the `EventStore/eventstore-docker` Dockerfile, the initial thing I'd point out (besides the reviews above) would be that `curl | bash` is not acceptable and should have some sort of verification of the download before being used (https://github.com/docker-library/official-images/tree/b7ec40a45da14cd6a2c2a251f5b56e59dbc6cc25#security), although in this case the 2-3 things it does should probably just be inlined instead since it appears to be just for setting up `/etc/apt/sources.list`?

tianon:
Closing this for now -- when the repository is updated with the requested changes and you're ready to recommit to maintaining this image, please reopen! :heart:

PR #5351, opened 2019-01-28
Bump SonarQube version to 7.6

julienlancelot:
Here are some new features bundled in SonarQube 7.6 : 
- Improve the drill down in the directory structure
- Simplify of the definition of Quality Gates
- Rework non-admin project provisioning
- A lot of new rules
- Some bug fixes

PR #5354, opened 2019-01-29
Improve bashbrew SharedTags handling (explicitly allowed for "bashbrew cat", etc)

tianon:
This fixes things like https://github.com/docker-library/official-images/pull/5351#issuecomment-458193467, and `./generate-stackbrew-library.sh` against `SharedTags` (without further changes, for now, since `.TagEntry` will default to the first entry, which is almost always Linux which is what we expect).

### DIFF
--- a/bashbrew/go/src/bashbrew/cmd-build.go
+++ b/bashbrew/go/src/bashbrew/cmd-build.go
@@ -44,7 +44,7 @@ func cmdBuild(c *cli.Context) error {
 				continue
 			}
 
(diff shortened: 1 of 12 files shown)


tianon:
Nice, https://doi-janky.infosiftr.net/job/reports/job/naughty-from/ is now just `jetty`, `rocket.chat` (https://github.com/RocketChat/Docker.Official.Image/pull/57), and `sentry` (https://github.com/getsentry/docker-sentry/issues/116). :muscle:

PR #5378, opened 2019-01-31
add enterprise releases from 1.2.0-ent to 1.4.2-ent

alvin-huang:
We are releasing Consul enterprise binaries and would like to package them as Docker containers for enterprise users. For the majority of users, they should be using the OSS release and therefore after these containers are built, we will be removing them from this list and remove them from the `Supported` list for clarity.

### DIFF
--- a/library/consul
+++ b/library/consul
@@ -5,3 +5,63 @@ Architectures: amd64, arm32v6, arm64v8, i386
 GitRepo: https://github.com/hashicorp/docker-consul.git
 GitCommit: 9bd2aa7ecf2414b8712e055f2374699148e8941c
 Directory: 0.X
+
+Tags: 1.2.0-ent
+Architectures: amd64, arm32v6, arm64v8, i386
+GitRepo: https://github.com/hashicorp/docker-consul.git
+GitCommit: ecefe0ba6acff8554147e65f312a29417e3b1573
+Directory: 0.X
+
+Tags: 1.2.1-ent
+Architectures: amd64, arm32v6, arm64v8, i386
+GitRepo: https://github.com/hashicorp/docker-consul.git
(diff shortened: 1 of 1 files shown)


tianon:
I'm a little bit wary of/confused by this -- if these aren't going to be maintained here, what's the purpose of pushing them through this repository?  Wouldn't it make more sense to maintain these via a new `consul/ent` repo (or `hashicorp/consul-ent` or something similar)?  It would be totally reasonable to link to another image like that from the documentation on this image to point out to users where to go if they want the version they can buy Enterprise support / licensing for.

alvin-huang:
@tianon, I think I confused you when I mentioned the versions wouldn't be supported. We view the enterprise and OSS variants as just different versions of Consul and you should be able to switch between images easily. We are backfilling older enterprise versions now because we made the previously private enterprise binaries, public, so we want to capture those releases in containers. As for the supported releases, we will support both 1.4.2+ent and 1.4.2.

alvin-huang:
@tianon Feel free to reach out if you need more details from me 😄

tianon:
Sorry for the delay here.  This is the bit that confuses me:

> For the majority of users, they should be using the OSS release and therefore after these containers are built, we will be removing them from this list and remove them from the `Supported` list for clarity.

To me, that reads as though you're going to remove all of them, but regardless if the majority of users aren't intended to use them, won't it be potentially confusing to users that they exist here in the same namespace?  Is "Consul Ent" something that searches well enough that users will be able to determine what these are and whether they should use them?

What is the user experience like for users who accidentally (or intentionally) use one of these tags without realizing there's something special or different about them that makes them potentially undesirable to use?

yosifkit:
Closing old issue. Feel free to comment if you feel this was closed in error.

PR #5433, opened 2019-02-12
Add OpenJDK + Eclipse OpenJ9 images from AdoptOpenJDK.

dinogun:
        - Yes, this is from the [AdoptOpenJDK](https://adoptopenjdk.net/) community, and I'm the maintainer for the current AdoptOpenJDK docker images.
        - Language Runtime
        - The current [Eclipse OpenJ9 docker images](https://hub.docker.com/u/adoptopenjdk) have over 170k pulls across releases.
         - docs [PR](https://github.com/docker-library/docs/pull/1426).
         - The dockerfiles and scripts have been derived from the [ibmjava](https://hub.docker.com/_/ibmjava/) official images.

### DIFF
--- a/library/adoptopenjdk-openj9
+++ b/library/adoptopenjdk-openj9
@@ -1,0 +1,52 @@
+# AdoptOpenJDK official images for OpenJDK with Eclipse OpenJ9
+
+Maintainers: Dinakar Guniguntala <dinakar.g@in.ibm.com> (@dinogun)
+GitRepo: https://github.com/AdoptOpenJDK/openjdk-docker.git
+
+Tags: latest, 8, 8-jdk, jdk, 8u202-b08_openj9-0.12.1-jdk
+Architectures: amd64, ppc64le, s390x
+GitCommit: 66c5d418831dae532a965b184309c8b3f58d65e5
+Directory: 8/jdk/ubuntu
+File: Dockerfile.openj9.releases.full
+
+Tags: 8-jdk-alpine, jdk-alpine, 8u202-b08_openj9-0.12.1-jdk-alpine
+Architectures: amd64
+GitCommit: 66c5d418831dae532a965b184309c8b3f58d65e5
+Directory: 8/jdk/alpine
+File: Dockerfile.openj9.releases.full
+
+Tags: 8-slim, slim, 8-jdk-slim, 8u202-b08_openj9-0.12.1-jdk-slim
+Architectures: amd64, ppc64le, s390x
+GitCommit: 66c5d418831dae532a965b184309c8b3f58d65e5
+Directory: 8/jdk/ubuntu
+File: Dockerfile.openj9.releases.slim
+
+Tags: 8-slim-alpine, slim-alpine, 8-jdk-slim-alpine, 8u202-b08_openj9-0.12.1-jdk-slim-alpine
+Architectures: amd64
+GitCommit: 66c5d418831dae532a965b184309c8b3f58d65e5
+Directory: 8/jdk/alpine
+File: Dockerfile.openj9.releases.slim
+
+Tags: 11, 11-jdk, 11.0.2.9_openj9-0.12.1-jdk
+Architectures: amd64, ppc64le, s390x
+GitCommit: 66c5d418831dae532a965b184309c8b3f58d65e5
+Directory: 11/jdk/ubuntu
+File: Dockerfile.openj9.releases.full
+
+Tags: 11-jdk-alpine, 11.0.2.9_openj9-0.12.1-jdk-alpine
+Architectures: amd64
+GitCommit: 66c5d418831dae532a965b184309c8b3f58d65e5
+Directory: 11/jdk/alpine
+File: Dockerfile.openj9.releases.full
+
+Tags: 11-slim, 11-jdk-slim, 11.0.2.9_openj9-0.12.1-jdk-slim
+Architectures: amd64, ppc64le, s390x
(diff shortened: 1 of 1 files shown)


tianon:
IMO this and https://github.com/docker-library/official-images/pull/5434 would make a lot more sense under a single repository -- what's the rationale for keeping them separated so explicitly under entirely separate image names?

dinogun:
@tianon Thank you for your immediate response ! Even though the binaries are being produced by a single entity, the VMs themselves (OpenJ9 and Hotspot) have distinct source repos and communities behind them. Having them as separate images here will help recognize that. Do you have a strong preference towards merging them ?

tianon:
That makes some sense, although still seems a little weird that they're
both under the same umbrella otherwise.

To put this another way, having two this way is eyebrow raising.  If there
were to be a third, it would be more than just raising eyebrows, which is
why I'm trying to head that off.

Looking at the download page, how are users expected to choose between
these implementations?  What differentiates them?  Why are they both called
"AdoptOpenJDK" ?

dinogun:
Thanks again for the Questions. I am talking to the Adopt community to see if they are ok with a single image with a different set of tags for each. (That brings up, which one will be the default, nomenclature of the tags, and how easily can they can differentiated). Will get back to you shortly on that.

> Looking at the download page, how are users expected to choose between these implementations? 
> What differentiates them? Why are they both called "AdoptOpenJDK" ?
In short, HotSpot images should ideally be the same as the current official "openjdk" images, except that we do not cover all of the arches as yet. OpenJ9 comes from the IBM "J9" JVM (the current ibmjava official images).

The main point here being, AdoptOpenJDK is now one single entity that helps to build, test, publish and support these images. As the AdopOpenJDK community grows, we have in our plans to talk to the distro vendors to start using these binaries as well. As can be seen by the downloads from the Adopt site (nearly 10M binary downloads and 972k docker image downloads), the community produced artifacts are already in wide use.

tianon:
> In short, HotSpot images should ideally be the same as the current official "openjdk" images, except that we do not cover all of the arches as yet. OpenJ9 comes from the IBM "J9" JVM (the current ibmjava official images).

So, this is interesting.  What makes this Hotspot image _different_ from `openjdk:11-oracle`, for example?  Likewise, what makes the OpenJ9 image _different_ from the existing `ibmjava` images?

To put that another way, why the shared "AdoptOpenJDK" name?  What do they actually have in common? (To me, they sound like entirely separate implementations of OpenJDK, and I'm still having trouble seeing what being under the AdoptOpenJDK umbrella adds for them over just consuming the official OpenJDK releases and/or IBM Java releases?)

> As the AdopOpenJDK community grows, we have in our plans to talk to the distro vendors to start using these binaries as well.

A bit unrelated to this PR, but it's worth noting that I'm not aware of _any_ of the major distributions that would be OK with consuming external binaries directly (even Firefox doesn't get that privilege in many places that I've seen, and they get the most slack out of anyone with the distributions AFAIK).

dinogun:
Let me try to answer the "why AdoptOpenJDK" question. (Thank you for the patience as we try to figure the best way forward here !)

The OpenJDK binaries (Both HotSpot and OpenJ9) are being built by using the AdoptOpenJDK infrastructure. The main USP for AdoptOpenJDK has been the completely Open Source and community way to build, test, distribute and support the OpenJDK binaries being built by it. There have been a number of users who have been building OpenJDK themselves for whatever reasons and now have taken to AdoptOpenJDK to get consistent binaries across a wide variety of platforms that has gone through a lot of testing. AdoptOpenJDK has several big names as sponsors for this community including distros (Azul, IBM, Jclarity, LJC etc). Another important aspect is support and AdoptOpenJDK has a defined roadmap to support OpenJDK builds, more details can be found [here](https://adoptopenjdk.net/support.html).

As regards to ibmjava, IBM is throwing its weight behind AdoptOpenJDK both for current and future [releases](https://dzone.com/articles/recent-java-updates-from-ibm).

I'll ask Martijn Verburg, owner of AdoptOpenJDK to chime in on this as well and specifically related to distros.

karianna:
Hi @tianon - thanks for looking into our request!  AdoptOpenJDK has now become the collaboration point for all of the vendors involved in OpenJDK (Amazon, Azul, GoDaddy, IBM, jClarity, Microsoft, Pivotal, Red Hat et al are all sponsors / major backers and more are rolling in).  With Red Hat leading Java 8 and Java 11 LTS code streams that Adopt will be the defacto OpenJDK provider for those releases and any other LTS release after the first 6 months (Oracle hands over leadership at that point).

You're correct in that distros will build from source.  We've started the early conversations for them to do so from Adopt (as that's where many of the other vendors are going to build and test from).

Hope that helps!  I appreciate there's a lot of Java ecosystem nuances at play here so happy to explain over a call and write down the content of that discussion here.

dinogun:
@tianon Do you think having a call as suggested by @karianna makes sense ? Do you want me to replace the current two distinct PRs for OpenJ9/Hotspot into just one ? Thank you helping to work through this !

arthurdm:
Adding my +1 to this PR.   We would really like to have a tag for WebSphere Liberty that used the AdoptOpenJDK + OpenJ9 image as its base.

mstoodle:
Wondering how to tease this one forward as the lack of official docker images for OpenJDK with OpenJ9 is starting to cause some headaches.

@tianon have your concerns been addressed at this point? Or would they be addressed if the naming were adjusted so that the JVM variant identifier into the image tags? (even though that wouldn't be my preferred solution)

dinogun:
Closing both #5433 and #5434 in favour of #5710.

PR #5434, opened 2019-02-12
Add OpenJDK +  HotSpot images from AdoptOpenJDK.

dinogun:
        - Yes, this is from the [AdoptOpenJDK](https://adoptopenjdk.net/) community, and I'm the maintainer for the current AdoptOpenJDK docker images.
        - Language Runtime
        - The current [OpenJDK docker images from AdoptOpenJDK](https://hub.docker.com/u/adoptopenjdk) have over 260k pulls across releases.
         - docs [PR](https://github.com/docker-library/docs/pull/1427).
         - The dockerfiles and scripts have been derived from the [ibmjava](https://hub.docker.com/_/ibmjava/) official images.


dinogun:
Closing both #5433 and #5434 in favour of #5710.

PR #5435, opened 2019-02-12
Updated README.md to add comment for the ‘ARMv5 32-bit (arm32v5)’ and Raspberry 2x

gforghetti:
Added `hint` for the ARMv5 32-bit that it can be used for the Raspberry 2x

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ Some images have been ported for other architectures, and many of these are offi
 	-	Linux x86-64 (`amd64`): https://hub.docker.com/u/amd64/
 	-	Windows x86-64 (`windows-amd64`): https://hub.docker.com/u/winamd64/
 -	Other architectures built by official images: (but *not* officially supported by Docker, Inc.)
-	-	ARMv5 32-bit (`arm32v5`): https://hub.docker.com/u/arm32v5/
+	-	ARMv5 32-bit (`arm32v5`): https://hub.docker.com/u/arm32v5/ (Raspberry 2x) 
 	-	ARMv6 32-bit (`arm32v6`): https://hub.docker.com/u/arm32v6/ (Raspberry Pi 1, Raspberry Pi Zero)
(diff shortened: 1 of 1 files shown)


tianon:
I'm not aware of any Raspberry Pi devices with an ARMv5 chip -- as far as I know they're ALL v6 and up? (The Pi2 is a v7 and the Pi3 is a v8.)

gforghetti:
A Docker user posted this to the Docker Community 

https://forums.docker.com/t/docker-in-raspberry-pi/68511/9

```For my raspberrypi 2+: ‘ARMv6-compatible processor rev 7 (v6l)’
the following image works
‘arm32v5/hello-world’ :star_struck:

Maybe a hint under https://github.com/docker-library/official-images#architectures-other-than-amd64 1
for ‘ARMv5 32-bit (arm32v5)’ and Raspberry 2x would be helpful.```

tianon:
Sure, ARMv5 images will work on any of the Raspberry Pi variants, but they definitely won't be fast or efficient (since ARMv5 requires emulating floating point instructions instead of using the dedicated hardware support).

I really would rather not have our documentation turn into a generic support document for the quirks and support modes of ARM -- is there some other official, semi-official, or even just accurate document we could point to for details like this? :disappointed:

(Maybe Raspbian has a document that describes some of this reasonably?)

gforghetti:
@tianon I'm not the "owner" here, so I'll defer the decision to you or someone else at Docker who "owns" this.

tianon:
https://en.wikipedia.org/wiki/Raspberry_Pi#Specifications seems like a reasonable table for pointing Pi users at for seeing the differences (especially the "Instruction set" row), but isn't great for anything else (and really isn't even applicable to many users -- using something like `uname -m` is going to tell them more accurately what their current _kernel_ is willing/able to actually run, because having a v7 chip doesn't mean anything if you're running a v6 kernel).

gforghetti:
@tianon Good point.  I don't want to do.

PR #5445, opened 2019-02-14
Add logic to tests for ignoring images that have zero tests

yosifkit:
Even if the image does not exist, like windows images.

Yay, no more test failures for windows images.

```console
$ ./test/run.sh golang:1.12rc1-windowsservercore-1803
testing golang:1.12rc1-windowsservercore-1803
	image has no tests...skipping
$ echo $?
0
```


tianon:
I still have lofty dreams of getting some of these tests working on Windows (like the Go tests, for example), but this seems sane for now (and realistically, for the foreseeable future).

PR #5460, opened 2019-02-21
[ros] remove images for EOL ROS Ardent

mikaelarguedas:
@ruffsl as discussed in https://github.com/osrf/docker_images/pull/227#pullrequestreview-205630609


tianon:
Deletions only, merging 👍

Issue #5371, opened 2019-01-30
Please remove openSUSE from the official library


sysrich:
As much as it is nice to have the attention produced by having openSUSE in the official library, the reality is that none of the images on https://hub.docker.com/_/opensuse/ are valid offerings of the openSUSE Project

The official library curation process has proven to be far too slow for the actual rate of change of openSUSE Tumbleweed and Leap, and issues like https://github.com/docker-library/official-images/issues/527 which hoped to rectify the situation have gone unanswered and unresolved for years

The images at https://hub.docker.com/r/opensuse/ are the only valid source of truth for valid openSUSE images, but this is not well reflected on hub.docker.com, with searches prioritising the invalid images as 'official'

The status quo is causing serious practical problems for users, with anyone using `opensuse:tumbleweed` getting an image that is phenomenally out of date and potentially vulnerable compared to `opensuse/tumbleweed`

As Chairman of the openSUSE Board I am also concerned about the Trademark and Copyright implications of Docker Hub redistributing out-of-date and unsupported openSUSE images as "official".

Therefore please can the openSUSE images be removed from the official library?

tianon:
It's a little bit harsh to blame us open source maintainers for the outdated images there (so I'd appreciate if we toned down the heavy-handed wording here; we're all doing our best :heart:) -- our process turned out to not be a good fit for the openSUSE image maintainers (which as far as we understand, are representing openSUSE appropriately while maintaining that image), which is what led to https://github.com/docker-library/docs/pull/1224 (which is where the notice at the top of https://hub.docker.com/_/opensuse/ came from).  From the looks of the deprecation notice that @Vogtinator wrote in that documentation update, the `opensuse:42.3` image should've still been receiving updates until June 30, 2019, right?

If the deprecation notice is considered insufficient and the image maintainers agree, we would be happy to remove the tags on that repository, which I believe is probably the best compromise for the user experience we could offer -- if we delete the repository completely, users who are currently using `opensuse:xxx` won't have anywhere to look for why it's gone or where it went, where if we just delete the tags, the user experience of trying to pull the images is the same (`Error response from daemon: manifest for opensuse:xxx not found`), but the Hub page would remain with the pointer to https://hub.docker.com/u/opensuse/ so users can figure out what they need to do to adjust.

cc @flavio @jordimassaguerpla @davidcassany @Vogtinator

[...]

PR #5477, opened 2019-02-27
Remove "opensuse:tumbleweed" from supported tags

tianon:
Refs https://github.com/docker-library/official-images/issues/5371

cc @Vogtinator

This doesn't remove the tag, but does stop us from "rebuilding" / re-pushing the artifacts *and* removes it from the "Supported tags and respective `Dockerfile` links" list (which is my main motivation here, so that list matches what the deprecation notice above it claims). This would also allow us to remove the tag now, if you're satisfied with where the image description is.

Once `leap` is also no longer supported (~June 30?), we can look to remove the `library/opensuse` file altogether and delete the actual tag contents from Docker Hub.

Make sense?

### DIFF
--- a/library/opensuse
+++ b/library/opensuse
@@ -17,14 +17,3 @@ GitCommit: 0611a04c49ad2d19cf4765455549e028f8a8015d
 amd64-Directory: x86_64/
 arm64v8-Directory: aarch64/
 ppc64le-Directory: ppc64le/
-
-# This is the multiarch section specific for openSUSE-Tumbleweed
-Tags: tumbleweed
-Architectures: amd64, arm64v8, ppc64le, s390x
-GitFetch: refs/heads/openSUSE-Tumbleweed
-GitCommit: 2b8e47140b2a964ac827d17d4b5edd52ee22997d
-amd64-Directory: x86_64/
-arm64v8-Directory: aarch64/
-ppc64le-Directory: ppc64le/
-s390x-Directory: s390x/
-


Vogtinator:
> This would also allow us to remove the tag now, if you're satisfied with where the image description is.

Sounds like a plan - the image is meanwhile outdated enough that removing it entirely is a good idea.

> Once `leap` is also no longer supported (~June 30?), we can look to remove the `library/opensuse` file altogether and delete the actual tag contents from Docker Hub.
>
> Make sense?

Yup!

tianon:
Ok, before I go and do something destructive, to be totally clear/explicit, I'm good to delete *all* of the following tags? (or just `tumbleweed`?)

- `opensuse:tumbleweed`
- `opensuse:42.2`
- `opensuse:harlequin`
- `opensuse:13.2`
- `opensuse:42.1`
- `opensuse:13.1`
- `opensuse:bottle`

Meaning the following should be the full set of tags which are *not* deleted yet?


- `opensuse:leap`
- `opensuse:latest`
- `opensuse:42.3`

Vogtinator:
The lists look correct to me - all of those tags are in `opensuse/archive` so nothing will get irrevocably lost.

tianon:
Awesome, thanks for confirming. :+1: :heart:

Tags are now deleted: https://hub.docker.com/_/opensuse?tab=tags

Vogtinator:
Thanks!

PR #5486, opened 2019-02-28
Update Debian (especially for 9.8)

tianon:


yosifkit:
Why empty manifests, something broke? https://github.com/debuerreotype/docker-debian-artifacts/blob/af6dd39556d43da10b3d197471adca7db2263ddc/jessie/slim/rootfs.manifest

tianon:
Neato, must've broken something in https://github.com/debuerreotype/debuerreotype/pull/55; I'll dig in.

tianon:
Ok, found a fix: https://github.com/debuerreotype/debuerreotype/pull/57

I'll close this for now since it's going to be a day or so before I can get all the bits lined up properly to get these regenerated across all arches. :+1: :sweat_smile:

PR #5491, opened 2019-02-28
test/override-cmd Fix first echo invocation

yannoff:
The line `[ "$testOutput" = "Hello $hello" ]` can't be logically evaluated to true if the docker run command output ends with a new line (`\n`)


tianon:
See https://www.gnu.org/software/bash/manual/bashref.html#Command-Substitution, especially:

> ... with any trailing newlines deleted.

tianon:
To illustrate slightly more clearly how this works:

```console
$ echo hi | hexdump -C
00000000  68 69 0a                                          |hi.|
00000003
$ test="$(echo hi)"
$ echo -n "$test" | hexdump -C
00000000  68 69                                             |hi|
00000002
```

PR #5527, opened 2019-03-08
Update photon image

gerrit-photon:
Update photon image

PR #5528, opened 2019-03-09
Update photon image

gerrit-photon:
Update photon image


tianon:
This looks like an overzealous bot: https://github.com/docker-library/official-images/pull/5527

PR #5545, opened 2019-03-14
bashbrew: store commit in DOCKER_BUILD_COMMIT build argument

7AC:
This is required in multistage builds when the binary running in the docker image is built in one of the stages and gets its version at compile-time from the SHA-1 of the git commit. This is not possible in Bashbrew as a git archive is passed around and it doesn't have the `.git` directory, so we need something to fallback to.

For example:

```
FROM golang:alpine as bundle
RUN apk add git make
COPY ./ /go/src/project/
ARG DOCKER_BUILD_COMMIT
# This makefile normally grabs the version from the git commit SHA-1, but in bashbrew it can fallback to  DOCKER_BUILD_COMMIT
RUN make -C /go/src/project/cmd/command

FROM alpine
COPY --from=bundle /go/src/project/cmd/command/command-* /usr/bin/command
ENTRYPOINT ["command"]
```

### DIFF
--- a/bashbrew/go/src/bashbrew/cmd-build.go
+++ b/bashbrew/go/src/bashbrew/cmd-build.go
@@ -90,7 +90,7 @@ func cmdBuild(c *cli.Context) error {
 					}
 					defer archive.Close()
 
-					err = dockerBuild(cacheTag, entry.ArchFile(arch), archive)
+					err = dockerBuild(cacheTag, entry.ArchFile(arch), archive, commit)
 					if err != nil {
 						return cli.NewMultiError(fmt.Errorf(`failed building %q (tags %q)`, r.RepoName, entry.TagsString()), err)
 					}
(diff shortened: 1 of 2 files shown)


7AC:
@tianon any feedback on this?

tianon:
Unfortunately, the best I've got is similar to the comments I've left
previously -- this isn't something we'd plan to use, nor is it something
we'd want Official Images maintainers using or relying on, so it's
unfortunately not something we're interested in maintaining (and having
show up as a warning in all our build logs for being "unconsumed").

7AC:
That's understandable, but how do you recommend people access `git` metadata during the build (or at least the SHA-1)? I think it's reasonable to give some way to access that since it is what's given as input to the manifest.

7AC:
Would adding the `.git` dir to the archive be a better solution for you?

7AC:
Superseded by https://github.com/docker-library/bashbrew/pull/14

PR #5583, opened 2019-03-21
add arm32v7 architecture for erlang/top 20+, erlang 21.3.2

getong:
erlang/docker-erlang-otp#82 erlang/docker-erlang-otp#211
Wait for erlang/docker-erlang-otp#212

PR #5593, opened 2019-03-22
add otp-21.2 for Elixir v1.8 & v1.7

c0b:
recently the OTP-21.3 release actually has an incompatibility
in SSL dependencies, which has caused some production deployment
failing, so we need to suppose a special otp-21.2 branch;

also add arm32v7 architecture support for alpine variant, because
of an upstream alpine3.9 fixed compilation issue

for c0b/docker-elixir#110

### DIFF
--- a/library/elixir
+++ b/library/elixir
@@ -1,4 +1,4 @@
-# this file is generated via https://github.com/c0b/docker-elixir/blob/0b5ffaee45186626b09863a3a91138d9ad63b812/generate-stackbrew-library.sh
+# this file is generated via https://github.com/c0b/docker-elixir/blob/e406adb866c4ca5381b46ca01b501917465c71ed/generate-stackbrew-library.sh
 
 Maintainers: . <c0b@users.noreply.github.com> (@c0b)
 GitRepo: https://github.com/c0b/docker-elixir.git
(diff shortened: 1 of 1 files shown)


getong:
This pr should be built again after #5583 .

c0b:
I see; the CI failure unable to fetch `erlang:21.2` looks like CI's own problem at this moment

```console
$ ./bashbrew/travis.sh
error: failed fetching repo "erlang:21.2"
tag not found in manifest for "erlang": "21.2"
error: failed fetching repo "erlang:21.2"
tag not found in manifest for "erlang": "21.2"
[...]
```

tianon:
The image `erlang:21.2` is no longer supported (it's only `erlang:21.3` now), so that's the primary thing Travis is currently complaining about here.

tianon:
To expand on that a little, `21.2` is not listed in https://github.com/docker-library/official-images/blob/master/library/erlang, which means it is not marked as supported, and we no longer have a pointer to the source of that image.  What that means functionally is that it will not receive any updates or rebuilds, especially for things like OS updates (as in, security updates for related packages like OpenSSL).

tianon:
Is `21.2` going to be added back to `erlang`?  What's the resolution here?

c0b:
it depends;   in case of Erlang,  once a `21.X+1` tag released doesn't always mean the previous minor `21.X` version deprecated,   this time the last `21.2` series is `OTP-21.2.7` sometimes upstream may decide to continue support it for an extended period, then there might be `OTP-21.2.7.1` or `OTP-21.2.8` `OTP-21.2.8.2` all are possible;   it happens sometimes just not very often, though

for this one as far as I know, there is no intention to maintain later `OTP-21.2.7.X.Y.ZZZZ` series, so I am intended to just close this PR.     will be more careful to get better upstream visibility before opening PR here

https://github.com/erlang/otp/releases

```
on Mar 5      OTP-21.2.7  …
                     0ec68a9   zip   tar.gz
```

c0b:
will re-open if there will be later `OTP-21.2.X+1` series

PR #5626, opened 2019-03-28
Port bashbrew to use Go modules instead of GB

tianon:
Something in here also fixes the "missing build output on error" bug we've been seeing on Jenkins (see https://doi-janky.infosiftr.net/job/multiarch/job/arm32v7/job/rabbitmq/184/console for example).


tianon:
Sorry Dave, I still love you! (and GB!)

image

PR #5683, opened 2019-04-04
Update photon image

gerrit-photon:
Update photon image


tianon:
Something isn't right here. :confused:

michellew-vmware:
Discard it

PR #5839, opened 2019-04-30
Add tests for NextCloud

J0WI:
Follow up of #5834
cc @tilosp @tianon

Once again this is inspired by the WordPress image :innocent:

### DIFF
--- a/test/config.sh
+++ b/test/config.sh
@@ -128,6 +128,17 @@ imageTests+=(
 		mysql-initdb
 		mysql-log-bin
 	'
+	[nextcloud]='
+		nextcloud-cli-mysql
+		nextcloud-cli-postgres
+		nextcloud-cli-sqlite
+	'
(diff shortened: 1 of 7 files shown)


tilosp:
looks like this makes ci unstable. take a look at https://travis-ci.org/nextcloud/docker/builds/526845666

i am not sure why this happens, maybe it's taking to long to start up.

J0WI:
The tests are actually right. I can reproduce the issue locally using the i386 images. Before I have only tested amd64.

J0WI:
This is actually an issue in the `i386/php:7.3-apache` image: https://github.com/docker-library/php/issues/822

J0WI:
@tianon thanks for merging this. Could you take a look at the broken i386 image (docker-library/php#822)? This causes our tests to fail on i386 too.

PR #5842, opened 2019-05-01
Add apk test for Alpine

J0WI:


J0WI:
Is this actually useful for you?
@tianon @yosifkit

tianon:
Doh, sorry!

If I were maintaining the Alpine image, I would want a test like this (verifies repository urls, etc), but I'll have to defer to @ncopa here: is this something you'd like having us run on each Alpine update to do a rough smoke test?

J0WI:
@ncopa FYI, there is a similar check for Debian/APT: https://github.com/docker-library/official-images/tree/master/test/tests/debian-apt-get

J0WI:
:ping_pong: @ncopa

yosifkit:
No response; closing old PR.

(@ncopa, you are still welcome to open a similar PR if this is something you want us to test)

PR #5878, opened 2019-05-08
Test wordpress against MariaDB

J0WI:
Motivation: most distros (including CentOS, Debian, Fedora, openSUSE, RedHat, SLES and Ubuntu) have moved to MariaDB.
Upstream requirements: https://wordpress.org/about/requirements/

### DIFF
--- a/test/tests/wordpress-apache-run/run.sh
+++ b/test/tests/wordpress-apache-run/run.sh
@@ -6,7 +6,7 @@ dir="$(dirname "$(readlink -f "$BASH_SOURCE")")"
 # Use a client image with curl for testing
 clientImage='buildpack-deps:buster-curl'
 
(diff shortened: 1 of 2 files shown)


J0WI:
Is this actually useful for you?
@tianon @yosifkit

yosifkit:
Sorry for leaving this so long :bowing_man:

I'm not sure what this adds since MySQL is a valid target as well.  With issues like https://github.com/docker-library/mariadb/issues/262, MySQL currently seems the better choice (though we have yet to reproduce that failure on our official-images build and test machines).

tianon:
Agreed -- this is interesting, but I don't think it provides significant value for the purposes of testing WordPress (unless they specifically recommend using MariaDB instead of MySQL). :heart:

Issue #3835, opened 2017-12-21
"no supported platform found in manifest list" / "no matching manifest for XXX in the manifest list entries"


tianon:
**TLDR:** Not all architectures are created equal, but perhaps even more importantly, not all build servers we have access to are equal in performance, power, or ability to process builds reliably.

**Important:** Please do *not* post here with reports of individual image issues -- we're aware of the overall problem, and this issue is a discussion of solving it generally.  Off-topic comments will be deleted.

---

When we merge an update PR to https://github.com/docker-library/official-images, it triggers Jenkins build jobs over in https://doi-janky.infosiftr.net/job/multiarch/ (see https://github.com/docker-library/official-images/issues/2289 for more details on our multiarch approach).

Sometimes, we'll have non-`amd64` image build jobs finish before their `amd64` counterparts, and due to the way we push the manifest list objects to the `library` namespace on the Docker Hub, that results in `amd64`-using folks (our primary target users) getting errors of the form "no supported platform found in manifest list" or "no matching manifest for XXX in the manifest list entries" (see linked issues below for several reports from users of this variety).

Thus, manifest lists under the `library` are "eventually consistent" -- once all arches complete successfully, the manifest lists get updated to include all the relevant sub-architectures.

Our current method for combating the main facet of this problem (missing `amd64` images while other arches are successfully built and available) is to trigger `amd64` build jobs within an hour after the update PR is merged, and all other arches only within 24 hours.  This helps to some degree in ensuring that `amd64` builds first, but not always.  For example, our `arm32vN` servers are *significantly* faster than our AWS-based `amd64` server, so if those jobs happen to get queued at the same time as existing `amd64` jobs are, they'll usually finish a lot more quickly.  Additionally, given the slow IO speed of our AWS-based `amd64` build server, the queue for `amd64` build jobs piles up really quickly (which also doesn't help with keeping our build window low).

As for triggering jobs more directly, the GitHub webhooks support in Jenkins makes certain assumptions about how jobs and pipelines are structured/triggered, and thus we can't use GitHub's webhooks to effectively trigger these jobs (without doing additional custom development to sit between the two systems), and thus rely on the built-in Jenkins polling mechanism.  This has been fine (we haven't noticed any scalability issues with how often we're polling), and even if we were triggering builds more aggressively, that's only half the problem (since then our build queues would just pile up faster).

[...]

PR #5897, opened 2019-05-09
Add new "put-multiarch" script

tianon:
This is a replacement for `bashbrew put-shared` (without `--single-arch`) that takes some shortcuts to get a pretty decent speed gain which should allow us to go back to the naïve solution of "push everything from each arch `:xyz-tag` to the multiarch repo's `:xyz-tag` and still have reasonable performance.

This was tested by pushing all ~3700 tags we currently have to https://hub.docker.com/u/trollin and for a full no-op we were around 8.5 minutes, and for a full forced push of every tag manifest (with no necessary blob mounts) we were around 45 minutes.

This is the first half of fixing https://github.com/docker-library/official-images/issues/3835 -- the second half will be to implement it over in https://github.com/docker-library/oi-janky-groovy for our two `put-shared` jobs which should be really trivial.


tianon:
To be fully clear: this will allow us to move to triggering all architectures at the same frequency, and likely trigger this roughly hourly so updated images should be available from `library/xxx` tags within ~an hour of them finishing build+push to each architecture namespace. :metal:

tianon:
For anyone curious what my original Bash-based prototype looked like: https://gist.github.com/tianon/e22655f5753abfdaa2fc3bff6a9669e7 :sweat_smile:

PR #5904, opened 2019-05-13
WIP: Add Slackware Linux base image

oxr463:
- [ ] The current slackware repository [1], should be forked at the slackware organization [2], e.g, `vbatts/slackware-container` -> `slackware/slackware-docker`
- [ ] Create metadata for official-images [3], (See: https://github.com/oxr463/official-images/commit/df93eefd540705bfae467bc35524009ea10e35d9).
- [ ] Create documentation for official-images [4].

[1]: https://github.com/vbatts/slackware-container
[2]: https://github.com/slackware
[3]: https://github.com/docker-library/official-images
[4]: https://github.com/docker-library/docs

### DIFF
--- a/library/slackware
+++ b/library/slackware
@@ -1,0 +1,16 @@
+Maintainers: Vincent Batts <vbatts@slackware.com> (@vbatts)
+GitRepo: https://github.com/vbatts/slackware-container.git
+
+Tags: current
+Architectures: arm32v7, i386, amd64
+GitFetch: refs/heads/master
+GitCommit: 9e6bff30f84ecb8b573808ec799e7d49e5010fa6
+arm32v7-Directory: slackwarearm-current/
+i386-Directory: slackware-current/
+amd64-Directory: slackware64-current/
+
+Tags: 14.2, latest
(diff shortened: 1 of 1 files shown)


tianon:
My first question would be whether you've been in contact with Slackware upstream to see if they're either interested in being involved, maintaining the image, providing their blessing for your ownership/maintainership of it, or would even rather there weren't an official image for Slackware at all (which has happened before)? :pray: :sweat_smile:

(Essentially/additionally, what's your relationship to the Slackware project?)

oxr463:
@tianon, I emailed a couple of slackware developers today. In fact, the `vbatts/slackware` images are created by an official dev, (e.g., has an `@slackware.com` email address).

I am a slackware user, myself; and while I would be willing to maintain these images, I would like to see them maintained by upstream devs if possible.

oxr463:
@tianon it looks like the latter most option is indeed the response that I have been given by the upstream developers. I agree in a sense with some, but not all, of the concerns expressed via email; unfortunately, I do not think I should to continue with this PR.

PR #5937, opened 2019-05-22
Releasing Neo4j 3.4.14

jennyowen:


yosifkit:
>
> \- https://github.com/docker-library/official-images/tree/89a7d20b3d060a10776c1fd782238edd1850f474#repeatability

If you want to move to `adoptopenjdk`, then it will have to wait until https://github.com/docker-library/official-images/pull/5710 is done.

jennyowen:
ah ok, we should be able to wait for that PR then, it looks as though it'll close soon ish.
Thanks for the info!

jennyowen:
closing PR because I introduced a bug that breaks Neo4j + kubernetes.
Now that #5710 is merged, is there anything special I'll need to do?

PR #5968, opened 2019-05-28
Add tox-node

NamsooCho:
It would be a service.
It serve p2p communication between peers.
It is open sourced project.

### DIFF
--- a/library/tox-node
+++ b/library/tox-node
@@ -1,0 +1,7 @@
+Maintainers: Namsoo CHO <nscho66@gmail.com> (@NamsooCho)
+GitRepo: https://github.com/tox-rs/tox-node.git
+
+Tags: 0.0.8, latest
+GitCommit: 751a81bfe7ccccf1b94ba4a237b32f10cf840306
+Architectures: amd64, i386
+Directory: ./


yosifkit:
Apologies for the delays. :bowing_man: New images unfortunately get the least priority but often take the most time to review, so they often fall by the wayside.

Since every change of every image is reviewed by humans, one of the criteria is that new images are "reasonably popular" to be able to get the most out of the available time we have.

In reviewing the project, I'm afraid I don't think it currently fits the criteria of this program. The current [Docker Hub image](https://hub.docker.com/r/toxrust/tox-node) has a small number of pulls and the repos within https://github.com/tox-rs don't yet have much community activity.

We recommend continuing keeping the `toxrust/tox-node` image up-to-date for your users. We are open to a future discussion

-----

Future considerations:

Unfortunately having the Dockerfile in the code repo (while useful for local development) doesn't work well as the target for official-images. Since the Dockerfile uses `ADD . /tox-node`, it essentially means the entire tox-node source code is now part of the Docker build context, and thus becomes part of the reviewed artifacts. (see `diff-pr.sh` for how we generate our diff; it is basically taking all docker build contexts from the starting `library/tox-node` file and diffing them to the new contexts from the PR). Since you have releases available as `.deb`, it would make sense to just use those (https://github.com/docker-library/faq/#why-do-so-many-official-images-build-from-source).

Regarding multi-stage builds for official-images, see our [FAQ entry on multi-stage builds](https://github.com/docker-library/faq/#multi-stage-builds) and the beginning of this [comment (docker-library/ghost#209)](https://github.com/docker-library/ghost/issues/209#issuecomment-582188371) for my current thoughts.

PR #6054, opened 2019-06-07
bashbrew: account for namespaces when sorting repos

7AC:
When sorting repos sort.go looks at the `FROM` lines in the Dockerfiles. If `--namespace` is set it needs to be taken into account as the line will be in the form `FROM <namespace>/<image>[:<tag>]`. This also adds `--namespace` to `bashbrew list` as it also sorts.

### DIFF
--- a/bashbrew/go/src/bashbrew/cmd-build.go
+++ b/bashbrew/go/src/bashbrew/cmd-build.go
@@ -12,8 +12,7 @@ func cmdBuild(c *cli.Context) error {
 		return cli.NewMultiError(fmt.Errorf(`failed gathering repo list`), err)
 	}
 
-	namespace := c.String("namespace")
-	repos, err = sortRepos(repos, true, namespace)
+	repos, err = sortRepos(repos, true)
 	if err != nil {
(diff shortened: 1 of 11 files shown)


tianon:
Heh, I debated doing this exact thing back and forth a lot when I originally wrote `bashbrew list` (hence why there's a `namespace := ""` in there as a placeholder of sorts).

The amount of places we pass this around has me wondering if maybe it's time to instead move `--namespace` to be a global option on par with `--arch` so we don't have to pass it around so much (and because it needs to affect basically everything anyhow).  The only real exception is `bashbrew tag`, which needs to take the non-namespaced images and tag them into a different namespace, so it would likely just adjust to have a `--target-namespace` parameter or something similar.

I need to noodle on this a little more. :+1: :heart:

tianon:
I gave it more thought and took a stab at porting your implementation (especially now that #5929 is finally resolved). :+1:

7AC:
Thanks @tianon ! Anything else I need to do to merge?

tianon:
I'd just like Joe to take a look as well, which he'll hopefully do today 👍

PR #6067, opened 2019-06-11
Update Clear Linux OS to 29850

clrbuilder:
Update Clear Linux OS latest+base images to release 29850

@bryteise @gtkramer


clrbuilder:
Will let automation submit one later this weekend

PR #6141, opened 2019-06-21
[percona] Update psmdb-36 version

EvgeniyPatlan:


tianon:
Unfortunately, we definitely can't accept something that's `FROM registry.access.redhat.com/ubi7/ubi-minimal` (https://github.com/docker-library/official-images#repeatability).

PR #6268, opened 2019-07-11
WIP: Add OpenWrt base image

oxr463:
- [ ] Associated with or contacted upstream? Yes, (See: https://github.com/openwrt/docker/issues/6).
- [ ] Does it fit into one of the common categories? base distribution
- [ ] Is it reasonably popular, or does it solve a particular use case well? Yes.
- [ ] Does a documentation PR exist? TODO
- [ ] Dockerization review for best practices and cache gotchas/improvements (ala the official review guidelines)? WIP
- [ ] Existing official images have been considered as a base? N/A
- [ ] If FROM scratch, tarballs only exist in a single commit within the associated history? TODO
- [ ] Passes current tests? Any simple new tests that might be appropriate to add?

### DIFF
--- a/library/openwrt
+++ b/library/openwrt
@@ -1,0 +1,7 @@
+Maintainers: Paul Spooren <mail@aparcar.org> (@aparcar)
+GitRepo: https://github.com/openwrt/docker.git
+
+Tags: latest
+Architectures: amd64
+GitFetch: refs/heads/master
+GitCommit: 29e08aee3d96e860f292a83ed42f6d144762bae2


tianon:
Nice, I am definitely a fan of this! :tada:

Unfortunately, we don't have any way for our build scaffolding to run a "build" command external to Docker (hopefully understandably given the potential issues with that), so this will need to be a pre-generated rootfs checked into Git somewhere.  I'd recommend applying some compression (I'm a big fan of `.tar.xz` given the compression ratio it achieves), and I like to use `dist-xxx` branches like in https://github.com/debuerreotype/docker-debian-artifacts/branches which get force-pushed with just the rootfs tarball (so that we don't end up with tarballs in the Git history).

Hopefully this is why you labelled this `WIP:`? :sweat_smile:

oxr463:
@tianon, hello again!

Yes, we are trying to figure out how to best proceed with this. I don't like the idea of storing binaries in git, but the force-push part makes more sense now. I'm reading through the documentation, but it would be nice to have a template repository that could be forked and modified. I'm currently using Alpine as a reference.

tianon:
The problem with providing a "template" is that it really isn't a one-size-fits-all kind of thing.  Some projects have official "chroot" or "rootfs" artifacts that get published and consumed, for example, while others have some kind of build process that simply gets rerun periodically to generate new tarballs.

Here are several examples though which might be helpful:

- https://github.com/alpinelinux/docker-alpine
- https://github.com/debuerreotype/docker-debian-artifacts
- https://github.com/tianon/docker-brew-ubuntu-core
- https://github.com/docker-library/busybox

At least in the case of OpenWrt, at least the tarballs should be pretty tiny, right? :smile:

tianon:
Any progress/update?  Anything we can help with?

oxr463:
None so far. Been a bit busy with other things. Thanks for the reminder though.

tianon:
FWIW, that diff command is failing because of what I mentioned up in https://github.com/docker-library/official-images/pull/6268#issuecomment-511059747 (the tarballs aren't part of the Git context, so it doesn't know what to diff just like our build servers wouldn't know what to build :sweat_smile:).

Is there anything I can do to help here?  Any particular stumbling blocks I could help explain or maybe help brainstorm around?

oxr463:
@aparcar do you have any bandwidth for this right now?

aparcar:
I'll give it a try as this PR is already so old. To be able testing with the Docker CI I created my own PR https://github.com/docker-library/official-images/pull/7975. The OpenWrt upstream scripts are still missing for automatic updates of the GitHub repository, however this should be easily integrated.

tianon:
Alright, closing in favor of https://github.com/docker-library/official-images/pull/7975. :+1: :heart:

PR #6541, opened 2019-08-26
Bump Traefik to v2.0.0-rc1

ldez:
Hello,

This PR updates Traefik to v2.0.0-rc1.

/cc @mmatur @emilevauge @dtomcej

Cheers


ldez:
@yosifkit I updated the PR, could you take another look.

emilevauge:
@tianon @yosifkit 
>I'm honestly wondering why there is still even a scratch-based image; is the 7.2 MB difference that important?

Yes it is. It's not about the size, but the the fact that there is no other binary in the image.
Plus, changing this at the last minute because of how the multi-stage build is implemented here is inconvenient. Is that OK to merge this one?

PR #6543, opened 2019-08-27
Update Flink images

patricklucas:
Missed a spot.

cc @fhueske


fhueske:
Thanks @patricklucas!

Looks good to me

PR #6588, opened 2019-09-05
Fix typo in heavy-hitters.txt

ernetas:


tianon:
Perhaps it's not a common word, but "pare" was intended here; https://www.google.com/search?q=define%3Apare:

> reduce (something) in size, extent, quantity, or number

ernetas:
Ohh.

PR #6621, opened 2019-09-12
rebar3 3.12.0, erlang/otp 21.3.8.7, erlang/otp 22.1

getong:
erlang/docker-erlang-otp#241  erlang/docker-erlang-otp#242

### DIFF
--- a/library/erlang
+++ b/library/erlang
@@ -3,34 +3,34 @@
 Maintainers: Mr C0B <denc716@gmail.com> (@c0b)
 GitRepo: https://github.com/erlang/docker-erlang-otp.git
 
-Tags: 22.0.7, 22.0, 22, latest
+Tags: 22.1, 22, latest
 Architectures: amd64, arm32v7, arm64v8, i386, s390x, ppc64le
-GitCommit: 3765be4573e249004e48a5de9c94127a214b1e6f
+GitCommit: daa8c85d275ac3b555f7bdadc8c5df24bc6a75a0
 Directory: 22
 
-Tags: 22.0.7-slim, 22.0-slim, 22-slim, slim
+Tags: 22.1-slim, 22-slim, slim
 Architectures: amd64, arm32v7, arm64v8, i386, s390x, ppc64le
-GitCommit: 3765be4573e249004e48a5de9c94127a214b1e6f
+GitCommit: daa8c85d275ac3b555f7bdadc8c5df24bc6a75a0
 Directory: 22/slim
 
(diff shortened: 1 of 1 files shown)


getong:
How about I change the dockerfile like below:
```
-LABEL otp.version=$OTP_VERSION
-LABEL rebar3.version=$REBAR3_VERSION
+LABEL org.opencontainers.image.version=$OTP_VERSION \
+ rebar3.version=$REBAR3_VERSION
```
@yosifkit

getong:
How do you think of this:
```
LABEL org.opencontainers.image.version=$OTP_VERSION \
 rebar3.version=$REBAR3_VERSION \
 alpine.version=3.10
```
If no problem, I will merge the erlang/docker-erlang-otp#234 and update this pr.
@tianon  @yosifkit

tianon:
`org.opencontainers.image.version=$OTP_VERSION` is fine, but both of `rebar3.version=$REBAR3_VERSION` and `alpine.version=3.10` are not compliant with the OCI standard (reverse DNS syntax).

What's the intended goal of these labels?  Is there a particular product that's looking for / consuming them?

tianon:
(In my experience, human-maintained metadata like labels is a poor replacement for real image introspection, especially because it has a tendency to be inaccurate or out of date.)

tsloughter:
The labels are set to the same variables used to install the software that they are the version for. Meaning they can not be out of date, they will be in sync with whatever is installed in that image.

The added labels are meant to make it easier to know what version of tools (like rebar3) is included in the image. I don't know how this would be done with inspecting the image besides running it and calling `rebar3 version`.

The only other option I know of would be to include all versions in the tag, resulting in `erlang:22.1-3.10-3.12`.

tianon:
If the point is introspection, why not introspect the environment variables directly? (which are as accessible as labels, and are extracted in exactly the same way, just under a different key)

```console
$ docker image inspect --format '{{- range .Config.Env -}}{{- . -}}{{- "\n" -}}{{- end -}}' erlang
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
OTP_VERSION=22.0.7
REBAR_VERSION=2.6.4
REBAR3_VERSION=3.11.1
```

getong:
@tsloughter If you agree, I will update the  erlang/docker-erlang-otp#243 , and then update this pr.

tsloughter:
Yea, that is fine.

PR #6772, opened 2019-10-11
Update rabbitmq

yosifkit:
Changes:

- https://github.com/docker-library/rabbitmq/commit/ffe636b: Update to 3.8.1-beta.1, Erlang/OTP 22.1.3, OpenSSL 1.1.1d
- https://github.com/docker-library/rabbitmq/commit/1e090d5: Update to 3.7.20-beta.1, Erlang/OTP 22.1.3, OpenSSL 1.1.1d
- https://github.com/docker-library/rabbitmq/commit/b8ca2ef: Update to 3.8.0, Erlang/OTP 22.1.3, OpenSSL 1.1.1d
- https://github.com/docker-library/rabbitmq/commit/e1a86a5: Update to 3.7.19, Erlang/OTP 22.1.3, OpenSSL 1.1.1d

PR #6779, opened 2019-10-14
Merge pull request #6772 from docker-library-bot/rabbitmq

micktie:
Update rabbitmq

Signed-off-by: michthom <michael.a.thompson@oracle.com>


tianon:
This seems a bit odd; what's the intention here?

micktie:
It is related to Oracle Linux 8 container images & pointers to the new images

tianon:
Ok sure, but then why is the PR's title "Merge pull request #6772 from docker-library-bot/rabbitmq", and why are there changes to `library/rabbitmq` included here as well?  There's also only one commit in this PR, and it's a merge commit?

micktie:
Ah, seems I did a commit amend rather than commit.  Corrected, so it should be ok now.

PR #6829, opened 2019-10-21
Update golang

tianon:
Changes:

- https://github.com/docker-library/golang/commit/12bd40a: Remove backfilled versions from https://github.com/docker-library/golang/pull/309


tianon:
Removals only, merging.

PR #6830, opened 2019-10-21
Update buildpack-deps

tianon:
Changes:

- https://github.com/docker-library/buildpack-deps/commit/05d1698: Remove cosmic (EOL)


tianon:
Removals only, merging.

PR #6843, opened 2019-10-23
Add Official Newrelic PHP Daemon Image

murphpdx:
Hello, 
We, the New Relic PHP Agent team, would like to add an official image for our New Relic PHP Agent daemon. 

-	[ ] associated with or contacted upstream? Yes https://github.com/newrelic/newrelic-php-daemon-docker
-	[ ] does a [documentation](https://github.com/docker-library/docs/blob/master/README.md) PR exist? (should be reviewed and merged at roughly the same time so that we don't have an empty image page on the Hub for very long) https://github.com/docker-library/docs/pull/1591

### DIFF
--- a/library/newrelic-php-daemon
+++ b/library/newrelic-php-daemon
@@ -5,7 +5,7 @@ Maintainers: Amanda Murphy <amurphy@newrelic.com> (@murphpdx),
              Kyle Kneitinger <kkneitenger@newrelic.com> (@kneitinger)
 GitRepo: https://github.com/newrelic/newrelic-php-daemon-docker.git
 
-Tags: latest, 9.2.0.247 9.2.0, 9.2
+Tags: latest, 9.2.0.247, 9.2.0, 9.2
 Architectures: amd64
 GitCommit: 8acf95eae363e3dcc75ffd8b803e08bed75ee2b9
 Directory: 9.2.0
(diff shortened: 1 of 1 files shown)


tianon:
I'm guessing this is the image we were hypothetically discussing over in https://github.com/docker-library/faq/issues/9? :sweat_smile:

murphpdx:
> I'm guessing this is the image we were hypothetically discussing over in [docker-library/faq#9](https://github.com/docker-library/faq/issues/9)? 😅

Yes, it sure is.

tianon:
So, I think what's discussed there is definitely still relevant here -- while I'm sure this image is useful in combination with New Relic, it's not exactly "generally useful" right?  For example, if someone isn't using the New Relic SaaS product, they wouldn't get any value from this image?

murphpdx:
> So, I think what's discussed there is definitely still relevant here -- while I'm sure this image is useful in combination with New Relic, it's not exactly "generally useful" right? For example, if someone isn't using the New Relic SaaS product, they wouldn't get any value from this image?

That is correct, is your concern with the cost? We do offer a trial period and once that ends the account becomes a free lite account.

yosifkit:
> My initial reaction is that it doesn't sound like the image would be "generally useful"

My reasoning for my reaction is that this image is tied to a single specific SAAS offering. It is not a generic service that you would plug into another system.  All the other images here are a self-contained service, language stack, or base OS; they don't require a hosted service to use.

yosifkit:
Closing given my comment just above: https://github.com/docker-library/official-images/pull/6843#issuecomment-554158766.

PR #6846, opened 2019-10-23
kong: bump Kong to 1.4.0

kikito:

PR #6855, opened 2019-10-24
kong: move plain tags to 1.4 images

kikito:
Move tags alpine, latest, ubuntu, and centos tags up from the 1.3 images to the 1.4 images.

Related with #6846


yosifkit:
Just tag changes; merging.

PR #7094, opened 2019-12-09
Update mongo

tianon:
Changes:

- https://github.com/docker-library/mongo/commit/6ac52e6: Remove EOL Windows 1803-based (SAC) images
- https://github.com/docker-library/mongo/commit/a107815: Update to 4.2.2

PR #7098, opened 2019-12-09
Add EOL Windows 1803-based (SAC) images to naughty-from

tianon:
See https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/base-image-lifecycle (EOL 11/12/2019)

### DIFF
--- a/naughty-from.sh
+++ b/naughty-from.sh
@@ -23,6 +23,11 @@ _is_naughty() {
 		| *=microsoft/*:latest \
 		) return 0 ;;
 		# https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/base-image-lifecycle
+		# "11/12/2019"
+		*=mcr.microsoft.com/windows/*:1803* \
+		| *=microsoft/*:1803* \
+		) return 0 ;;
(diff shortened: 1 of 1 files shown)


tianon:
FYI @ relevant image maintainers:

- `adoptopenjdk`: @dinogun
- `haxe`: @andyli
- `hylang`: @tianon
- `mongo`: https://github.com/docker-library/official-images/pull/7094 :disappointed:
- `nats`: @derekcollison @kozlovic @wallyqs @variadico
- `nats-streaming`: @kozlovic

kozlovic:
@tianon Thanks, we will remove the `-1803` images from our repos. Looks like `-1809` will be in May of next year...
We also use `mcr.microsoft.com/windows/servercore:ltsc2016`, which I still find in the list of this [page](https://hub.docker.com/_/microsoft-windows-servercore), but no idea if this is EOL or not.. would you know?

yosifkit:
@kozlovic `ltsc2016` (aka `1607`) should be until 01/11/2022 and `1809` is Long term supported for Sever Core, so it should be until 01/09/2024. Nano Server support is shortened and so `1809` for that ends next May.

That's how I read the chart linked from their hub page (https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/base-image-lifecycle).

PR #7144, opened 2019-12-17
Update SonarQube 8.1 and 7.9.2

julienlancelot:


library/sonarqube
```diff
@@ -4,17 +4,17 @@ Maintainers: Janos Gyerik <janos.gyerik@sonarsource.com> (@janos-ss),
...
-Tags: 7.9.1-community, 7.9-community, latest, lts
+Tags: 7.9.2-community, 7.9-community, latest, lts
 Directory: 7/community
 
-Tags: 8.0-community-beta, 8-community-beta, community-beta
+Tags: 8.1-community-beta, 8-community-beta, community-beta
 Directory: 8/community
 
-Tags: 8.0-developer-beta, 8-developer-beta, developer-beta
+Tags: 8.1-community-beta, 8-developer-beta, developer-beta
```
jeff-zapotoczny-sonarsource: Shouldn't this tag be "8.1-developer-beta" and not "community"?

library/sonarqube
```diff
@@ -4,17 +4,17 @@ Maintainers: Janos Gyerik <janos.gyerik@sonarsource.com> (@janos-ss),
...
 Directory: 8/developer
 
-Tags: 8.0-enterprise-beta, 8-enterprise-beta, enterprise-beta
+Tags: 8.1-community-beta, 8-enterprise-beta, enterprise-beta
```
jeff-zapotoczny-sonarsource: Like above, I think you accidentally pasted "community" instead of "enterprise"

library/sonarqube
ast3150: I think this is fixed in a newer PR #7159

ast3150:
@jeff-zapotoczny-sonarsource I think this is fixed in a newer PR #7159

PR #7166, opened 2019-12-19
Add some tests for bundler finding binstubs

deivid-rodriguez:
Hello! :wave:

We released bundler 2.1 a few days ago, and unfortunately we broke some stuff regarding official docker ruby images :disappointed:. See https://github.com/bundler/bundler/issues/7494.

Since I want to make sure this doesn't happen again, I'd like to add some tests to make sure it's all good with our releases :)

These tests currently fail against bundler 2.1.1, but I'm planning to fix this, of course.

Currently I'm slightly leaning towards fixing this inside the official images, but I'm open to discussing more, of course. There's a discussion about this in the linked bundler issue, with a proposed fix of setting `$GEM_PATH` and modifying `$PATH`.

PR #7208, opened 2019-12-27
Update run.sh to use the latest version of bundler (2.1.2)

JulienDefrance:
This is causing test/failures in other repositories such as https://github.com/docker-library/ruby/pull/301

### DIFF
--- a/test/tests/ruby-binstubs/run.sh
+++ b/test/tests/ruby-binstubs/run.sh
@@ -4,5 +4,4 @@ set -eo pipefail
 testDir="$(readlink -f "$(dirname "$BASH_SOURCE")")"
 runDir="$(dirname "$testDir")"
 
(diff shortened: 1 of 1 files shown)


tianon:
See https://github.com/docker-library/official-images/pull/7166 -- this was an intentional change by @deivid-rodriguez from Bundler upstream to try and help avoid https://github.com/bundler/bundler/issues/7494 from cropping back up when we get either the images or Bundler fixed accordingly. :sweat_smile:

deivid-rodriguez:
I'm thinking maybe these tests should be changed to test whatever version of bundler comes with each ruby by default? I think that would be consistent with the official images not installing any specific bundler version.

yosifkit:
Closing in favor of #7229.

PR #7233, opened 2020-01-04
Add official monica image

asbiin:
We would like to add the [Monica](https://github.com/monicahq/monica) image to the official image repository.

Our current image [monicahq/monicahq](https://hub.docker.com/r/monicahq/monicahq) has reached 5 Million downloads, and I will be glad if we have some review on the best practice or how to improve this image.

---
# Checklist for Review

    - https://github.com/monicahq/docker
    - "service"
    - Previous [image](https://hub.docker.com/r/monicahq/monicahq) hit 5M+ pulls
    - See https://github.com/docker-library/docs/pull/1663
    - base image is `php`


tianon:
Sorry for the delay!

Unfortunately, the way these `Dockerfile`s are structured, they cannot be consumed by our build tooling (especially using `ARG` for `VERSION` and/or `COMMIT`, which we don't have any way to specify in our build manifest here).

There are a lot of little things I notice about them that will need to be changed/improved, but I don't want to get into minutia of `Dockerfile` review without getting that primary higher-level issue out of the way. :sweat_smile:

To move forward, I would suggest net-new `Dockerfile` sources which pin to a specific explicit release of Monica.  The best way I've found to think of this is like downstream packaging for a Linux distribution -- the packaging files for those will typically live outside your main Git repository (so that they can be updated/maintained independently of the packaged product's lifecycle), and from a technical perspective what happens here in the official images is very similar to that.  What gets released/published here will typically correspond to whatever versions of a given software are still supported by the project.

There are a **lot** of examples of this in the full corpus of images, but here are some that might be more relevant than others for you to compare/consider (these are existing PHP-based projects within the library):

- https://hub.docker.com/_/adminer: https://github.com/TimWolla/docker-adminer
- https://hub.docker.com/_/backdrop: https://github.com/backdrop-ops/backdrop-docker
- https://hub.docker.com/_/drupal: https://github.com/docker-library/drupal
- https://hub.docker.com/_/friendica: https://github.com/friendica/docker
- https://hub.docker.com/_/joomla: https://github.com/joomla/docker-joomla
- https://hub.docker.com/_/matomo: https://github.com/matomo-org/docker
- https://hub.docker.com/_/mediawiki: https://github.com/wikimedia/mediawiki-docker
- https://hub.docker.com/_/nextcloud: https://github.com/nextcloud/docker
- https://hub.docker.com/_/postfixadmin: https://github.com/postfixadmin/docker
- https://hub.docker.com/_/wordpress: https://github.com/docker-library/wordpress
- https://hub.docker.com/_/yourls: https://github.com/YOURLS/docker-yourls

asbiin:
Thank you so much @tianon 
I've created a separate repository https://github.com/monicahq/docker with a brand new Dockerfiles 
It's way better and clean this way.

I still need to add some documentation, and create a PR in docker-library/docs as well.

asbiin:
The image has been updated. If you have the time to have a look, I'll really appreciate to have a feeback. I've completely change the image build from the first commit, and I think it's now more even as an official docker image.
Thanks!

tianon:
A few initial thoughts:

---

```dockerfile
LABEL maintainer="Alexis Saettler <alexis@saettler.org> (@asbiin)" \
```

Unfortunately this isn't an OCI annotation, nor does it follow the reverse-DNS syntax defined by the OCI, so I'd recommend just removing it entirely in favor of `org.opencontainers.image.authors` (which you've already got).

---

```dockerfile
    apk add --virtual .monica-phpext-rundeps $runDeps; \
    apk del .build-deps; \
```

Not a blocker, but I'd suggest adding `--no-network` to both of these (since neither of them should need to fetch updated package lists).

---

```dockerfile
    rm -rf /var/cache/apk/*
```

Given we're appropriately using `--no-cache` in all our `apk add` invocations, this shouldn't be necessary (or even useful), right?

---

```dockerfile
ENV PHP_OPCACHE_VALIDATE_TIMESTAMPS="0" \
    PHP_OPCACHE_MAX_ACCELERATED_FILES="20000" \
    PHP_OPCACHE_MEMORY_CONSUMPTION="192" \
    PHP_OPCACHE_MAX_WASTED_PERCENTAGE="10"
```

I'm slightly concerned that setting these as environment variables this way might give users the impression they can set them on their images to adjust the settings (which won't work because these get baked into `$PHP_INI_DIR/conf.d/opcache-recommended.ini` at build-time) -- thoughts?

---

```dockerfile
RUN { \
...
    } > $PHP_INI_DIR/conf.d/opcache-recommended.ini; \
    \
    echo 'apc.enable_cli=1' >> $PHP_INI_DIR/conf.d/docker-php-ext-apcu.ini; \
...
```

This run line is using `;` as an instruction separator, but without `set -e`, so in the (admittedly unlikely) event that one of these file writes fails, the image will still build successfully (and will simply be missing bits we expected to be there).

---

```dockerfile
RUN if [ $(uname -m) = "x86_64" ]; then \
        PLATFORM=$(uname -s); \
        if [ "$PLATFORM" = "Linux" -o "$PLATFORM" = "Windows" -o "$PLATFORM" = "Darwin" ]; then \
            mkdir -p /root/.local/bin; \
            curl -sL https://sentry.io/get-cli/ | INSTALL_DIR=/root/.local/bin bash; \
        fi \
    fi
```

Using `uname -m` for platform detection is very strongly discouraged, as it tests the host kernel's architecture, when what you really wanted to test was the image's userspace architecture, so `apk --print-arch` or `dpkg --print-architecture` are going to be safer/more reliable options.

Beyond that however, doing `curl | bash` is definitely not going to be acceptable (*especially* given that script in question also doesn't perform any download verification of the `sentry-cli` binaries it fetches).  See https://github.com/docker-library/official-images#security for at least one applicable section around that (although I think there are several that apply in spirit to this, even if none mention it explicitly).

Looking at https://github.com/getsentry/sentry-cli/releases, it doesn't seem like the `sentry-cli` project publishes checksums or signatures for these binaries. :disappointed:

---

```dockerfile
VOLUME /var/www/html
```

```shell
echo "Syncing sources..."
...
```

This section seems like it's been (conceptually) copied from the Wordpress image -- I have to wonder if the underlying intent for why Wordpress does this actually applies to Monica?

For context, this pattern was adopted in Wordpress specifically because Wordpress has a tendency to treat the code of Wordpress itself as "mutable state" -- the configuration file can be updated/modified to a minor extent via the web UI (especially the initial configuration), plugins and themes get downloaded and installed directly inside the installation directory, but most relevantly Wordpress strongly recommends allow it to do self-auto-updates.

In this sense, the image is a "seed" for your Wordpress install, after which it's self-maintaining to some degree (minus system dependencies like PHP version / updates, etc).

From the way this image is designed, it appears that Monica updates will only be applied to the volume when the container image is replaced by a newer version.

Given all that, I wonder why Monica is not simply being installed directly to `/var/www/html` in the image itself and thus not needing any copying behavior in the entrypoint?

The following would then simply imply the need for `VOLUME /var/www/html/storage`, since it's the explicitly mutable bits of the installation, from what I can tell?

```shell
    # Ensure storage directories are present
    STORAGE=${MONICADIR}/storage
    mkdir -p ${STORAGE}/logs
    mkdir -p ${STORAGE}/app/public
    mkdir -p ${STORAGE}/framework/views
    mkdir -p ${STORAGE}/framework/cache
    mkdir -p ${STORAGE}/framework/sessions
    chown -R www-data:www-data ${STORAGE}
    chmod -R g+rw ${STORAGE}
```

---

```shell
#!/bin/bash
```

The entrypoint script seems written to assume `set -e` (`... || true`, etc), but the script does not include `set -e`.

Personally, I'd suggest going with something really explicit like `set -Eeuo pipefail` (since this is Bash), but that does require more thorough testing to make sure you don't run into any unitialized variable errors (where the script was relying on the assumption that they'd simply be empty).

---

```shell
exec $@
```

This will turn something like `xyz-command 'foo bar baz'` into `xyz-command 'foo' 'bar' 'baz'`.

Here's what I'd suggest instead, which will preserve quoting properly:

```shell
exec "$@"
```

---

```dockerfile
    apt-get -y autoremove; \
```

This shouldn't be necessary, because the `php` image shouldn't leave any dangling dependencies.  If it does, we should figure out why and fix it there, but more importantly removing them here won't actually do anything except add image bloat (because they already exist in previous layers, so this won't do much of anything but add more whiteout files at best).

asbiin:
@tianon Thank you so much for your review!
I'll update my script with your recommandations.
There is still 2 points to address:

## sentry
I could get the sentry-cli binary from their [npm package](https://www.npmjs.com/package/@sentry/cli), using an other container 'node', and copy the binary from this container.

## `VOLUME /var/www/html` and syncing sources
Actually it was copied from Nextcloud image :)

What I would like to achieve here is to use the fpm variant, and use another container with an nginx server.
The nginx server will need to access the static files in the /public directory, like css, images, etc.
The only way I found to do that is to copy rsync the content of the website. Is there a more elegant way to use fpm + nginx?

Thank you again

tianon:
> ## sentry
> I could get the sentry-cli binary from their [npm package](https://www.npmjs.com/package/@sentry/cli), using an other container 'node', and copy the binary from this container.

That's not very likely to be an acceptable solution (https://github.com/docker-library/faq#multi-stage-builds) -- if you need `sentry-cli`, you'll probably have to resort to downloading it in some secure way yourself for the cross section of architectures that it supports that you also support.

I'm curious though, why does Monica need `sentry-cli`?  From the current installation method, it appears that the dependency is optional?

> ## `VOLUME /var/www/html` and syncing sources
> Actually it was copied from Nextcloud image :)
> 
> What I would like to achieve here is to use the fpm variant, and use another container with an nginx server.
> The nginx server will need to access the static files in the /public directory, like css, images, etc.
> The only way I found to do that is to copy rsync the content of the website. Is there a more elegant way to use fpm + nginx?

See https://github.com/phpmyadmin/docker/issues/284#issuecomment-625468269 for some very recent discussion around that exact point.

asbiin:
@tianon Ok I agree with you. I'll remove the sentry get. I don't think it's used so much, and I can find another way to get it from a monica script only if activated, so not include the cli in the docker image.

Thank you for the tip for fpm+nginx, it completely make sense.

asbiin:
@tianon again, thanks a lot.
I fix all the feedback from you.

```
ENV PHP_OPCACHE_VALIDATE_TIMESTAMPS="0" \
    PHP_OPCACHE_MAX_ACCELERATED_FILES="20000" \
    PHP_OPCACHE_MEMORY_CONSUMPTION="192" \
    PHP_OPCACHE_MAX_WASTED_PERCENTAGE="10"
```
This was actually intended, and I make the ini file use the variable, instead of fixing them in the Dockerfile. I hope it's ok like that.

asbiin:
Last update use now gpg to check upstream sources instead of sha512

asbiin:
OK, I think the image is now stable. If anyone has a chance to look at it and make other feedback, I would very appreciate it.
Thank you!

yosifkit:
Just some minor feedback from me:

To maximize image layer cache, these labels (and any that might change as often) ought to be defined next to the `ENV MONICA_VERSION` (i.e. as late as possible to avoid a premature docker build cache bust):

```
      org.opencontainers.image.revision="186f336799df2740df28198bfd2431f206668ff2" \
      org.opencontainers.image.version="v2.18.0"
```

A few of the modules are already enabled the base php image, so they don't need to be run through `docker-php-ext-install`: `json` `mbstring` `iconv` and `sodium`. (Though, it is usually harmless as is.) We also discovered https://github.com/docker-library/wordpress/pull/511 as I've been working on this comment, so `opcache` just needs `docker-php-ext-enable` :smile:

asbiin:
Thanks you for the update @yosifkit !
I've applied all your feedbacks: moving the LABEL and removing all modules already included in the base php image.

test/tests/monica-apache-run/run.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+set -eo pipefail
+
+dir="$(dirname "$(readlink -f "$BASH_SOURCE")")"
+
+image="$1"
+
+mysqlImage='mysql:5.7'
```
yosifkit, line 8: The only feedback I have is a recent improvement to the tests (https://github.com/docker-library/official-images/pull/8040), but I can add it later. :smile_cat:

https://github.com/docker-library/official-images/blob/39708ad55fab9b12c0c21afbc0a57a78e2e02a24/test/tests/wordpress-apache-run/run.sh#L14-L17

asbiin: I can add it now if you prefer?

yosifkit: Sure! :heart:

asbiin: ... and done!

tianon:
Sorry for the delay!  Overall, this is looking real good -- just a couple really minor comments left and I think we're good to go!

---

```diff
+      org.opencontainers.image.version="%%VERSION%%"
```

:wink:

---

```diff
+ENV APACHE_DOCUMENT_ROOT /var/www/html/public
+RUN set -ex; \
+    \
+    sed -ri -e "s!/var/www/html!${APACHE_DOCUMENT_ROOT}!g" $APACHE_CONFDIR/sites-available/*.conf; \
+    sed -ri -e "s!/var/www/!${APACHE_DOCUMENT_ROOT}!g" $APACHE_CONFDIR/apache2.conf $APACHE_CONFDIR/conf-available/*.conf
```

I think I have similar concerns to this as to the previous opcache bits -- the way this is set, the document root is going to get set to `/var/www/html/public` but the variable will still exist.  Perhaps this should just be something like the following instead?

```diff
+RUN set -ex; \
+    APACHE_DOCUMENT_ROOT=/var/www/html/public; \
+    sed -ri -e "s!/var/www/html!${APACHE_DOCUMENT_ROOT}!g" $APACHE_CONFDIR/sites-available/*.conf; \
+    sed -ri -e "s!/var/www/!${APACHE_DOCUMENT_ROOT}!g" $APACHE_CONFDIR/apache2.conf $APACHE_CONFDIR/conf-available/*.conf
```

asbiin:
@tianon Thanks!

Indeed I first though APACHE_DOCUMENT_ROOT was a common env variable, but it does not exist in the [upstream Dockerfile](https://github.com/docker-library/php/blob/master/7.3/stretch/apache/Dockerfile)!

asbiin:
@tianon I have a last (?) change I would like to merge: https://github.com/monicahq/docker/pull/25 to output queue and cron command to `stderr` so message appear on docker logs. What do think of this?

tianon:
It seems fine to me, although I don't necessarily understand what problem it's solving? :sweat_smile:

(It can be included either here or as a PR after we merge -- it doesn't really make much difference overall, especially since the `COPY` of the changed scripts is so low so most of the build will be cached anyhow :+1:)

asbiin:
OK fine, we can do this as a review process later, when I have some arguments to this change ;)
Thanks!

PR #7284, opened 2020-01-16
Adjust apt + gpg example to use gpg --export instead of apt-key

yosifkit:
Also swapped order to prefer download + gpg over `apt` since it is more common.

Edit: moved to using MySQL as example since MongoDB was using Xenial (though i could have moved to 4.2 :disappointed:).


yosifkit:
> **add** *filename*
>
> Note: Instead of using this command a keyring should be placed directly in the /etc/apt/trusted.gpg.d/ directory with a descriptive name and either "gpg" or "asc" as file extension.
>
> **adv**
>
> Note that there are no checks performed, so it is easy to completely undermine the apt-secure(8) infrastructure if used without care.

https://manpages.debian.org/buster/apt/apt-key.8.en.html

Issue #3540, opened 2017-10-05
Standard set of labels for official images


Djelibeybi:
Hey librarians!

Is there a documented set of labels we should be using in the `Dockerfile` of an official library image? Currently, we don't have any labels for the Oracle Linux image and my (admittedly brief) scan of other official base OS (`FROM scratch`) images shows almost zero consistency.

Using labels seems to be 👍 so I just want to make sure I'm using either the _right_ labels or at least the _best_ labels.  😄

tianon:
Honestly, I find labels on base images to be irritating, especially since they're all inherited by any image `FROM` the base (just like `MAINTAINER` of old, which they descended from), and prefer to consider them something useful for users to organize their images and/or their deployments.

As for the official stance, we don't have one -- if an image maintainer chooses to apply `LABEL`s to their image, that's their prerogative (a few examples that do are [CentOS](https://github.com/CentOS/sig-cloud-instance-images/blob/16dab97b0ce72b1db7a2f9b02c76e452cb0a63cb/docker/Dockerfile#L4-L7), [Silverpeas](https://github.com/Silverpeas/docker-silverpeas-prod/blob/03bd2800e83bf262d07b53f067758115fa6f05fc/Dockerfile#L70), and [Træfɪk](https://github.com/containous/traefik-library-image/blob/d807e0f3692209f960bb2567e7890453a190c899/scratch/amd64/Dockerfile#L7-L13)).

I think the argument for them is slightly stronger in the case of a single bit of software (like Træfɪk), but even there I'm indifferent (and the added need to then update both the software contained in the image and the metadata of the image concurrently is irritating, although trivially scriptable -- basically just one more place for information to potentially become out-of-sync).

tianon:
I literally just noticed your "Hey librarians!" here :100: :laughing: :heart:

tianon:
Closing since this isn't something we plan to implement any time soon.  If that ever changes, we're happy to re-open. :+1:

alexbozhenko:
Hi, @tianon and @yosifkit.
For images that do set oci labels, the image is self-describing, e.g.
`# docker  pull solr && docker image inspect solr --format='{{json .Conf
ig.Labels}}' | jq '.'`
<details>

```
Using default tag: latest
latest: Pulling from library/solr
Digest: sha256:60f3fbc0911eabd3ad0a6efb2f465b6e68d35ed487bda7b6bbf6339b9b5a39e6
Status: Image is up to date for solr:latest
[...]

PR #7288, opened 2020-01-16
Add phpMyAdmin

williamdes:
Closes: https://github.com/phpmyadmin/docker/issues/56

# Checklist for Review

    - https://github.com/phpmyadmin/docker
    - "service"
    - https://github.com/docker-library/docs/pull/1642
    - `FROM php:xxx`

### DIFF
--- a/library/phpmyadmin
+++ b/library/phpmyadmin
@@ -1,0 +1,19 @@
+# This file is generated via https://github.com/phpmyadmin/docker/blob/52d6fe4cafdb81b97c5ebb9cbf990456efe5eb9f/generate-stackbrew-library.sh
+Maintainers: Isaac Bennetch <bennetch@gmail.com> (@ibennetch),
+             William Desportes <williamdes@wdes.fr> (@williamdes)
+GitRepo: https://github.com/phpmyadmin/docker.git
+
+Tags: 5.0.2-apache, 5.0-apache, 5-apache, apache, 5.0.2, 5.0, 5, latest
+Architectures: amd64, arm32v7, arm64v8, i386, ppc64le, s390x
+GitCommit: f464cf5a67f8eae125d7f098892c03dcb5b9858b
+Directory: apache
+
+Tags: 5.0.2-fpm, 5.0-fpm, 5-fpm, fpm
+Architectures: amd64, arm32v7, arm64v8, i386, ppc64le, s390x
+GitCommit: f464cf5a67f8eae125d7f098892c03dcb5b9858b
+Directory: fpm
+
+Tags: 5.0.2-fpm-alpine, 5.0-fpm-alpine, 5-fpm-alpine, fpm-alpine
+Architectures: amd64, arm32v7, arm64v8, i386, ppc64le, s390x
+GitCommit: f464cf5a67f8eae125d7f098892c03dcb5b9858b
+Directory: fpm-alpine


williamdes:
Should be okay now, review can start

williamdes:
@yosifkit could you please review this one

williamdes:
@yosifkit could you please review the new phpMyAdmin image ?

J0WI:
_@williamdes you can use the [request review](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/requesting-a-pull-request-review) feature._

tianon:
Sorry for the delay (and thanks for your patience) :pray:

---

Taking an initial pass on the Dockerization, I've got a few comments:

> ```dockerfile
> FROM php:7.4-apache
> ```

Given the number of packages being installed, I'd recommend adjusting this to be explicit about the base OS version (so `FROM php:7.4-apache-buster` instead), which will allow you to control when it updates to `bullseye`, for example (because it will likely require more changes when that occurs).

The same goes for `FROM php:7.4-fpm-alpine` -> `FROM php:7.4-fpm-alpine3.11` and `FROM php:7.4-fpm` -> `FROM php:7.4-fpm-buster`.

(This is just a suggestion -- if you'd rather just have the builds potentially fail for the benefit of them maybe not failing and update them only if necessary, that's fine too. :+1:)

---

> ```dockerfile
> RUN { \
> ... \
>     } > $PHP_INI_DIR/conf.d/opcache-recommended.ini; \
> ...
> ```

This `RUN` line is using `;` as a separator, but doesn't include `set -e` like the other `RUN` lines doing the same (it's not likely that any of those `echo`s are going to fail, but we'd rather have the build fail than silently publish an image that's missing some bit of intended configuration).

> ```dockerfile
> LABEL version=$VERSION
> ```

---

> ```dockerfile
> # Download tarball, verify it using gpg and extract
> RUN set -ex; \
>     fetchDeps=" \
> ...
>     apt-get purge -y --auto-remove -o APT::AutoRemove::RecommendsImportant=false $fetchDeps; \
> ```

I would recommend adjusting this to use the `apt-mark` behavior used in the lines above so that if one of the packages in `fetchDeps` was already installed, it doesn't get accidentally removed.

---

> ```dockerfile
>     curl --output phpMyAdmin.tar.xz --location $URL; \
> ```

Any invocation of `curl` should include `--fail` / `-f` (at a minimum).

---

> ```dockerfile
>     tar -xf phpMyAdmin.tar.xz -C /usr/src; \
> ...
>     mv /usr/src/phpMyAdmin-$VERSION-all-languages /usr/src/phpmyadmin; \
> ```

It's not critical, but this could be written more simply using `tar`'s `--strip-components` flag to avoid needing to use `$VERSION` at all here:

```dockerfile
     mkdir /usr/src/phpmyadmin; \
     tar -xf phpMyAdmin.tar.xz -C /usr/src/phpmyadmin --strip-components=1; \
```

---

> ```dockerfile
> # Add directory for sessions to allow session persistence
>     mkdir /sessions; \
>     mkdir -p /var/nginx/client_body_temp; \
> ```

> ```bash
>     # Custom folders for sessions
>     chown ${user}:${group} /sessions /var/nginx/client_body_temp
> ```

Is this still relevant, or is this a layover from a previous version of the `Dockerfile`?

---

> ```dockerfile
>     apk del .build-deps
> ```

It's not strictly required, but I would recommend adding `--no-network` to any `apk del` instruction.

---

> ```bash
>         echo >&2 "phpMyAdmin not found in $PWD - copying now..."
> ...
> ```

This section seems like it's been copied from the Wordpress image -- I have to wonder if the underlying intent for why Wordpress does this actually applies to phpMyAdmin?

For context, this pattern was adopted in Wordpress specifically because Wordpress has a tendency to treat the code of Wordpress itself as "mutable state" -- the configuration file can be updated/modified to a minor extent via the web UI (especially the initial configuration), plugins and themes get downloaded and installed directly inside the installation directory, but most relevantly Wordpress strongly recommends allow it to do self-auto-updates.

In this sense, the image is a "seed" for your Wordpress install, after which it's self-maintaining to some degree (minus system dependencies like PHP version / updates, etc).

Looking at https://github.com/phpmyadmin/phpmyadmin/wiki/Automatic_updates, it does not appear that phpMyAdmin has similar functionality/behavior (yet?), and looking further at https://github.com/phpmyadmin/phpmyadmin/issues/6155 it does not appear to be something that seems to be a high priority either.

Given all that, I wonder why phpMyAdmin is not simply being installed directly to `/var/www/html` in the image itself and thus not needing any copying behavior in the entrypoint?

williamdes:
Thank you @tianon for the review
@J0WI made the changes in https://github.com/phpmyadmin/docker/pull/274 :muscle: 
It is merged and I am waiting for the next version of phpMyAdmin to update the PR

Everything seems right now ?

(I apologise for the delay to fix our image after the feedback)

williamdes:
I also made a PR to apply your recommendation on labels
https://github.com/phpmyadmin/docker/pull/275

J0WI:
@williamdes what do you think about the `echo >&2 "phpMyAdmin not found in $PWD - copying now..."`? Will phpMyAdmin become mutable in near future?

williamdes:
> Will phpMyAdmin become mutable in near future?

What do you mean by mutable?

J0WI:
> What do you mean by mutable?

That the files in `/var/www/html` change during runtime

williamdes:
> > What do you mean by mutable?
> 
> That the files in `/var/www/html` change during runtime

No, I do not see any chance of such a feature in the future

Should I remove the line y mentioned ?

ibennetch:
I wasn't the one who wrote the original Docker code, but strongly suspect that what tianon said is correct: downloading to `/usr/src/phpmyadmin` and then copying to `/var/www/html` is simply copied from some other project that had need of such functionality.

I see no reason why phpMyAdmin can't directly extract the files to `/var/www/html`. I've opened https://github.com/phpmyadmin/docker/issues/277 to put that in our to do list.

williamdes:
@tianon could you please re-review this request ?

williamdes:
Did somebody at Docker have some time to review our changes?

J0WI:
@tianon this made some progress since your last review and it would be great to get an update on this :smiley:

tianon:
> ```dockerfile
> ARG BUILD_DATE
> ARG VCS_REF
> ```

These aren't intended for us to set, right? (because our build processes will not provide any build args, and these will thus be empty on every build)

---

> ```dockerfile
> LABEL maintainer="The phpMyAdmin Team <developers@phpmyadmin.net>" \
> ```

---

> ```dockerfile
> # Add directory for sessions to allow session persistence
> ```

:smile: :wink:

---

> ```sh
>     if [ "$(id -u)" = '0' ]; then
>         case "$1" in
>             apache2*)
>                 user="${APACHE_RUN_USER:-www-data}"
>                 group="${APACHE_RUN_GROUP:-www-data}"
>                 ;;
>             *) # php-fpm
>                 user='www-data'
>                 group='www-data'
>                 ;;
>         esac
>     else
>         user="$(id -u)"
>         group="$(id -g)"
>     fi
> ```

This section appears to be unused?

williamdes:
@J0WI do I have to update the file now on this PR ?

J0WI:
> @J0WI do I have to update the file now on this PR ?

Right after reviewing https://github.com/phpmyadmin/docker/pull/286 you can update the `GitCommit` here.

williamdes:
@tianon I did update the file and rebase after the latest suggestions
Thank you for the review, it should be all right now

ibennetch:
> Given all that, I wonder why phpMyAdmin is not simply being installed directly to /var/www/html in the image itself and thus not needing any copying behavior in the entrypoint?

After I modified the code to install directly to /var/www/html in the image, there were some reports from people with advanced configurations. The primary report [uses /var/www/html as a volume](https://github.com/phpmyadmin/docker/issues/284#issuecomment-619002215). There's an additional report that I'm not sure is related. The user [has a k8s install](https://github.com/phpmyadmin/docker/issues/284#issuecomment-617258021) where nginx won't serve the static parts of phpMyAdmin.

Because of the trouble with getting updates when mounting a volume, I'm likely to revert the change to downloading directly to /var/www/html, or modify it a bit so that the code tests if /usr/share/phpmyadmin is the same version as /var/www/html and overwrites it if it's not.

I'm mentioning this here because I didn't want it to even appear like I'm trying to trick you by fixing something you mentioned, then reverting it after you'd looked at it.

emanuelb:
The change to use `--strip-components=1` instead of previous $VERSION usage, enable downgrade attacks (attacker change the content of downloaded file via curl command to be content of old version via MITM or malicious mirror, hacked server, etc...) which will pass the GPG test (they both signed with same key), the usage of $VERSION is the fix (as the mv operation will fail because the file don't exists if versions don't match between what in variable vs folder name inside downloaded archive)
another solution is to add hash checks (using sha256sum or b2sum commands)
I reported this issue a while ago, probably to security@ email, you can see some of the content here:
https://github.com/phpmyadmin/docker/issues/43

tianon:
*We* would definitely love to see a checksum verification on top of the existing PGP verification. :smile:

J0WI:
> attacker change the content of downloaded file via curl command to be content of old version via MITM or malicious mirror, hacked server, etc...

I tend to agree, but if you manage to break `https://files.phpmyadmin.net/` they are in huge trouble anyway...

emanuelb:
> > attacker change the content of downloaded file via curl command to be content of old version via MITM or malicious mirror, hacked server, etc...
> 
> I tend to agree, but if you manage to break `https://files.phpmyadmin.net/` they are in huge trouble anyway...

The usage of GPG signing is for the reason that we don't trust the network, and it can return us everything in different scenarios (SSL MITM, malicious mirror/CDN/cloud-provider, or turned to be malicious as it's hacked by attacker)
see: https://www.qubes-os.org/faq/#what-does-it-mean-to-distrust-the-infrastructure
Thus attacker must steal the signing key in order to return malicious content instead (of course anyone who download files without verification will be owned in any of the above situations)
I suggested to add verification to downloads from the website in all places which was implemented in:
https://github.com/phpmyadmin/website/issues/24
The usage of mv with $VERSION was due to my report of downgrade attacks, as the new change to use `--strip-components=1` revert it(is regression) a fix is needed, adding checksum verification as well is the best course of action with using the new `--strip-components=1` method.

tianon:
> ... adding checksum verification as well is the best course of action ...

Especially since https://files.phpmyadmin.net appears to have officially-published SHA256 values that should be pretty trivial to consume/embed! :smile:

J0WI:
I created https://github.com/phpmyadmin/docker/pull/289 to add the checksum, but It not a big deal as long as the checksums are served from the same server.

@tianon is this a blocker for you?

williamdes:
@tianon Is there something blocking this PR ?

williamdes:
@yosifkit @tianon 

All the suggestions have been applied, and I merged the SHA checksum checking part.
Would you agree to merge this one soon ?

There is one comment to read before merging: https://github.com/docker-library/official-images/pull/7288#issuecomment-625320576

J0WI:
:ping_pong: @tianon :wink:

williamdes:
Hi dear reviewers ;)
I rebased my PR if you could have a look ?

cc @tianon

williamdes:
Hi @tianon 
Could you let us know what you think about this PR ?

tianon:
Sorry for the delay -- I've just re-reviewed this, and I think it looks like it's in a good state.

In other words, LGTM

cc @yosifkit

williamdes:
Hi @yosifkit and @tianon 
I just had a look to https://hub.docker.com/search?image_filter=official&type=image&category=database and could not find phpMyAdmin but Adminer is listed. Could you add our image to the category please ?

tianon:
Unfortunately, that's not something we currently can control -- I'd
suggest you add your "thumbs up" on
https://github.com/docker/roadmap/issues/76

PR #2917, opened 2017-05-02
Add CentOS variant and new template system

dmetzler:
This just adds a CentOS variant. For that we also introduce a new templating system that allows us to generate more easily the Dockerfiles etc... plus a `generate-stackbrew` script.

Complete story of development can be found here : https://github.com/nuxeo/docker-nuxeo/pull/10

No rush to merge, I did that in parallel of our 9.1 release, so PR happens after it.

PR #7307, opened 2020-01-21
Add buster and centos images

gabor-boros:
Extend the list of available tags with Debian and Centos8. This request was initiated by our community in issue https://github.com/rethinkdb/rethinkdb-dockerfiles/issues/47.

### DIFF
--- a/library/rethinkdb
+++ b/library/rethinkdb
@@ -1,6 +1,10 @@
-Maintainers: Gábor Boros <gabor.brs@gmail.com> (@gabor-boros)
+Maintainers: RethinkDB <services@rethinkdb.com> (@gabor-boros)
 GitRepo: https://github.com/rethinkdb/rethinkdb-dockerfiles.git
-GitCommit: 54d3eebf6409b196264c193e0cbad027061739b3
 
-Tags: 2.4.0, 2.4, 2, latest
-Directory: bionic/2.4.0
+Tags: 2.4.0-buster-slim, 2.4-buster-slim, 2-buster-slim, buster-slim, 2.4.0, 2.4, 2, latest
+Directory: buster/2.4.0
+GitCommit: 444f7754822cb3e96cb2b66042d8e87d2227be0b
+
+Tags: 2.4.0-centos, 2.4-centos, 2-centos, centos
+Directory: centos8/2.4.0
+GitCommit: 444f7754822cb3e96cb2b66042d8e87d2227be0b


tianon:
Isn't RethinkDB a database that folks run and use externally via TCP, etc?  What functionally changes for a user by using a different base image, especially one that's comparable in size (or larger in the case of `centos`)?

(Trying to understand the indented use case of these variants. :confused:)

daveisfera:
Different base OSes have different libraries and security support processes. I'm more familiar with RedHat/CentOS than Debian/Ubuntu and I'm appreciative that I can use that as a base

yosifkit:
So, I think this [comment](https://github.com/docker-library/official-images/pull/2917#issuecomment-300637486) is my most comprehensive response on adding "image Y based on Linux distro X"

-------------------------

Response to the following comment to hopefully explain how the official-images are built (and one difficulty introduced when basing an image on the `centos` image)

>Different base OSes have different libraries and security support processes. I'm more familiar with RedHat/CentOS than Debian/Ubuntu and I'm appreciative that I can use that as a base

> Tags in the library file are only built through an update to that library file or as a result of its base image being updated
>
> \- https://github.com/docker-library/official-images/tree/3dd9ba40463965e74597520f568e06ebf6da376a#library-definition-files

The official-images build servers make heavy use of docker build cache, so only changed lines (and build context) or a new base will update an image.

While I don't know if CentOS packages are updated more frequently than Debian or Ubuntu, one problem is that the `centos` images get updated less frequently than `debian` or `ubuntu`. The respective maintainers strive to update `debian` and `ubuntu` images at least monthly. See the links at end to view the cadence of each image.  What this means for images based off of `centos` is that they will not be rebuilt to include any of those new packages unless they change their Dockerfile to break docker build cache. We recommend against changing the Dockerfile to update packages that are included in the base image (like `centos`) to get security fixes, since it is duplicating the space used by those package (since the old version still exists in the other docker image layer) and should instead be addressed in the base image itself.

- [`centos` prs](https://github.com/docker-library/official-images/pulls?utf8=%E2%9C%93&q=is%3Apr+label%3Alibrary%2Fcentos) (note that not all versions are updated in each PR)
- [`debian` prs](https://github.com/docker-library/official-images/pulls?q=is%3Apr+label%3Alibrary%2Fdebian)
- [`ubuntu` prs](https://github.com/docker-library/official-images/pulls?utf8=%E2%9C%93&q=is%3Apr+label%3Alibrary%2Fubuntu)

daveisfera:
I agree that layer caching is crucial for docker to work well and that base images should be updated appropriately to pull in security fixes on a reasonable (if not an aggressive schedule), but I don't see why that impacts adding a `centos` version of RethinkDB images. There are [real reasons to chose RHEL/CentOS over other base images](https://www.elastic.co/blog/docker-base-centos7).

tianon:
I'm reading that post, and I can definitely agree with the argument of something like CentOS when compared to Alpine (musl libc is definitely odd and has many "weird" edge behaviors when coming from glibc, which is what most software is tested against), but I'm having trouble seeing how anything there applies to CentOS vs Ubuntu (or Debian), where both are glibc based and one base image is updated frequently while the other is updated much less frequently.

The size arguments there are also confusing to me because I'm not seeing the same math -- `centos:7` is ~70MB compressed download size, but the on-disk size is closer to ~200MB.  Contrast that to `ubuntu:bionic` which is ~25MB compressed and ~65MB on-disk.  If we're comparing compressed size (as the referenced CentOS image size in that post indicates), Alpine is really closer to 2.5MB than it is to 5MB.

Multi-architecture support is another point that post doesn't even touch (because the Elastic folks aren't interested in touching it, from my understanding) -- `centos:8` supports `amd64`, `arm64v8`, and `ppc64le`, while `ubuntu:bionic` has a much longer list.  Additionally, the package manager in Ubuntu has a native concept of "system architecture" which it uses to fetch packages, while CentOS (and alas, all RPM-based distributions) use the equivalent of `uname -m` for choosing package architecture, which is unusable when you're doing something like `arm32v7` builds on an `arm64v8` machine (even if using `setarch` to run with 32bit `uname -m`, it will report `armv8l` and thus fail to fetch `armv7l` packages).

gabor-boros:
Hello @tianon and @yosifkit,

I would check this question from a different angle. The main reason of this change is not to provide alternatives for the users because of image size. As a former RedHat system administrator I can clearly state, that for those who need to operate own services and third parties it is much simpler to work in the same environment. Although the compiler is the same on both OS which compiled the database, for the end users a CentOS based os could mean a huge help what I would not like to block for them.

An other important note: The debian <> ubuntu related comments are valid, so I updated to PR to use slim version of debian images.

gabor-boros:
@yosifkit or @tianon, do you have any opinion on this?

yosifkit:
I would point out that one of the major benefits of containers is to not have to worry about what OS (libraries, tools, etc) that the (database) service is packaged with. It should be using whatever is best for the service and not necessarily match what the user is running on the host. It could be the same OS; it could be older or newer; or it may not even include a shell. The interface to the service is Docker and the host OS. I realize that this is the ideal and users are often unable to operate in the ideal, but it is something that should be encouraged wherever possible.

With that, I am not against the CentOS variant if that is what y'all want to support. I am still hesitant on having both Ubuntu and Debian versions since anyone used to a Debian environment would have no trouble debugging in an Ubuntu-based container.

gabor-boros:
Hello @yosifkit,

>  I am still hesitant on having both Ubuntu and Debian versions since anyone used to a Debian environment would have no trouble debugging in an Ubuntu-based container.

I fully understand this. I would say lets only have Debian, the size is smaller and user would have no trouble as you pointed out. I just removed the ubuntu version.

gabor-boros:
@tianon or @yosifkit could you please help what is the issue with the commit? (Travis build is failing) I can not see any error message in the output.

tianon:
Doh, there was a bug in the script -- I've rebased this on master and pushed a fix to the script which should give us the correct output.

tianon:
Success: (https://travis-ci.org/docker-library/official-images/builds/650211258#L354-L361)

```console
$ ./bashbrew/travis.sh

Invalid 'FROM' + 'Constraints' combinations detected:

 - rethinkdb:2.4.0-centos -- missing constraints (FROM centos:8):
   - !aufs

The command "./bashbrew/travis.sh" exited with 1.
```

library/rethinkdb
```diff
@@ -1,6 +1,10 @@
...
 GitRepo: https://github.com/rethinkdb/rethinkdb-dockerfiles.git
-GitCommit: 54d3eebf6409b196264c193e0cbad027061739b3
 
-Tags: 2.4.0, 2.4, 2, latest
-Directory: bionic/2.4.0
+Tags: 2.4.0-buster-slim, 2.4-buster-slim, 2-buster-slim, buster-slim, 2.4.0, 2.4, 2, latest
+Directory: buster/2.4.0
+GitCommit: b65c9817c67bf6738bb083c1da1dca76714ddf06
+
+Tags: 2.4.0-centos, 2.4-centos, 2-centos, centos
```
tianon:
```suggestion
Tags: 2.4.0-centos, 2.4-centos, 2-centos, centos
Constraints: !aufs
```

tianon:
<details>
<summary>Diff:</summary>

```diff
diff --git a/_bashbrew-arches b/_bashbrew-arches
index 16c952c..9914cbb 100644
--- a/_bashbrew-arches
+++ b/_bashbrew-arches
@@ -1 +1,2 @@
+rethinkdb:centos @ amd64
 rethinkdb:latest @ amd64
diff --git a/_bashbrew-list b/_bashbrew-list
index ce0c8df..2d9e02e 100644
--- a/_bashbrew-list
+++ b/_bashbrew-list
@@ -1,4 +1,12 @@
 rethinkdb:2
+rethinkdb:2-buster-slim
+rethinkdb:2-centos
 rethinkdb:2.4
+rethinkdb:2.4-buster-slim
+rethinkdb:2.4-centos
 rethinkdb:2.4.0
+rethinkdb:2.4.0-buster-slim
+rethinkdb:2.4.0-centos
+rethinkdb:buster-slim
+rethinkdb:centos
 rethinkdb:latest
diff --git a/rethinkdb_centos/Dockerfile b/rethinkdb_centos/Dockerfile
new file mode 100644
index 0000000..305145e
--- /dev/null
+++ b/rethinkdb_centos/Dockerfile
@@ -0,0 +1,17 @@
+FROM centos:8
+
+RUN curl -L https://download.rethinkdb.com/centos/8/`uname -m`/rethinkdb.repo > /etc/yum.repos.d/rethinkdb.repo
+
+ENV RETHINKDB_PACKAGE_VERSION 2.4.0
+
+RUN yum install -y rethinkdb-$RETHINKDB_PACKAGE_VERSION \
+	&& yum clean all
+
+VOLUME ["/data"]
+
+WORKDIR /data
+
+CMD ["rethinkdb", "--bind", "all"]
+
+# process cluster webui
+EXPOSE 28015 29015 8080
diff --git a/rethinkdb_latest/Dockerfile b/rethinkdb_latest/Dockerfile
index 441ed5d..8edf00a 100644
--- a/rethinkdb_latest/Dockerfile
+++ b/rethinkdb_latest/Dockerfile
@@ -1,13 +1,13 @@
-FROM ubuntu:bionic
+FROM debian:buster-slim
 
 RUN apt-get -qqy update \
     && apt-get install -y --no-install-recommends ca-certificates gnupg2 \
     && rm -rf /var/lib/apt/lists/*
 
 RUN apt-key adv --keyserver keys.gnupg.net --recv-keys "539A 3A8C 6692 E6E3 F69B 3FE8 1D85 E93F 801B B43F" \
-    && echo "deb https://download.rethinkdb.com/apt bionic main" > /etc/apt/sources.list.d/rethinkdb.list
+    && echo "deb https://download.rethinkdb.com/apt buster main" > /etc/apt/sources.list.d/rethinkdb.list
 
-ENV RETHINKDB_PACKAGE_VERSION 2.4.0~0bionic
+ENV RETHINKDB_PACKAGE_VERSION 2.4.0~0buster
 
 RUN apt-get -qqy update \
 	&& apt-get install -y rethinkdb=$RETHINKDB_PACKAGE_VERSION \
```

</details>

tianon:
```diff
+RUN curl -L https://download.rethinkdb.com/centos/8/`uname -m`/rethinkdb.repo > /etc/yum.repos.d/rethinkdb.repo
```

This has me concerned -- I downloaded https://download.rethinkdb.com/centos/8/x86_64/rethinkdb.repo to verify, and it's only five simple lines:

```console
$ wget -qO- 'https://download.rethinkdb.com/centos/8/x86_64/rethinkdb.repo'
[rethinkdb]
name=RethinkDB
baseurl=https://download.rethinkdb.com/centos/8/x86_64/
gpgcheck=0
enabled=1
```

Beyond downloading this tiny file over the internet, it strikes me as odd that `gpgcheck=0` -- is that because it's a hassle to set up and manage, or just because it's easier for users to use that way?  Does the repo have a key which could be added/verified?

gabor-boros:
@tianon The download server and the setup of that is suboptimal, and it is a hassle to set up and manage the things there. Right now I'm working on to refactor that and make sure we are using the best practices (across the whole project including the download server). 

Once I reach that point, I'll set up everything properly for repos too, such like `gpgcheck`.

gabor-boros:
@tianon @yosifkit is there any blocker to merge this?

gabor-boros:
Update: We are preparing the new download solution, but we will sign RPM packages after the migration happened. The gpg check will be introduced in a separate PR (I wouldn't like to spam this PR).

gabor-boros:
@tianon @yosifkit any updates on this?

tianon:
I'd go back to https://github.com/docker-library/official-images/pull/7307#issuecomment-586532518 -- the GPG bit just strikes me as odd (given you control the repo); the blocker is that it's doing essentially `curl ... > xyz.repo` without any sort of verification of any kind for what amounts to a 5-line file that really could've and should've just been embedded in the `Dockerfile` instead.

gabor-boros:
@tianon I’m in the migration of the download server. By the beginning of the next week (around Monday or Tuesday) I’ll will finish the migration and I’ll be able to sign the rom repo as well.

Is it ok to ping you when I’m done with that?

gabor-boros:
As a quick follow up: The download server is migrated. We will change the download server at the end of the month (by 29th of May) and I'll update this PR too, to include the GPG check as well.

Thanks for pushing that change, I believe it will be better on the long (and even short) run.

gabor-boros:
Dear @tianon and @yosifkit,

The download server is replaced and the GPG checking is setup for the CentOS images as well. Also, the `.repo` file is embedded in the Dockerfile now.

The only failing test is because we had a missing package for buster, which is uploaded now, so everything should be fine now if you can rerun the tests.

Could you please help me resolve this issue? 😇

tianon:
GItHub wouldn't restart properly for some reason, so I rebased/squashed and force pushed to trigger it. :+1:

gabor-boros:
Thank you so much @tianon !

PR #7320, opened 2020-01-22
Triggering Oracle Linux 7 rebuild.

Djelibeybi:
After the merge this morning, Docker's build systems didn't rebuild the `oraclelinux:7-slim` tag for `amd64` but did rebuild the tag 7.7 tag as well as both the 7-slim and 7.7 tags for `arm64v8`.

Checking the image build times shows that the image wasn't updated at all, so this is essentially a no-op PR to trigger that build.

### DIFF
--- a/library/oraclelinux
+++ b/library/oraclelinux
@@ -4,10 +4,10 @@ GitCommit: 01a15ec99c7470a3391c691509db1759b41eaf66
 GitFetch: refs/heads/master
 # https://github.com/oracle/container-images/tree/dist-amd64
 amd64-GitFetch: refs/heads/dist-amd64
-amd64-GitCommit: 550fefbaaceb2d46e9af820214e803b81ea60bde
+amd64-GitCommit: ef45135c78b4f9441c67fb10b41e992ec4c37c59
 # https://github.com/oracle/container-images/tree/dist-arm64v8
 arm64v8-GitFetch: refs/heads/dist-arm64v8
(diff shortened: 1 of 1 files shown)


Djelibeybi:
I figured triggering a new build is probably easier than trying to work out why Docker's official library build system skipped that image this morning.

tianon:
We can hit rebuild but if the tarballs are the same we'll hit Docker build cache again (which is likely what happened here).  I tried to clear the older cache before merging the previous PR to avoid this, but it looks like I need to be more aggressive.  Give me a minute to try a manual force.

(See https://github.com/docker-library/official-images/issues/5216 for more details on the underlying issue.)

tianon:
Got it: https://doi-janky.infosiftr.net/job/put-shared/job/light/job/oraclelinux/1124/console

```
Pushed oraclelinux:7-slim (sha256:45b291ffe89c7e53a6764ff58aa3c4c2ac34f2ba1c95fa16145b7a318a935ac9)
```

tianon:
(we can still look to merge this if it's convenient, but hopefully that
decreases the urgency 😅)

Djelibeybi:
Nope, that worked. Though weirdly it nows shows an image build of 4 weeks ago:

```
➜  ~ docker images
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
oraclelinux         7-slim              0d49b70bc432        4 weeks ago         118MB
```

It's the right image though.

PR #7377, opened 2020-01-30
Update Lightstreamer image due to hotfix

gfinocchiaro:


tianon:
<details>
<summary>Diff:</summary>

```diff
diff --git a/lightstreamer_7-jdk8/Dockerfile b/lightstreamer_7-jdk8/Dockerfile
index 393e531..9053509 100644
--- a/lightstreamer_7-jdk8/Dockerfile
+++ b/lightstreamer_7-jdk8/Dockerfile
@@ -2,6 +2,7 @@
 FROM openjdk:8-jdk
 
 LABEL maintainer="Lightstreamer Server Development Team <support@lightstreamer.com>"
+LABEL build="1951"
 
 # Import Lighstreamer's public key
 RUN gpg --keyserver hkp://p80.pool.sks-keyservers.net:80 --recv-keys 9B90BFD14309C7DA5EF58D7D4A8C08966F29B4D2
@@ -30,7 +31,7 @@ RUN set -ex; \
 # Add new user and group
         && groupadd -r -g 10000 lightstreamer \
         && useradd --no-log-init -r -g lightstreamer -u 10000 lightstreamer \
-# Change ownership of tye lightstreamer folder
+# Change ownership of the lightstreamer folder
         && chown -R lightstreamer:lightstreamer ../lightstreamer \
 # Finally, remove no longer needed files
         && rm Lightstreamer.tar.gz Lightstreamer.tar.gz.asc
diff --git a/lightstreamer_latest/Dockerfile b/lightstreamer_latest/Dockerfile
index e3b8bf5..fecdae0 100644
--- a/lightstreamer_latest/Dockerfile
+++ b/lightstreamer_latest/Dockerfile
@@ -2,6 +2,7 @@
 FROM openjdk:11-jdk
 
 LABEL maintainer="Lightstreamer Server Development Team <support@lightstreamer.com>"
+LABEL build="1951"
 
 # Import Lighstreamer's public key
 RUN gpg --keyserver hkp://p80.pool.sks-keyservers.net:80 --recv-keys 9B90BFD14309C7DA5EF58D7D4A8C08966F29B4D2
@@ -30,7 +31,7 @@ RUN set -ex; \
 # Add new user and group
         && groupadd -r -g 10000 lightstreamer \
         && useradd --no-log-init -r -g lightstreamer -u 10000 lightstreamer \
-# Change ownership of tye lightstreamer folder
+# Change ownership of the lightstreamer folder
         && chown -R lightstreamer:lightstreamer ../lightstreamer \
 # Finally, remove no longer needed files
         && rm Lightstreamer.tar.gz Lightstreamer.tar.gz.asc
```

</details>

tianon:
What's the goal of this PR?  What is this new `build` label intended to accomplish?

If the goal is to have a point to force an image rebuild, we should just update the `openjdk:XX-jdk` base image (or even further down the chain) instead and then the `lightstreamer` image will naturally get rebuilt.

gfinocchiaro:
I'm sorry but It is not clear to me the workflow: how can I trigger an update of the `openjdk:XX-jdx` base image?

tianon:
If you look at https://hub.docker.com/_/lightstreamer?tab=tags, you can see that it already happened ~3 days ago. :smile:

gfinocchiaro:
Oh, cool! Could you expand please more on this? Is there any link you can point me out which explains how and when already published images get rebuilt?

tianon:
Sure, the primary reference is https://github.com/docker-library/official-images#library-definition-files:

> Tags in the library file are only built through an update to that library file or as a result of its base image being updated (ie, an image `FROM debian:jessie` would be rebuilt when `debian:jessie` is built).

It's also referenced at https://github.com/docker-library/faq#why-does-my-security-scanner-show-that-an-image-has-cves (with a link back to that primary reference):

> These refreshed base images also means that any other image in the Official Images program that is `FROM` them will also be rebuilt (as described in [the project `README.md` file](https://github.com/docker-library/official-images#library-definition-files)).

gfinocchiaro:
Ok, thank you very much.

PR #7459, opened 2020-02-11
Make amazon corretto use corretto yum repositories.

alvdavi:
We have updated our dockerfile images to make use of our new yum repos. This does not change the corretto version in the container, but it will allow people to update containers with yum update.

### DIFF
--- a/library/amazoncorretto
+++ b/library/amazoncorretto
@@ -6,11 +6,11 @@ Maintainers: Amazon Corretto Team (@corretto),
 Tags: 8, 8u242, 8-al2-full, latest
 GitRepo: https://github.com/corretto/corretto-8-docker.git
 GitFetch: refs/heads/8-al2-full
-GitCommit: ae31a3e5106d3d16e0a3154bca567c10879522af
+GitCommit: 222fba51be1858ab30f32ffff762a6909ef0e0e8
 Architectures: amd64, arm64v8
 
 Tags: 11, 11.0.6, 11-al2-full
 GitRepo: https://github.com/corretto/corretto-11-docker.git
 GitFetch: refs/heads/11-al2-full
-GitCommit: a797c24219a9262e581ee70fc928c57acacde331
+GitCommit: e6b006191f79582abc672bde1ecddffa6417639a
 Architectures: amd64, arm64v8


yosifkit:
<details>
<summary>Diff:</summary>

```diff
diff --git a/amazoncorretto_11-al2-full/Dockerfile b/amazoncorretto_11-al2-full/Dockerfile
index 91972e8..297f463 100644
--- a/amazoncorretto_11-al2-full/Dockerfile
+++ b/amazoncorretto_11-al2-full/Dockerfile
@@ -1,16 +1,7 @@
 FROM amazonlinux:2
 
-# x86_64 args
-ARG rpm_x64=java-11-amazon-corretto-devel-11.0.6.10-1.x86_64.rpm
-ARG path_x64=https://corretto.aws/downloads/resources/11.0.6.10.1
-ARG key_x64=6DC3636DAE534049C8B94623A122542AB04F24E3
-
-# aarch64 args
-ARG rpm_aarch64=java-11-amazon-corretto-devel-11.0.6.10-1.aarch64.rpm
-ARG path_aarch64=https://corretto.aws/downloads/resources/11.0.6.10.1
-ARG key_aarch64=6DC3636DAE534049C8B94623A122542AB04F24E3
-
-# In addition to installing the RPM, we also install
+ARG version=11.0.6.10-1
+# In addition to installing the Amazon corretto, we also install
 # fontconfig. The folks who manage the docker hub's
 # official image library have found that font management
 # is a common usecase, and painpoint, and have
@@ -18,21 +9,11 @@ ARG key_aarch64=6DC3636DAE534049C8B94623A122542AB04F24E3
 #
 # See:
 #  https://github.com/docker-library/official-images/blob/master/test/tests/java-uimanager-font/container.java
-RUN set -eux; \
-    case "$(uname -p)" in \
-        x86_64) rpm=$rpm_x64; path=$path_x64; key=$key_x64 ;; \
-        aarch64) rpm=$rpm_aarch64; path=$path_aarch64; key=$key_aarch64 ;; \
-        *) echo >&2 "Unsupported architecture $(uname -p)."; exit 1 ;; \
-    esac; \
-    \
-    curl -O $path/$rpm \
-    && export GNUPGHOME="$(mktemp -d)" \
-    && gpg --batch --keyserver ha.pool.sks-keyservers.net --recv-keys $key \
-    && gpg --armor --export $key > corretto.asc \
-    && rpm --import corretto.asc \
-    && rpm -K $rpm \
-    && rpm -i $rpm \
-    && rm -r $GNUPGHOME corretto.asc $rpm \
+
+RUN set -eux \
+    && rpm --import https://yum.corretto.aws/corretto.key \
+    && curl -L -o /etc/yum.repos.d/corretto.repo https://yum.corretto.aws/corretto.repo \
+    && yum install -y java-11-amazon-corretto-devel-$version \
     && yum install -y fontconfig \
     && yum clean all
 
diff --git a/amazoncorretto_latest/Dockerfile b/amazoncorretto_latest/Dockerfile
index 9affa5d..d284130 100644
--- a/amazoncorretto_latest/Dockerfile
+++ b/amazoncorretto_latest/Dockerfile
@@ -1,15 +1,7 @@
 FROM amazonlinux:2
 
-# x86_64 args
-ARG rpm_x64=java-1.8.0-amazon-corretto-devel-1.8.0_242.b08-1.x86_64.rpm
-ARG path_x64=https://corretto.aws/downloads/resources/8.242.08.1
-ARG key_x64=6DC3636DAE534049C8B94623A122542AB04F24E3
-
-# aarch64 args
-ARG rpm_aarch64=java-1.8.0-amazon-corretto-devel-1.8.0_242.b08-1.aarch64.rpm
-ARG path_aarch64=https://corretto.aws/downloads/resources/8.242.08.1
-ARG key_aarch64=6DC3636DAE534049C8B94623A122542AB04F24E3
-# In addition to installing the RPM, we also install
+ARG version=1.8.0_242.b08-1
+# In addition to installing the Amazon corretto, we also install
 # fontconfig. The folks who manage the docker hub's
 # official image library have found that font management
 # is a common usecase, and painpoint, and have
@@ -19,21 +11,10 @@ ARG key_aarch64=6DC3636DAE534049C8B94623A122542AB04F24E3
 #  https://github.com/docker-library/official-images/blob/master/test/tests/java-uimanager-font/container.java
 
 
-RUN set -eux; \
-    case "$(uname -p)" in \
-        x86_64) rpm=$rpm_x64; path=$path_x64; key=$key_x64 ;; \
-        aarch64) rpm=$rpm_aarch64; path=$path_aarch64; key=$key_aarch64 ;; \
-        *) echo >&2 "Unsupported architecture $(uname -p)."; exit 1 ;; \
-    esac; \
-    \
-    curl -O $path/$rpm \
-    && export GNUPGHOME="$(mktemp -d)" \
-    && gpg --batch --keyserver ha.pool.sks-keyservers.net --recv-keys $key \
-    && gpg --armor --export $key > corretto.asc \
-    && rpm --import corretto.asc \
-    && rpm -K $rpm \
-    && rpm -i $rpm \
-    && rm -r $GNUPGHOME corretto.asc $rpm \
+RUN set -eux \
+    && rpm --import https://yum.corretto.aws/corretto.key \
+    && curl -L -o /etc/yum.repos.d/corretto.repo https://yum.corretto.aws/corretto.repo \
+    && yum install -y java-1.8.0-amazon-corretto-devel-$version \
     && yum install -y fontconfig \
     && yum clean all
 
```

</details>

yosifkit:
We usually recommend using the fingerprint version and pulling from a keyserver as previously done, since it provides more transparency into the build (rather than a remote `key` file that can change at any time).

alvdavi:
We agree that having a fingerprint and pulling from the keyserver is more transparent. However, there were occasions when the keyservers were unresponsive during the build, failing to create the container. Can you suggest a keyserver that is dependable? We would be willing to go back if needed

yosifkit:
> Can you suggest a keyserver that is dependable?

Not exactly. Doing `gpg --recv-key` on many images builds, we haven't found any that are much more reliable than any others (`ha.pool.sks-keyservers.net` is "better" but since it is a DNS round robin of many servers, it can fail too). Often maintainers will just loop over a couple keyservers in their Dockerfile. What we do for the official-images build servers is the following (and for travis builds on the ones we maintain):

> Our solution to the problem of flaky keyservers is to use a single keyserver in our Dockerfiles (we use `ha.pool.sks-keyservers.net` because it has less downsides than pinning to a single server), and instead hijack DNS for our builds to point that at a local instance of [github.com/tianon/pgp-happy-eyeballs](https://github.com/tianon/pgp-happy-eyeballs), which is a project that takes incoming HTTP GET requests for keys and in turn forwards them out to several keyservers at once, returning back to the client the fastest successful response (which has been working very successfully for us since early 2018).
>
> \-https://github.com/docker-library/faq/#openpgp--gnupg-keys-and-verification

alvdavi:
Thanks for the answer, we are taking it into consideration. However, what if we were to include a fingerprint in the dockerfile and check that the file we download from our own servers matches that fingerprint. That way, we still use our endpoint, which we consider reliable, but we still have the same guarantee that the key has not been changed.

yosifkit:
That sounds reasonable. It keeps the transparency and still verifies the gpg fingerprint of the key. You'll want to also make sure that it doesn't have any unexpected extra keys.

tianon:
I took a pass at this, in case it's helpful:

```dockerfile
FROM amazonlinux:2

ARG version=1.8.0_242.b08-1
# In addition to installing the Amazon corretto, we also install
# fontconfig. The folks who manage the docker hub's
# official image library have found that font management
# is a common usecase, and painpoint, and have
# recommended that Java images include font support.
#
# See:
#  https://github.com/docker-library/official-images/blob/master/test/tests/java-uimanager-font/container.java


RUN set -eux \
    && curl -fL -o correto.key https://yum.corretto.aws/corretto.key \
    && gpg --batch --import correto.key \
    && gpg --batch --export --armor '6DC3636DAE534049C8B94623A122542AB04F24E3' > correto.key \
    && rpm --import correto.key \
    && rm -r "$GNUPGHOME" correto.key \
    && rpm --import https://yum.corretto.aws/corretto.key \
    && curl -fL -o /etc/yum.repos.d/corretto.repo https://yum.corretto.aws/corretto.repo \
    && grep -q '^gpgcheck=1' /etc/yum.repos.d/corretto.repo \
    && yum install -y java-1.8.0-amazon-corretto-devel-$version \
    && yum install -y fontconfig \
    && yum clean all

ENV JAVA_HOME=/usr/lib/jvm/java-1.8.0-amazon-corretto
```

alvdavi:
Thanks, sorry for slacking on this, we will make use of that Dockerfile

alvdavi:
Closed, as we are getting a new PR with the 8.252.09.1 and 11.0.7.10.1 update that includes these changes

PR #7526, opened 2020-02-26
Add Turing, a probabilistic programming library

KDr2:
# Checklist for Review

### DIFF
--- a/library/turing
+++ b/library/turing
@@ -1,0 +1,8 @@
+Maintainers: The Turing team <https://turing.ml> (@TuringLang/turing),
+             ZHUO Qingliang <zhuo.dev@gmail.com> (@KDr2)
+GitRepo: https://github.com/TuringLang/Turing.jl.git
+GitCommit: 3588417f772290efc010b3bb8992c398c6e33529
+
+Tags: latest, 0.8.3
+Architectures: amd64, i386, arm64v8, arm32v7
+Directory: docker/latest


tianon:
I'm a little bit confused here, is Turing a separate language implemented in Julia, or is it simply a library for use with Julia?

How do you envision users consuming this image (`FROM turing:xxx`, interactive only, etc)?

The tags here reference version 0.8.3, but the `Dockerfile` will install latest (which appears to be 0.9.0 at this time of writing), which doesn't comply with https://github.com/docker-library/official-images#repeatability.  I'm also curious what the reason for setting `CMD` again to the same value as the base image does?  Is `WORKDIR` actually relevant, or just preference?

I was able to build Julia 1.3 + Turing 0.8.3 with the following:

```dockerfile
FROM julia:1.3-buster

RUN set -eux; \
	julia -e 'using Pkg; Pkg.add(Pkg.PackageSpec(name="Turing", version="0.8.3"))'; \
	julia -e 'using Turing'; \
	mkdir -p ~/.julia/config; \
	echo 'using Turing' >> ~/.julia/config/startup.jl
```

When built/run, it drops me into the interactive Julia shell (although the shell loads and runs *significantly* slower than the standard Julia shell, both with and without `WORKDIR`, and my machine is pretty beefy so this seems strange).

Wouldn't it be more common for users who are using this to make images that are `FROM julia:xxx` and then include `Turing` in their `Project.toml` files?

KDr2:
Turing.jl is a Julia library and a probabilistic programming language (a DSL for probabilistic programming). 

Yes, it is more common for users who are using this to make images that are FROM julia:xxx and then include Turing in their Project.toml files. But sometimes someone failed to install it and asked us if there's a docker image.  This one is intended to be used interactively to try Turing.jl.

The non-repeatability is caused by a new release after this PR, I will update the dockerfiles with all the issues you mentioned fixed then update this library file.

We also maintain an unofficial image at https://hub.docker.com/r/turinglang/turing, if you think it is not proper to incorporate this image officially, feel free to close this PR. Thanks.

tianon:
Thanks for the clarification (and for being generally amenable)!  That helps me understand better.

A concern I've got is around how generic the name is, especially the fact that there is another programming language that's also called [Turing](https://en.wikipedia.org/wiki/Turing_(programming_language)), but even more that it's a general term/name used in the industry at large.

Maybe our next step here should be to wait and see whether your image at https://hub.docker.com/r/turinglang/turing gets some traction/use from your user base, and reconsider if/when it is more popular or well-used? (perhaps users will also help determine the best ways to use it, so that it can both improve what comes here and improve the eventual documentation for this image?)

KDr2:
> Maybe our next step here should be to wait and see whether your image at https://hub.docker.com/r/turinglang/turing gets some traction/use from your user base, and reconsider if/when it is more popular or well-used? (perhaps users will also help determine the best ways to use it, so that it can both improve what comes here and improve the eventual documentation for this image?)

Fair enough, thanks.

yosifkit:
Thanks for your understanding. I'll close this for now and hopefully we'll see you in the future. :rocket: Good Luck!

PR #7603, opened 2020-03-10
Add EspoCRM official image

tmachyshyn:
We would like to add the [EspoCRM](https://github.com/espocrm/docker) to the official image repository.

--------------------

# Checklist for Review

-   [x] associated with or contacted upstream?
    - https://github.com/espocrm/docker
    - service
-   [x] is it reasonably popular, or does it solve a particular use case well?
-   [x] does a [documentation](https://github.com/docker-library/docs/blob/master/README.md) PR exist? (should be reviewed and merged at roughly the same time so that we don't have an empty image page on the Hub for very long)
    - https://github.com/docker-library/docs/pull/1672
-   [x] dockerization review for best practices and cache gotchas/improvements (ala [the official review guidelines](https://github.com/docker-library/official-images/blob/master/README.md#review-guidelines))?
-   [x] 2+ dockerization review?
-   [x] existing official images have been considered as a base? (ie, if `foobar` needs Node.js, has `FROM node:...` instead of grabbing `node` via other means been considered?)
-   [x] if `FROM scratch`, tarballs only exist in a single commit within the associated history?
-   [x] passes current tests? any simple new tests that might be appropriate to add? (https://github.com/docker-library/official-images/tree/master/test)

### DIFF
--- a/library/espocrm
+++ b/library/espocrm
@@ -1,19 +1,19 @@
-# this file is generated via https://github.com/espocrm/espocrm-docker/blob/12856eaa14480e6adec6420e9a294ff1a6ab41ea/generate-stackbrew-library.sh
+# this file is generated via https://github.com/espocrm/espocrm-docker/blob/1029b7a4cbc7bce1494c97383e04fc96d9c29ae0/generate-stackbrew-library.sh
 
 Maintainers: Taras Machyshyn <docker@espocrm.com> (@tmachyshyn)
 GitRepo: https://github.com/espocrm/espocrm-docker.git
 
-Tags: apache, 8.3.1-apache, 8.3-apache, 8-apache, latest, 8.3.1, 8.3, 8
+Tags: apache, 8.3.2-apache, 8.3-apache, 8-apache, latest, 8.3.2, 8.3, 8
 Architectures: amd64, i386, arm32v7, arm64v8, s390x
-GitCommit: 12856eaa14480e6adec6420e9a294ff1a6ab41ea
+GitCommit: 1029b7a4cbc7bce1494c97383e04fc96d9c29ae0
 Directory: apache
 
-Tags: fpm, 8.3.1-fpm, 8.3-fpm, 8-fpm
+Tags: fpm, 8.3.2-fpm, 8.3-fpm, 8-fpm
 Architectures: amd64, i386, arm32v7, arm64v8, s390x
-GitCommit: 12856eaa14480e6adec6420e9a294ff1a6ab41ea
+GitCommit: 1029b7a4cbc7bce1494c97383e04fc96d9c29ae0
 Directory: fpm
 
-Tags: fpm-alpine, 8.3.1-fpm-alpine, 8.3-fpm-alpine, 8-fpm-alpine
+Tags: fpm-alpine, 8.3.2-fpm-alpine, 8.3-fpm-alpine, 8-fpm-alpine
 Architectures: amd64, i386, arm32v6, arm32v7, arm64v8, ppc64le, s390x
-GitCommit: 12856eaa14480e6adec6420e9a294ff1a6ab41ea
+GitCommit: 1029b7a4cbc7bce1494c97383e04fc96d9c29ae0
 Directory: fpm-alpine
--- a/library/espocrm
+++ b/library/espocrm
@@ -1,19 +1,19 @@
-# this file is generated via https://github.com/espocrm/espocrm-docker/blob/c75f6deaef3e685f6767a83014aba433b58e07a4/generate-stackbrew-library.sh
+# this file is generated via https://github.com/espocrm/espocrm-docker/blob/7b933d4fc7036f1a037e5c20b5e88fd2537c669f/generate-stackbrew-library.sh
 
 Maintainers: Taras Machyshyn <docker@espocrm.com> (@tmachyshyn)
 GitRepo: https://github.com/espocrm/espocrm-docker.git
 
-Tags: apache, 8.4.0-apache, 8.4-apache, 8-apache, latest, 8.4.0, 8.4, 8
+Tags: apache, 8.4.1-apache, 8.4-apache, 8-apache, latest, 8.4.1, 8.4, 8
 Architectures: amd64, i386, arm32v7, arm64v8, s390x
-GitCommit: c75f6deaef3e685f6767a83014aba433b58e07a4
+GitCommit: 7b933d4fc7036f1a037e5c20b5e88fd2537c669f
 Directory: apache
 
-Tags: fpm, 8.4.0-fpm, 8.4-fpm, 8-fpm
+Tags: fpm, 8.4.1-fpm, 8.4-fpm, 8-fpm
 Architectures: amd64, i386, arm32v7, arm64v8, s390x
-GitCommit: c75f6deaef3e685f6767a83014aba433b58e07a4
+GitCommit: 7b933d4fc7036f1a037e5c20b5e88fd2537c669f
 Directory: fpm
 
-Tags: fpm-alpine, 8.4.0-fpm-alpine, 8.4-fpm-alpine, 8-fpm-alpine
+Tags: fpm-alpine, 8.4.1-fpm-alpine, 8.4-fpm-alpine, 8-fpm-alpine
 Architectures: amd64, i386, arm32v6, arm32v7, arm64v8, ppc64le, s390x
-GitCommit: c75f6deaef3e685f6767a83014aba433b58e07a4
+GitCommit: 7b933d4fc7036f1a037e5c20b5e88fd2537c669f
 Directory: fpm-alpine
--- a/library/espocrm
+++ b/library/espocrm
@@ -1,19 +1,19 @@
-# this file is generated via https://github.com/espocrm/espocrm-docker/blob/541e2ead27a98bffa809f9302bf4e78262c902c6/generate-stackbrew-library.sh
+# this file is generated via https://github.com/espocrm/espocrm-docker/blob/062527a437bbabb8d5712970938d5defdc937a04/generate-stackbrew-library.sh
 
 Maintainers: Taras Machyshyn <docker@espocrm.com> (@tmachyshyn)
 GitRepo: https://github.com/espocrm/espocrm-docker.git
 
-Tags: apache, 8.2.5-apache, 8.2-apache, 8-apache, latest, 8.2.5, 8.2, 8
+Tags: apache, 8.3.0-apache, 8.3-apache, 8-apache, latest, 8.3.0, 8.3, 8
 Architectures: amd64, i386, arm32v7, arm64v8, s390x
-GitCommit: 541e2ead27a98bffa809f9302bf4e78262c902c6
+GitCommit: 062527a437bbabb8d5712970938d5defdc937a04
 Directory: apache
 
-Tags: fpm, 8.2.5-fpm, 8.2-fpm, 8-fpm
+Tags: fpm, 8.3.0-fpm, 8.3-fpm, 8-fpm
 Architectures: amd64, i386, arm32v7, arm64v8, s390x
-GitCommit: 541e2ead27a98bffa809f9302bf4e78262c902c6
+GitCommit: 062527a437bbabb8d5712970938d5defdc937a04
 Directory: fpm
 
-Tags: fpm-alpine, 8.2.5-fpm-alpine, 8.2-fpm-alpine, 8-fpm-alpine
+Tags: fpm-alpine, 8.3.0-fpm-alpine, 8.3-fpm-alpine, 8-fpm-alpine
 Architectures: amd64, i386, arm32v6, arm32v7, arm64v8, ppc64le, s390x
-GitCommit: 541e2ead27a98bffa809f9302bf4e78262c902c6
+GitCommit: 062527a437bbabb8d5712970938d5defdc937a04
 Directory: fpm-alpine
--- a/library/espocrm
+++ b/library/espocrm
@@ -1,0 +1,19 @@
+# this file is generated via https://github.com/espocrm/espocrm-docker/blob/b0b555425c5aa0b4e357a8fe431e06249f1489f5/generate-stackbrew-library.sh
+
+Maintainers: Taras Machyshyn <docker@espocrm.com> (@tmachyshyn)
+GitRepo: https://github.com/espocrm/espocrm-docker.git
+
+Tags: apache, 8.2.4-apache, 8.2-apache, 8-apache, latest, 8.2.4, 8.2, 8
+Architectures: amd64, i386, arm32v7, arm64v8, s390x
+GitCommit: b0b555425c5aa0b4e357a8fe431e06249f1489f5
+Directory: apache
+
+Tags: fpm, 8.2.4-fpm, 8.2-fpm, 8-fpm
+Architectures: amd64, i386, arm32v7, arm64v8, s390x
+GitCommit: b0b555425c5aa0b4e357a8fe431e06249f1489f5
+Directory: fpm
+
+Tags: fpm-alpine, 8.2.4-fpm-alpine, 8.2-fpm-alpine, 8-fpm-alpine
+Architectures: amd64, i386, arm32v6, arm32v7, arm64v8, ppc64le, s390x
+GitCommit: b0b555425c5aa0b4e357a8fe431e06249f1489f5
+Directory: fpm-alpine
--- a/library/espocrm
+++ b/library/espocrm
@@ -1,19 +1,19 @@
-# this file is generated via https://github.com/espocrm/espocrm-docker/blob/33718c149de426626382a54bd4b4ae9911df72f2/generate-stackbrew-library.sh
+# this file is generated via https://github.com/espocrm/espocrm-docker/blob/9345bf125c1b7f0ced021c89ec12e3deecc79ac1/generate-stackbrew-library.sh
 
 Maintainers: Taras Machyshyn <docker@espocrm.com> (@tmachyshyn)
 GitRepo: https://github.com/espocrm/espocrm-docker.git
 
-Tags: apache, 8.3.3-apache, 8.3-apache, 8-apache, latest, 8.3.3, 8.3, 8
+Tags: apache, 8.3.4-apache, 8.3-apache, 8-apache, latest, 8.3.4, 8.3, 8
 Architectures: amd64, i386, arm32v7, arm64v8, s390x
-GitCommit: 33718c149de426626382a54bd4b4ae9911df72f2
+GitCommit: 9345bf125c1b7f0ced021c89ec12e3deecc79ac1
 Directory: apache
 
-Tags: fpm, 8.3.3-fpm, 8.3-fpm, 8-fpm
+Tags: fpm, 8.3.4-fpm, 8.3-fpm, 8-fpm
 Architectures: amd64, i386, arm32v7, arm64v8, s390x
-GitCommit: 33718c149de426626382a54bd4b4ae9911df72f2
+GitCommit: 9345bf125c1b7f0ced021c89ec12e3deecc79ac1
 Directory: fpm
 
-Tags: fpm-alpine, 8.3.3-fpm-alpine, 8.3-fpm-alpine, 8-fpm-alpine
+Tags: fpm-alpine, 8.3.4-fpm-alpine, 8.3-fpm-alpine, 8-fpm-alpine
 Architectures: amd64, i386, arm32v6, arm32v7, arm64v8, ppc64le, s390x
-GitCommit: 33718c149de426626382a54bd4b4ae9911df72f2
+GitCommit: 9345bf125c1b7f0ced021c89ec12e3deecc79ac1
 Directory: fpm-alpine
--- a/library/espocrm
+++ b/library/espocrm
@@ -1,19 +1,19 @@
-# this file is generated via https://github.com/espocrm/espocrm-docker/blob/9184c12c27b05014826bdae389691841b234fa15/generate-stackbrew-library.sh
+# this file is generated via https://github.com/espocrm/espocrm-docker/blob/ffa773c6600a24e4025220c741da159e4f79518f/generate-stackbrew-library.sh
 
 Maintainers: Taras Machyshyn <docker@espocrm.com> (@tmachyshyn)
 GitRepo: https://github.com/espocrm/espocrm-docker.git
 
-Tags: apache, 9.0.3-apache, 9.0-apache, 9-apache, latest, 9.0.3, 9.0, 9
+Tags: apache, 9.0.4-apache, 9.0-apache, 9-apache, latest, 9.0.4, 9.0, 9
 Architectures: amd64, i386, arm32v5, arm32v7, arm64v8
-GitCommit: 9184c12c27b05014826bdae389691841b234fa15
+GitCommit: ffa773c6600a24e4025220c741da159e4f79518f
 Directory: apache
 
-Tags: fpm, 9.0.3-fpm, 9.0-fpm, 9-fpm
+Tags: fpm, 9.0.4-fpm, 9.0-fpm, 9-fpm
 Architectures: amd64, i386, arm32v5, arm32v7, arm64v8
-GitCommit: 9184c12c27b05014826bdae389691841b234fa15
+GitCommit: ffa773c6600a24e4025220c741da159e4f79518f
 Directory: fpm
 
-Tags: fpm-alpine, 9.0.3-fpm-alpine, 9.0-fpm-alpine, 9-fpm-alpine
+Tags: fpm-alpine, 9.0.4-fpm-alpine, 9.0-fpm-alpine, 9-fpm-alpine
 Architectures: amd64, i386, arm32v6, arm32v7, arm64v8
-GitCommit: 9184c12c27b05014826bdae389691841b234fa15
+GitCommit: ffa773c6600a24e4025220c741da159e4f79518f
 Directory: fpm-alpine
--- a/library/espocrm
+++ b/library/espocrm
@@ -1,19 +1,19 @@
-# this file is generated via https://github.com/espocrm/espocrm-docker/blob/29cf2cf3bc4b865d3b5e92d4b3ca9e24f503e8c7/generate-stackbrew-library.sh
+# this file is generated via https://github.com/espocrm/espocrm-docker/blob/9184c12c27b05014826bdae389691841b234fa15/generate-stackbrew-library.sh
 
 Maintainers: Taras Machyshyn <docker@espocrm.com> (@tmachyshyn)
 GitRepo: https://github.com/espocrm/espocrm-docker.git
 
-Tags: apache, 9.0.2-apache, 9.0-apache, 9-apache, latest, 9.0.2, 9.0, 9
+Tags: apache, 9.0.3-apache, 9.0-apache, 9-apache, latest, 9.0.3, 9.0, 9
 Architectures: amd64, i386, arm32v5, arm32v7, arm64v8
-GitCommit: 29cf2cf3bc4b865d3b5e92d4b3ca9e24f503e8c7
+GitCommit: 9184c12c27b05014826bdae389691841b234fa15
 Directory: apache
 
-Tags: fpm, 9.0.2-fpm, 9.0-fpm, 9-fpm
+Tags: fpm, 9.0.3-fpm, 9.0-fpm, 9-fpm
 Architectures: amd64, i386, arm32v5, arm32v7, arm64v8
-GitCommit: 29cf2cf3bc4b865d3b5e92d4b3ca9e24f503e8c7
+GitCommit: 9184c12c27b05014826bdae389691841b234fa15
 Directory: fpm
 
-Tags: fpm-alpine, 9.0.2-fpm-alpine, 9.0-fpm-alpine, 9-fpm-alpine
+Tags: fpm-alpine, 9.0.3-fpm-alpine, 9.0-fpm-alpine, 9-fpm-alpine
 Architectures: amd64, i386, arm32v6, arm32v7, arm64v8
-GitCommit: 29cf2cf3bc4b865d3b5e92d4b3ca9e24f503e8c7
+GitCommit: 9184c12c27b05014826bdae389691841b234fa15
 Directory: fpm-alpine
--- a/library/espocrm
+++ b/library/espocrm
@@ -1,19 +1,19 @@
-# this file is generated via https://github.com/espocrm/espocrm-docker/blob/bd481f48a90373b2850a73494b63d63144c7793d/generate-stackbrew-library.sh
+# this file is generated via https://github.com/espocrm/espocrm-docker/blob/29cf2cf3bc4b865d3b5e92d4b3ca9e24f503e8c7/generate-stackbrew-library.sh
 
 Maintainers: Taras Machyshyn <docker@espocrm.com> (@tmachyshyn)
 GitRepo: https://github.com/espocrm/espocrm-docker.git
 
-Tags: apache, 9.0.1-apache, 9.0-apache, 9-apache, latest, 9.0.1, 9.0, 9
+Tags: apache, 9.0.2-apache, 9.0-apache, 9-apache, latest, 9.0.2, 9.0, 9
 Architectures: amd64, i386, arm32v5, arm32v7, arm64v8
-GitCommit: bd481f48a90373b2850a73494b63d63144c7793d
+GitCommit: 29cf2cf3bc4b865d3b5e92d4b3ca9e24f503e8c7
 Directory: apache
 
-Tags: fpm, 9.0.1-fpm, 9.0-fpm, 9-fpm
+Tags: fpm, 9.0.2-fpm, 9.0-fpm, 9-fpm
 Architectures: amd64, i386, arm32v5, arm32v7, arm64v8
-GitCommit: bd481f48a90373b2850a73494b63d63144c7793d
+GitCommit: 29cf2cf3bc4b865d3b5e92d4b3ca9e24f503e8c7
 Directory: fpm
 
-Tags: fpm-alpine, 9.0.1-fpm-alpine, 9.0-fpm-alpine, 9-fpm-alpine
+Tags: fpm-alpine, 9.0.2-fpm-alpine, 9.0-fpm-alpine, 9-fpm-alpine
 Architectures: amd64, i386, arm32v6, arm32v7, arm64v8
-GitCommit: bd481f48a90373b2850a73494b63d63144c7793d
+GitCommit: 29cf2cf3bc4b865d3b5e92d4b3ca9e24f503e8c7
 Directory: fpm-alpine
--- a/library/espocrm
+++ b/library/espocrm
@@ -1,19 +1,19 @@
-# this file is generated via https://github.com/espocrm/espocrm-docker/blob/04e2777d95b3e4a21f8d1ffca8d5b7c59f4b052c/generate-stackbrew-library.sh
+# this file is generated via https://github.com/espocrm/espocrm-docker/blob/c75f6deaef3e685f6767a83014aba433b58e07a4/generate-stackbrew-library.sh
 
 Maintainers: Taras Machyshyn <docker@espocrm.com> (@tmachyshyn)
 GitRepo: https://github.com/espocrm/espocrm-docker.git
 
-Tags: apache, 8.3.6-apache, 8.3-apache, 8-apache, latest, 8.3.6, 8.3, 8
+Tags: apache, 8.4.0-apache, 8.4-apache, 8-apache, latest, 8.4.0, 8.4, 8
 Architectures: amd64, i386, arm32v7, arm64v8, s390x
-GitCommit: 04e2777d95b3e4a21f8d1ffca8d5b7c59f4b052c
+GitCommit: c75f6deaef3e685f6767a83014aba433b58e07a4
 Directory: apache
 
-Tags: fpm, 8.3.6-fpm, 8.3-fpm, 8-fpm
+Tags: fpm, 8.4.0-fpm, 8.4-fpm, 8-fpm
 Architectures: amd64, i386, arm32v7, arm64v8, s390x
-GitCommit: 04e2777d95b3e4a21f8d1ffca8d5b7c59f4b052c
+GitCommit: c75f6deaef3e685f6767a83014aba433b58e07a4
 Directory: fpm
 
-Tags: fpm-alpine, 8.3.6-fpm-alpine, 8.3-fpm-alpine, 8-fpm-alpine
+Tags: fpm-alpine, 8.4.0-fpm-alpine, 8.4-fpm-alpine, 8-fpm-alpine
 Architectures: amd64, i386, arm32v6, arm32v7, arm64v8, ppc64le, s390x
-GitCommit: 04e2777d95b3e4a21f8d1ffca8d5b7c59f4b052c
+GitCommit: c75f6deaef3e685f6767a83014aba433b58e07a4
 Directory: fpm-alpine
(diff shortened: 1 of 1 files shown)


tmachyshyn:
@tianon please have a look. 
Thanks

tmachyshyn:
@tianon, @yosifkit, please take a look.

I would like to say that while creating the docker image we followed all the recommendations. For example:
- we created the Dockerfile easy to understand/read. 
- using SHA256 checksum in the Dockerfile while image build.
- added and tested on different architectures.
- implemented recommendation about tags.
- updating the docker image after new versions are released.

tmachyshyn:
@tianon, @yosifkit
Is there any chance to review the image?

tmachyshyn:
Is there any chance to review the image?

AlexAv:
Hi @tianon, @yosifkit, thank you for your attention, if there something wrong, please write and we will fix it.
Is there any possibility to speed up the process?

AlexAv:
@tianon, @yosifkit,  look the unofficial image has been installed 50K+ times,  it makes sense to create **official image**

Docker

tianon:
```dockerfile
    apk update; \
    apk add --no-cache --virtual .build-deps \
```

The `apk update` here is redundant -- `--no-cache` implies updating the package lists.

---

```dockerfile
        php7-gd \
        php7-exif \
```

This is definitely not right -- when you're `FROM php:xxx`, installing `php`-related packages via `apk` or `apt-get` will give you two versions of PHP (the version provided by the base image, and the version from Alpine or Debian).  What you intended instead is probably to make use of `docker-php-ext-install` (and possibly `docker-php-ext-configure`) to get `gd` and `exif` installed for the PHP provided by the base image (exactly as is done lower down in this same `Dockerfile`).

---

```dockerfile
# Install php-zmq
    cd /usr; \
    curl -fSL https://github.com/zeromq/libzmq/releases/download/v4.3.2/zeromq-4.3.2.tar.gz -o zeromq-4.3.2.tar.gz; \
    tar -xvzf zeromq-4.3.2.tar.gz; \
    rm zeromq-4.3.2.tar.gz; \
    cd /usr/zeromq-4.3.2; \
    ./configure; \
    make; \
    make install; \
    cd /usr; \
    git clone https://github.com/zeromq/php-zmq.git; \
    cd /usr/php-zmq; \
    phpize && ./configure; \
    make; \
    make install; \
```

This seems odd to me -- why are we building `zeromq` from source?  Alpine has a `libzmq` package (and the corresponding `zeromq-dev` package), and in Alpine 3.12 the version is the same 4.3.2 being installed here.

Additionally, why are we grabbing the latest `master` branch of https://github.com/zeromq/php-zmq.git ?  I see the latest release on that repository is from 2013, so I could understand if that's not quite new enough, but just pulling directly from `master` is not going to be very reliable over time, and pulling from a specific commit would be better (likely also giving https://github.com/zeromq/php-zmq/issues/203 some attention -- it hasn't had a release because they're looking for new maintainers, so it might not be the best platform to be based on right now).

---

```dockerfile
ENV DOCUMENT_ROOT /var/www/html
ENV DEFAULT_OWNER www-data
ENV DEFAULT_GROUP www-data
```

What is the intent of these variables?  As a user reading these, it feels like I can adjust these and expect the webserver configuration to adjust as well, but I don't think that's going to be the case.  It seems like perhaps a local variable is what was intended instead?  In the entrypoint, you probably don't want to use `DEFAULT_OWNER` but rather do some detection based on `id -u` / `id -g` so that users can run the image as an arbitrary user (especially since OpenStack and some Kubernetes configurations will require that ability).

--- 

```dockerfile
ENV ESPOCRM_UPGRADE_VERSION 5.8.5
ENV ESPOCRM_UPGRADE_URL https://www.espocrm.com/downloads/upgrades/EspoCRM-upgrade-5.8.5-to-5.9.3.zip
ENV ESPOCRM_UPGRADE_SHA256 a04d4fdadd877d684abd885dc7356ea2f0593e8a805851d79d4a6f68b6168700
```

This is interesting -- can you elaborate on what this is for, especially since it's only downloaded/used in the entrypoint and doesn't appear to be part of the image itself?

---

```dockerfile
VOLUME ${DOCUMENT_ROOT}
```

This and the (fairly complex) entrypoint script have me a little bit concerned/confused.  This appears to be copied from the general pattern of the WordPress official image (or from another PHP-based image which copied the WordPress pattern), wherein WordPress considers the code of itself to be mutable state that it manages (by doing automatic upgrades especially for security fixes, for example).  From what I can find, EspoCRM is *not* similar in the sense that upgrades to EspoCRM are always performed manually (https://www.espocrm.com/documentation/administration/upgrading/), and updating from the UI even appears to be highly discouraged.

Given this, I have to wonder why EspoCRM is not instead copied/installed directly to `/var/www/html` in the image, and the `VOLUME` then defined to be only the specific directory or directories where EspoCRM actually stores mutable state, thus encouraging users to adopt the appropriate container-upgrade paradigms to update their instance of EspoCRM from version to version by updating to newer versions of the container image instead of having the image include complex logic to try and maintain and old-style installation of EspoCRM within a `VOLUME` for them?

---

```dockerfile
CMD ["sh", "-c", "cron && php-fpm"]
```

I understand the intent of this (to run both Cron and PHP simultaneously), but this is not acceptable in the official images.  What I would recommend is documenting for users how to run `crond` as a completely separate process/container (there are several other PHP-based images that are good examples of what I mean, such as `nextcloud`).

---

```dockerfile
# Set timezone
RUN echo "UTC" > /etc/timezone && dpkg-reconfigure --frontend noninteractive tzdata
```

This is odd to me -- the default timezone of all images in the official images is already UTC, and we even [have an integration test to validate that](https://github.com/docker-library/official-images/tree/805f7dcdb2d00770e0748fb1e0d73af83adfbc47/test/tests/utc).

yosifkit:
Related to `VOLUME`, we've seen that many users have been fairly vocal about them existing in many images since they can't be removed, so it could make sense to not have any volume and just document the mutable folders that should be saved -- e.g. [drupal](https://github.com/docker-library/docs/tree/04b9e53e5ac8b2c5db54e3a2165f40cac41fb486/drupal#volumes).

tmachyshyn:
Thanks @tianon, @yosifkit for your reviews. We will implement your recommendations soon.

tmachyshyn:
@tianon, @yosifkit thanks again for your reviews. We made changes to your recommendations. Please check it again. Thanks

> ```dockerfile
>     apk update; \
>     apk add --no-cache --virtual .build-deps \
> ```
> 
> The `apk update` here is redundant -- `--no-cache` implies updating the package lists.

Fixed

> ```dockerfile
>         php7-gd \
>         php7-exif \
> ```
> 
> This is definitely not right -- when you're `FROM php:xxx`, installing `php`-related packages via `apk` or `apt-get` will give you two versions of PHP (the version provided by the base image, and the version from Alpine or Debian). What you intended instead is probably to make use of `docker-php-ext-install` (and possibly `docker-php-ext-configure`) to get `gd` and `exif` installed for the PHP provided by the base image (exactly as is done lower down in this same `Dockerfile`).

Fixed

> ```dockerfile
> # Install php-zmq
>     cd /usr; \
>     curl -fSL https://github.com/zeromq/libzmq/releases/download/v4.3.2/zeromq-4.3.2.tar.gz -o zeromq-4.3.2.tar.gz; \
>     tar -xvzf zeromq-4.3.2.tar.gz; \
>     rm zeromq-4.3.2.tar.gz; \
>     cd /usr/zeromq-4.3.2; \
>     ./configure; \
>     make; \
>     make install; \
>     cd /usr; \
>     git clone https://github.com/zeromq/php-zmq.git; \
>     cd /usr/php-zmq; \
>     phpize && ./configure; \
>     make; \
>     make install; \
> ```
> 
> This seems odd to me -- why are we building `zeromq` from source? Alpine has a `libzmq` package (and the corresponding `zeromq-dev` package), and in Alpine 3.12 the version is the same 4.3.2 being installed here.
> 
> Additionally, why are we grabbing the latest `master` branch of https://github.com/zeromq/php-zmq.git ? I see the latest release on that repository is from 2013, so I could understand if that's not quite new enough, but just pulling directly from `master` is not going to be very reliable over time, and pulling from a specific commit would be better (likely also giving [zeromq/php-zmq#203](https://github.com/zeromq/php-zmq/issues/203) some attention -- it hasn't had a release because they're looking for new maintainers, so it might not be the best platform to be based on right now).

The installation of php-zmq is optimized. We cannot completely abandon its use, because it's required in EspoCRM.

> ```dockerfile
> ENV DOCUMENT_ROOT /var/www/html
> ENV DEFAULT_OWNER www-data
> ENV DEFAULT_GROUP www-data
> ```
> 
> What is the intent of these variables? As a user reading these, it feels like I can adjust these and expect the webserver configuration to adjust as well, but I don't think that's going to be the case. It seems like perhaps a local variable is what was intended instead? In the entrypoint, you probably don't want to use `DEFAULT_OWNER` but rather do some detection based on `id -u` / `id -g` so that users can run the image as an arbitrary user (especially since OpenStack and some Kubernetes configurations will require that ability).

Fixed. Now, detection of user/group is moved to the entrypoint as you recommended, https://github.com/espocrm/docker/blob/master/docker-entrypoint.sh#L188. 

> ```dockerfile
> ENV ESPOCRM_UPGRADE_VERSION 5.8.5
> ENV ESPOCRM_UPGRADE_URL https://www.espocrm.com/downloads/upgrades/EspoCRM-upgrade-5.8.5-to-5.9.3.zip
> ENV ESPOCRM_UPGRADE_SHA256 a04d4fdadd877d684abd885dc7356ea2f0593e8a805851d79d4a6f68b6168700
> ```
> 
> This is interesting -- can you elaborate on what this is for, especially since it's only downloaded/used in the entrypoint and doesn't appear to be part of the image itself?

This is required for automatic upgrade EspoCRM depending on the version of EspoCRM docker image and the version of EspoCRM files / database. For example, if the version of EspoCRM docker container is 5.9.3, but the EspoCRM files / database is 5.8.3, then the entrypoint will upgrade EspoCRM files / database to the version of the docker container 5.9.3. EspoCRM has own upgrade system, but this `ESPOCRM_UPGRADE_VERSION` and `ESPOCRM_UPGRADE_URL` is required only if we need to upgrade the EspoCRM files / database to the specified version `ESPOCRM_VERSION`.
Currently, we are saving a link of this upgrade. Also, we can download this upgrade while building the docker image itself. What way would you recommend?

> ```dockerfile
> VOLUME ${DOCUMENT_ROOT}
> ```
> 
> This and the (fairly complex) entrypoint script have me a little bit concerned/confused. This appears to be copied from the general pattern of the WordPress official image (or from another PHP-based image which copied the WordPress pattern), wherein WordPress considers the code of itself to be mutable state that it manages (by doing automatic upgrades especially for security fixes, for example). From what I can find, EspoCRM is _not_ similar in the sense that upgrades to EspoCRM are always performed manually (https://www.espocrm.com/documentation/administration/upgrading/), and updating from the UI even appears to be highly discouraged.
> 
> Given this, I have to wonder why EspoCRM is not instead copied/installed directly to `/var/www/html` in the image, and the `VOLUME` then defined to be only the specific directory or directories where EspoCRM actually stores mutable state, thus encouraging users to adopt the appropriate container-upgrade paradigms to update their instance of EspoCRM from version to version by updating to newer versions of the container image instead of having the image include complex logic to try and maintain and old-style installation of EspoCRM within a `VOLUME` for them?

> Related to `VOLUME`, we've seen that many users have been fairly vocal about them existing in many images since they can't be removed, so it could make sense to not have any volume and just document the mutable folders that should be saved -- e.g. [drupal](https://github.com/docker-library/docs/tree/04b9e53e5ac8b2c5db54e3a2165f40cac41fb486/drupal#volumes).

Based on that fact that EspoCRM files / database can be automatically upgraded to the version of EspoCRM docker image, we need to mount the entire EspoCRM directory. What would you recommend this this case?

> ```dockerfile
> CMD ["sh", "-c", "cron && php-fpm"]
> ```
> 
> I understand the intent of this (to run both Cron and PHP simultaneously), but this is not acceptable in the official images. What I would recommend is documenting for users how to run `crond` as a completely separate process/container (there are several other PHP-based images that are good examples of what I mean, such as `nextcloud`).

Fixed, thanks for the example.

> ```dockerfile
> # Set timezone
> RUN echo "UTC" > /etc/timezone && dpkg-reconfigure --frontend noninteractive tzdata
> ```
> 
> This is odd to me -- the default timezone of all images in the official images is already UTC, and we even [have an integration test to validate that](https://github.com/docker-library/official-images/tree/805f7dcdb2d00770e0748fb1e0d73af83adfbc47/test/tests/utc).

Fixed

AlexAv:
We highly appreciate the feedback and recommendations you’ve previously shared with us. Since we have made all the changes you’ve mentioned, we would like to ask you to review the image again @tianon, @yosifkit.

tianon:
Thank you for your patience. :bow:

The `php-zmq` installation looks a lot better now; I'd just add that it looks like `/usr/php-zmq` is unnecessary at the end of it, and could probably be deleted. :+1:

(You could also possibly optimize a tiny bit by downloading a source archive from GitHub at that commit such as https://github.com/zeromq/php-zmq/archive/e0db82c3286da81fa8945894dd10125a528299e4.tar.gz instead of using `git clone`, but that's not strictly necessary.)

> Based on that fact that EspoCRM files / database can be automatically upgraded to the version of EspoCRM docker image, we need to mount the entire EspoCRM directory. What would you recommend this this case?

Let me explain this slightly differently to see if I can do a better job.  Right now, the upgrade behavior is not something EspoCRM itself does automatically, but rather is a function of this image based on how EspoCRM is deployed and upgraded outside containers.  What I'm proposing is that if EspoCRM itself was *not* stored in the data volume, then perhaps that upgrade bundle wouldn't even be necessary (nor a large part of the entrypoint code/logic), because the way to upgrade EspoCRM itself would instead just be recreating the container.

In other words, the current flow (oversimplified) is something like:

1. run container
    1. copies EspoCRM code to volume
    1. performs database/application initialization
    1. runs Apache/EspoCRM
1. recreate container with newer version
    1. downloads upgrade delta
    1. applies upgrade delta patching to code in volume (ignoring the already upgraded code that exists in the image)
    1. performs applicable database upgrades
    1. runs Apache/EspoCRM

Proposed flow:

1. run container
    1. performs database/application initialization
    1. runs Apache/EspoCRM
1. recreate container with newer version
    1. performs applicable database upgrades
    1. runs Apache/EspoCRM

In this simpler flow, instead of having all of `/var/www/html` be a volume, you would have that be part of the image and only specific subdirectories with content that needs to be persisted (user uploads, etc) would be explicit volumes.  This way, the image provides a more "container-native" workflow of the image being the source-of-truth for which version of the application we're running instead of our data volume (which in Wordpress's case is a consequence of how Wordpress works / expects to work).

Does that make more sense?  Do you think that workflow makes sense for EspoCRM?  If it's possible to accomplish, I think it could pretty dramatically simplify the logic in your container entrypoint script, and the way that users use/reason about your container image.

tmachyshyn:
Thanks for your reply.
We will try to implement your recommendations soon.

tmachyshyn:
@tianon thanks again for your thoughts.

The php-zmq installation is optimized due to your recommendations.

The current upgrade process is optimized and moved to EspoCRM side. The "entrypoint" script just checking the version and run the upgrade on EspoCRM side if needed.

EspoCRM has the directories for user's customizations:
-  `application/Espo/Modules` 
- `client/modules`
- `custom/Espo/Custom`
- `client/custom`

At this point I see 2 issues with the new process. For successful upgrade EspoCRM we have to do:

1. Updating files. Copy new files and delete old ones.

Upgrade process deletes files that have been removed from EspoCRM. This will work fine only if a user will not have any additional extensions of EspoCRM. Add-ons can be installed on EspoCRM. All they are stored in `application/Espo/Modules` and `client/modules` directories. By default these directories are not empty and contain internal module `Crm`. So, when only mount the volumes of `modules` directories, the files of `Crm` module might not be updated or deleted.

2. Updating database, configuration parameters via upgrade script.

In some cases, we need to run some actions with the database, configuration parameters, etc. For such purposed we can run the upgrade script which is included in the upgrade package of EspoCRM. So, when only mount the volumes of  `custom` and `modules` directories, the script will not be executed.

For example, the new process can be:
- We have espocrm-docker v6.0.0 container.
- Run espocrm-docker v6.1.0 container. In this case, our files in `Crm` module and data in the database can be out of date.

The current process implemented in "entrypoint" script with the mounted volume `/var/www/html` are:
- We have espocrm-docker v6.0.0 container
- Run espocrm-docker v6.1.0 container. Files will be checked at the start and run an upgrade 6.1.0. The files will be updated and all scripts will be executed.

It would be great if you have any suggestions for resolving these issues.
Thanks

AlexAv:
@tianon , @yosifkit  is there anything you need?

LaurentGoderre:
@tmachyshyn I opened a PR to optimize the layers and the size a bit: https://github.com/espocrm/espocrm-docker/pull/42

LaurentGoderre:
@tmachyshyn I added some comments to the docs. Would be possible to address those and then squash the commits in both PRs?

tmachyshyn:
@LaurentGoderre thanks for your recommendations. That's all fixed.

LaurentGoderre:
Would you be able to squash down to 1 commit for each repo?

tmachyshyn:
@LaurentGoderre Done for this and for documentation repositories.
If any other action is required, please let me know.

LaurentGoderre (dismissed):

tianon:
First, thank you -- genuinely. This PR has been open since 2020 and you've kept it updated through six years of EspoCRM releases without losing patience with the review process. That kind of sustained effort deserves more than a form response, and I'm sorry it's taken this long to get back to a substantive review. 🙇 The Dockerization has visibly improved across that time, and it shows. 👍

Great to see so many of the 2020 concerns resolved -- the duplicate PHP installation is gone, the explicit `VOLUME` is gone, and `CMD ["sh", "-c", "cron && php-fpm"]` is gone. 👍

Still a few things worth discussing:

---

```diff
+FROM php:8.4-fpm-alpine
+FROM php:8.4-fpm
+FROM php:8.4-apache
```

These are square-wheel tags -- when the underlying base image updates its default OS, the tag rolls forward and can break things. `php:8.4-apache` and `php:8.4-fpm` currently resolve to Debian **trixie**; `php:8.4-fpm-alpine` resolves to **Alpine 3.23**. Pinning the OS suffix means you choose when that square wheel rotates:

```dockerfile
FROM php:8.4-apache-trixie
FROM php:8.4-fpm-trixie
FROM php:8.4-fpm-alpine3.23
```

---

```diff
+Tags: apache, 9.3.4-apache, 9.3-apache, 9-apache, latest, 9.3.4, 9.3, 9
```

Similarly, the tag list doesn't include a versioned OS alias (like `9.3.4-apache-trixie` / `9.3.4-fpm-alpine3.23`). That's not a blocker, but it's worth having -- when `php:8.4-apache` eventually rolls to a new Debian release, the `9.3.4-apache` tag will silently change its base. A versioned alias gives users an escape hatch (`9.3.4-apache-trixie`, `9.3.4-fpm-alpine3.23`), if you think that's a problem worth solving.

---

```diff
+LABEL org.opencontainers.image.description="EspoCRM is an Open Source CRM. Try for Free."
```

The OCI-namespaced keys are fine, but `"Try for Free."` reads more like a marketing tagline than a technical image description, and something neutral like `"EspoCRM open source CRM application."` seems more appropriate for image metadata. Also worth noting that labels in general have [unfortunate inheritance behavior](https://github.com/docker-library/official-images/issues/3540#issuecomment-2715158488) -- child images built `FROM espocrm` inherit all of these, which is rarely what anyone wants. Dropping them entirely is the usual recommendation, but I'll leave that call to you.

---

```diff
+    apk add --no-cache --virtual .build-deps \
+        ...
+        bash \
+        $PHPIZE_DEPS \
```

Two things here. First, `bash` is in the `.build-deps` virtual group, but all three scripts are `#!/bin/bash` -- so `bash` is a runtime dependency, not a build dependency. Runtime tools like `bash` that aren't detected via dynamic linking should be installed before the virtual group (or alongside the runtime deps after `apk del`), so they aren't swept up by the cleanup step. Second -- there's no `apk del --no-network .build-deps` at the end of this `RUN` block, which means all the `-dev` packages (`postgresql-dev`, `openssl-dev`, `zeromq-dev`, `freetype-dev`, etc.) and `$PHPIZE_DEPS` are left in the final image. For the shared library runtime deps, the `php` base image itself uses `scanelf` to auto-detect which `.so` files the built extensions actually need -- the same approach would work well here and avoids the risk of manually missing one.

---

```diff
+RUN set -ex; \
```

All three `RUN` blocks that do non-trivial work use `set -ex` -- strongly recommend adding `-u` to get `set -eux`. The `-u` flag treats unset variable references as errors, which catches a surprising number of bugs (especially in the copy/extract block where `${ESPOCRM_SHA256}` and `${ESPOCRM_VERSION}` are referenced -- if either is ever accidentally undefined, `-u` makes the build fail loudly instead of silently continuing).

---

```diff
+    curl -fSL "https://www.espocrm.com/downloads/EspoCRM-9.3.4.zip" -o EspoCRM.zip; \
+    ...
+    mv "/usr/src/EspoCRM-9.3.4" /usr/src/espocrm; \
```

Just a minor note: the version string in the download URL (and extracted directory name) is hardcoded as `9.3.4` rather than referencing `${ESPOCRM_VERSION}`. Might be worth wiring up if the upstream templates allow it, so there's only one place to update on version bumps.

---

```sh
SOURCE_FILES="/usr/src/espocrm"

actionInstall() {
    ...
    cp -a "$SOURCE_FILES/." /var/www/html/
```

```sh
CUSTOM_RESOURCE_LIST=(
    "/var/www/html/data"
    "/var/www/html/custom"
    "/var/www/html/client/custom"
    "/var/www/html/install/config.php"
)
```

Picking up the thread from the 2020 discussion: back then you explained that `application/Espo/Modules` and `client/modules` needed to be mutable because they contain both the internal `Crm` module and user-installed extensions, making a simple volume split difficult. That's a fair constraint -- but looking at the current `CUSTOM_RESOURCE_LIST`, those module directories are no longer in it. The only genuinely mutable paths now are `data/`, `custom/`, `client/custom/`, and `install/config.php`.

Given that, I want to revisit the question: does the full-webroot seed pattern still earn its keep? With only those four paths as true mutable state, it seems like the application code could live directly in `/var/www/html` in the image and only those four paths would need to be bind-mounted or managed by the user. The upgrade path (`php command.php upgrade`) would still run at startup to handle database migrations from the old schema to the new one -- that part doesn't require the files to have come from `/usr/src/espocrm` rather than the image layer.

The seed pattern is genuinely useful when an application manages and modifies its own code at runtime (as WordPress does). But if EspoCRM's mutable state is really just `data/`, `custom/`, and `client/custom/`, then copying the full application into the webroot at every container start is adding a lot of startup work and entrypoint complexity for something the image layer could just... provide directly. It would also make the upgrade story simpler for users to reason about: "update the image tag, recreate the container, EspoCRM runs its DB migrations."

Not a hard blocker, but worth considering whether the 2020 reasoning still holds given where `CUSTOM_RESOURCE_LIST` has landed. 😅

---

There is quite a bit of shell scripting beyond the `Dockerfile`s themselves -- three scripts per variant (~1300 lines total per variant, ~3900 lines across all three). This is worth stepping back and asking: is all of this logic necessary at the image level, and is the Docker packaging layer the right place to maintain it?

The distinction matters here. Images like `postgres` have long, complex entrypoints because the upstream PostgreSQL project doesn't own or maintain the Docker packaging -- so behavioral things like initialization, config templating, and locale setup have to be reimplemented in shell by whoever is doing the packaging. That's a genuine constraint. EspoCRM, on the other hand, directly maintains this Docker repository, which means you have the option to push startup logic *into EspoCRM itself* rather than reimplementing it in the packaging layer. Database migration orchestration and config-from-environment both feel like things the EspoCRM application could handle natively (or via a small dedicated CLI utility), rather than being implemented in ~350 lines of `bash` that shim into PHP via `php -r "..."` strings.

Some of the notes below are illustrative of that broader concern -- but the injection issue, the `local` exit-code masking, and the missing `exec` in the sidecar scripts are worth addressing regardless of where that broader conversation lands.

---

```sh
runUpgradeStep() {
    local result=$(php command.php upgrade -y --toVersion="$ESPOCRM_VERSION")

    if [[ "$result" == *"Error:"* ]]; then
        ...
        return 1 #false
    fi

    return 0 #true
}
```

```sh
    if ! runUpgradeStep ; then
        return
    fi
```

The upgrade path has a silent failure mode -- when `runUpgradeStep` returns non-zero, `runUpgradeProcess` swallows it with `if ! runUpgradeStep; then return; fi` (returning 0), so `actionUpgrade` continues to call `setPermissions` and then `exec`s the server. A partially-upgraded database gets a running application with no indication anything went wrong. It's also worth noting that `runUpgradeStep` downloads the upgrade package from espocrm.com at container startup -- so a restart during a network hiccup means a failed download, which triggers this same silent-failure path. This is exactly the kind of fragility that would largely disappear if the seed pattern weren't needed: if the application files were baked into the image rather than seeded and then upgraded in place, the entrypoint would only need to run database migrations at startup -- no file download, no `php command.php upgrade`, no upgrade loop, no silent partial-upgrade state. The complexity is downstream of that one architectural choice. 😬

---

```sh
settingsTestResult=$(runInstallationStep "settingsTest" \
    "dbPlatform=${ESPOCRM_DATABASE_PLATFORM}&hostName=${databaseHost}&dbName=${ESPOCRM_DATABASE_NAME}&dbUserName=${ESPOCRM_DATABASE_USER}&dbUserPass=${ESPOCRM_DATABASE_PASSWORD}" \
    true 2>&1)
```

These values are interpolated directly into query-string format without any URL encoding. If a password contains `&`, `=`, `+`, `#`, or `%`, it will silently corrupt the data being sent to the installer -- `ESPOCRM_ADMIN_PASSWORD=hunter2&user-name=evil` would add a spurious second `user-name` key. Strong passwords are likely to contain exactly these characters; the values need URL encoding before concatenation.

---

```sh
isDatabaseReady() {
    php -r "
        require_once('/var/www/html/bootstrap.php');
        \$app = new \Espo\Core\Application();
        ...
        try {
            \$helper->createPDO();
        }
        catch (Exception \$e) {
            echo false;
            exit;
        }
        echo true;
    "
}

verifyDatabaseReady() {
    for i in {1..40}
    do
        isReady=$(isDatabaseReady 2>&1)
        if [ -n "$isReady" ]; then
            return 0 #true
        fi
```

A few layers of concern here. The readiness check works by testing whether `$isReady` is non-empty -- but `echo false` in PHP outputs nothing (`false` casts to `""`, not the string `"false"`), so you need to hold PHP type-casting rules in your head to verify at a glance that this is correct. That's a smell on its own.

The deeper issue is the approach: using `$(isDatabaseReady 2>&1)` and testing for non-empty output is fail-open. Any PHP warning, deprecation notice, or output from the bootstrap/Application initialization lands in `$isReady` and signals "ready" even when the DB isn't. `catch (Exception $e)` also misses PHP 7+ `Error` objects, but even fixing that doesn't address the output-pollution risk.

Exit codes are the natural Unix mechanism here, and they fix everything at once:

```php
try {
    $helper->createPDO();
} catch (\Throwable $e) {
    exit(1);
}
// implicit exit(0) on success
```

```sh
if isDatabaseReady; then
    return 0 #true
fi
```

This is easier to read, requires no output capture, and is fail-closed: a PHP crash, uncaught error, or unexpected failure produces a non-zero exit and correctly signals "not ready" with no special handling required.

---

```sh
isCustomPath() {
    local path="$1"
    for customDir in "${CUSTOM_RESOURCE_LIST[@]}"; do
        if [[ "$path" == "$customDir"* ]]; then
            return 0 # true
        fi
    done
    return 1 # false
}
```

```sh
    local configParamName
    local configParamValue
```

There is some dead code: `isCustomPath` is defined in these scripts but called nowhere, and `configParamName`/`configParamValue` are declared in `applyConfigEnvironments` but never used. 😅

---

```sh
declare -A DEFAULTS=(
    ['ESPOCRM_DATABASE_PASSWORD']='password'
    ['ESPOCRM_ADMIN_PASSWORD']='password'
    ...
)
```

Both the database password and the admin password default to `password`. Any instance started without explicitly setting these gets a well-known credential in production. At minimum printing a loud startup warning when the default passwords are in use -- something like the pattern the MySQL image uses -- would make this much less likely to bite users. 😬

---

```sh
compareVersion() {
    local version1="$1"
    local version2="$2"
    local operator="$3"

    echo $(php -r "echo version_compare('$version1', '$version2', '$operator');")
}
```

A couple of things here. First, `echo $(...)` is a useless construction -- `echo` captures the subshell output, strips trailing newlines, and adds one back, which is almost always equivalent to just running the command directly (and `echo $(...)` without quotes additionally does word-splitting and globbing on the result). Just call `php -r "..."` directly.

More importantly: shell variables are interpolated directly into PHP code throughout these scripts. `$version1`, `$version2`, `$operator`, and later `$name`, `$value`, `$key1`, `$key2`, `$prefix` are all substituted raw into `php -r "..."` strings. An attacker who can influence an `ESPOCRM_CONFIG_*` environment variable (which flows through `applyConfigEnvironments` -> `saveConfigValue` -> `normalizeConfigParamName`/`isValueQuoted`/`saveConfigParam`) can inject arbitrary PHP code. For example, if `ESPOCRM_CONFIG_SITE_URL` is set to a value containing `'); system("..."); //`, it will execute. This pattern appears in these scripts.

The fix is to pass values as PHP variables rather than letting shell strings compose PHP code -- keeping the shell-to-PHP interface clean and not letting user-supplied strings escape their intended context. 👀

---

```sh
local isInstalled=$(getConfigParamFromFile "isInstalled")
local maintenanceMode=$(getConfigParamFromFile "maintenanceMode")
local key=$(normalizeConfigParamName "$envName")
local value=$(normalizeConfigParamValue "$envValue")
local isValueQuoted=$(isValueQuoted "$value")
local isVersionGreater=$(compareVersion "$ESPOCRM_VERSION" "$installedVersion" ">")
```

`local foo="$(cmd)"` (and `local foo=$(cmd)`) is a well-known shell pitfall: `local` is itself a command that returns exit code 0 when the assignment is syntactically valid, masking the exit code of the subshell entirely. A failing `getConfigParamFromFile` or `compareVersion` call will silently appear to succeed. The fix is to split the declaration from the assignment:

```sh
local isInstalled
isInstalled=$(getConfigParamFromFile "isInstalled")
```

This appears in many places in these scripts.

---

```sh
/usr/local/bin/php /var/www/html/daemon.php

exec "$@"
```

(Same pattern in `docker-websocket.sh`.) When these scripts run in a container, `bash` is PID1 (or at least the parent PID) and `php daemon.php` is a child process. Signals from `docker stop` go to `bash`, not to the PHP process -- so `SIGTERM` may never reach the daemon. This should be `exec /usr/local/bin/php /var/www/html/daemon.php` to make the PHP process PID1 directly. (The trailing `exec "$@"` would then be unreachable and can be removed.)

---

None of this is meant as a comprehensive line-by-line critique -- most of the above is illustrative of why pushing startup behavior upstream into the application tends to produce more robust outcomes than implementing it in the packaging layer. The packaging layer is a hard place to maintain complex stateful logic: it lacks proper error handling primitives, the test surface is limited, and bugs tend to be silent.

This PR has been iterating for six years and the Dockerization has genuinely improved considerably in that time -- reducing the surface area further seems like a natural next step. 👍

tmachyshyn:
@tianon thank you very much for reviewing the code. We will update it based on your recommendations and get back to you soon.

tmachyshyn:
@tianon thanks again for your detailed comments.

For convenience, we have numbered each of your points below:

1\. `FROM php:8.4-apache-trixie`  
Done. We have implemented "versions.json" to load and store release information.

2\. `Tags`  
Done.

3\. `LABEL`  
Done.

4\. `apk add --no-cache --virtual .build-deps`  
Done.

5\. `+RUN set -ex;`  
Done.

6\. `curl -fSL "https://www.espocrm.com/downloads/EspoCRM-9.3.4.zip" -o EspoCRM.zip;`  
Done.

7\. `Full-webroot mapping, CUSTOM_RESOURCE_LIST`  
We have already implemented the ability to run upgrade scripts by copying only the files within EspoCRM itself.

We are currently working on the next major EspoCRM release, v10, which will include these changes. As a result, running "php command.php upgrade" will no longer be required in the future.

We are also working on a solution for mapping only the custom directories. However, this change requires more time, so it has not yet been implemented.

8\. `There is quite a bit of shell scripting beyond the Dockerfiles themselves -- three scripts per variant (~1300 lines total per variant, ~3900 lines across all three). This is worth stepping back and asking: is all of this logic necessary at the image level, and is the Docker packaging layer the right place to maintain it?`  

Thank you for this recommendation. We are planning to move "docker-entrypoint.sh", "docker-daemon.sh", and "docker-websocket.sh" into EspoCRM itself. This is planned for the next major release, v10.

9\. `Function runUpgradeStep()`  
This function will be removed once the changes mentioned in item 7 are implemented.  
For now, we have improved the current code.

10\. `These values are interpolated directly into query-string format without any URL encoding.`  
Fixed, thanks.

11\. `Functions isDatabaseReady(), verifyDatabaseReady()`  
Fixed. 

12\. `Function isCustomPath and local variables`  
Removed.

13\. `Both the database password and the admin password default to password.`  
We have added a warning when the container starts.

14\. `More importantly: shell variables are interpolated directly into PHP code throughout these scripts.`  
We have patched the code by changing how "echo" is used and how variables are passed into PHP. In the next major release, this logic will be moved into EspoCRM itself, avoiding the use of "php -r".

15\. `local foo="$(cmd)"`  
Changed.

16\. `exec "$@"`  
Fixed.


----

Just to summarize, the following changes are planned for the next major release:

- Moving `docker-entrypoint.sh`, `docker-daemon.sh`, and `docker-websocket.sh` into EspoCRM itself.
- Implementing a new upgrade flow so that running `php command.php upgrade` will no longer be required.
- Removing `runUpgradeStep()` as part of the upgrade flow changes.
- Refactoring the current Docker-related PHP execution logic to avoid using `php -r`.
- Adding support for mapping only custom directories.

tianon:
Genuinely impressive follow-through on almost everything from the last round -- the `FROM` tags are OS-pinned, the versioned OS aliases are in the tag list, `bash` is correctly split out of `.build-deps` on Alpine (with `apk del --no-network` properly in place), `set -eux` is everywhere, URL encoding is wired into the installer steps, `isDatabaseReady` uses exit codes directly with `\Throwable`, the `local foo="$(cmd)"` masking is fixed throughout, PHP code no longer has shell variables interpolated into it, the default-password warning is there, and `compareVersion` drops the `echo $(...)` wrapper. That's a substantial amount of careful work. 👍

One item from the last round was listed as "Fixed" in the response but only half-landed:

---

```diff
+/usr/local/bin/php /var/www/html/daemon.php
```

```diff
+/usr/local/bin/php /var/www/html/websocket.php
```

These still need `exec` -- all six calls across all three variants. Removing the trailing `exec "$@"` that followed each call was correct (it was unreachable since the PHP process runs synchronously and blocks), but the other half of the fix -- adding `exec` to the invocation itself -- didn't land:

```sh
exec /usr/local/bin/php /var/www/html/daemon.php
```

(Same for `websocket.php`.)

---

On the v10 roadmap -- moving the scripts into EspoCRM itself, the new upgrade flow, dropping `runUpgradeStep`, refactoring away from `php -r`, mapping only custom directories -- all of that sounds great and none of it is standing in the way here. Those items were flagged as things worth improving over time, not as merge blockers, and the roadmap is welcome context.

The `exec` fix is the one thing that can't wait for v10, because it affects every user of the image from day one. Everything else looks good.

tmachyshyn:
@tianon thanks for your comments. The fix has been applied.
We really appreciate your effort and feedback.
We are already working on v10, so I hope the pull request can be closed soon.

Have a nice weekend.

tianon (dismissed):
overall I think this looks great; thank you again for your patience and your persistence :bow: :heart:

a couple notes; the first is blocking but minor, and the second is just a suggestion that doesn't really matter much at all:

- if you look at the CI output, there are some failing tests that are legit -- it looks like there's some initialization behavior in the entrypoint that needs to be gated on the command being run (I can point to some examples of doing that, if you need them)

- `awk '{print $2}' /proc/mounts | grep -qxE "/var/www/html"` could be implemented in pure `awk` (without the pipe to `grep`):

  ```bash
  awk '$2 == "/var/www/html" { found = 1; exit 0 } END { if (!found) { exit 1 } }' /proc/mounts
  ```

  and if you need it to accept any arbitrary path to check for:

  ```bash
  awk '$2 == ENVIRON["MOUNT"] { found = 1; exit 0 } END { if (!found) { exit 1 } }' /proc/mounts
  ```

  (this pattern appears at least one other time throughout the scripts)

I'm also making this an explicit approval because I think we're there, just one more minor thing to button back up 👍

tmachyshyn:
@tianon thank you very much for the review.

> CI output issues

It would be very helpful if you could point me to some examples of how to handle that.


> awk

Fixed, thanks

tianon:
WordPress is a good PHP-specific example: https://github.com/docker-library/wordpress/blob/abc2d453bf04d94816dc2fef3038e7a08a45dac0/docker-entrypoint.sh#L4 (although you'd probably only want/need the first half of that)

PostgreSQL has a decent example: https://github.com/docker-library/postgres/blob/4f9ced003ba58a854656ba150d146243d27ae3ac/docker-entrypoint.sh#L337

Cassandra is probably on the simpler end, but such a simple approach is less applicable to you given that you're PHP-based: https://github.com/docker-library/cassandra/blob/7b9da8f4f58f48af79feb0fc2f7a1e2c639965f1/docker-entrypoint.sh#L43

```bash
if [[ "${1-}" == apache2* ]] || [ "${1-}" = 'php-fpm' ]; then
```

tmachyshyn:
@tianon thanks for the examples. 
We’ll try to include this in the next release. I’ll keep you posted. 
Have a great weekend!

tmachyshyn:
@tianon thank you for your help.
The fixes have been added in the latest release. 

If you have any other recommendations, please let me know.

tianon (dismissed):

yosifkit:
- blocking: these two layers should be combined into one (applicable to all three variants). Otherwise, the second layer is modifying files added in an earlier layer and increasing the size of the image. It should save ~24 MB on the Debian images.

    ```diff
    +RUN set -eux; \
    +    curl -fSL "https://www.espocrm.com/downloads/EspoCRM-10.0.3.zip" -o EspoCRM.zip; \
    +    echo "${ESPOCRM_SHA256} *EspoCRM.zip" | sha256sum -c -; \
    +    unzip -q EspoCRM.zip -d /usr/src; \
    +    cp -a "/usr/src/EspoCRM-${ESPOCRM_VERSION}/." /var/www/html/; \
    +    rm -rf /usr/src/EspoCRM*; \
    +    mkdir -p /usr/src/espocrm; \
    +    cp -a ./client/ /usr/src/espocrm/; \
    +    cp -a ./public/ /usr/src/espocrm/
    +
    +RUN set -eux; \
    +    rm -rf ./install; \
    +    find . -type d -exec chmod 755 {} +; \
    +    find . -type f -exec chmod 644 {} +; \
    +    chown -R root:root . /usr/src/espocrm; \
    +    chown -R www-data:www-data ./data ./custom ./client/custom; \
    +    chmod +x ./bin/command
    ```

- non-blocking:
  - I recommend just including `entrypoint-utils.sh` directly in the images so that the other scripts can just `source entrypoint-utils.sh`. This might simplify maintenance and will cut down on the duplication in the generated diffs on pull requests to official-images. Alternatively, just making one of the other scripts source-able (so that you get the functions, but it doesn't start a daemon or anything) could work too. For example, the `postgres` images have a source-able entrypoint along with [`docker-ensure-initdb.sh`](https://github.com/docker-library/postgres/blob/62a714f93cc32220de46fd12235c9d509e3b1ad6/docker-ensure-initdb.sh) that sources the main entrypoint script to use its functions to do startup a little different.

  - The `posix` php module is already installed by default (checking `php -m`), so it doesn't need to be installed with `docker-php-ext-*` scripts.

tmachyshyn:
@yosifkit thank you for your comments and for taking the time to review and investigate the project code.

All of your recommendations have been implemented: https://github.com/espocrm/espocrm-docker/commits/master/. They will be included in the next release.

If any additional changes are required, please let me know.

yosifkit (approved):

PR #7611, opened 2020-03-11
Add Hitch

ThijsFeryn:
# Context

I'm a [Varnish Software](https://www.varnish-software.com/) employee and I help @gquintard maintain the official [Varnish Docker image](https://hub.docker.com/_/varnish).

[Hitch](https://hitch-tls.org/) is a very powerful TLS proxy. It is Varnish Software's preferred way to terminate TLS, because it is extremely stable and lightweight. Hitch *only* does TLS termination and doesn't even speak HTTP.

In this day and age TLS is everywhere, and the fact that [Varnish Cache](https://varnish-cache.org/) doesn't provide native TLS support makes a very Hitch relevant service.

[The Hitch image](https://github.com/varnish/docker-hitch) was built using the same methodology as the official [Varnish Docker image](https://github.com/varnish/docker-varnish) that was accepted by you.

At this point we don't provide publicly available packages for Hitch. We might do this in the near future. Meanwhile, we're relying on Debian Buster's Hitch package which does offer a quite recent version.

And just like [Varnish](https://hub.docker.com/_/varnish), we consider this to be a *service*.

# Checklist for Review

### DIFF
--- a/library/hitch
+++ b/library/hitch
@@ -2,6 +2,6 @@
 Maintainers: Thijs Feryn <thijs@varni.sh> (@thijsferyn)
 GitRepo: https://github.com/varnish/docker-hitch.git	
 
-Tags: 1, 1.5, 1.5.0, 1.5.0-1 latest
+Tags: 1, 1.5, 1.5.0, 1.5.0-1, latest
 Architectures: amd64
 GitCommit: eefa62b493b7dc4370f30dd278f3f37043769148
--- a/library/hitch
+++ b/library/hitch
@@ -1,0 +1,7 @@
+# this file was generated using https://github.com/varnish/docker-hitch/blob/eefa62b493b7dc4370f30dd278f3f37043769148/populate.sh
+Maintainers: Thijs Feryn <thijs@varni.sh> (@thijsferyn)
+GitRepo: https://github.com/varnish/docker-hitch.git	
+
+Tags: 1, 1.5, 1.5.0, 1.5.0-1 latest
+Architectures: amd64
+GitCommit: eefa62b493b7dc4370f30dd278f3f37043769148


ThijsFeryn:
Docs PR created: docker-library/docs#1675

gquintard:
`failed fetching repo "hitch"
unable to find a manifest named "hitch" (in "/tmp/tmp.lVOMEhzXPf/oi/library" or as a remote URL)`

isn't very reassuring, is it our fault?

tianon:
@gquintard sorry, that bit's just a quirk of how `diff-pr.sh` works for "new-image" PRs (there's no real error there) -- I'm looking to make that more friendly in #7966 :+1:

Sorry for the delay on this; the world has been a little more crazy than usual. :sweat_smile:

Here are a few thoughts on my initial look over the Dockerization:

---

> ```dockerfile
> 	apt-get install -y --no-install-recommends openssl hitch=1.5.0-1; \
> ```

As a Debian Developer, I love to see Debian packages being used.  However, for the purposes of official images, I'd exercise caution taking this route -- packages in Debian often require/use some amount of customization or default configuration that may not be what you as upstream would want (such as the auto-created user being called `_hitch` instead of `hitch`) or may not be suited for use in Docker.  More critically though, this means that you're stuck with whatever packages are in Buster or backports, and in this case that concretely means you're at least 2 minor releases behind (Sid has 1.5.2).

~~Are there not any upstream-maintained packages available for Hitch (similar to what Varnish has) such that the distribution packages are preferred?~~

> At this point we don't provide publicly available packages for Hitch. We might do this in the near future. Meanwhile, we're relying on Debian Buster's Hitch package which does offer a quite recent version.

(To be clear, this isn't a blocker -- think of it more as a warning of issues that might be in your future taking this route. Ultimately the choice is up to you whether this older version is "recent enough" to be considered useful. :smile: :heart:)

---

> ```dockerfile
> CMD hitch --config=/etc/hitch/hitch.conf --frontend="[$FRONTEND_HOST]:$FRONTEND_PORT" --backend="[$BACKEND_HOST]:$BACKEND_PORT" $PROXY_PROTOCOL
> ```

Using the shell syntax here means that there are potentially cases where `/bin/sh` will stick around; see https://github.com/docker-library/official-images/pull/7662#issuecomment-602856540 for more on this and how it was fixed in `varnish` :smile: (I think the note there about "generic" environment variables names apply here to some extent, but not quite to the level of `$SIZE` :sweat_smile:).

Also, with so many flags being specified by default (and controlled via environment variables), this becomes very difficult for users to add additional command-line flags, especially if they want their additional flags to interact with these default settings in some way.

---

> ```dockerfile
> COPY example.com /etc/hitch/certs
> ```

Having this example certificate pair pre-baked into the image feels a little strange IMO -- what're the details of that certificate?  How does it get regenerated?  What's it valid for, how long, etc?  Is it signed by any other party?  (These are all questions that are hard to answer from just an opaque certificate.)

Does having this certificate pre-baked in the image help users get started using Hitch faster?  Does it hamper folks who are already set up and ready to go and just want to get Hitch running?  (For example, do they have to disable or delete this certificate before using Hitch?  Could it cause any ill/unexpected effect if they don't do so?)

---

> ```ini
> user = "_hitch"
> ```

If users accidentally omit this bit from their configuration files, will Hitch run as root instead, or does it have some kind of protection against that?

Is it worth considering using `USER _hitch` or similar instead to enforce this at a higher level?  (Usually filesystem permissions and/or "privileged" ports are the reasons setting `USER` becomes a hassle.)

Did you want to create a dedicated `hitch` user separate from the one auto-created by the Debian package? (Think about whether that might be something you'll want to consider doing later too, because it's *much* easier on you to make that sort of user-breaking change now than after we merge this, *especially* if it ends up getting embedded in user's configuration files too.)

---

All in all, I think one of my biggest concerns here is the overall simplicity of it -- I frankly don't see a ton of benefit of this over a user doing the bare-minimum `FROM debian` + `RUN apt-get install hitch` themselves (minus a tiny amount of sugar) beyond it being a pre-built image, which IMO isn't a very compelling sell for maintaining it here.  Thoughts?

gquintard:
this can be closed in favor of https://github.com/docker-library/official-images/pull/8674

ThijsFeryn:
I'll close this one, but please have a look at #8674, which takes the criticism of @tianon into account.

PR #7659, opened 2020-03-20
Add new wrlx-image for Wind River Linux images

robertlinux:
Signed-off-by: Robert Yang <liezhi.yang@windriver.com>

### DIFF
--- a/library/wrlx-image
+++ b/library/wrlx-image
@@ -1,0 +1,7 @@
+Maintainers: Robert Yang <liezhi.yang@windriver.com> (@robertlinux)
+GitRepo: https://github.com/Wind-River/wrlx-docker
+
+Tags: 10.18.0011, latest
+Architectures: amd64
+GitCommit: 9485253219c9b9203691e3b38ad0e34d38a22218
+amd64-Directory: 10.18.0011/x86_64/


loot-king:
👍

tianon:
Hi!  Thanks for your interest in contributing, and for your patience.

There are some issues with our ability to consume this Dockerization as-is, but I'd prefer to start by discussing more high level items before you spend too much time on technical changes.

I'm not familiar with Wind River Linux, so I spent a little time trying to research it, and the information I was able to find was pretty sparse (so I guess more links about what makes it unique would be good).  Additionally, what I was able to find only confused me more -- there were some mentions of Yocto, which is more of a framework for building your own distribution than a proper distribution itself, and really wouldn't make much sense as a Docker base image (although is used successfully by many organizations as a toolkit to *create* minimal base images).

If what I'm reading about Wind River Linux is accurate, it seems that it's similar, and as such I'm honestly kind of confused what this image contains, and how a user of Wind River Linux would be expected to consume it?

More broadly/importantly, I have some concerns about the popularity of Wind River Linux -- we have to have a minimal "popularity" requirement to keep our limited available time focused on things that will benefit a larger set of users.  The normal metrics we check to try and gauge popularity are things like GitHub stars (while not perfect, they do serve as *a* metric of user traffic and interest which we can combine with other sources to get a better sense -- users do tend to star repositories that they use or enjoy).  Looking over https://github.com/Wind-River, I don't see any repositories with more than a handful of stars, and from what I've found elsewhere it seems like Wind River Linux may be a bit of a niche offering?

To be clear, I don't mean to be discouraging -- I think creating Docker images for Wind River Linux is a great idea and something that absolutely should continue, but I'm merely questioning whether *this* is the right place to maintain them right now.  Perhaps you should start with a public repo on Docker Hub (something like `windriverlinux/windriverlinux`?) and use that to gauge both your own ability to maintain such an image long-term, and the interest from users in using it?

robertlinux:
Hi Tianon,

Thanks for detailed reply and I'm sorry for the late response since I didn't get an email notification. And please see my reply inline.

> Hi! Thanks for your interest in contributing, and for your patience.
> 
> There are some issues with our ability to consume this Dockerization as-is, but I'd prefer to start by discussing more high level items before you spend too much time on technical changes.
> 
> I'm not familiar with Wind River Linux, so I spent a little time trying to research it, and the information I was able to find was pretty sparse (so I guess more links about what makes it unique would be good). Additionally, what I was able to find only confused me more -- there were some mentions of Yocto, which is more of a framework for building your own distribution than a proper distribution itself, and really wouldn't make much sense as a Docker base image (although is used successfully by many organizations as a toolkit to _create_ minimal base images).
> 

"Wind River Linux is one of the leading commercially supported Linux distributions for the edge. It has been around for over 10 years, and is used by 100's of customers in all types of industries including aerospace and defense, telecom, and industrial. Attached is the deck we use for new customers. You can also get more information about Wind River and Wind River Linux at https://www.windriver.com and https://www.windriver.com/products/linux

> If what I'm reading about Wind River Linux is accurate, it seems that it's similar, and as such I'm honestly kind of confused what this image contains, and how a user of Wind River Linux would be expected to consume it?

The image is built from Wind River Linux, not Wind River Linux itself. It is a light weight image, By default, the size is 11.3M and mainly contains glibc, busybox and opkg. Image size is key in embedded and edge applications. It can be expanded by package manager opkg, which gives the user the ultimate flexibility to create a minimally sized base container image with only the packages they need. There are a lot of packages such as mysql, lighttpd and
apache2 in the packages feeds:

https://distro.windriver.com/release/wrlinux-lts18-base/public-feeds/container-qemux86-64/core2-64/
https://distro.windriver.com/release/wrlinux-lts18-base/public-feeds/container-qemux86-64/all/
https://distro.windriver.com/release/wrlinux-lts18-base/public-feeds/container-qemux86-64/qemux86_64/

It is already on docker hub, you can try it if you like:

$ docker pull windriver/wrlinux-image-glibc-small-lts18-base

We don't only expect Wind River Linux users to use the image, but everyone who wants a light weight glibc image can use it. The size would be smaller than other glibc images if user builds their images based on this one, this is because Wind River Linux is original aimed for embedded linux where size matters, so its binary packages have a more detailed split, and the packages dependencies are well handled (handled automatically by package manager opkg).

> 
> More broadly/importantly, I have some concerns about the popularity of Wind River Linux -- we have to have a minimal "popularity" requirement to keep our limited available time focused on things that will benefit a larger set of users. The normal metrics we check to try and gauge popularity are things like GitHub stars (while not perfect, they do serve as _a_ metric of user traffic and interest which we can combine with other sources to get a better sense -- users do tend to star repositories that they use or enjoy). Looking over https://github.com/Wind-River, I don't see any repositories with more than a handful of stars, and from what I've found elsewhere it seems like Wind River Linux may be a bit of a niche offering?

The github repository for Wind River Linux is https://github.com/WindRiver-Labs/wrlinux-x. As you can see it has 100 stars.
Wind River Linux is used by 100s of customers across all industries, including some of the biggest companies in the world. Our customers include Intel, Ericsson, Nokia, Schneider, Bombardier and GE Healthcare and our annual revenue is over $30M and growing rapidly. Please see slide 21 of attached deck for a more expansive list of our customers.

> 
> To be clear, I don't mean to be discouraging -- I think creating Docker images for Wind River Linux is a great idea and something that absolutely should continue, but I'm merely questioning whether _this_ is the right place to maintain them right now. Perhaps you should start with a public repo on Docker Hub (something like `windriverlinux/windriverlinux`?) and use that to gauge both your own ability to maintain such an image long-term, and the interest from users in using it?

Wind River is recognized as the market leader in embedded Linux (https://www.bloomberg.com/press-releases/2019-03-12/wind-river-ranked-global-leading-provider-of-embedded-operating-systems).  Wind River is also one of the top two contributors to the Yocto project. Wind River Linux has existed for over 10 years. We release new versions yearly and they are supported for 5 years. We fix 1000's of CVEs and customer defects every year.  We run >100K tests on our product before release and have the capacity to release weekly.  So we absolutely have the capability to maintain such an image. We also have a wide customer base, many of who are moving to CNCF and are looking to have a edge optimized container base image from a known supplier such as Wind River.

robertlinux:
@tianon kindly ping.

tianon:
Thanks for your patience and your explanations, that's helpful! :pray: :heart:

I think I'm satisfied with your description, but we're definitely going to have to make some changes together to get this in, if it's still something you're interested in. :smile: :bow:

For starters, I don't think `library/wrlx-image` is the file you want to create -- I would suggest something like `library/windriverlinux` but even `library/wrlx` would be acceptable if you believe it to be suitably unique to your solution (searching Google for "wrlx linux" definitely does a decent job at pointing me to Wind River Linux).  Ultimately, this filename determines the name of the image that users will run (`docker run ... wrlx:xxx ...` or `FROM wrlx:xxx` so IMO `windriver` or `windriverlinux` or even `wind-river` or `wind-river-linux` looks a little better and gives you more name recognition, but any are OK).  The fact that it's a Docker image is already implied by the way it's being received, so the `-image` suffix is redundant.

I'll additionally echo the same note I put over in https://github.com/docker-library/official-images/pull/11081#issuecomment-1079513975, because it's very relevant here:

> I'll try to explain this more clearly -- what I expect to see for a base image like this is a repository with a branch on it (can be the main branch if you want, more on that in a second) that has a `Dockerfile` that isn't much more than `FROM scratch` + `ADD xyz-foo-bar.tar.gz /` + `CMD ["bash"]` side-by-side with the committed tarball.  The primary challenge is that the tarball needs to not be modified by any commits in the history, so what I usually do is have source code (sometimes the `Dockerfile` too) on a `main` or `master` branch, and then have a separate `dist` branch that is only ever one commit above my `main` branch which adds the necessary tarball (with a force push that includes the new tarball in the same single commit for every image update).  For supporting multiple architectures, I usually rename that `dist` branch after the architecture, such as `dist-amd64`.
>
> If there's anything I can do to help set that up, please let me know!

tianon:

yosifkit:

PR #7740, opened 2020-04-06
Add mountebank image

mijailr:
Hey docker community, I will like to maintain the official image for [Mountebank](https://github.com/bbyars/mountebank)

Mountebank is is the first open source tool to provide cross-platform, multi-protocol test doubles over the wire.

This is the docs PR docker-library/docs#1693

# Checklist for Review

### DIFF
--- a/library/mountebank
+++ b/library/mountebank
@@ -1,0 +1,18 @@
+Maintainers: Mijail Rondon <mijail@riseup.net> (@MijailR)
+GitRepo: https://github.com/bbyars/mountebank.git
+
+Tags: latest, 2, 2.2.0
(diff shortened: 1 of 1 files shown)


tianon:
Unfortunately having the `Dockerfile` in the root of the upstream repository (while useful for local development) doesn't work well as the target for official-images. Since the `Dockerfile` uses `COPY . ...`, it essentially means the entire upstream source code is now part of the Docker build context, and thus becomes part of the reviewed artifacts. (see `diff-pr.sh` for how we generate our diff; it is basically taking all docker build contexts from the starting `library/foo` file and diffing them to the new contexts from the PR).

yosifkit:

PR #7748, opened 2020-04-06
bump Kong to 2.0.3

hutchic:


hutchic:
I guess I'll find an alternative way of achieving this

yosifkit:
Unfortunately our build systems cannot handle variables in `COPY` or `FROM` statements.

We also recommend against hiding build steps inside scripts since it makes the Dockerfile overly opaque (https://github.com/docker-library/official-images/tree/ee670022653b800c8e74451d1b9c3098b5f38902#clarity).

I'd also recommend to be careful with the layers; in these new images, the kong artifact is downloaded in a separate layer than it is installed, so it can't be deleted to save space in the resulting image.

hutchic:
We're trying to use the same Dockerfile for our community edition as we do with our (private) enterprise edition which is why we relocated some of the logic to a build script. Though it increases the resulting image size the benefits to us might outweigh that con unless there's an alternative approach available.

The variable based `FROM` intention was to let us use our official Dockerfile for testing build assets in a 1:1 way representative of their installation environments.

PR #7749, opened 2020-04-07
Add Apache Flink Stateful Functions

tzulitai:
Associated docs PR: docker-library/docs#1694.

# Checklist for Review


tzulitai:
Comments for some of the checklist points:

- **associated with or contacted upstream?**
I'm a committer / PMC member in the Apache Flink community, the maintainers of this project.
There are also various [upstream discussion threads](http://apache-flink-mailing-list-archive.1008284.n3.nabble.com/DISCUSS-Creating-a-new-repo-to-host-Stateful-Functions-Dockerfiles-td39342.html) about publishing official Docker Hub images for the project.

- **does it fit into one of the common categories?**
This would probably fall under the _base distribution_ category. Building Docker images is the [recommended deployment mode](https://ci.apache.org/projects/flink/flink-statefun-docs-master/deployment-and-operations/packaging.html#images) for Flink Stateful Functions applications, and this image would serve as a base for that.

- **is it reasonably popular, or does it solve a particular use case well?**
Please see https://flink.apache.org/stateful-functions.html for an overview of what use cases the project applies to.

- **existing official images have been considered as a base?**
We're using the `flink` official image as base.

loot-king:
👍

patricklucas:
Hi @tianon / @yosifkit, just wondering if we could get some feedback on this. As @tzulitai mention, this is an extension of the relatively popular Flink image (page 3 of the official images list these days!), but with a different deployment model that lends itself to being a separate image.

We're looking into alternative ways to build and host this ourselves, but we think it's a good candidate to coexist with Flink as an official image.

Thanks!

tianon:
Sorry for the delay, and thank you for your patience with us. :pray:

I have to admit, I'm struggling here with the idea of this being a separate image, especially since it seems to essentially be a framework within Flink, officially supported by the Flink project, and even shares the same official webpage (https://flink.apache.org/).  Can you share some additional detail on why you believe it's a good candidate for a separate repository?

tzulitai:
@tianon Thank you for getting back to this PR! We're still very excited to have this as an official image.

We believe that this should be a separate repository for the following reasons:

- Stateful Functions (StateFun) is a framework with a runtime built on top of Apache Flink, and has a completely different deployment / configuration / programming model. Moreover, they target very different use cases; StateFun is meant as a framework for typical applications development, while Flink is targeted for data pipelines and analytics. Ideally, StateFun users would not be exposed to any Flink concerns or interfaces. I say "ideally" here because there are still some rough edges that are being worked on, but the main point is that they lend themselves to be separate systems.
- StateFun is maintained by the Flink community as a separate sub-project, with its own versioning scheme and release schedules separate from Flink. Trying to manage both versioning schemes and image variants with same set of tags could potentially start to get confusing for users.
- As for the website, there is actually on-going discussion for separating the content to have a standalone webpage for Stateful Functions. The two projects are currently sharing the same main webpage due to historical reasons; StateFun was a project that was donated to the Flink community early this year, and at the time to smoothen the integration it was decided that both projects would live within the same website (download links, documentation, tutorials etc.). Later on, given the vast differences between the frameworks and user confusions, it gradually became clear that content separation would be beneficial. Once we separate the website, all contents of both websites will still be maintained by the Flink community.

I hope that clarifies some things for you!

PR #7758, opened 2020-04-08
Adding myself as co-maintainer to docker-haskell

hvr:
as per https://github.com/haskell/docker-haskell/issues/5#issuecomment-551191030

### DIFF
--- a/library/haskell
+++ b/library/haskell
@@ -1,5 +1,6 @@
 Maintainers: Peter Salvatore <peter@psftw.com> (@psftw),
              Christopher Biscardi <chris@christopherbiscardi.com> (@ChristopherBiscardi)
+             Herbert Valerio Riedel <hvr@gnu.org> (@hvr)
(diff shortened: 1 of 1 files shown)


library/haskell
```diff
@@ -1,5 +1,6 @@
 Maintainers: Peter Salvatore <peter@psftw.com> (@psftw),
              Christopher Biscardi <chris@christopherbiscardi.com> (@ChristopherBiscardi)
```
tianon, line 2: These values are comma-separated, so this line needs a comma: :sweat_smile:

```suggestion
             Christopher Biscardi <chris@christopherbiscardi.com> (@ChristopherBiscardi),
```

(Also, welcome! :tada:)

cc @psftw @ChristopherBiscardi (FYI)

hvr: yikes... fixed :-)

yosifkit:
LGTM

PR #7791, opened 2020-04-14
consul 1.6.5

i0rek:


library/consul
```diff
@@ -6,10 +6,10 @@ GitRepo: https://github.com/hashicorp/docker-consul.git
 GitCommit: 20657edd8a3d66ee155b1d050a7a9400e8ab3bdd 
 Directory: 0.X
 
-Tags: 1.6.4, 1.6
+Tags: 1.6.5, 1.6
 Architectures: amd64, arm32v6, arm64v8, i386
 GitRepo: https://github.com/hashicorp/docker-consul.git
-GitCommit: 59bdf0ae8863e3166aadc81e18f8781a52a07136
+GitCommit: 5fdce521f44a1cd56d3f5fb92e5732f2ae516d3e
```
tianon, line 12: In order to pull this commit from a branch other than `master`, we'll need to inform `bashbrew` how/where to fetch it from (GitHub doesn't support fetching arbitrary commits, unfortunately):

```suggestion
GitFetch: refs/heads/consul-1.6.5
GitCommit: 5fdce521f44a1cd56d3f5fb92e5732f2ae516d3e
```

library/consul
```diff
@@ -6,10 +6,10 @@ GitRepo: https://github.com/hashicorp/docker-consul.git
...
 Directory: 0.X
 
 Tags: 1.5.3, 1.5
```
tianon, line 15: Any particular reason to keep all these older versions listed here?  Removing tags here will remove them from the "Supported" section on the Hub readme (and will prevent us from spending cycles rebuilding them on the official build servers), but the tags will still be available to users who want them.  (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)

PR #7814, opened 2020-04-16
Update Nextcloud image

J0WI:
Downgraded MariaDB due https://github.com/nextcloud/server/pull/19328/commits/f36ba8bb011bcb869cc41f1788ea0b7dbaa3f5d6
Manually stripped `mips64le` for now.

### DIFF
--- a/library/nextcloud
+++ b/library/nextcloud
@@ -5,60 +5,105 @@ GitRepo: https://github.com/nextcloud/docker.git
 
 Tags: 16.0.9-apache, 16.0-apache, 16-apache, 16.0.9, 16.0, 16
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, ppc64le, s390x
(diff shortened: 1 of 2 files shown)


tianon:
If we merge this instead of #7812, the bot will just keep trying (and ideally we're going to fix the issues with that PHP image build or disable the architecture for PHP), so I think we're better off merging that one, right?  That or updating the `nextcloud` generate script to exclude `mips64le` explicitly until it's successful on PHP.

J0WI:
> ideally we're going to fix the issues with that PHP image build or disable the architecture for PHP

+1

How does @docker-library-bot handle this?

> I think we're better off merging that one, right?

Well, we still want to push the [recent changes](https://github.com/docker-library/docs/pull/1698) to the library.

tianon:
Merged in #7812 (although there's still issues with the test even after e445d29 that should be addressed separately). :+1:

PR #7830, opened 2020-04-20
bashbrew:go:cmd-cat: remove extra v in Sprintf

SantiagoTorres:
The panic Sprintf for cmd-cat.go has an extra 'v' argument, which
triggers a failure when running go test. Remove the extra 'v' argument.


tianon:
Hah, good catch!  I think the intent there was to print out the entry object itself that caused the validation issue, but this is a good fix in the meantime. :+1: :sweat_smile:

PR #7875, opened 2020-04-24
update chronograf image

hoorayimhelping:


hoorayimhelping:
made a mistake on my end, will re-open after I resolve it.

PR #7888, opened 2020-04-27
Adjust test harness to run many tests successfully on Windows

tianon:
I've tested this on `python`, `julia`, `openjdk`, and `golang` successfully.  The `mongo` tests require more adjustments to work on Windows, so I've excluded them for now.

### DIFF
--- a/test/clean.sh
+++ b/test/clean.sh
@@ -1,4 +1,12 @@
-#!/bin/bash
-set -e
+#!/usr/bin/env bash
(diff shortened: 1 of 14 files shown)


tianon:
(Many of these I was testing from Linux pointed at a Windows Docker via `ssh://` or other means of connecting to a Docker host, but all of them have been tested directly on a Windows host via Git Bash with "real" symlinks enabled.)

PR #7945, opened 2020-05-05
Consul 1.7.3

i0rek:


tianon:
Actions is confused thanks to what I've now fixed in 3e21779c20adcd12a8e94cc9ab163aae3f403983 -- any chance you could rebase to include that?

Also, something to consider to make your life easier WRT https://github.com/docker-library/official-images/pull/7791 and friends -- you might want to consider just maintaining each version in a separate directory (or branch, if you'd rather) so you don't have to do complicated commit dancing every time they change. :sweat_smile:

i0rek:
@tianon thanks for helping out! I will take a look at this tomorrow!

i0rek:
I created a new PR based on the new master. I will discuss internally if we can get rid of the old versions.

PR #7946, opened 2020-05-06
caddy: Add s390x and ppc64le architectures

hairyhenderson:
From https://github.com/caddyserver/caddy-docker/pull/81

Note that we've been able to test the build and the binaries seem to work OK, but I haven't had a (good) way to test the images out yet. I presume they'll work _just fine_ 😉


tianon:
Nice, let's find out together! :smile: :+1:

PR #7953, opened 2020-05-06
arangodb 3.6.3.1

KVS85:
Updated ArangoDB 3.6 to 3.6.3.1 (3.6.3 with Hotfix 1).

### DIFF
--- a/library/arangodb
+++ b/library/arangodb
@@ -12,6 +12,6 @@
 3.4.9: https://github.com/arangodb/arangodb-docker@4aa774dd973b77f958fbe7a90285ca2452ca22aa alpine/3.4.9
 3.5: https://github.com/arangodb/arangodb-docker@8fa0c5d9b74ca36cf3e7ed6e8df8954c87cf04a6 alpine/3.5.4
 3.5.4: https://github.com/arangodb/arangodb-docker@8fa0c5d9b74ca36cf3e7ed6e8df8954c87cf04a6 alpine/3.5.4
-3.6: https://github.com/arangodb/arangodb-docker@ef37ecea1c60befdf513eb2cad81d01b447dc28a alpine/3.6.3
-3.6.3: https://github.com/arangodb/arangodb-docker@ef37ecea1c60befdf513eb2cad81d01b447dc28a alpine/3.6.3
-latest: https://github.com/arangodb/arangodb-docker@ef37ecea1c60befdf513eb2cad81d01b447dc28a alpine/3.6.3
+3.6: https://github.com/arangodb/arangodb-docker@8f7d55580407838958ae8e7677de8c9803280e18 alpine/3.6.3
+3.6.3: https://github.com/arangodb/arangodb-docker@8f7d55580407838958ae8e7677de8c9803280e18 alpine/3.6.3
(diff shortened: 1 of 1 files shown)


KVS85:
@yosifkit can you please help with checks results. There were no changes in older versions from our side and for 3.2 and 3.3 we got
```
gpg: directory '/root/.gnupg' created
gpg: keybox '/root/.gnupg/pubring.kbx' created
gpg: keyserver receive failed: Cannot assign requested address
Removing intermediate container a55758d30e45
The command '/bin/sh -c gpg --keyserver hkps://hkps.pool.sks-keyservers.net --recv-keys CD8CB0F1E0AD5B52E93F41E7EA93F5E56E751E9B' returned a non-zero code: 2
```
while for 3.4, 3.5 and 3.6 we got
```
npm ERR! code EUIDLOOKUP
npm ERR! lifecycle could not get uid/gid
npm ERR! lifecycle [ 'nobody', 0 ]
npm ERR! lifecycle 
npm ERR! 
npm ERR! Failed to look up the user/group for running scripts.
npm ERR! 
npm ERR! Try again with a different --user or --group settings, or
npm ERR! run with --unsafe-perm to execute scripts as root.
```

tianon:
For the first issue, you're unfortunately running into the issues described at https://github.com/docker-library/faq#openpgp--gnupg-keys-and-verification, but with the additional hurdle that you've switched to the `hkps` pool (which is a very small set of servers -- see the "hkps" column in https://sks-keyservers.net/status/), for which we're not able to apply our man-in-the-middle reliability shim (due to the transport encryption).  What I'd suggest is swapping that for any of the non-`hkps` pools in https://sks-keyservers.net/overview-of-pools.php such that our transparent proxy can auto-distribute the load, or better yet to evaluate https://keys.openpgp.org/about and see about getting the key email address verified there (since their service is much more reliable and would allow you to keep using `hkps`).

Regarding that `npm` error, it _looks_ like a case of bitrot, since the closet issues I can find are things like https://github.com/npm/npm/issues/20861, https://github.com/npm/uid-number/issues/3, https://github.com/npm/uid-number/issues/7 -- all implying the underlying issue should be resolved with updated environments/dependencies, so perhaps there's a dependency somewhere in those older versions that's too old or too specific?

One "solution" you could consider is simply removing those tags from this file, if they're no longer actively supported revisions (I use "supported" in the sense that if we have a release numbered say 1.2.3, and there's an issue found, it's considered "supported" if there will be a 1.2.4 released to fix the issue).  To be clear, removing tags here will remove them from the "Supported" section on the Hub readme (and will prevent us from spending cycles rebuilding them on the official build servers), but the tags will still be available to users who want them.  (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)

Additionally, in the future, feel free to reach out to us sooner in issues like this (we can restart the jobs directly), or just do `git commit --amend` and force push back to your PR branch and it will re-trigger the CI (instead of closing and opening a new PR, which causes churn in all our inboxes :sweat_smile:). :+1: :heart:

KVS85:
@tianon thank you!

I will address these issues in separate PR with more ArangoDB versions.

PR #7971, opened 2020-05-08
Haskell GHC 8.10.1 and Debian variants.

psftw:
The Cabal test gets an update since "install"
is now the default "new-install". We also drop
8.6.5 per "two latest minor releases" policy.


tianon:
Nice!  LGTM

PR #7972, opened 2020-05-08
Updates for 20.0.0.5 release of websphere-liberty

crpotter:
Updates for the release of version 20.0.0.5 for websphere-liberty.

PR #7973, opened 2020-05-08
20.0.0.5 open liberty

crpotter:
Updates for the 20.0.0.5 release of open-liberty.


arthurdm:
@crpotter - this PR should remove the WebSphere Liberty updates, since they are done in a separate PR.

tianon:
It also looks like the commit for #7972 snuck into this one (which is fine if you want to go that route -- we've suggested previously sending one PR for both images since they're so similar and co-maintained, but it's up to you). :smile:

tianon:
Closing in favor of https://github.com/docker-library/official-images/pull/7989

PR #7974, opened 2020-05-08
Add some new jvm versions to jruby

cpuguy83:
Adds support for:

- jre11
- jdk11
- jdk14
- jdk15

New tags for jdk/jre 8 as well.

### DIFF
--- a/library/jruby
+++ b/library/jruby
@@ -1,20 +1,31 @@
 Maintainers: Brian Goff <cpuguy83@gmail.com> (@cpuguy83)
 GitRepo: https://github.com/cpuguy83/docker-jruby.git
 GitFetch: refs/heads/master
-GitCommit: 2701333ca7d7bf910be70fb103bdaadcdd874979
+GitCommit: 669ad8f91df152ae83deef766569e0b1f5ec6b2f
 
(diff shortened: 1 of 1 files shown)


library/jruby
```diff
@@ -1,20 +1,31 @@
...
+GitCommit: 669ad8f91df152ae83deef766569e0b1f5ec6b2f
 
-Tags: latest, 9, 9.2, 9.2.11, 9.2-jre, 9.2.11-jre, 9.2.11.1, 9.2.11.1-jre
+Tags: latest, 9, 9.2, 9.2.11, 9.2-jre, 9.2-jre8, 9.2.11-jre, 9.2.11-jre8, 9.2.11.1, 9.2.11.1-jre, 9.2.11.1-jre8
 Architectures: amd64
-Directory: 9000/jre
+Directory: 9000/jre8
 
-Tags: 9-jdk, 9.2-jdk, 9.2.11-jdk, 9.2.11.1-jdk
+Tags: 9-jdk, 9.2-jdk, 9.2.11-jdk, 9.2.11-jdk8, 9.2.11.1-jdk, 9.2.11.1-jdk8
```
tianon, line 10: Did you want this one to have more aliases like `9.2-jdk8` matching some of the others? (maybe that's the only one that seems to be missing)

cpuguy83: good catch, updated.

PR #7975, opened 2020-05-08
OpenWrt: add initial support

aparcar:
This commit is a follow up on a previous PR #6268. While multiple
architectures are supported, initally only amd64 is added for testing.

Signed-off-by: Paul Spooren <mail@aparcar.org>


tianon:
Just FYI, that test failure is CVE-2019-5021, so probably worth fixing elsewhere if it's not isolated to the way this image is built. :grimacing:

tianon:
There also appear to be a lot of files under `/lib/modules` that are probably ripe for image size reduction. :smile:

aparcar:
Thanks @tianon for the quick response! Regarding the CVE, as most routers don't offer anything but network access after the first installation we can't setup a password. However I changed that for Docker.

How can I remove the modules? I remember running a command like `rm` on the *wrong* architecture would cause failures, are all imags created *natively*?

aparcar:
> if FROM scratch, tarballs only exist in a single commit within the associated history?

What kind of history is needed? No force pushes I guess?

yosifkit:
> > if FROM scratch, tarballs only exist in a single commit within the associated history?
> 
> What kind of history is needed? No force pushes I guess?

Most likely it would be exclusively force pushes to the branches that have the tar files, so that the tarball is never a changed file in git history. For example, see the `dist-*` branches in https://github.com/debuerreotype/docker-debian-artifacts.

> How can I remove the modules?

They would need to be removed before the tar file is committed to git. For example, this Dockerfile wouldn't save space in the resulting docker image (since previous docker layers already take space):

```dockerfile
FROM scratch
ADD same-base.tar.gz /
RUN rm /lib/modules/*
```

> I remember running a command like rm on the wrong architecture would cause failures, are all imags created natively?

Yes they are built on architecture appropriate hardware, no qemu.

Amacorp:
Thank you, It was great!!!

tianon:
I think instead of copying/maintaining a full copy of the tarball's `/etc/inittab` and `/etc/shadow` files (which then have the possibility of drifting from the distribution copy), I'd recommend using `RUN` to modify them, as in:

```dockerfile
FROM scratch
ADD openwrt-x86-64-rootfs.tar.gz /
RUN set -eux; \
# make sure "/sbin/init" as a command will give us a shell when run interactively
	echo 'console::askfirst:/usr/libexec/login.sh' >> /etc/inittab; \
# adjust/remove empty "root" password to account for CVE-2019-5021
	passwd -l root
CMD ["ash", "--login"]
```

Also, I'm not sure it makes sense for the *default* command to be `/sbin/init`, since most users likely don't actually need `init`, which is why my suggestion above includes a switch to just `ash --login`, which is what `/etc/inittab` ends up running when `/sbin/init` is invoked and the user presses `[enter]` to activate the console.

Otherwise this rootfs tarball seems pretty good for a first pass -- size improvements like culling `/lib/module` from the generated rootfs can be incremental improvements later.

It might be a good idea to add a simple `opkg` test here (similar to https://github.com/docker-library/official-images/tree/8c95adca7690747a702dfaa43f0b0c20f14be79e/test/tests/debian-apt-get) to ensure that works / stays working over time, but that's up to your discretion.

Do you already have a PR for https://github.com/docker-library/docs in-progress? :smile:

aparcar:
I'm happy to look into that and continue the work. My main issue right now with the docker images is that `ctl c` quits the running container and not only the running program within OpenWrt. Do you have an idea how this could be fixed?

westurner:
Could the ENTRYPOINT be a shell script that traps SIGINT?

https://www.shellscript.sh/trap.html

Does anything else wneed to be done to solve for zombie PIDs and handles?
This says docker uses `tini` when `--init` is specified; which should be documented if that's necessary:
https://stackoverflow.com/questions/49162358/docker-init-zombies-why-does-it-matter

aparcar:
The idea using `ash --login` is actually related. It instantly fires up a login shell and automatically traps SIGINT. To have OpenWrt usable it usually requires `ubusd` and some extra folders. I added a minimal `run.sh` script:

```shell
mkdir -p /var/run/
mkdir -p /var/lock/
ubusd &
ash --login
```

tianon:
Sorry for the delay :pray:

I'm really confused here -- what values of "usable" are we talking about?  For example, does `opkg` not work without `ubusd` running?

In my view, the most interesting thing about OpenWrt is that it's a very minimal distribution explicitly focused on size that also happens to have a decent collection of pre-built packages available for easy install / use (very similar to Alpine Linux, in many regards).

What specific things are not possible without `ubusd` running that would actually be interesting/useful/necessary to run in a container?  From what I can tell, it's mostly for "system" services that don't actually make any sense for the container to run, so it's probably mostly fine to not have it running in the general case?  (It certainly won't be running when users create their own images based off this one with `FROM openwrt:xxx` and use `RUN opkg ...`, for example.)

FWIW, the "container-native" way to handle this would be one container running `ubusd` and other containers sharing that via shared volume, shared networking, etc, but I realize that's likely a bit complicated to implement (and that `ubusd` was likely never designed with that usage mode in mind).

westurner:
IIRC, in order to run ansible within the container (for testing OpenWRT
configs in a container before deploying to a device running an OpenWRT
firmware (all of which expect ubusd to be running)).

Newer builds of gpg-agent spawn their own GPG agent if necessary (IIRC, if
the socket doesn't yet exist). Could ubusd be auto-spawned similarly? Or,
at the very least, could the entrypoint.sh print how to start the container
with ubusd running?

Supervisord and s6 work well in containers; though I agree that's not
maximal containerisation.

tianon:
Honestly that all seems like downstream concerns -- there are use cases for having a simple, unadorned, uncomplex `openwrt` image which provides just a basic OpenWrt rootfs ready to build more complex things on top of (such as a test environment for running Ansible scripts, etc but also just for users who want a simple, small image that includes a good package manager for downloading/packaging their own software).

So what I would recommend here is that you go with the smallest possible rootfs you can still call "OpenWrt" (likely a shell + appropriately configured opkg?), set the default command to a basic shell, and let further enhancement happen downstream so that the official base is something everyone interested in OpenWrt can use and expand off of without needing to worry about the details of how certain services need to run (unless they actually need those services, which they should discover pretty quickly).

Anything involving supervisord, s6, or even a process that backgrounds itself before starting whatever process was requested by the user will not be acceptable at this level (https://github.com/docker-library/official-images#review-guidelines).

To put this in a slightly different perspective, the Debian image doesn't include/default to running `systemd` (or `sysvinit`) and the Alpine image doesn't default to running `openrc`.  The Debian image is the bare minimum that can still be called Debian (essential packages like `bash` + `apt`), and users who want to do crazy things like run `systemd` in a container can easily build on top of it and install `systemd` just as easily as users who want to do smaller things like just install `nginx` and run it directly.

westurner:
So the instructions for running init and dropping into a shell on top of
the image could be in the README.

tianon:

yosifkit:

PR #5644, opened 2019-03-30
update EulerOS-2.3.1806 and EulerOS-2.3.1803

liusdu:
Signed-off-by: Liu Hua <sdu.liu@huawei.com>

PR #7980, opened 2020-05-11
add openeuler base image for amd64, arm64v8

openeulerHW:
add openeuler base image for amd64, arm64v8
The root filesystem tar package is larger than 100M, we had to use git-lfs to upload it to github, completely pull of dist-amd64 and dist-arm64 branch also need git-lfs.

### DIFF
--- a/library/openeuler
+++ b/library/openeuler
@@ -1,0 +1,15 @@
+# see https://openeuler.org/en/releases.html
+Maintainers: caihaomin <caihaomin@huawei.com> (@openeulerHW)
+GitRepo: https://github.com/openeulerHW/openeuler-image.git
+GitCommit: 712590bf3fee6c362da4fb0a47af760faf603d5c
+
+Tags: 20.03-LTS, 20.03, 20, latest
+Architectures: amd64, arm64v8
+# https://github.com/openeulerHW/openeuler-image/tree/dist-amd64
+amd64-GitFetch: refs/heads/dist-amd64
+amd64-GitCommit: 26c33f81afeebe4ff0b007ef2b0561589c77930c
+amd64-Directory: build
+# https://github.com/openeulerHW/openeuler-image/tree/dist-arm64
+arm64v8-GitFetch: refs/heads/dist-arm64
+arm64v8-GitCommit: 3ede232c35033864091a543321089ea2f8ae03cb
+arm64v8-Directory: build


tianon:
Replying to https://github.com/docker-library/official-images/issues/7982 here (since the PR with the issue is the appropriate place to have a discussion about it :smile:):

> I'm not sure if it is suitable to ask the question through filing a issue here, the guideline doesn't seem to metion the point. 
> As the pull request https://github.com/docker-library/official-images/pull/7980 shows, our image have a rootfs package larger than 100M. Since github does not allow push such a big file directly, we use git-lfs to push it and thus causing the file cannot be pulled completely using git (it need to be pulled also using git-lfs). It seem to be the reason why this pull request failed the automatic check.
> So is it means any official image should not including a package larger than 100M?

As you've discovered, we unfortunately don't yet support Git LFS (https://github.com/docker-library/bashbrew/issues/10).

I'm very curious *why* the rootfs is over 100MB though -- isn't EulerOS an RPM-based distribution?  What's included in that tarball which makes it so large?  Surely there's room for improvement on the size (packages that don't make sense / aren't necessary in containers, kernel-related files, etc)?

openeulerHW:
> Replying to #7982 here (since the PR with the issue is the appropriate place to have a discussion about it 😄):
> 
> > I'm not sure if it is suitable to ask the question through filing a issue here, the guideline doesn't seem to metion the point.
> > As the pull request #7980 shows, our image have a rootfs package larger than 100M. Since github does not allow push such a big file directly, we use git-lfs to push it and thus causing the file cannot be pulled completely using git (it need to be pulled also using git-lfs). It seem to be the reason why this pull request failed the automatic check.
> > So is it means any official image should not including a package larger than 100M?
> 
> As you've discovered, we unfortunately don't yet support Git LFS ([docker-library/bashbrew#10](https://github.com/docker-library/bashbrew/issues/10)).
> 
> I'm very curious _why_ the rootfs is over 100MB though -- isn't EulerOS an RPM-based distribution? What's included in that tarball which makes it so large? Surely there's room for improvement on the size (packages that don't make sense / aren't necessary in containers, kernel-related files, etc)?

Got it, We will try to reduce the size of base image at next release.

tianon:
On the next PR, you should also spend some time explaining what makes openEuler different from EulerOS (perhaps what they have in common as well), or link to some relevant documentation about the project that might help explain what makes it unique.

As additional context, the EulerOS image is already not being well-maintained (https://github.com/docker-library/official-images/pull/5644#issuecomment-565200222), and I'd hate to see a repeat of that.

martin-g:
[openEuler](https://www.openeuler.org/en/) is an open-source Linux distribution with contributors from many organizations. It is governed by OpenAtom Foundation - similar to Linux Foundation but for Chinese OSS.
Some people say that it is based on RHEL but this is not really true. It just uses the RPM package management (DNF/YUM).
I am not sure how much technical details about openEuler I should share here, so I will just add a link to the latest [technical white paper](https://www.openeuler.org/whitepaper/en/openEuler%2023.09%20Technical%20White%20Paper.pdf).

[EulerOS](https://developer.huaweicloud.com/intl/en-us/euleros/euleros-introduction.html) is a commercial distro by Huawei. It is based on openEuler but adds proprietary software that optimizes it for Huawei's ARM64 hardware ([Kunpeng](https://www.hisilicon.com/en/products/Kunpeng/Huawei-Kunpeng/Huawei-Kunpeng-920) and [Ascend](https://e.huawei.com/en/products/computing/ascend))

Disclaimer: I work for Huawei in the past few years!

I am pretty excited to work on and maintain the official Docker image for openEuler! 
Unless something bad happens to me I plan to be around in the next several years!

PR #8003, opened 2020-05-13
Update gcc

tianon:
Changes:

- https://github.com/docker-library/gcc/commit/7f1fc5e: Fix architecture exclusion
- https://github.com/docker-library/gcc/commit/febea33: Remove mips64le, per https://github.com/docker-library/gcc/issues/67


tianon:
Arch changes only, merging.

PR #8006, opened 2020-05-14
Add the Consul 1.8.0-beta1 image

mkeeler:
This adds the Consul 1.8.0-beta1 build.


mkeeler:
@tianon I went looking on [docker hub](https://hub.docker.com/_/consul?tab=tags) as well as attempting a `docker pull consul:1.8.0-beta1` and it looks like only the armv6 image is there and not the other 3 architectures. Did something go wrong in the build process?

PR #2578, opened 2017-01-24
[nats] Adding Windows variants

kozlovic:
This PR is following the work done by @sixeyed
on this [PR](https://github.com/nats-io/nats-docker/pull/5)

PR #8010, opened 2020-05-14
[nats-server] Release v2.1.7

kozlovic:
Details can be found [here](https://github.com/nats-io/nats-server/releases/tag/v2.1.7)

### DIFF
--- a/library/nats
+++ b/library/nats
@@ -3,31 +3,31 @@ Maintainers: Derek Collison <derek@synadia.com> (@derekcollison),
              Waldemar Salinas <wally@synadia.com> (@wallyqs),
              Jaime Piña <jaime@synadia.com> (@variadico)
 GitRepo: https://github.com/nats-io/nats-docker.git
-GitCommit: c8c9f5e84405f71d3300d42879de67c3f32f4b98
+GitCommit: 8f8045b5c6c6f17a28b5bc12f0f18eb6380f435f
 
-Tags: 2.1.6-alpine3.11, 2.1-alpine3.11, 2-alpine3.11, alpine3.11, 2.1.6-alpine, 2.1-alpine, 2-alpine, alpine
+Tags: 2.1.7-alpine3.11, 2.1-alpine3.11, 2-alpine3.11, alpine3.11, 2.1.7-alpine, 2.1-alpine, 2-alpine, alpine
 Architectures: amd64, arm32v6, arm32v7, arm64v8
-Directory: 2.1.6/alpine3.11
+Directory: 2.1.7/alpine3.11
 
-Tags: 2.1.6-scratch, 2.1-scratch, 2-scratch, scratch, 2.1.6-linux, 2.1-linux, 2-linux, linux
-SharedTags: 2.1.6, 2.1, 2, latest
+Tags: 2.1.7-scratch, 2.1-scratch, 2-scratch, scratch, 2.1.7-linux, 2.1-linux, 2-linux, linux
+SharedTags: 2.1.7, 2.1, 2, latest
 Architectures: amd64, arm32v6, arm32v7, arm64v8
-Directory: 2.1.6/scratch
+Directory: 2.1.7/scratch
 
-Tags: 2.1.6-windowsservercore-1809, 2.1-windowsservercore-1809, 2-windowsservercore-1809, windowsservercore-1809
-SharedTags: 2.1.6-windowsservercore, 2.1-windowsservercore, 2-windowsservercore, windowsservercore
+Tags: 2.1.7-windowsservercore-1809, 2.1-windowsservercore-1809, 2-windowsservercore-1809, windowsservercore-1809
+SharedTags: 2.1.7-windowsservercore, 2.1-windowsservercore, 2-windowsservercore, windowsservercore
 Architectures: windows-amd64
-Directory: 2.1.6/windowsservercore-1809
+Directory: 2.1.7/windowsservercore-1809
 Constraints: windowsservercore-1809
 
-Tags: 2.1.6-nanoserver-1809, 2.1-nanoserver-1809, 2-nanoserver-1809, nanoserver-1809
-SharedTags: 2.1.6-nanoserver, 2.1-nanoserver, 2-nanoserver, nanoserver, 2.1.6, 2.1, 2, latest
+Tags: 2.1.7-nanoserver-1809, 2.1-nanoserver-1809, 2-nanoserver-1809, nanoserver-1809
+SharedTags: 2.1.7-nanoserver, 2.1-nanoserver, 2-nanoserver, nanoserver, 2.1.7, 2.1, 2, latest
 Architectures: windows-amd64
-Directory: 2.1.6/nanoserver-1809
+Directory: 2.1.7/nanoserver-1809
 Constraints: nanoserver-1809, windowsservercore-1809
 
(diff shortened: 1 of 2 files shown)


kozlovic:
Doc: https://github.com/docker-library/docs/pull/1718

kozlovic:
@tianon PR failed, but not sure what that means: https://github.com/docker-library/official-images/pull/8010/checks?check_run_id=675630393#step:3:31

yosifkit:
We're looking into it. :+1:

tianon:
The inter-dependency graph for `nats` is unique enough that you managed to find an edge case bug in `bashbrew`! :smile:
(fixed via https://github.com/docker-library/bashbrew/commit/7c78a8a385a7adf9f10ec8541ab423e5b8b8ff61, and I've retriggered the builds here which look like they're going successfully now :+1:)

kozlovic:
@tianon Well, I get some failures now:
```
testing nats:2.1.7-windowsservercore-1809
nats-server: unrecognized command: "cmd"
	'override-cmd' [1/1]...failed
```
And all our Dockerfile have this at the end (except the alpine one for some reason, probably why it works):
```
COPY nats-server.conf C:\\nats-server.conf

EXPOSE 4222 8222 6222
ENTRYPOINT ["C:\\nats-server.exe"]
CMD ["--config", "nats-server.conf"]
```
Is that not the proper way? It used to work in the past.

kozlovic:
Actually the scratch has the same content and it passes, so it seem to be an issue for Windows images only?

tianon:
Aha, you're finding all the edge cases today. :sweat_smile:

This hearkens back to what we discussed in https://github.com/docker-library/official-images/pull/2578#issuecomment-274889851, but up until very recently we didn't have a way to run the tests on Windows-based images.

For now, I think we should update `test/config.sh` under `globalExcludeTests+=(` to include the following, probably right after the `# windows!` block of configuration (so, right before `# clearlinux ...`):

```bash
	# https://github.com/docker-library/official-images/pull/2578#issuecomment-274889851
	[nats:nanoserver_override-cmd]=1
	[nats:windowsservercore_override-cmd]=1
```

... and that should make the tests pass. :crossed_fingers:

kozlovic:
> Aha, you're finding all the edge cases today. 

Sorry :-)

kozlovic:
@tianon I missed the fact that you were asking me to make that change in test/config.sh. Should I make this change as part of this PR?

kozlovic:
@tianon I made the change, let's see if that works.

kozlovic:
@tianon Nope, that did not work, or I did not do it right.

tianon:
Doh, it's using the tests from master instead of from this PR, which is a little more complicated to fix.  I think we'll merge this as-is for now and I'll look to fix separately.  Thanks for your patience!

FYI, we merged security updates to PHP and Debian yesterday. Build queues are still quite full, so this may take a while to be fully built.

kozlovic:
@tianon Thanks! I will monitor to see when images are avail. As always, thanks for your help.

kozlovic:
@tianon @yosifkit I have been made aware that the image for ARM64v8 is not available for the latest release v2.1.7 (https://github.com/nats-io/nats-server/issues/1540). I now realize that this image has been missing since after v2.1.0 when we did a reorg of nats-docker with use of alpine, etc..
The manifest for v2.1.0 shows the arm64:
```
docker run --rm -ti mplatform/mquery nats:2.1.0
Image: nats:2.1.0
 * Manifest List: Yes
 * Supported platforms:
   - linux/amd64
   - linux/arm/v6
   - linux/arm/v7
   - linux/arm64
   - windows/amd64:10.0.17763.802
```
But any version after that does not:
```
docker run --rm -ti mplatform/mquery nats:2.1.2
Image: nats:2.1.2
 * Manifest List: Yes
 * Supported platforms:
   - linux/amd64
   - linux/arm/v6
   - linux/arm/v7
   - windows/amd64:10.0.17763.973
   - windows/amd64:10.0.17134.1246
```
with current v2.1.7 also missing it:
```
docker run --rm -ti mplatform/mquery nats:2.1.7
Image: nats:2.1.7
 * Manifest List: Yes
 * Supported platforms:
   - linux/amd64
   - linux/arm/v6
   - linux/arm/v7
   - windows/amd64:10.0.17763.1339
```
Coincidentally, we also miss it with the nats-streaming image for v0.18.0 (the latest) after with switch to similar layout than nats-server:
```
docker run --rm -ti mplatform/mquery nats-streaming:0.18.0
Image: nats-streaming:0.18.0
 * Manifest List: Yes
 * Supported platforms:
   - linux/amd64
   - linux/arm/v6
   - linux/arm/v7
   - windows/amd64:10.0.17763.1339
```
the previous has it:
```
docker run --rm -ti mplatform/mquery nats-streaming:0.17.0
Image: nats-streaming:0.17.0
 * Manifest List: Yes
 * Supported platforms:
   - linux/amd64
   - linux/arm/v6
   - linux/arm/v7
   - linux/arm64
   - windows/amd64:10.0.17763.1282
```
So it clearly has something to do with the new way of doing things. As I type that I notice this in Jenkins: https://doi-janky.infosiftr.net/job/multiarch/job/arm64v8/job/nats/6165/execution/node/55/log/
It says that the build failed:
```
---> 8bca330227af
Step 3/8 : RUN set -eux; 	apkArch="$(apk --print-arch)"; 	case "$apkArch" in 		arm64) natsArch='arm64' ;; 		armhf) natsArch='arm6' ;; 		armv7) natsArch='arm7' ;; 		x86_64) natsArch='amd64' ;; 		x86) natsArch='386' ;; 		*) echo >&2 "error: $apkArch is not supported!"; exit 1 ;; 	esac; 		wget -O nats-server.zip "https://github.com/nats-io/nats-server/releases/download/v${NATS_SERVER}/nats-server-v${NATS_SERVER}-linux-${natsArch}.zip"; 		apk add --no-cache ca-certificates; 	apk add --no-cache --virtual buildtmp unzip; 		unzip nats-server.zip "nats-server-v${NATS_SERVER}-linux-${natsArch}/nats-server"; 	rm nats-server.zip; 	mv "nats-server-v${NATS_SERVER}-linux-${natsArch}/nats-server" /usr/local/bin; 	rmdir "nats-server-v${NATS_SERVER}-linux-${natsArch}"; 		apk del --no-cache --no-network buildtmp
 ---> Running in 65e40346fa61
[91m+ apk --print-arch
[0m[91m+ apkArch=aarch64
+ echo 'error: aarch64 is not supported!'
error: aarch64 is not supported!
```
So I guess `apkArch` for arm64v8 is `aarch64` and not `arm64`, right?

kozlovic:
Yet, Golang does use arm64 and seem to be ok: https://github.com/docker-library/golang/blob/master/1.14/stretch/Dockerfile#L23

kozlovic:
I see some other repos using `aarch64` or `aarch64|arm64`, so I guess this is what I should be doing..

yosifkit:
The difference is Alpine vs Debian, so you just need to match the correct one on the particular OS:
`apk --print-arch` -> `aarch64`
`dpkg --print-architecture` -> `arm64`

kozlovic:
I see, I have updated nats-docker/nats-streaming-docker with `aarch64|arm6` but I guess I should resubmit with your solution instead.

yosifkit:
Seems more correct to just use `aarch64`, since you don't have a Debian based image.

kozlovic:
Sorry, did not know at all about that. So `aarch64|arm64` will work, but I should really remove `arm64`.

PR #8011, opened 2020-05-14
Update open-liberty

arthurdm:
Some edge cases were discovered, which required a small update to our build script.


arthurdm:
@tianon - when you get a chance, can you please take a peak into this one?  small PR, but blocking some usage unfortunately.

yosifkit:
Apologies for the delay and for not mentioning this earlier. We merged security updates to PHP and Debian yesterday.  Build queues are still quite full, so this may take a while to be built.

LGTM

arthurdm:
thanks @yosifkit!

PR #8012, opened 2020-05-14
Add Corretto alpine images

alvdavi:
Added new Amazon Corretto images for Alpine

### DIFF
--- a/library/amazoncorretto
+++ b/library/amazoncorretto
@@ -1,16 +1,32 @@
-Maintainers: Amazon Corretto Team (@corretto),
-             James Guo (@jguo11),
-             Ziyi Luo (@ziyiluo),
-             David Alvarez (@alvdavi)
+Maintainers: Amazon Corretto Team <corretto-team@amazon.com> (@corretto),
+             James Guo <junguoj@amazon.com> (@jguo11),
+             Ziyi Luo <ziyiluo@amazon.com> (@ziyiluo),
+             David Alvarez <alvdavi@amazon.com> (@alvdavi)
+GitRepo: https://github.com/corretto/corretto-docker.git
+GitFetch: refs/heads/master
+GitCommit: 619edf4eceed8bc4392509cfb29a4cdaac4787d8
 
-Tags: 8, 8u252, 8-al2-full, latest
-GitRepo: https://github.com/corretto/corretto-8-docker.git
-GitFetch: refs/heads/8-al2-full
-GitCommit: a1e2c7e4fb1d00430b76963020aa2861bba7658f
+Tags: 8, 8u252, 8u252-al2, 8-al2-full,8-al2-jdk, latest
 Architectures: amd64, arm64v8
+Directory: 8/jdk/al2
 
-Tags: 11, 11.0.7, 11-al2-full
-GitRepo: https://github.com/corretto/corretto-11-docker.git
-GitFetch: refs/heads/11-al2-full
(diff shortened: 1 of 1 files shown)


tianon:
My understanding is that Alpine support is being worked on within the OpenJDK / Java community under Project Portola (https://wiki.openjdk.java.net/display/portola/Main), and that it's currently officially unsupported and pretty desperate for contributions; see https://github.com/docker-library/openjdk/pull/235#issuecomment-424599754:

> Yeah, the alpine port is not in a supported release, since it's not in the mainline code base, so it's only available as early access builds of OpenJDK Project Portola http://openjdk.java.net/projects/portola.  
> 
> It works, but at this stage it basically needs community contributors to help maintain it and get it into mainline - see https://twitter.com/MikaelVidstedt/status/984920426655526912 for details. Interested contributors should drop by portola-dev to discuss specifics.

So honestly, I'm a little confused here, and hesitant to accept this without further information on what makes Corretto different in this regard (preferably ending up in the image description, so users can find it easily).

library/amazoncorretto
```diff
@@ -1,16 +1,32 @@
-Maintainers: Amazon Corretto Team (@corretto),
-             James Guo (@jguo11),
-             Ziyi Luo (@ziyiluo),
-             David Alvarez (@alvdavi)
+Maintainers: Amazon Corretto Team <corretto-team@amazon.com> (@corretto),
+             James Guo <junguoj@amazon.com> (@jguo11),
+             Ziyi Luo <ziyiluo@amazon.com> (@ziyiluo),
+             David Alvarez <alvdavi@amazon.com> (@alvdavi)
```
tianon, line 5: This is the source of the CI failures -- you've got an extra blank line here that shouldn't exist (if you want a "global" `GitRepo:`, etc value, they need to be part of the global block/"entry", which is what `Maintainers:` is part of).

alvdavi: Thanks, I'm fixing this

alvdavi:
For why Amazon Corretto now has alpine images, I will defer to my colleague @cliveverghese

cliveverghese:
Hi

Corretto is a downstream distribution of OpenJDK and we are working on providing and supporting Corretto builds for Alpine. The current versions for alpine are based on works from the Portola Project and Alpine aports Community. While we are working on testing the builds for Alpine, We would like to provide early access builds to customers to get their feedback.

tianon:
Very interesting!  I definitely agree that getting some community testing on this makes sense, but I think my main concern with this being the place that those exist is going to be messaging to users.

In our experience on other images, when a variant is labelled "`alpine`", users end up choosing it believing that it is supported entirely equally to the non-Alpine variants (through the full stack), and that it just gives them free image size benefits with no functional difference, and then have confusion when it is not, in spite of whatever warnings in the image description we provide to that effect.  There are some things that are fundamentally different, like the way musl handles locales (ie, not at all) versus how glibc handles them (very actively), but in most cases the differences (and incompatibilities) end up going deeper.

So if the goal is to create this as a "consumer preview", my suggestion would be to put them in an organizational namespace on Docker Hub labelled explicitly as such, and communicate to users that Alpine-based images are available as a preview at `https://hub.docker.com/r/amazoncorretto/alpine-preview` or something like that.

alvdavi:
We will move the alpine images to a different place. However, this PR also included a change on how we organize our Dockerfiles (moving from multiple repositories to a single one). I will raise a PR with just that change.

PR #8008, opened 2020-05-14
[security] Update php

tianon:
Changes:

- https://github.com/docker-library/php/commit/a3cf31c: Update to 7.3.18
- https://github.com/docker-library/php/commit/057b438: Update to 7.2.31
- https://github.com/docker-library/php/commit/8c6d1f4: Update to 7.4.6

PR #8016, opened 2020-05-15
Update bash

tianon:
Changes:

- https://github.com/tianon/docker-bash/commit/f167842: Update to commit 48394b08a7aab1b86bf721b8bb95979824cd8e9f


yosifkit:

PR #8021, opened 2020-05-15
Update redis

tianon:
Changes:

- https://github.com/docker-library/redis/commit/3914a02: Update to 6.0.2


yosifkit:

PR #8022, opened 2020-05-15
Update Github Actions to use our local clone for running tests

tianon:
This allows test changes to be tested too, if they live in the PR with the image change.

(This was discovered over in https://github.com/docker-library/official-images/pull/8010 :sweat_smile:)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# https://github.com/actions/checkout/issues/135#issuecomment-613361104
+* text eol=lf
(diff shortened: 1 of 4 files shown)


tianon:
(I've included `library/adoptopenjdk` in fd7f717 so we can verify that this change actually works. :sweat_smile:)

tianon:
For posterity (now that I've removed the temporary `library/adoptopenjdk` changes), https://github.com/docker-library/official-images/actions/runs/106075722 was very successful both on Windows and on the test config changes here. :+1:

PR #8031, opened 2020-05-19
[ros] move build tools to ros-base image

mikaelarguedas:
This PR moves the installation of the build and compilation tools from the `ros-core` to `ros-base`
It also adds the flag `--no-install-recommends` to all apt installation calls to reduce image size

Related to https://github.com/osrf/docker_images/pull/357

### DIFF
--- a/library/ros
+++ b/library/ros
@@ -9,22 +9,22 @@ GitRepo: https://github.com/osrf/docker_images.git
 
 Tags: kinetic-ros-core, kinetic-ros-core-xenial
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: b075c7dbe56055d862f331f19e1e74ba653e181a
+GitCommit: 0b33e61b5bbed5b93b9fba2d5bae5db604ff9b58
 Directory: ros/kinetic/ubuntu/xenial/ros-core
 
 Tags: kinetic-ros-base, kinetic-ros-base-xenial, kinetic
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: f2b13092747c0f60cf7608369b57ea89bc01e22d
+GitCommit: 0b33e61b5bbed5b93b9fba2d5bae5db604ff9b58
 Directory: ros/kinetic/ubuntu/xenial/ros-base
 
 Tags: kinetic-robot, kinetic-robot-xenial
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: f2b13092747c0f60cf7608369b57ea89bc01e22d
+GitCommit: 7ff09c2a75e902bc2bb25a1f1ae748ec4e9c7a4b
 Directory: ros/kinetic/ubuntu/xenial/robot
 
 Tags: kinetic-perception, kinetic-perception-xenial
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: f2b13092747c0f60cf7608369b57ea89bc01e22d
+GitCommit: 7ff09c2a75e902bc2bb25a1f1ae748ec4e9c7a4b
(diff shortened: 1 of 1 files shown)


yosifkit:
Was `build-essential` added to `ros-base` images on purpose?  I don't see it discussed on https://github.com/osrf/docker_images/pull/357.

Also, I usually recommend just installing the tools you need rather than `build-essential`:

> If you do not plan to build Debian packages, you don't need this package.
>
> https://packages.debian.org/buster/build-essenti

mikaelarguedas:
My bad I forgot to link the original discussion https://github.com/osrf/docker_templates/pull/82#issuecomment-589595139

Long story short: `build-essential` was installed as a recommended package before in the `ros-core` images. With this change build-essential (and any C/C++ compilers) were not installed anymore. To retain user ability to build ROS packages out-of-the-box when using the `ros:<distro_name>` images (alias for `ros:<distro_name>-ros-base` we decided to explicitly add `build-essential` along with the build tools installed in ros-base.

tianon:
The point @yosifkit is making (with his broken link that I've now fixed :sweat_smile:) is that `build-essential` is for building Debian packages (`dpkg-buildpackage` and friends), and that's why the description for that package mentions that it's probably not the right package otherwise.  It also has a very short list of dependencies that are easily audited to determine which ones might be useful:

- dep: dpkg-dev (>= 1.17.11)
  Debian package development tools

- dep: g++ (>= 4:8.3)
  GNU C++ compiler

- dep: gcc (>= 4:8.3)
  GNU C compiler

- dep: libc6-dev
  GNU C Library: Development Libraries and Header Files
  or libc-dev
  virtual package provided by libc6-dev

- dep: make
  utility for directing compilation

Of these, I imagine you probably only need `make`, `libc6-dev`, and `gcc`?  Possibly `g++`, depending on how common C++ is within ROS? (but then again, `g++` then depends on `gcc`, so if that's the case you're looking at `make`, `libc6-dev`, and `g++`)

You probably *don't* need `dpkg-dev`, which pulls in a host of additional dependencies including `perl`. :sweat_smile:

mikaelarguedas:
That's a fair point. In our case we needed all but `dpkg-dev`.

When doing the original testing on ubuntu bionic, it didnt make any difference with regards to the final package list installed as parts of the core of ROS rely on `pkg-config` that in turn depends on `dpkg-dev` https://packages.ubuntu.com/bionic/pkg-config

On more recent ubuntu versions that dependency on dpkg-dev was relaxed (though it still depends on perl and friends), so we can save some package installation.
Here is the difference if trying to install build-essential on top of a ROS image that installs `gcc`,`g++`,`libc6-dev`,`make`:
```
The following NEW packages will be installed:
  build-essential bzip2 dpkg-dev patch xz-utils
0 upgraded, 5 newly installed, 0 to remove and 0 not upgraded.
Need to get 2028 kB of archives.
After this operation, 2876 kB of additional disk space will be used.
```

The impact will be minor but I can make that change for the sake of preciseness in the dependency declaration.
If this is the preferred route, feel free to reject this PR and I'll open a new one with the changes.

ruffsl:
> When doing the original testing on ubuntu bionic, it didnt make any difference with regards to the final package list installed as parts of the core of ROS rely on pkg-config that in turn depends on dpkg-dev

Given this'll take some more downstream work, could we still get this merged as is? We'd like to update the docs in preparation of Foxy next week to reflect the tooling changes, but we'd like all the supported images to be consistent with image tag contents by then: https://github.com/docker-library/docs/pull/1381

tianon:
Ok, fair enough.

PR #8032, opened 2020-05-19
R 4.0.0 can now be installed from unstable, add arm64v8 back in

eddelbuettel:
Now that R 4.0.0 (via source package `r-base`) is back in unstable (with the r-4.0 transition in full swing having done more than 800 of the 950 or so packages) we can revert back to using unstable only (and not experimental).  

We also get to add arm64v8 back in.


eddelbuettel:
Thanks guys!

PR #8065, opened 2020-05-26
Updated mainline nginx to 1.19.0

thresheek:
While at it, build for more architectures.

### DIFF
--- a/library/nginx
+++ b/library/nginx
@@ -1,26 +1,26 @@
-# this file is generated via https://github.com/nginxinc/docker-nginx/blob/7956a3fed9b57f4796881c99ec23c03c05945d43/generate-stackbrew-library.sh
+# this file is generated via https://github.com/nginxinc/docker-nginx/blob/8ec60b8d3132eb78a2c9ec120490095142f9bcab/generate-stackbrew-library.sh
 
 Maintainers: NGINX Docker Maintainers <docker-maint@nginx.com> (@nginxinc)
 GitRepo: https://github.com/nginxinc/docker-nginx.git
 
-Tags: 1.17.10, mainline, 1, 1.17, latest
-Architectures: amd64, arm32v7, arm64v8, i386, ppc64le, s390x
-GitCommit: 594ce7a8bc26c85af88495ac94d5cd0096b306f7
+Tags: 1.19.0, mainline, 1, 1.19, latest
+Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
+GitCommit: 0b607f324d208664f37f7c1136ada6b249ddcd9b
 Directory: mainline/buster
 
-Tags: 1.17.10-perl, mainline-perl, 1-perl, 1.17-perl, perl
-Architectures: amd64, arm32v7, arm64v8, i386, ppc64le, s390x
-GitCommit: 594ce7a8bc26c85af88495ac94d5cd0096b306f7
+Tags: 1.19.0-perl, mainline-perl, 1-perl, 1.19-perl, perl
+Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
+GitCommit: 0b607f324d208664f37f7c1136ada6b249ddcd9b
 Directory: mainline/buster-perl
 
-Tags: 1.17.10-alpine, mainline-alpine, 1-alpine, 1.17-alpine, alpine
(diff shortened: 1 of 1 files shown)


yosifkit:
```sh
## 10-listen-on-ipv6-by-default.sh
if [ -f "/proc/net/if_inet6" ]; then
    continue
else
    echo "$ME: ipv6 not available, exiting" 1>&2
    exit 0
fi
```

The `if` statements in this file are a bit weird. `continue` outside of a loop is undefined behavior ("If there is no enclosing loop, the behavior is unspecified." \-[posix shell spec](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#continue)).  Not a blocker, but would be cleaner with just a `if [ ! -f "/proc/net/if_inet6" ]; then ...` since the action only happens in the else clause.

---

```
## 10-listen-on-ipv6-by-default.sh
echo "$CHECKSUM /$DEFAULT_CONF_FILE" | md5sum -c - >/dev/null 2>&1
```

This one should have two spaces in the echo to md5sum or space with asterisk. 

> When checking, the input should be a former output of this program. The default mode is to print a line with checksum, a space, a character indicating input mode ('*' for binary, ' ' for text or where binary is insignificant), and name for each FILE.
>
> \- https://manpages.debian.org/buster/coreutils/md5sum.1.en.html

---

The echo warnings on changed default conf will print on container restart. Not sure what a great fix would be here; maybe grep the config file for the ipv6 listen (`[::]:80`) and skip the rest of the script?

---

```sh
## 20-envsubst-on-templates.sh
for template in $(find "$template_dir" -follow -name "*$suffix"); do
```

This loop won't handle spaces in file/folder names. Not sure that there is a great posix `sh` compatible way to do so and still get the recursiveness. It seems that [this answer](https://unix.stackexchange.com/a/418992) is the most compatible with sh in Alpine + Debian. It should probably also have a `-type f` to ensure that directories matching the suffix are ignored.  These same improvements would also be needed in `docker-entrypoint.sh`.

```sh
find "$template_dir" -follow -type f -name "*$suffix" -print | while read -r template; do
```

---

```sh
## docker-entrypoint.sh

# this if doesn't really prevent running if the user replaced the content of `docker-entrypoint.d/`
# with a directory that is empty or unreadable (`/docker-entrypoint.d/mydir/`).
if /usr/bin/find "/docker-entrypoint.d/" -mindepth 1 -print -quit 2>/dev/null | /bin/grep -q .; then

# are these supposed to run every container start or only the first time?
echo "$0: /docker-entrypoint.d/ is not empty, will attempt to perform an initial configuration"

# these would be a little safer to make a single loop with a "case" to prevent edge cases
# or drift between the four similar find invocations in this file
for f in $(/usr/bin/find /docker-entrypoint.d/ -type f -name "*.sh" -executable | sort -n); do
for f in $(/usr/bin/find /docker-entrypoint.d/ -type f -name "*.sh" -not -executable); do
for f in $(/usr/bin/find /docker-entrypoint.d/ -type f -not -name "*.sh"); do

# so the inner loop would be something like this:
case "$f" in
    *.sh)
        if [ -x "$f" ]; then
            echo "$0: Launching $f"
            "$f"
        else
            # warn on shell scripts without exec bit
            echo "$0: Ignoring $f, not executable"
        fi
        ;;
    *) echo "$0: Ignoring $f";;
esac

```

thresheek:
Thanks for the review @yosifkit - much appreciated!

I agree on all the points, will implement fixes for that.

One thing I'm not sure on how to proceed is the flag for docker-entrypoint.sh not to run things on subsequent container runs - I envision those to only run on the first start of container.  I have no good ideas other than a file on a / filesystem, something along the lines of docker-entrypoint.done, but I'm not sure if that's a good idea or something usually done.  I think of it as something like initdb for postgre, but we don't really create any db paths like that as initdb does.

Thanks again!

yosifkit:
We don't have any first-start-done files (since we only do this sort of startup configuration on databases and wordpress, which both have state that gets saved indefinitely). I don't have strong opinions for or against a file marking that it ran; putting it in `/run/` or `/etc/nginx/` seems a sane enough (since those are paths we know it needs to have write access to if someone starts the container as a user other than root).

On the other hand, `20-envsubst-on-templates.sh` would definitely be useful to run on every start (in case the templates are from a mount and have changed) so it could make sense for each script to be idempotent to check if they need to run (or have no side-effect if they do). So you could just grep for the expected config and bail if it already exists?

thresheek:
Thanks, I have pushed some updates to the scripts to address your review.

PR #8093, opened 2020-05-30
[Swift] Add support for Ubuntu 20.04, CentOS 8, and Amazon Linux 2. A…

shahmishal:
…lso, update the hash for Swift 5.2.4 release

### DIFF
--- a/library/swift
+++ b/library/swift
@@ -32,11 +32,11 @@ Directory: 5.2/ubuntu/20.04
 
 Tags: 5.2.4-amazonlinux2, 5.2-amazonlinux2, amazonlinux2
 GitCommit: bcaa665ab185e9af7d82612c9436950e92466451
-Directory: 5.2/ubuntu/amazonlinux/2
+Directory: 5.2/amazonlinux/2
 
 Tags: 5.2.4-centos8, 5.2-centos8, centos8
 GitCommit: bcaa665ab185e9af7d82612c9436950e92466451
-Directory: 5.2/ubuntu/centos/8
+Directory: 5.2/centos/8
 
 Tags: 5.1.5, 5.1, 5.1.5-bionic, 5.1-bionic
 GitCommit: 05538e13a3015675d83e2553cdce5d1d67e17235
--- a/library/swift
+++ b/library/swift
@@ -6,21 +6,37 @@ Maintainers: Ted Kremenek <kremenek@apple.com> (@tkremenek),
 GitRepo: https://github.com/apple/swift-docker.git
 
 
-Tags: 5.2.3, 5.2, 5.2.3-bionic, 5.2-bionic, bionic, latest
-GitCommit: 38f179345ace24236d6c09de84e77d91384014cd
+Tags: 5.2.4, 5.2, 5.2.4-bionic, 5.2-bionic, bionic, latest
+GitCommit: bcaa665ab185e9af7d82612c9436950e92466451
 Directory: 5.2/ubuntu/18.04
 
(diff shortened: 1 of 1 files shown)


tianon:
We typically discourage this sort of (all predominantly glibc-based) base OS diversity within a single image because for most users it's not functionally all that different or useful (and usually just makes choosing an appropriate base more difficult), and instead encourage maintainers to focus on the "upstream-preferred" base distribution, especially since Docker makes distribution differences matter quite a bit less -- are there concrete functional differences/advantages to users choosing one of these over the other besides their own personal preferences?

tomerd:
hi @tianon, swift is architecture and glibc specific and there are use cases that require specific distros - for example aws lambda requires al2 and many enterprise envs (including many of the large companies) require centos. we expect to support 2-3 more distros namely debian and fedora to offer our users complete coverage

shahmishal:
@tianon Please let us know if you any additional questions. Thanks!

shahmishal:
cc: @yosifkit

yosifkit:
Apologies, I think @tianon was waiting for me to add my thoughts. Related comment for context:

> see https://github.com/docker-library/golang/pull/139#issuecomment-2824499 for a fairly up-to-date comment about where we stand on Ubuntu variants for images we maintain -- whether or not to add things like this is up to the discretion of the image maintainers, but do be prepared to explain why it's useful and meaningful over what's already provided and not simply variants for variants sake
>
> \- https://github.com/nodejs/docker-node/pull/590#issuecomment-346065995

> swift is architecture and glibc specific 

So, does a swift executable not work if compiled in Ubuntu and run in Debian (or any other Linux cross-distribution combination assuming all same CPU + Kernel + userspace architecture)? Are users really unable to use CentOS when they "require" Amazon Linux or Fedora?

I don't mind if you need to support multiple OS distributions, it just seems excessive to have multiple yum-based (and eventually apt-based) when one should work the same as any other.

---

Unrelated other item that should eventually be addressed:

> Image label names should adhere to the image spec: https://github.com/opencontainers/image-spec/blob/v1.0.1/annotations.md.
>
> \- https://github.com/docker-library/official-images/issues/3540#issuecomment-530925319

shahmishal:
> Apologies, I think @tianon was waiting for me to add my thoughts. Related comment for context:
> 
> > see [docker-library/golang#139 (comment)](https://github.com/docker-library/golang/pull/139#issuecomment-2824499) for a fairly up-to-date comment about where we stand on Ubuntu variants for images we maintain -- whether or not to add things like this is up to the discretion of the image maintainers, but do be prepared to explain why it's useful and meaningful over what's already provided and not simply variants for variants sake
> > - [nodejs/docker-node#590 (comment)](https://github.com/nodejs/docker-node/pull/590#issuecomment-346065995)
> 
> > swift is architecture and glibc specific
> 
> So, does a swift executable not work if compiled in Ubuntu and run in Debian (or any other Linux cross-distribution combination assuming all same CPU + Kernel + userspace architecture)? Are users really unable to use CentOS when they "require" Amazon Linux or Fedora?
> 
> I don't mind if you need to support multiple OS distributions, it just seems excessive to have multiple yum-based (and eventually apt-based) when one should work the same as any other.
> 
> Unrelated other item that should eventually be addressed:
> 
> > Image label names should adhere to the image spec: https://github.com/opencontainers/image-spec/blob/v1.0.1/annotations.md.
> > - [#3540 (comment)](https://github.com/docker-library/official-images/issues/3540#issuecomment-530925319)

Unfortunately, swift executable can not be executed on different platforms currently.

I will look into the annotations issue, thanks for pointing it out.

PR #8096, opened 2020-05-31
Update swipl devel and stable images to 8.2.0

ninjarobot:


tianon:
FWIW, if you moved `ENV LANG C.UTF-8` up before your `RUN` line, you could drop the `LANG=C.UTF8` on your build lines. :smile:

PR #8097, opened 2020-06-01
added yetiforcecrm official image

mariuszkrzaczkowski:
added yetiforcecrm official image

### DIFF
--- a/library/yetiforcecrm
+++ b/library/yetiforcecrm
@@ -1,6 +1,7 @@
 Maintainers: Mariusz Krzaczkowski <m.krzaczkowski@yetiforce.com> (@mariuszkrzaczkowski)
-GitRepo: https://github.com/YetiForceCompany/YetiForceCRM.git
+GitRepo: https://github.com/YetiForceCompany/YetiForceCRM-docker.git
 
-Tags: developer, latest, 5.3.75
+Tags: developer, latest
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
-GitCommit: 59e43effb8b22c30d32a75a43d7e4339cbcc41a1
+GitCommit: 19c32e02dfaac4f1c70ebd29d2adbc2064c9470a
+Directory: developer
--- a/library/yetiforcecrm
+++ b/library/yetiforcecrm
@@ -2,6 +2,5 @@ Maintainers: Mariusz Krzaczkowski <m.krzaczkowski@yetiforce.com> (@mariuszkrzacz
 GitRepo: https://github.com/YetiForceCompany/YetiForceCRM-docker.git
 
 Tags: developer, latest
-Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
 GitCommit: b1730a63466be118650307fe73c0536373008619
 Directory: developer
--- a/library/yetiforcecrm
+++ b/library/yetiforcecrm
@@ -1,0 +1,6 @@
+Maintainers: Mariusz Krzaczkowski <m.krzaczkowski@yetiforce.com> (@mariuszkrzaczkowski)
+GitRepo: https://github.com/YetiForceCompany/YetiForceCRM.git
+
+Tags: developer, latest, 5.3.75
+Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
+GitCommit: 59e43effb8b22c30d32a75a43d7e4339cbcc41a1
--- a/library/yetiforcecrm
+++ b/library/yetiforcecrm
@@ -3,5 +3,5 @@ GitRepo: https://github.com/YetiForceCompany/YetiForceCRM-docker.git
 
 Tags: developer, latest
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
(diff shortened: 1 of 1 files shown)


mariuszkrzaczkowski:
I don't get this error?
image

tianon:
Unfortunately having the `Dockerfile` in the root of the upstream repository (while useful for local development) doesn't work well as the target for official-images. Since the `Dockerfile` uses `COPY . ...`, it essentially means the entire upstream source code is now part of the Docker build context, and thus becomes part of the reviewed artifacts. (see `diff-pr.sh` for how we generate our diff; it is basically taking all docker build contexts from the starting `library/foo` file and diffing them to the new contexts from the PR).

mariuszkrzaczkowski:
@tianon I've separated the Dockerfile into a separate repository

mariuszkrzaczkowski:
@tianon  check test in checks  "Test yetiforcecrm:developer" , it lasts too long

mariuszkrzaczkowski:
any progress?

yosifkit:
Let me know if you have any questions.

1. The test "lasts too long" because the entrypoint doesn't follow https://github.com/docker-library/official-images/tree/5369615f65d4fee661f3057e5aa8248fae46f5bb#consistency (i.e. `docker run -it --rm image bash` should work).
   - The [redis entrypoint script](https://github.com/docker-library/redis/blob/3a0b4f1d370e4efd30a0db32b0e0cf878f3133d3/6.0/docker-entrypoint.sh) is a good example. If the image is switched to only have `php-fpm` (point 11), then an `entrypoint.sh` may not even be necessary.
1. https://github.com/YetiForceCompany/YetiForceCRM-docker is missing a license file.
1. Downloading keys without any verification of the key is not great; we instead recommend grabbing keys via `gnupg` via a full fingerprint (https://github.com/docker-library/official-images/tree/5369615f65d4fee661f3057e5aa8248fae46f5bb#image-build)
   - use `gpg --batch -recv-keys --keyserver ... && gpg --armor --export > /etc/apt/trusted.gpg.d/name-for-key.gpg.asc` instead of `apt-key add`
   - the current keys may not even be required (see point 5)
1. anything downloaded (even over `https`) should be verified in some way, usually via a detached gpg signature (https://github.com/docker-library/official-images/tree/5369615f65d4fee661f3057e5aa8248fae46f5bb#image-build)
1. PHP is installed from `packages.sury.org`, why not just use the packages from Debian or some of the [`php` images](https://hub.docker.com/_/php/)?
   - similar question for `yarn`, but also it doesn't seem to even be installed or used in the image.
1. Changing the same files across many layers causes image size bloat. Since installing things is almost always more than running a single executable, we recommend a pattern similar to the following:
    ```
    RUN set -ex; \
    	[prepare/download steps]; \
    	[installation steps]; \
    	[cleanup steps]

    # concrete apt-get example
    RUN set -ex; \
    	apt-get update; \
    	apt-get install -y php-fpm; \
    	rm -rf /var/lib/apt/lists/*

    # concrete download example
    RUN set -ex; \
    	apt-get update; \
    	apt-get install -y --no-install-recommends wget; \
    	rm -rf /var/lib/apt/lists/*; \
    	wget https://download.yetiforce.com/developer-for-docker.zip; \
    	# TODO gpg --batch --verify
    	unzip developer-for-docker.zip -d /var/www/html/; \
    	chown -R www-data:www-data /var/www/html/; \
    	# TODO other cleanup/setup/symlinks
    	apt-get purge -y --auto-remove wget; \
    	rm developer-for-docker.zip
    ```
   1. `chmod`, `mv`, `rm` and similar commands need to be run in the layer that initially creates the file
   1. `apt-get update` should not be in separate layer from `install` ([docs.docker.com](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#apt-get))
      - in order to keep image size lean, `rm -rf /var/lib/apt/lists/*` is the rough reverse equivalent to `apt-get update` (also at [docs.docker.com](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#apt-get)).
1. I don't understand adding anything to `/etc/environment`. Unless the PHP is reading that file, it won't have any effect. Also, that is what Dockerfile `ENV` is for.
1. We often recommend against setting `DEBIAN_FRONTEND=noninteractive` and only using it on the `apt-get install` line that requires it.  (I've only seen a handful of packages that fail without it):
    ```bash
    DEBIAN_FRONTEND=noninteractive apt-get install -y ...
    ```
1. `MAINTAINER` is deprecated. If you want to have it as a `LABEL`, that is fine, but Image label names should adhere to the image spec: https://github.com/opencontainers/image-spec/blob/v1.0.1/annotations.md.
1. Why is `openssh-server` server installed? This service shouldn't be run within a container; users should `ssh` to the Docker host if they need to manage the php-fpm process (see also next point)
1. `cron`, `mysqld`, and `nginx`
   - running multiple unrelated processes in one container is almost always a no-go for official-images. This is especially true since there is no process supervisor in the container. In your case if anything besides `mysqld` crashes, nothing happens except a broken website (or non-functioning `cron`).  We instead recommend that official-images stick to the one concern/process per container and use an orchestrator, like docker-compose, to deploy multi-container solutions.
   - See https://github.com/docker-library/official-images/tree/5369615f65d4fee661f3057e5aa8248fae46f5bb#init

tianon:
I'm closing this for now -- please feel free to let us know when this is something you're interested in picking back up and we'll be happy to reopen!

PR #8107, opened 2020-06-02
mosquitto 1.6.10

ralight:
Update to alpine:3.12
Use openssl instead of libressl.
Enable TLS-PSK support.

Thanks for the ping on alpine:3.8 being EOL.


tianon:
LGTM :+1:

PR #8123, opened 2020-06-04
[percona] Update psmdb-42 version

vorsel:


yosifkit:
I don't recommend including clutter in the official-image Dockerfile that will not be used.  A separate debug image could be accomplished in its own Dockerfile and also not require a build arg locally:

```Dockerfile
FROM percona:psmdb-4.2.7
RUN sed -i ...
```

vorsel:
> I don't recommend including clutter in the official-image Dockerfile that will not be used. A separate debug image could be accomplished in its own Dockerfile and also not require a build arg locally:
> 
> ```dockerfile
> FROM percona:psmdb-4.2.7
> RUN sed -i ...
> ```

Done

yosifkit:
LGTM (ignoring build failures for `psmdb-3.6` and `psmdb-4.0` since they aren't being updated in this PR; 4.0 has its own update in #8242)

PR #8132, opened 2020-06-04
Nextcloud 19

J0WI:


yosifkit:
CI failure is only flaky nextcloud tests; merging.

PR #8137, opened 2020-06-05
[nats] Fixed Alpine Dockerfile that was not using config file

kozlovic:

### DIFF
--- a/library/nats
+++ b/library/nats
@@ -3,7 +3,7 @@ Maintainers: Derek Collison <derek@synadia.com> (@derekcollison),
              Waldemar Salinas <wally@synadia.com> (@wallyqs),
              Jaime Piña <jaime@synadia.com> (@variadico)
 GitRepo: https://github.com/nats-io/nats-docker.git
-GitCommit: 8f8045b5c6c6f17a28b5bc12f0f18eb6380f435f
+GitCommit: a24499a62a425656d693954cd507ff3360f7cb32
 
 Tags: 2.1.7-alpine3.11, 2.1-alpine3.11, 2-alpine3.11, alpine3.11, 2.1.7-alpine, 2.1-alpine, 2-alpine, alpine
 Architectures: amd64, arm32v6, arm32v7, arm64v8


kozlovic:
@tianon @yosifkit We had the Alpine Dockerfile not specifying entrypoint and using config file (see https://github.com/nats-io/nats-docker/commit/ea417afe07d36ec3c0b20032c395bc9e54dedd14).
I am wondering if we update the official image or not (hence why this PR is in draft). Say that we do and I have single commit at the top of the library file. Will that regenerate all images or simply the alpine? If if updates all, would that be better that I am git commit for each image and change only for the alpine one? Thanks!

kozlovic:
oh.. but then now the image fails with the override-cmd, so I would need to change test/config.sh to exclude with what, that `[nats:alpine_override-cmd]=1`?

yosifkit:
No, the `override-cmd` test should not be disabled.  Since the image has a shell and other commands that can be used to explore the container, it needs to follow the "consistency" guideline: 

> A beginning user should be able to `docker run official-image bash` (or `sh`) without needing to learn about `--entrypoint`
>
> \- https://github.com/docker-library/official-images/tree/c39fe063841e2fecbb89c8125bbdaeff86730cd3#consistency

(there is an example there of providing an entrypoint script so that `--flag` as well as `sh` work)

yosifkit:
To answer the building question: the build servers make use of docker build cache so since the `windowsservercore` images have unchanged Dockerfiles they won't be pushed again.

kozlovic:
@yosifkit Thanks a bunch! I think I understand what I have to do. I am going to update/test the nats-docker repo and will update the PR here to see if that fixes this.

kozlovic:
@yosifkit Ok, I think I got it. Thank you for your continued patience with me :-)

yosifkit:
Did you want your entrypoint to have automatic `--config /etc/nats/nats-server.conf` added even if users add their own flags, like: `docker run -d nats:alpine -D`? I can go ahead and merge as-is if you need to think on it.

kozlovic:
> Did you want your entrypoint to have automatic --config /etc/nats/nats-server.conf added even if users add their own flags, like: docker run -d nats:alpine -D?

Are you saying that in this case it would automatically use config AND the flags? If so, I don't think so because then there would be no way of NOT using what is in the config file I would think? Meaning that one would have to override with flag all settings in the config file that one does not want. If one day we extend this config, it may not even be possible since not everything can be set from the command line. But maybe I am not understanding what you were saying?

Basically the goal is that if one runs `docker run nats:alpine`, it uses the default configuration. After that, if one provides any flag, then the configuration is ignored and the server is configured from the command line flags. Of course, one of the flag can be `-c <config file>` and combine flags.

kozlovic:
Also, with current approach one can always do `docker run nats:alpine -c /etc/nats/nats-server.conf -D` which would combine provided config file + debug flag. But again, at least we are not forcing the config file.

kozlovic:
> This section is not strictly necessary (since exec "$@" is what follows), but is harmless.

I see, in the example it was there because of using `gosu`.. I may update that for the next NATS Server release, but ok for now

PR #8138, opened 2020-06-05
[ros] Release ROS foxy images and make them latest

mikaelarguedas:
Images for the new ROS 2 release: [Foxy Fitzroy](https://github.com/ros-infrastructure/official-images/pull/new/foxy)
This is a new LTS and will be the target of the latest tag.

### DIFF
--- a/library/ros
+++ b/library/ros
@@ -39,7 +39,7 @@ Architectures: amd64, arm32v7, arm64v8
 GitCommit: 0b33e61b5bbed5b93b9fba2d5bae5db604ff9b58
 Directory: ros/melodic/ubuntu/bionic/ros-core
 
-Tags: melodic-ros-base, melodic-ros-base-bionic, melodic, latest
+Tags: melodic-ros-base, melodic-ros-base-bionic, melodic
 Architectures: amd64, arm32v7, arm64v8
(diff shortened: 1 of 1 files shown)


yosifkit:
I have two concerns. Are the yml files from `colcon/*/master` branches going to ever cause an issue when images are rebuilt monthly when `ubuntu` images get updated?  I see that they are downloaded over https which is good, are they released files that could also be `gpp` or at least `sha256sum` verified (https://github.com/docker-library/official-images/tree/c39fe063841e2fecbb89c8125bbdaeff86730cd3#image-build)?

ruffsl:
> Are the yml files from colcon/*/master branches going to ever cause an issue when images are rebuilt monthly when ubuntu images get updated?

These files are simple aliases so users can specify generic build flags to the build system, e.g. `release` vs `debug`. e.g.:

https://github.com/ros-planning/navigation2/blob/9e304b1ad4c10d03f7a08991224fd9686c6bb26d/Dockerfile#L58-L63

> are they released files that could also be gpp or at least sha256sum verified

There is no formal release cycle for these community mixins, as they are relatively unchanging but I'd prefer not to pin them down as we're not pinning the other non-ros packages. It'd also take a lot more templating that we don't have the infra for currently.

Lastly, we've already been doing this for the last 4 ros2 distro tags.

PR #8143, opened 2020-06-08
Update Solr to 8.5.2

janhoy:
Release note: https://lucene.apache.org/solr/news.html#apache-solrtm-852-available

This version also includes

* New jattach tool https://github.com/docker-solr/docker-solr/pull/321 
* Update maintainer info https://github.com/docker-solr/docker-solr/pull/313
* Various bug fixes

### DIFF
--- a/library/solr
+++ b/library/solr
@@ -1,95 +1,97 @@
-# this file is generated via https://github.com/docker-solr/docker-solr/blob/96477231f35e50cc9bd8cb0c8e9b004e841a126a/generate-stackbrew-library.sh
+# this file is generated via https://github.com/docker-solr/docker-solr/blob/a71f2522d58b57ef9be9cdd42525379d980479b2/generate-stackbrew-library.sh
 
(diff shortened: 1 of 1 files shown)


janhoy:
Note, as the image is now maintained by Solr Project, we wanted to add the project as maintainer. But the project does not have a github user, and the `Maintainers` syntax requires one. So I added `asfbot` to get the build passing. Do you have any other preference? Ideally we'd be able to just use e-mail.

Hronom:
When it will be merged? Since 8.5.2 still not on docker hub...

yosifkit:
Using the bot handle is fine; many maintianers also use the name of the group (so @docker-solr is also valid).  It is usually for us to have someone to ping on issues here or in our `docs` repo (so having individual maintainers also listed is helpful :+1:)

Thanks @makuk66 for all your hard work maintaining the image.

For future reference, we recommend that Image label names should adhere to the image spec: https://github.com/opencontainers/image-spec/blob/v1.0.1/annotations.md.

PR #8147, opened 2020-06-08
Update Debian (esp. for ca-certificates 20200601)

tianon:
The `ca-certificates` 20200601 release has migrated to testing and been successfully uploaded in both buster and stretch:

> This release updates the Mozilla CA bundle to 2.40, blacklists distrusted Symantec roots, and blacklists expired "AddTrust External Root".

PR #8150, opened 2020-06-09
Fix a problem in GeoNetwork 4.0.0-alpha.1 entrypoint

juanluisrp:
Fix a problem with a `sed` replace command in the image entrypoint for 4.0.0-alpha.1 tag.

See https://github.com/geonetwork/docker-geonetwork/pull/49.


yosifkit:
LGTM.

PR #8151, opened 2020-06-09
Update openjdk

tianon:
Changes:

- https://github.com/docker-library/openjdk/commit/1c30e1a: Update to 15-ea+26


yosifkit:
LGTM.

PR #8152, opened 2020-06-09
Update bash

tianon:
Changes:

- https://github.com/tianon/docker-bash/commit/b678c0c: Update to commit ab309487d5c1d7fa1a40e2d6b2e77e413cef4933
- https://github.com/tianon/docker-bash/commit/01d8523: Update to commit 32ba27b400d4ad50f265ad6b05d614292b7a430c
- https://github.com/tianon/docker-bash/commit/a084a7b: Update to commit 6ebfddbf2e32828761111c5b1e77cecffa1769fd


yosifkit:
LGTM.

PR #8154, opened 2020-06-09
Update python

tianon:
Changes:

- https://github.com/docker-library/python/commit/44c7d20: Update to 3.9.0b2, pip 20.1.1


yosifkit:
LGTM.

PR #8155, opened 2020-06-09
Update haproxy

tianon:
Changes:

- https://github.com/docker-library/haproxy/commit/ff55928: Update to 2.1.7
- https://github.com/docker-library/haproxy/commit/66d6cd4: Update to 2.1.6


yosifkit:
LGTM.

PR #8158, opened 2020-06-09
Update redis

tianon:
Changes:

- https://github.com/docker-library/redis/commit/3a0b4f1: Update to 6.0.5


yosifkit:
LGTM.

PR #8159, opened 2020-06-09
Telegraf 1.14.4

danielnelson:
Tag 1.14.4 and set as latest.

### DIFF
--- a/library/telegraf
+++ b/library/telegraf
@@ -1,7 +1,7 @@
 Maintainers: Jonathan A. Sternberg <jonathan@influxdata.com> (@jsternberg),
              Daniel Nelson <daniel.nelson@influxdata.com> (@danielnelson)
 GitRepo: git://github.com/influxdata/influxdata-docker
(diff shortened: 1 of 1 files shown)


danielnelson:
I'm unsure what the cause of the test error, seems odd that this file would cause an error during `rm -rf` since we aren't specifically removing it:
```
+ rm -rf telegraf-1.12.6-static_linux_amd64.tar.gz telegraf-1.12.6-static_linux_amd64.tar.gz.asc /usr/src /root/.gnupg
107
rm: can't remove '/root/.gnupg/S.gpg-agent.extra': No such file or directory
```

yosifkit:
It is a common failure; gnupg runs multiple daemons automatically (`dirmngr` and `gpg-agent`) and so there is sometimes a race in deleting the file.  We've had slightly better success when adding `gpgconf --kill all` once gpg verification is done but it is not required. I can merge if you are fine with how it is (esp since the "failing" image hasn't changed)

danielnelson:
Yes, please merge and I'll add the suggested command for the next update.

yosifkit:
LGTM.

PR #8198, opened 2020-06-16
Update Ubuntu

tianon:
It's been ~40 days. :sweat_smile:

cc @mwhudson @philroche


philroche:
@tianon my bad. Thanks

PR #8209, opened 2020-06-18
[registry:2.7] update to alpine 3.11, remove apache2-utils, remove 2.6 tags

thaJeztah:
Scanning the image showed some vulnerabilities, and apache2-utils was not an essential dependency at runtime, only included for "convenience", so has been removed.

diff: https://github.com/docker/distribution-library-image/compare/0b6ea3ba50b65563600a717f07db4cfa6f18f957...ab00e8dae12d4515ed259015eab771ec92e92dd4

brings in the changes from https://github.com/docker/distribution-library-image/pull/92 / https://github.com/docker/distribution-library-image/pull/104

### DIFF
--- a/library/registry
+++ b/library/registry
@@ -5,7 +5,7 @@ GitRepo: https://github.com/docker/distribution-library-image.git
 Tags: 2.7.1, 2.7, 2, latest
 Architectures: amd64, arm64v8, arm32v6
 GitFetch: refs/heads/release/2.7
-GitCommit: 0b6ea3ba50b65563600a717f07db4cfa6f18f957
+GitCommit: ab00e8dae12d4515ed259015eab771ec92e92dd4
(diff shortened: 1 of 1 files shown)


thaJeztah:
@manishtomar @ddebroy @dmcgowan @tianon PTAL

/cc @andriisoldatenko @othomann

thaJeztah:
hm.. so it's failing because the 2.6 tag is using a deprecated alpine version;

```

 - registry:2.6.2 (FROM alpine:3.8) -- completely unsupported base!

Read https://github.com/docker-library/official-images#multiple-architectures for more details.
```

I guess those tags should be removed so that they're no longer rebuilt ?

thaJeztah:
Pushed a commit to remove the 2.6 tags (so that they're no longer rebuilt)

PR #8211, opened 2020-06-18
Kong 2.1.0 beta.1

hutchic:
Kong 2.1.0-beta.1 is available.
Also pruned some older versions

### DIFF
--- a/library/kong
+++ b/library/kong
@@ -2,7 +2,25 @@ Maintainers: Kong Docker Maintainers <support@konghq.com> (@thekonginc)
 GitRepo: https://github.com/Kong/docker-kong.git
 # needs "Constraints" to match "centos" (which this image is "FROM")
 
-Tags: 2.0.4-alpine, 2.0.4, 2.0, latest, alpine
+Tags: 2.1.0-beta.1-alpine, 2.1-alpine, 2.1.0-beta.1, 2.1
+GitCommit: 69c50b396440ab6434853b982c3295d54a9fb3f2
+GitFetch: refs/tags/2.1.0-beta.1
+Directory: alpine
+Architectures: amd64
+
+Tags: 2.1.0-beta.1-ubuntu, 2.1-ubuntu
+GitCommit: 69c50b396440ab6434853b982c3295d54a9fb3f2
+GitFetch: refs/tags/2.1.0-beta.1
+Directory: ubuntu
+Architectures: amd64, arm64v8
+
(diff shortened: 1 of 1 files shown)


yosifkit:
https://github.com/Kong/docker-kong/blob/69c50b396440ab6434853b982c3295d54a9fb3f2/ubuntu/Dockerfile#L18-L30

I am slightly confused as to why the `|| true` is needed for either of these packages on `ubuntu:xenial` (since they [both](https://packages.ubuntu.com/xenial/zlibc) [exist](https://packages.ubuntu.com/xenial/zlib1g-dev)).

In the unlikely event of failure, there *could* be a bug here since and (`&&`) binds tighter than or (`||`) so this block becomes roughly this:

```dockerfile
RUN ... \
    ( apt-get update \
    && apt-get install -y --no-install-recommends perl unzip git \
    && apt-get install -y --no-install-recommends zlibc ) || ( true \
    && apt-get install -y --no-install-recommends zlib1g-dev ) || ( true \
    && rm -rf /var/lib/apt/lists/* \
	&& dpkg -i /tmp/kong.deb \
	&& rm -rf /tmp/kong.deb \
	&& useradd -ms /bin/bash kong \
    && mkdir -p "/usr/local/kong" \
	&& chown -R kong:0 /usr/local/kong \
	&& chown kong:0 /usr/local/bin/kong \
	&& chmod -R g=u /usr/local/kong \
	&& kong version )
```

One solution is to switch all `&&` to semicolons like the if statement; the other solution is to use curly braces for grouping:  `&& { apt-get install -y --no-install-recommends zlibc || true; } \`.

hutchic:
we recycle the same Dockerfile for bionic and focal. I'll try to adjust the line in question

library/kong
```diff
@@ -38,62 +56,6 @@ GitFetch: refs/tags/1.5.1
...
-GitCommit: 196331b1e6b4798032af4d6c218a441e2c8db74d
-GitFetch: refs/tags/1.1.3
-Directory: alpine
-
-Tags: 1.1.3-centos, 1.1-centos
-GitCommit: 196331b1e6b4798032af4d6c218a441e2c8db74d
-GitFetch: refs/tags/1.1.3
-Directory: centos
-
 Tags: 1.0.4-alpine, 1.0.4, 1.0
```
gszr, line 59: Let's also remove these 1.0.4 from the list.

yosifkit: Should we delay this PR or can this removal be addressed in another PR?

PR #8220, opened 2020-06-19
Removing the latest tag from Oracle Linux

Djelibeybi:
Removal of the `latest` tag from the official Oracle Linux images to reduce confusion and ambiguity when selecting a version for downstream images.

Documentation updated via https://github.com/docker-library/docs/pull/1743


yosifkit:
Normally tags removed from the library file stay available on Docker Hub. To avoid users getting the old `latest` in the future, I'll make sure we delete it from the Docker Hub.

PR #8165, opened 2020-06-10
Update Docker image for Silverpeas version 6.1

mmoqui:

PR #8222, opened 2020-06-19
Update Silverpeas to 6.1 and improve silverpeas-basics test

yosifkit:
closes #8165 (commit 6a11aad48cdfab2870388f8ea04502cf1090d2d9 cherry-picked here)


yosifkit:
Not necessary to change, but this part can be simplified and won't have to muck about with `set +e/-e`:

```diff
-    set +e
-    ./silverpeas install
-    if [ $? -eq 0 ]; then
+    if ./silverpeas install; then

PR #8224, opened 2020-06-20
Actually update ZNC to 1.8.1

DarthGandalf:
I forgot one important detail in #8141

### DIFF
--- a/library/znc
+++ b/library/znc
@@ -3,8 +3,8 @@ GitRepo: https://github.com/znc/znc-docker.git
 GitCommit: cb2cabf8d63bc992f530b0fb86d0ad9cf955819d
 Architectures: amd64, arm32v6, arm64v8
 
-Tags: 1.8.0, 1.8, latest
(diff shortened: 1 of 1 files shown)


IceBear2k:
#8225 should be merged before this, unless there is some other process to fix currently wrong tags `1.8.0` and `1.8.0-slim`.

DarthGandalf:
I don't think reverting that commit as is is a good idea: that would mean that tags `latest`, `1.8-slim`, etc would refer to the version 1.8.0 again, instead of 1.8.1, even if temporarily.

tianon:
Indeed, if fixing the older tags is important, they should be added as new entries to the existing file (temporarily) next to the existing version.

This LGTM :+1:

tianon:
(If fixing the old tags is desirable to you, @DarthGandalf, that can be a separate PR. :+1:)

DarthGandalf:
@IceBear2k wanna add the separate entry temporarily?

PR #8225, opened 2020-06-20
Revert "Update ZNC to 1.8.1" to fix wrong tags

IceBear2k:
#8141 Updated ZNC to 1.8.1 but forgot to update the tags.

This should be merged before #8224 to fix currently wrong tags `1.8.0` and `1.8.0-slim` which incorrectly point to ZNC 1.8.1.

This reverts commit ea0cd533ec25a4999bcc6b4db7318ce47d5e2936.


tianon:
Closing given https://github.com/docker-library/official-images/pull/8224#issuecomment-647123763 and following discussion.

PR #8227, opened 2020-06-21
Update Perl for 5.32

zakame:
- https://github.com/Perl/docker-perl/pull/86
- Also removes Perl 5.26 per https://metacpan.org/pod/perlpolicy


zakame:
Thanks @tianon! :tada:

PR #8235, opened 2020-06-22
Fix ZNC tags for 1.8.0

IceBear2k:
Tags `1.8.0` and `1.8.0-slim` currently incorrectly point to ZNC 1.8.1. This will fix those tags to correctly point to previous version 1.8.0 again. This fix is temporary and the previous file layout may be used again in the future.

Note: I didn't know it was possible to add multiple tags here. I've since taken a look at a couple of other files in this repository and wonder, why ZNC only maintains the latest version and not all previous tags as well? Adding a new entry with every commit seems to be a transparent and clean solution.

### DIFF
--- a/library/znc
+++ b/library/znc
@@ -1,10 +1,19 @@
 Maintainers: Alexey Sokolov <alexey+znc@asokolov.org> (@DarthGandalf)
 GitRepo: https://github.com/znc/znc-docker.git
-GitCommit: cb2cabf8d63bc992f530b0fb86d0ad9cf955819d
 Architectures: amd64, arm32v6, arm64v8
 
 Tags: 1.8.1, 1.8, latest
+GitCommit: cb2cabf8d63bc992f530b0fb86d0ad9cf955819d
 Directory: full
 
 Tags: 1.8.1-slim, 1.8-slim, slim
+GitCommit: cb2cabf8d63bc992f530b0fb86d0ad9cf955819d
+Directory: slim
+
+Tags: 1.8.0
+GitCommit: 91f58af10b62197187af1f0a9695488e88279e98
+Directory: full
+
+Tags: 1.8.0-slim
+GitCommit: 91f58af10b62197187af1f0a9695488e88279e98
 Directory: slim


DarthGandalf:
This is follow-up to #8224 

> why ZNC only maintains the latest version and not all previous tags as well?

Do you want this file to be ever-growing?

IceBear2k:
> Do you want this file to be ever-growing?

I don't see a problem with that, tbh.

tianon:
"All previous tags" wouldn't be appropriate because they're not actually supported by ZNC upstream.  For example, there won't be a 1.8.0.1 to fix bugs in 1.8.0, because that's what 1.8.1 is.  If there were a 1.7 release series that were still supported, that would definitely be appropriate to list.

tianon:
(I'm currently cleaning out older images on the affected build servers to ensure this doesn't run into https://github.com/docker-library/bashbrew/issues/8; once that's done, I plan to merge, make sure this builds and pushes, then open the revert PR. :+1:)

DarthGandalf:
Thanks!

tianon:
Doh, and as I watch this build, I realize the fix for this is actually going to be quite a bit more complicated thanks to this line:

https://github.com/znc/znc-docker/blob/cb2cabf8d63bc992f530b0fb86d0ad9cf955819d/full/Dockerfile#L1

(`FROM znc:slim`)

tianon:
So this will at least fix the `slim` variants, but the non-slim variants will require deeper adjustment in order to rebuild.

DarthGandalf:
For the future, can there be some kind of check that tags are updated in the same PR as the version?

tianon:
We've tried to add checks like that in the past, but they are kind of difficult to make generic.

The solution we use (and what I would suggest) is a short script to generate `library/znc` from the contents of the `Dockerfile` itself so that the tag gets updated automatically.

tianon:
I might be able to jerry-rig a fix for this on the build server by hand -- it's only 3 architectures, so shouldn't be too bad.  Give me a minute to try massaging it.

tianon:
Alright, both `znc:1.8.0` and `znc:1.8.0-slim` should now be fixed. :+1:

PR #8254, opened 2020-06-25
[ros] add rolling distribution images

mikaelarguedas:
ROS(2) has now a rolling / unstable distribution.
Relevant links:
- Definition of the rolling release https://www.ros.org/reps/rep-2002.html
- Announcement https://discourse.ros.org/t/new-packages-for-ros-2-rolling-ridley-2020-06-23/15019
- dockerfiles creation PR: https://github.com/osrf/docker_images/pull/422

### DIFF
--- a/library/ros
+++ b/library/ros
@@ -193,3 +193,25 @@ Architectures: amd64, arm64v8
 GitCommit: c71d7177eaad534e42307b4e7ab7e30078c884de
 Directory: ros/foxy/ubuntu/focal/ros1-bridge
 
+
+################################################################################
+# Release: rolling
+
+########################################
+# Distro: ubuntu:focal
+
+Tags: rolling-ros-core, rolling-ros-core-focal
+Architectures: amd64, arm64v8
+GitCommit: f8cfc3fa7ee88d6b59ae18deedd179b8de4c3eb2
+Directory: ros/rolling/ubuntu/focal/ros-core
+
+Tags: rolling-ros-base, rolling-ros-base-focal, rolling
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
The new images fail the 'override-cmd' check on CI.
Surprisingly the other images (e.g. ros_foxy-ros-core-focal) that is almost identical does pass the test.

The CMD directive is currently used only in the lowest level dockerfile: https://github.com/osrf/docker_images/blob/5ee4389841e72474698cd2dce7378b54e3717459/ros/rolling/ubuntu/focal/ros-core/Dockerfile#L39
And I confirmed that overriding cmd with an echo command does work locally on Linux.

Is there any action that should be taken on our side to fix this ?

yosifkit:
I would guess that [this `source` line](https://github.com/osrf/docker_images/blob/5ee4389841e72474698cd2dce7378b54e3717459/ros/rolling/ubuntu/focal/ros-core/ros_entrypoint.sh#L5) is the problem. It is probably outputting stuff to `stdout` and so gets detected as not working correctly for our `override-cmd` check that is verifying the [consistency](https://github.com/docker-library/official-images/tree/b323b55462819028244522f820da1f7a7eb77d2d#consistency) requirement.

mikaelarguedas:
You're perfectly right :+1:

I'll report this upstream to get it fixed and put this PR on hold for now.

mikaelarguedas:
The packages have been synced and this is now passing CI

FYI @yosifkit

PR #8255, opened 2020-06-25
caddy: some Dockerfile updates

hairyhenderson:
A couple updates to the Dockerfile:
- Adding an `/etc/nsswitch.conf` file (https://github.com/caddyserver/caddy-docker/pull/96)
- Setting `WORKDIR` to `/srv` (https://github.com/caddyserver/caddy-docker/pull/97)


tianon:
Yep, aware of it, just waiting for the "Periodic Actions" bot to post the diff so we can review :+1:

PR #8257, opened 2020-06-26
Re-releasing Neo4j 4.1.0 docker image

jennyowen:
fixed Neo4j 4.1.0 docker bug where NEO4J_AUTH=none was not being honoured.
It is severe enough that I need to re-release the neo4j:4.1.0 image, if that's ok?

### DIFF
--- a/library/neo4j
+++ b/library/neo4j
@@ -11,12 +11,12 @@ Maintainers: Ben Butler-Cole <ben@neo4j.com> (@benbc),
 
 Tags: 4.1.0, 4.1, latest
 GitRepo: https://github.com/neo4j/docker-neo4j-publish.git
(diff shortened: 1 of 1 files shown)


yosifkit:
> re-release the `neo4j:4.1.0 image`, if that's ok?

Definitely! The image tags aren't considered static. Images listed in the `library/` files are rebuilt every time they are changed or when their base image changes (e.g. `debian` image is updated and so all images `FROM` debian are also rebuilt)

tianon:
To put that another way -- it's not only considered OK, it's the whole point of keeping the tags listed in this file, and is what we mean when we say "supported" :smile:

jennyowen:
Ah brilliant, that's good to know, thanks!
the more you know

PR #8265, opened 2020-06-26
caddy: 2.1.0 release

hairyhenderson:
Caddy v2.1 is released!

This also removes the 2.0 release.

See https://github.com/caddyserver/caddy/releases/tag/v2.1.0 for details.

### DIFF
--- a/library/caddy
+++ b/library/caddy
@@ -1,63 +1,33 @@
 # this file is generated with gomplate:
 # template: https://github.com/caddyserver/caddy-docker/blob/6355eb99a01d12c55a7e043129bda9cb90da81d6/stackbrew.tmpl
-# config context: https://github.com/caddyserver/caddy-docker/blob/6355eb99a01d12c55a7e043129bda9cb90da81d6/stackbrew-config.yaml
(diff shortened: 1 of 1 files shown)


library/caddy
```diff
@@ -1,60 +1,60 @@
...
 GitRepo: https://github.com/caddyserver/caddy-docker.git
 Directory: 2.1/windows/ltsc2016
-GitCommit: 6355eb99a01d12c55a7e043129bda9cb90da81d6
+GitCommit: 89777d071050735e2769ea3a48d0325adac96900
 Architectures: windows-amd64
 Constraints: windowsservercore-ltsc2016
 
-Tags: 2.0.0-alpine, 2-alpine, alpine
-SharedTags: 2.0.0, 2, latest
+Tags: 2.0.0-alpine
```
tianon, line 35: Is the 2.0.x series going to continue to be supported as well?

hairyhenderson: @tianon that was my plan at least short-term, but I'll double-check with @mholt 🤔

mholt: Our `master` is on the 2.2.x dev tree already; I'm not sure why we'd need to keep supporting 2.0.

hairyhenderson: ok, fair enough - I'll remove 2.0!

hairyhenderson:
PTAL @tianon @yosifkit 😉

PR #8282, opened 2020-06-30
Update Amazon Linux images

iliana:


iliana:
Looks like some Git LFS-related failures in the diff and the failed checks? :(

yosifkit:
Yeah, we still can't use LFS :disappointed: (https://github.com/docker-library/bashbrew/issues/10).

iliana:
Used to be it was just ignored, though... is it a blocker now?

yosifkit:
Oh, it was ignored when we ran locally because we didn't have lfs installed, but the github actions definitely do :cry:.  Thoughts @tianon?

iliana:
We could potentially modify the actions to run `git lfs uninstall` prior to cloning repos to remove the global hooks?

iliana:
I can try and figure that out and send a PR if y'all like, since I think we're the only repo doing these shenanigans.

tianon:
My initial thought was exactly the same before I saw your reply. :laughing: 

I did some research on LFS and it appears to support a `GIT_LFS_SKIP_SMUDGE` environment variable for disabling the smudge filters, so I think I'd rather we explore that before we get drastic and just purge LFS from the Actions environment.  I'll play a little bit and see if I can get this working properly. :+1: :heart:

iliana:
🎉

tianon:
Yay, success!! :smile:

PR #8289, opened 2020-07-02
fix(logging): fix default logging configuration

educhastenier:


tianon:
> ```diff
> +# extra lib required at Tomcat startup (for instance bonita juli extensions)
> +CLASSPATH="${CATALINA_HOME}/lib/ext/*"
> ```

Did this one need to be `export`ed also, or is this sufficient for how it gets used?

educhastenier:
> > ```diff
> > +# extra lib required at Tomcat startup (for instance bonita juli extensions)
> > +CLASSPATH="${CATALINA_HOME}/lib/ext/*"
> > ```
> 
> Did this one need to be `export`ed also, or is this sufficient for how it gets used?

Good point. I'll double-check that and come back to you. Thx.

educhastenier:
After checking, it is ok like that (even if would be better with the `export`, you are perfectly right).
So you can merge it like that.
Thx.

PR #8291, opened 2020-07-03
Update Sonarqube to 8.4.0

tobias-trabelsi-sonarsource:
Hey there :wave: 
we released a new version of sonarqube today

### DIFF
--- a/library/sonarqube
+++ b/library/sonarqube
@@ -9,11 +9,11 @@ GitCommit: 2618ed9a5371c30736de1e0d5dbcc9e7aa05f105
 Tags: 7.9.3-community, 7.9-community, lts
 Directory: 7/community
 
-Tags: 8.4.0-community, 8.3.1-community, 8.3-community, 8-community, community, latest
+Tags: 8.4.0-community, 8.4-community, 8-community, community, latest
 Directory: 8/community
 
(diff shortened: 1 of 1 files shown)


library/sonarqube
```diff
@@ -4,16 +4,16 @@ Maintainers: Janos Gyerik <janos.gyerik@sonarsource.com> (@janos-ss),
...
              Michal Duda <michal.duda@sonarsource.com> (@michal-duda-sonarsource)
 GitRepo: https://github.com/SonarSource/docker-sonarqube.git
-GitCommit: 2f7290e97a56f71603eff494db0d6b8d34ca426e
+GitCommit: 2618ed9a5371c30736de1e0d5dbcc9e7aa05f105
 
 Tags: 7.9.3-community, 7.9-community, lts
 Directory: 7/community
 
-Tags: 8.3.1-community, 8.3-community, 8-community, community, latest
+Tags: 8.4.0-community, 8.3.1-community, 8.3-community, 8-community, community, latest
```
tianon, line 12: I don't think this is quite what you intended -- this includes all of `8.4.0`, `8.3.1`, and `8.3` as tags for the same image.

tobias-trabelsi-sonarsource: this was indeed not the plan. Thanks for the catch (: 
corrected in my latest commit

PR #7238, opened 2020-01-07
update archlinux: 20200106

SantiagoTorres:

PR #8297, opened 2020-07-05
archlinux: add 20200705

SantiagoTorres:


tianon:
https://github.com/docker-library/official-images/pull/7238#issuecomment-571382011

> Just so we don't forget, we are still deciding on a policy on the Arch Linux side on how long we are keeping images for. I think by the end of Q1 of 2020 we'll have something encoded :)

Friendly ping :wink:

(Especially now that we're up to 10 snapshot releases, none of which will ever really be updated :sweat_smile:)

SantiagoTorres:
Thank you for the reminder. I really want to determine this soon. Do you want me to have settle on this *before* merging? (it's definitely possible, I just want to know how up in the list of priorities things are :)

tianon:
Naw, just don't want it to get lost/forgotten :+1: :heart:

PR #8303, opened 2020-07-06
haxe 4.1.2

andyli:


andyli:
That CI failure looks like a random error to me. A retry may "fix" it.

tianon:
I've retried it and it failed somewhere else, but it's a tag that isn't updated here (and all other checks are green), so I'm merging without it. :+1:

PR #8316, opened 2020-07-07
kong: bump 2.0 series to 2.0.5

gszr:


atikhono:
Please merge soon, we are really waiting for this!

vgrebenschikov:
Please

tianon:
I expect y'all mean well, but spamming the repository maintainers isn't a particularly effective way to get *positive* attention on a PR you care about (and ~an hour old is way too soon to be thinking we missed it).

If the reason for the spam is that it's a security update, I would suggest that the image maintainers should follow https://github.com/docker-library/official-images#security-releases next time so we can prioritize the review (even during our off hours, if necessary) appropriately.

If you're just antsy for the new release, please exercise some patience.

atikhono:
@tianon I'm very sorry to trouble you, it just contains a critical bug fix to us/our production.

tianon:
It's strange to me that you're keeping the `zlibc-dev` packages (instead of just `zlibc`) -- is there a specific reason for doing so?

hutchic:
@tianon we looked into it and it seems there is not a reason to include the `-dev` packages.

@atikhono FWIW we do publish unofficial builds at `kong/kong` ie https://hub.docker.com/layers/kong/kong/2.0.5/images/sha256-332a54719e1d370c989c4316185589d446aed6d5385d0c23b685ae6ef1112604?context=explore though I would advise against using those in a production scenario.
That dokcer hub location also includes daily images if you want to verify a specific bug fix 👍

yosifkit:
Alpine image still adds `zlib-dev`; did you want to also fix that one or just go ahead with this PR?

hutchic:
@yosifkit sorry missed that one. It's updated to not include it now

hutchic:
The ubuntu package is `zlib1g` please hold off on merging

hutchic:
ok this is prepped now -- sorry for the delay

PR #8337, opened 2020-07-10
bump kong to 2.1.0-rc.1

hutchic:
sorry for the flurry of commits (2.0.5 immediately followed by 2.1.0-rc.1)

Going forward we need the `zlib-dev` packages because https://github.com/hamishforbes/lua-ffi-zlib/blob/master/lib/ffi-zlib.lua#L98 tries to load the zlib.so via the symlink that's only included in the `-dev` packages.

### DIFF
--- a/library/kong
+++ b/library/kong
@@ -2,21 +2,21 @@ Maintainers: Kong Docker Maintainers <support@konghq.com> (@thekonginc)
 GitRepo: https://github.com/Kong/docker-kong.git
 # needs "Constraints" to match "centos" (which this image is "FROM")
 
-Tags: 2.1.0-beta.1-alpine, 2.1-alpine, 2.1.0-beta.1, 2.1
-GitCommit: d89a2a8f20ce55d537d20aa97a10203fd5bf3039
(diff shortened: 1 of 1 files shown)


tianon:
> Going forward we need the `zlib-dev` packages because https://github.com/hamishforbes/lua-ffi-zlib/blob/master/lib/ffi-zlib.lua#L98 tries to load the zlib.so via the symlink that's only included in the `-dev` packages.

Frankly this feels like a bug in Lua's `ffi.load` function not taking into account ABI and/or standard ABI filename suffixes. :disappointed:

If you wanted to avoid pulling in the dependencies of `zlib1g-dev` (like `libc6-dev`, `libc-dev-bin`, `libcrypt-dev`, etc), you could just create a symlink yourself (it would probably work fine if you added it to `/usr/local/lib` to avoid conflict if users then install `zlib1g-dev`), but it's up to you.  If the extra dependencies don't bother you, this solution is acceptable to us.

hutchic:
> you could just create a symlink yourself

Not sure how well this will work since we're setting `zlib1g-dev` as a dependency on our deb's (and equivalent for rpm). At this time the extra depdencies don't bother us but we'll look into this in the future if that works.

PR #8341, opened 2020-07-12
Friendica Release "Red Hot Poker" 2020.07

nupplaphil:


nupplaphil:
@yosifkit I see some troubles at jenkins with the mips64le image of Friendica.
Can I assist you in any way like adding different libraries? (Like the error suggests)

yosifkit:
As far as I can tell the root cause is that Debian Stretch is no longer supported on mips64le and s390x (as of June 30, https://wiki.debian.org/DebianReleases and https://wiki.debian.org/LTS) and so a few security package updates were already built into the php image and now some dev packages are not installable since the updated packages from `http://security.debian.org/debian-security stretch/updates` have been deleted as part of the EOL process.

While they should be able to (re)build one last time on the final Stretch release at the end of the week, I'd suggest moving to Buster based images if you want to keep all the current architectures.

PR #8342, opened 2020-07-13
add ppc64le to Architectures, support sudo in container

eddelbuettel:
This addresses https://github.com/rocker-org/rocker/issues/409 requesting ppc64le support by adding, and updates the Dockerfile by one commit adding `sudo` support which is useful if one launches container with `-u 1000:1000` as I now do more often :)

### DIFF
--- a/library/r-base
+++ b/library/r-base
@@ -3,7 +3,7 @@ Maintainers: Carl Boettiger <rocker-maintainers@eddelbuettel.com> (@cboettig),
 GitRepo: https://github.com/rocker-org/rocker.git
 
 Tags: 4.0.2, latest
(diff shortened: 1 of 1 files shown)


yosifkit:
We usually recommend against adding `sudo` to official-images, since the same can be accomplished more securely with `docker exec --user 0:0 -it ...`.

eddelbuettel:
I was wondering if there was a policy. In that case I will revert.

eddelbuettel:
I force-pushed an update removing sudo, leaving the other two changes.

eddelbuettel:
Thanks as always!

PR #8366, opened 2020-07-15
Add official images for Corretto on Alpine Linux

cliveverghese:
Add official images for Corretto on Alpine Linux 

Dockerfiles
* [11/jdk/alpine](https://github.com/corretto/corretto-docker/blob/master/11/jdk/alpine/Dockerfile)
* [11/jre/alpine](https://github.com/corretto/corretto-docker/blob/master/11/jre/alpine/Dockerfile)
* [8/jdk/alpine](https://github.com/corretto/corretto-docker/blob/master/8/jdk/alpine/Dockerfile)
* [8/jre/alpine](https://github.com/corretto/corretto-docker/blob/master/8/jre/alpine/Dockerfile)

Release announcement
[Corretto-8](https://github.com/corretto/corretto-8/releases/tag/8.262.10.2)
[Corretto-11](https://github.com/corretto/corretto-11/releases/tag/11.0.8.10.1)

### DIFF
--- a/library/amazoncorretto
+++ b/library/amazoncorretto
@@ -13,3 +13,19 @@ Directory: 8/jdk/al2
 Tags: 11, 11.0.8, 11.0.8-al2, 11-al2-jdk, 11-al2-full
 Architectures: amd64, arm64v8
 Directory: 11/jdk/al2
+
+Tags: 8-alpine, 8u262-alpine, 8-alpine-full, 8-alpine-jdk
+Architectures: amd64
+Directory: 8/jdk/alpine
+
+Tags: 8-alpine-jre, 8u262-alpine-jre
+Architectures: amd64
+Directory: 8/jre/alpine
+
+Tags: 11-alpine, 11.0.8-alpine, 11-alpine-full, 11-alpine-jdk
+Architectures: amd64
+Directory: 11/jdk/alpine
+
+Tags: 11-alpine-jre, 11.0.8-alpine-jre
+Architectures: amd64
+Directory: 11/jre/alpine
--- a/library/amazoncorretto
+++ b/library/amazoncorretto
@@ -4,7 +4,7 @@ Maintainers: Amazon Corretto Team <corretto-team@amazon.com> (@corretto),
              Clive Verghese <verghese@amazon.com> (@cliveverghese)
 GitRepo: https://github.com/corretto/corretto-docker.git
 GitFetch: refs/heads/master
-GitCommit: fa556a8d84f2d2f39b1925b15f4fb7ebd3e6e4ed
+GitCommit: 45521b3fed98030c45c4a7fb23af52c614b48994
 
 Tags: 8, 8u265, 8u265-al2, 8-al2-full,8-al2-jdk, latest
 Architectures: amd64, arm64v8
@@ -25,7 +25,3 @@ Directory: 8/jre/alpine
 Tags: 11-alpine, 11.0.8-alpine, 11-alpine-full, 11-alpine-jdk
 Architectures: amd64
 Directory: 11/jdk/alpine
-
-Tags: 11-alpine-jre, 11.0.8-alpine-jre
-Architectures: amd64
-Directory: 11/jre/alpine


tianon:
I think my comments on https://github.com/docker-library/official-images/pull/8012 are still relevant, specifically https://github.com/docker-library/official-images/pull/8012#issuecomment-628947058 and https://github.com/docker-library/official-images/pull/8012#issuecomment-629430209:

---

> My understanding is that Alpine support is being worked on within the OpenJDK / Java community under Project Portola (https://wiki.openjdk.java.net/display/portola/Main), and that it's currently officially unsupported and pretty desperate for contributions; see https://github.com/docker-library/openjdk/pull/235#issuecomment-424599754:
> 
> > Yeah, the alpine port is not in a supported release, since it's not in the mainline code base, so it's only available as early access builds of OpenJDK Project Portola http://openjdk.java.net/projects/portola.  
> > 
> > It works, but at this stage it basically needs community contributors to help maintain it and get it into mainline - see https://twitter.com/MikaelVidstedt/status/984920426655526912 for details. Interested contributors should drop by portola-dev to discuss specifics.
> 
> So honestly, I'm a little confused here, and hesitant to accept this without further information on what makes Corretto different in this regard (preferably ending up in the image description, so users can find it easily).

---

> Very interesting!  I definitely agree that getting some community testing on this makes sense, but I think my main concern with this being the place that those exist is going to be messaging to users.
> 
> In our experience on other images, when a variant is labelled "`alpine`", users end up choosing it believing that it is supported entirely equally to the non-Alpine variants (through the full stack), and that it just gives them free image size benefits with no functional difference, and then have confusion when it is not, in spite of whatever warnings in the image description we provide to that effect.  There are some things that are fundamentally different, like the way musl handles locales (ie, not at all) versus how glibc handles them (very actively), but in most cases the differences (and incompatibilities) end up going deeper.
> 
> So if the goal is to create this as a "consumer preview", my suggestion would be to put them in an organizational namespace on Docker Hub labelled explicitly as such, and communicate to users that Alpine-based images are available as a preview at `https://hub.docker.com/r/amazoncorretto/alpine-preview` or something like that.

---

Has something changed?

cliveverghese:
Hi @tianon,

* Regarding support of Corretto for musl libc. While this port is mainly derived from the Portola project and ported back to versions 8 and 11, Amazon OpenJDK team is supporting these builds of Corretto independently of the status of the Portola project on OpenJDK. These builds have been found to be stable both in internal testing and with AWS customers we have worked with. In any case, If we identify bugs or fix issues, we will be contributing them back to the Portola Project and/or OpenJDK tip once merged. 
* Regarding the difference in behavior between glibc based and musl based images, We agree that there are differences in  the underlying implementations of libc, we still support both images officially even though they are not 100% equivalent.

tianon:
Ok, that makes sense; thank you for clarifying.  This has been a pain point between us and the Java community in the past, so I want to make sure we're doing as much due diligence as we can.

I am kind of curious -- does Amazon have access to the TCK?  Have you run the TCK against these Alpine builds?

A few `Dockerfile` review comments:

---

```diff
+FROM alpine:latest
```

This should be more specific (`FROM alpine:3.12`, for example).

---

```diff
+RUN wget -O /etc/apk/keys/amazoncorretto.rsa.pub https://apk.corretto.aws/amazoncorretto.rsa.pub && \
```

Can this downloaded key be verified in some way?  Checksum, fingerprint, etc.

---

```diff
+FROM alpine:latest
...
+    jlink --endian little --release-info $JAVA_HOME/release \ 
```

This seems like a bit of an extreme way to create a strictly-JRE artifact -- I realize OpenJDK 11 doesn't have an official specification for what a "JRE" is (instead preferring `jlink`), but I would think that if you want to publish a JRE image, you would probably want to also publish an "official" JRE package, like the AdoptOpenJDK community does, right?

Otherwise, this should be adjusted to something like the following:

```dockerfile
FROM amazoncorretto:11.0.8-alpine-jdk AS jlink

RUN jlink ...

FROM alpine:3.12

COPY --from=jlink ...
```

(Although my preference is still pretty strongly in favor of creating proper separate JRE packages, especially given how poorly multi-stage builds like this will interact with the Docker build cache.)

cliveverghese:
Hi @tianon , 

Thank you for the valuable feedback. 

* Added specific version of alpine
* Added validation for public key
* We have currently removed 11-alpine-jre from the images. 

> I am kind of curious -- does Amazon have access to the TCK? Have you run the TCK against these Alpine builds?

Yes, Amazon [has access to TCK](https://openjdk.java.net/groups/conformance/JckAccess/jck-access.html). Amazon Corretto builds are all certified using the TCK and this includes our musl-libc builds.

tianon:
```dockerfile
ADD THIRD-PARTY-LICENSES /
```

This ends up being included in the image context (which is the bits we review) -- shouldn't this be either part of the `amazon-corretto-11` package `apk` file directly, or something you could download or synthesize from elsewhere?

cliveverghese:
> ```dockerfile
> ADD THIRD-PARTY-LICENSES /
> ```
> 
> This ends up being included in the image context (which is the bits we review) -- shouldn't this be either part of the `amazon-corretto-11` package `apk` file directly, or something you could download or synthesize from elsewhere?


This THIRD-PARTY-LICENSES is attributions to the packages included in the Alpine Distribution. The attributions for the Corretto is included as part of the amazon-corretto package.

tianon:
Is the information included with the distribution / the packages themselves not sufficient for some reason?  If not, couldn't the build do something to synthesize that content based on what's actually installed during build, instead of copying a static hand-maintained file which will then suffer bitrot?

cliveverghese:
> Is the information included with the distribution / the packages themselves not sufficient for some reason? If not, couldn't the build do something to synthesize that content based on what's actually installed during build, instead of copying a static hand-maintained file which will then suffer bitrot?

The distribution / packages include attribution to the third party software that is used within Amazon Corretto. This attribution is specific to the alpine base image. It include attribution to the software present in alpine base image. Yes, We understand that the hand maintained file will suffer bitrot. However, we plan on adding automation on our end to ensure that this does not happen.

cliveverghese:
Checking-in to see if you have further concerns regarding this pull request.

yosifkit:
We'd rather not have the license files in the build context; it makes simple updates much harder to review. See https://github.com/docker-library/official-images/pull/8313#issuecomment-654441025 for example and unfortunately quite a few of the [websphere-liberty PRs](https://github.com/docker-library/official-images/pulls?q=is%3Apr+label%3Alibrary%2Fwebsphere-liberty+is%3Aclosed).

It seems like this lIcense file doesn't really reflect the resultant image, since it is generated before the image exists. If the end result is to give end users access to the licenses of the software that is included then it would make better sense to either:
 - generate it within the image
   - accurately reflects installed packages
   - may not be possible to run the script/tool in the image?
 - or after the image is built
   - harder for users to correlate which commit of the licenses is from which image
   - needs periodic task to check if image was updated
     - and, when more than `amd64`, be careful if only one architecture is updated in the manifest list

---

Related information: Tags in the library file are only built through an update to that library file or as a result of its base image being updated. So the `amazoncorretto:8u262-alpine` can change without the `library/amazoncorretto` file changing if `alpine:3.12` is updated.

cliveverghese:
Hi, 

Sorry for the delay in the response while we were figuring out the next steps, 

We have removed the THIRD-PARTY-LICENSE from the build context,

We understand that the licenses would not accurately represent the package versions in the base image, The corretto team would work on keeping the file updated. However, there would be a delay of a few days while the team prepares the updated license when the base image changes.

PR #8369, opened 2020-07-16
Update CrateDB to 4.2.1.

kovrus:


yosifkit:
In our experience, there shouldn't need to be a separate Dockerfile for each architecture except on bases images where the image is `FROM scratch`.

> We strongly recommend that most images create a single `Dockerfile` per entry in the library file that can be used for multiple architectures. This means that each supported architecture will have the same `FROM` line (e.g. `FROM debian:jessie`). See [`golang`](https://github.com/docker-library/official-images/blob/master/library/golang), [`docker`](https://github.com/docker-library/official-images/blob/master/library/docker), [`haproxy`](https://github.com/docker-library/official-images/blob/master/library/haproxy), and [`php`](https://github.com/docker-library/official-images/blob/master/library/php) for examples of library files using one `Dockerfile` per entry and see their respective git repos for example `Dockerfile`s.
>
> \- https://github.com/docker-library/official-images/blob/master/README.md#multiple-architectures

kovrus:
Starting from `4.2.0` we distribute `CrateDB` with platform-specific JDK bunldes, e.g. https://github.com/crate/docker-crate/blob/master/arm64v8/crate/Dockerfile#L18 is bundled with `jdk-14.0.2+12 linux/aarch64` and https://github.com/crate/docker-crate/blob/master/amd64/crate/Dockerfile#L17 with `jdk-14.0.2+12 linux/x64`. We followed the `hello-world` example and introduced two separate Dockerfiles for platform-specific CrateDB distributions. Is it still possible to achieve the same result while having a single Dockerfile? Could you maybe provide some recommendations on how to do that? 

As a side note, we also tried to build the arm64v8 image with single `FROM centos:7` which didn't work https://github.com/crate/docker-crate/pull/183/checks?check_run_id=878265564

tianon:
> We followed the `hello-world` example and introduced two separate Dockerfiles for platform-specific CrateDB distributions. Is it still possible to achieve the same result while having a single Dockerfile? Could you maybe provide some recommendations on how to do that?

Unfortunately, `hello-world` is a pretty bad example image for anything that isn't a single (small) static binary like the `hello-world` image is.

I'd suggest checking the approach used by one of the images that @yosifkit linked above, such as [`golang`](https://github.com/docker-library/official-images/blob/master/library/golang), [`docker`](https://github.com/docker-library/official-images/blob/master/library/docker), [`haproxy`](https://github.com/docker-library/official-images/blob/master/library/haproxy), and [`php`](https://github.com/docker-library/official-images/blob/master/library/php).

> As a side note, we also tried to build the arm64v8 image with single `FROM centos:7` which didn't work https://github.com/crate/docker-crate/pull/183/checks?check_run_id=878265564 

That's expected given your usage -- if you just do `docker build` on that, it will resolve `centos:7` for the _host_ architecture (it's not going to look at the directory name to "cleverly" determine whether to switch architectures or something like that).  You'll need to build it on a native `arm64` host (like our official builders use), use `docker build --platform`, use `buildx`/BuildKit, etc if you want to build-test for ARM.

---

Here's the `diff` between your two current files:

```diff
$ git diff --no-index amd64/crate/Dockerfile arm64v8/crate/Dockerfile
diff --git a/amd64/crate/Dockerfile b/arm64v8/crate/Dockerfile
index 57e008f..1b80ab6 100644
--- a/amd64/crate/Dockerfile
+++ b/arm64v8/crate/Dockerfile
@@ -4,7 +4,7 @@
 # https://github.com/crate/docker-crate
 #
 
-FROM centos:7
+FROM arm64v8/centos:7
 
 RUN groupadd crate && useradd -u 1000 -g crate -d /crate crate
 
@@ -14,8 +14,8 @@ RUN yum install -y yum-utils \
     && yum install -y python36 openssl \
     && yum clean all \
     && rm -rf /var/cache/yum \
-    && curl -fSL -O https://cdn.crate.io/downloads/releases/cratedb/x64_linux/crate-4.2.1.tar.gz \
-    && curl -fSL -O https://cdn.crate.io/downloads/releases/cratedb/x64_linux/crate-4.2.1.tar.gz.asc \
+    && curl -fSL -O https://cdn.crate.io/downloads/releases/cratedb/aarch64_linux/crate-4.2.1.tar.gz \
+    && curl -fSL -O https://cdn.crate.io/downloads/releases/cratedb/aarch64_linux/crate-4.2.1.tar.gz.asc \
     && export GNUPGHOME="$(mktemp -d)" \
     && gpg --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 90C23FC6585BC0717F8FBFC37FAAE51A06F6EAEB \
     && gpg --batch --verify crate-4.2.1.tar.gz.asc crate-4.2.1.tar.gz \
@@ -56,7 +56,7 @@ COPY --chown=1000:0 config/crate.yml /crate/config/crate.yml
 COPY --chown=1000:0 config/log4j2.properties /crate/config/log4j2.properties
 
 LABEL maintainer="Crate.io <office@crate.io>" \
-    org.opencontainers.image.created="2020-07-16T09:33:12.106046" \
+    org.opencontainers.image.created="2020-07-16T09:33:11.717146" \
     org.opencontainers.image.title="crate" \
     org.opencontainers.image.description="CrateDB is a distributed SQL database handles massive amounts of machine data in real-time." \
     org.opencontainers.image.url="https://crate.io/products/cratedb/" \
```

What this amounts to is just a little architecture detection with an `if` or `case` to download the right file.  I think the `golang` image is probably going to be the closest to what you're trying to accomplish from the examples listed above (although the `docker` image is also very similar).

kovrus:
@tianon thank you for your help. We moved to back to the single Dockerfile for all supported architectures approach.

PR #6661, opened 2019-09-19
Add an ubuntu Kong image which includes arm64

hutchic:
We're buildling Ubuntu (xenial) arm64 assets so to make Kong available via docker on arm64 adding an Ubuntu based image.

I debated including an `arm64` tag on the Ubuntu image for convenience (ie `docker run kong:arm64`) but wasn't sure if that was advisable

PR #8376, opened 2020-07-16
kong: bump to 2.1.0

gszr:


tianon:
Are the 1.5.x releases still actively supported? (as in, if there's a problem found, will there be a 1.5.2 release?)

I'd also like to make sure https://github.com/docker-library/official-images/pull/6661#issuecomment-534314625 is still on your radar -- having both Ubuntu-based and CentOS-based images simultaneously was a initially intended to be a temporary stop-gap solution.

hutchic:
cicling back to alpine arm64 builds so we can remove the ubuntu builds is still on the radar. It hasn't found it's way to the top of the queue yet but its getting there.

gszr:
> Are the 1.5.x releases still actively supported? (as in, if there's a problem found, will there be a 1.5.2 release?)

1.5.x tags have been removed, thanks for the nudge!

PR #7962, opened 2020-05-07
arangodb 3.4.10, 3.5.5, 3.6.3.1

KVS85:
Updated ArangoDB 3.4 to 3.4.10, 3.5 to 3.5.5, 3.6 to 3.6.3.1 (3.6.3 with Hotfix 1) and remove unsupported versions.

PR #8383, opened 2020-07-17
arangodb 3.6.5

KVS85:
Updated ArangoDB 3.6 to 3.6.5 and fix foxx-cli version to 1.3.0 in 3.4.10 and 3.5.5.

Didn't put #7962 (comment) workaround yet (planned for the next version only) so npm ERR! lifecycle could not get uid/gid can happen.

### DIFF
--- a/library/arangodb
+++ b/library/arangodb
@@ -2,10 +2,10 @@
 # maintainer: Wilfried Goesgens <w.goesgens@arangodb.org> (@dothebart)
 # maintainer: Vadim Kondratyev <vadim@arangodb.org> (@KVS85)
 
-3.4: https://github.com/arangodb/arangodb-docker@63dc866beb0546ba3b172e76f64759abfe1a0a28 alpine/3.4.10
-3.4.10: https://github.com/arangodb/arangodb-docker@63dc866beb0546ba3b172e76f64759abfe1a0a28 alpine/3.4.10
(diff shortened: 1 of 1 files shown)


KVS85:
@tianon can you please merge this?

tianon:
```diff
 FROM alpine:3.10
...
+    apk add --no-cache --repository http://dl-cdn.alpinelinux.org/alpine/v3.11/main/ nodejs npm && \
```

In our experience, mixing Alpine version packages like this ends up causing sometimes subtle (sometimes obvious) bugs that are difficult for users to diagnose.  If you need Node.js from Alpine 3.11, why not just update the image to Alpine 3.11?  Given it's a database image, and users won't often be doing `FROM arangodb:xxx`, the downstream effect should be pretty minimal, right?

KVS85:
> ```diff
>  FROM alpine:3.10
> ...
> +    apk add --no-cache --repository http://dl-cdn.alpinelinux.org/alpine/v3.11/main/ nodejs npm && \
> ```
> 
> In our experience, mixing Alpine version packages like this ends up causing sometimes subtle (sometimes obvious) bugs that are difficult for users to diagnose. If you need Node.js from Alpine 3.11, why not just update the image to Alpine 3.11? Given it's a database image, and users won't often be doing `FROM arangodb:xxx`, the downstream effect should be pretty minimal, right?

We wanted to avoid raising Alpine version in stable release. Here NodeJS isn't required for main functionality: it's just for foxx-cli 2.0.0 ability. That's why we didn't care that much.

But yes, we will think about this in the next release.

PR #8393, opened 2020-07-21
Updating oraclelinux:8 and oraclelinux:8-slim images 

Djelibeybi:
This change is purely structural. The underlying tarball names have been made less ambiguous and documentation is retained within the generated image.

### DIFF
--- a/library/oraclelinux
+++ b/library/oraclelinux
@@ -4,10 +4,10 @@ GitCommit: 01a15ec99c7470a3391c691509db1759b41eaf66
 GitFetch: refs/heads/master
 # https://github.com/oracle/container-images/tree/dist-amd64
 amd64-GitFetch: refs/heads/dist-amd64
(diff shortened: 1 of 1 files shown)


tianon:
Wheeeee

Djelibeybi:
Could you check to see if the push got wedged again? I see the builds on https://doi-janky.infosiftr.net/job/multiarch/view/images/view/oraclelinux/ are finished, but Docker Hub is still serving the image(s) from 6 weeks ago.

tianon:
It was probably waiting for https://doi-janky.infosiftr.net/job/put-shared/job/light/job/oraclelinux/ -- we need to make that one auto-trigger more often (like after builds finish :sweat_smile:).

It should be up now.

Djelibeybi:
> It should be up now.

Yep, thanks!

PR #8399, opened 2020-07-21
Update Vault to 1.5.0

sarahethompson:


library/vault
```diff
@@ -3,9 +3,9 @@ Maintainers: Chris Hoffman <choffman@hashicorp.com> (@chrishoffman),
              Matthew Irish <matthew@hashicorp.com> (@meirish)
 GitRepo: https://github.com/hashicorp/docker-vault.git
 
-Tags: 1.5.0-rc
+Tags: 1.5.0
```
tianon, line 6: Did you want to also move `latest` to this tag?

cc @chrishoffman @jefferai @meirish

chrishoffman: Thank you for the review and attention to detail. This is now fixed!

PR #8409, opened 2020-07-22
Add liquibase as official image

r2-lf:
# Checklist for Review

    - https://github.com/liquibase
    - https://github.com/docker-library/docs/pull/1803
    - `FROM openjdk:xxx`


r2liquibase:
Hi, @tianon and @yosifkit ! Thanks for the insight via IRC. 

I think we're at a point to start working through those 9 issues now that the tests are successful. What is the process you'd like me to work through?

Thanks again!

Robert

r2liquibase:
Hey, gang! Got some feedback from Chad on this PR. How can I get these tests to start working? Thanks!!!

metcalfc:
@r2liquibase I looked at the failed test and it took me a bit to figure out that the test thats failing is: https://github.com/docker-library/official-images/blob/master/test/tests/override-cmd/run.sh 

Like we talked about, entrypoint is a PITA to override so what the official images have done is set a specific expectation for how this works. TIL, https://github.com/docker-library/official-images/blob/d28cb89e79417cac50c2a8ae163a9b3b79167f79/README.md#consistency

Short story is to make it work you'll need to write a `docker-enterpoint.sh` script. I'm not sure if there was one previously that you can just reuse? Its going to be a bit of project because you'll need to know the difference between folks asking for something liquidbase can do and literally anything else...

For example: https://github.com/docker-library/docker/blob/master/19.03-rc/docker-entrypoint.sh

Docker basically checks if the arg is actually docker or something that docker has a help command for and then anything else gets passed directly through.

Redis does something simpler and just checks for flags (any dash will do) and conf. https://github.com/docker-library/redis/blob/master/6.0/docker-entrypoint.sh

So write a `docker-entrypoint.sh` and do a little:

```
CMD ["--help"]
```

Since in your entrypoint script you detected flags and options the default `--help` flag would get picked up and this would all work.

This means you'll also want to make sure liquidbase is on the path. Easiest solution from `/liquidbase/liquidbase` would be you link the binary into /usr/local/bin.

r2liquibase:
Well, I guess that didn't solve the test case issue. Would love some feedback on how to fix this. Thanks, @metcalfc! Pretty sure we did as suggested, but obviously we didn't!

tianon:
You need to apply `chmod +x` in your local copy, then commit that change so that the script is executable in the container. :+1:

r2liquibase:
Thanks, @tianon ! I'm on it...

r2liquibase:
Neat! Tests pass now. Thanks!

r2liquibase:
Hi, @tianon and @metcalfc ! Would love to work on those 9 requriements. I think we're good on that last one so I updated it. Thanks!

tianon:
> - [ ] does it fit into one of the common categories? ("service", "language stack", "base distribution")
> - [ ] is it reasonably popular, or does it solve a particular use case well?

We talked about this a little bit on IRC, and I'm copying the logs here for posterity:

> `<tianon>` r2liquibase: ... from a cursory look at https://www.liquibase.org/, I'd also suggest you spend some time thinking about / articulating why liquibase makes sense as a separate Docker image, when it seems like it would be most useful / used within another application as part of the application startup (similar to "rake db:seed" and friends in other languages/frameworks)
> `<r2liquibase>` roger that. we thought the same thing. we released it as a lark and it turns out that a lot of folks are using it via docker. i think that is driven by people using more and more db's in docker containers for dev/testing purposes. so, if you're using mysql, say, in a docker container, you might not want to install liquibase on the host.
> `<r2liquibase>` ironically, liquibase came out of developer frustration that the db tools in ror weren't available for non-ruby projects.
> `<r2liquibase>` does that justification make sense?
> `<r2liquibase>` keep in mind, the liquibase team is willing to put in the work. don't want to task anyone else with that.
> `<tianon>` yeah, I suppose so -- I think "thrift" is *somewhat* of a parallel in that sense (although I don't think that Dockerization is going to be a particularly useful example for you)
> `<tianon>` it's somewhat unique in that users won't be using it as a long-running service, or even very often via "FROM liquibase"
> `<tianon>` it's really more of an interactive tool
> `<r2liquibase>` yeah, it's similar enough. thanks!
> `<tianon>` and we don't have a lot of examples of that category in the official images
> `<tianon>` everything typically falls into "service", "language stack", or "base image"
> `<r2liquibase>` i can see that, but we're seeing on our side a lot more use of utilities via docker.
> `<tianon>` even thrift is closer to "language stack" than liquibase is

So, I'm still on the fence as to whether this fits within the norm of the program generally, and I'd love to hear more opinions on this (perhaps @metcalfc has some additional thoughts?)

I think I'm open to accepting it conditionally, but I want to be clear that if we do so, this is a one-time exception (please don't open the floodgates of one-off tools here -- anyone doing so with this image as justification should be receiving a link to this comment :sweat_smile:).

---

> - [ ] does a [documentation](https://github.com/docker-library/docs/blob/master/README.md) PR exist? (should be reviewed and merged at roughly the same time so that we don't have an empty image page on the Hub for very long)

I'm not seeing a companion PR over at https://github.com/docker-library/docs for `liquibase` yet -- is that something that's still in-progress?  Seeing how you intend for users to use the image might be helpful on the previous point.

---

> ```
> Tags: 3.10.x, latest
> ```

This should be more specific (especially as what's currently proposed is actually 4.0.0); here's what I'd suggest:

```
Tags: 4.0.0, 4.0, 4, latest
```

(Depending on how useful it is for users to be able to pin to a particular bucket of "updates" -- is "Liquibase 4" a useful thing to say/use, or is "Liquibase 4.0" the more appropriate level of specificity?)

---

I have a few more notes on the Dockerization itself, but I think many of them are fair to be addressed in a later image update (and don't necessarily need to be part of an initial merge):

(I've bolded the items that I see as outright blockers in this list, for clarity.)

- `apt-get update` should usually be combined with `rm -rf /var/lib/apt/lists/*` so that the APT list files don't linger, since they end up getting stale quickly (and thus just add ~17MB of useless bloat to the image)

- `addgroup` and `adduser` should probably be done in a single layer, since they end up touching a similar set of files

- `chown liquibase /liquibase` / `COPY --chown=liquibase:liquibase ...` -- since `/liquibase` is where the actual software is installed, does it make sense for this to be owned by the unprivileged user?  Typically, we see maintainers instead install the software only modifiable by `root`, and make just the data directory user-modifiable, which also helps prevent the runtime from modifying the application files (likely less relevant here, since this is not a long-running service but more of a oneshot tool, but still relevant IMO).

- **`RUN GNUPGHOME="$(mktemp -d)"` -- having this in a line by itself doesn't actually do anything, because there's a new shell spawned for each `RUN` line; you'll need to either join all these `RUN` lines into a single `RUN` line, or add setting/cleanup to each (also, that value needs `export` if you want the `gpg` tool to pick it up)**

- **`gpg --batch --keyserver ha.pool.sks-keyservers.net --recv-keys ...` should use a *full* key fingerprint, not the short or long ID (for example, `79752DB6C966F0B8` becomes `0F07D1201BDDAB67CFB84EB479752DB6C966F0B8`, which appears to be an expired key)**

- **`gpg --batch --verify -fSLo` -- this looks like a copy/pasta error from a `curl` invocation; I don't think you intended to pass `-fSLo` to `gnupg`**

- we've recently updated https://github.com/docker-library/official-images#security to clarify that our *ideal* is both a checksum *and* GnuPG, since they verify different things (provenance vs transport/versioning)

- `RUN chmod 0755 /usr/local/bin/docker-entrypoint.sh` -- we've seen a lot of issues with this round-tripping through graph drivers properly due to the changed permissions; what you should do instead is commit the executable bit on the file in Git, so that what gets `COPY`-ed into the image can naturally be executable (Git will track the `+x` bit on the file properly)

- is there a particular reason to not put `/liquibase/liquibase` into the path somewhere, like a symlink in `/usr/local/bin/liquibase` so that users can start an interactive shell and just use `liquibase` directly without needing to know that it's under `/liquibase`?

r2liquibase:
Thanks, @tianon ! We're on it. Expect updates soon. Thanks!

r2liquibase:
I created a doc PR: https://github.com/docker-library/docs/pull/1803

r2liquibase:
Can we address the version numbers at a later date? I'm reticent to change this until we have an official image as the current users of liquibase/liquibase got a bit heated (rightfully so!) with the changes we have made to date. Frankly, I'd like to make those changes to the official image prior to users migrating over to that. Sound fair? Let me know your thoughts, @tianon .

Also, GREAT feedback on the other issues we should address. We appreciate you, @tianon !!!

dhwanitsshah:
Hi, @tianon any feedbacks/issues with the PR - https://github.com/liquibase/docker/pull/21 ?

tianon:
Running `apt-get clean` will be a no-op in all the Debian and Ubuntu base images -- it happens automatically (https://github.com/debuerreotype/debuerreotype/blob/d29dd5e030525d9a5d9bd925030d1c11a163380c/scripts/debuerreotype-minimizing-config#L85-L109).

dhwanitsshah:
> Running `apt-get clean` will be a no-op in all the Debian and Ubuntu base images -- it happens automatically (https://github.com/debuerreotype/debuerreotype/blob/d29dd5e030525d9a5d9bd925030d1c11a163380c/scripts/debuerreotype-minimizing-config#L85-L109).

Hi @tianon - indeed, have updated the code, thanks.

r2liquibase:
Hi, @tianon, @yosifkit , and @metcalfc ! What are the blockers on releasing this official image? We're ready to apply changes for better tagging and labeling, but we're reticent to make the changes to the liquibase/liquibase image. We'd prefer to make them immediately to the official image due to some feedback from the community. Happy to have a conversation about it!!!

tianon:
What I'd suggest is putting the source for "liquibase/liquibase" on a
separate branch (or separate Dockerfiles/directories) so that changes you
make for the official image no longer propagate to that other image, and so
that users only have to adapt to the changes in the official image when
they migrate to the new namespace.

r2liquibase:
Thanks, @tianon . Just made a change that we think will help. We've totally changed to using tags and I think this inline with what you are expecting.

r2liquibase:
@nvoxland is awesome!!! 

@tianon @yosifkit What do you think?

r2liquibase:
Hi, @tianon and @yosifkit . How can I help with this PR? Any questions? Happy to help!!!

tianon:
Thank you for your patience :pray:

I think there's still some items from https://github.com/docker-library/official-images/pull/8409#issuecomment-696340529 that aren't completely resolved here and/or aren't resolved consistently between the two proposed versions (several of which are definitely blockers).

```diff
+# Install GPG for package vefification
+RUN apt-get update \
+	&& apt-get -y install gnupg wget
```

> `apt-get update` should usually be combined with `rm -rf /var/lib/apt/lists/*` so that the APT list files don't linger, since they end up getting stale quickly (and thus just add ~17MB of useless bloat to the image)

```diff
+# Setup GPG
+RUN GNUPGHOME="$(mktemp -d)" 
```

> **`RUN GNUPGHOME="$(mktemp -d)"` -- having this in a line by itself doesn't actually do anything, because there's a new shell spawned for each `RUN` line; you'll need to either join all these `RUN` lines into a single `RUN` line, or add setting/cleanup to each (also, that value needs `export` if you want the `gpg` tool to pick it up)**

```diff
+    && gpg --batch --keyserver ha.pool.sks-keyservers.net --recv-keys 38F47D3E410C47B1 \
+    && gpg --batch --keyserver ha.pool.sks-keyservers.net --recv-keys 01B0B092A6925976 \
+    && gpg --batch --keyserver ha.pool.sks-keyservers.net --recv-keys A18EEA8E3426F280 \
+    && gpg --batch --keyserver ha.pool.sks-keyservers.net --recv-keys EC218558EABB25A1 \
+    && gpg --batch --keyserver ha.pool.sks-keyservers.net --recv-keys 79752DB6C966F0B8 \
```

> **`gpg --batch --keyserver ha.pool.sks-keyservers.net --recv-keys ...` should use a *full* key fingerprint, not the short or long ID (for example, `79752DB6C966F0B8` becomes `0F07D1201BDDAB67CFB84EB479752DB6C966F0B8`, which appears to be an expired key)**

```diff
+    && gpg --batch --verify -fSLo ...
```

> **`gpg --batch --verify -fSLo` -- this looks like a copy/pasta error from a `curl` invocation; I don't think you intended to pass `-fSLo` to `gnupg`**

```diff
+RUN wget -O ... \
+	&& wget -O ...asc \
+    && gpg --batch --keyserver ha.pool.sks-keyservers.net --recv-keys ... \
+    && gpg --batch --verify ...asc ...
```

> we've recently updated https://github.com/docker-library/official-images#security to clarify that our *ideal* is both a checksum *and* GnuPG, since they verify different things (provenance vs transport/versioning)

```diff
+RUN chmod 0755 /usr/local/bin/docker-entrypoint.sh
```

> `RUN chmod 0755 /usr/local/bin/docker-entrypoint.sh` -- we've seen a lot of issues with this round-tripping through graph drivers properly due to the changed permissions; what you should do instead is commit the executable bit on the file in Git, so that what gets `COPY`-ed into the image can naturally be executable (Git will track the `+x` bit on the file properly)

```diff
+    /liquibase/liquibase "$@"
```

> is there a particular reason to not put `/liquibase/liquibase` into the path somewhere, like a symlink in `/usr/local/bin/liquibase` so that users can start an interactive shell and just use `liquibase` directly without needing to know that it's under `/liquibase`?

---

Additionally, just noticed this one:

```diff
+# Download, verify, extract
+ARG LB_SHA256=ef8e0b8f7f0cabc34a61b8a1e7f4feac46652b6f5e62b58a9f2d2cfc98f0033f
+RUN set -x \
+  && wget -O liquibase-${LIQUIBASE_VERSION}.tar.gz "https://github.com/liquibase/liquibase/releases/download/v${LIQUIBASE_VERSION}/liquibase-${LIQUIBASE_VERSION}.tar.gz" \
+  && echo "$LB_SHA256  liquibase-${LIQUIBASE_VERSION}.tar.gz" | sha256sum -c - \
+  && tar -xzf liquibase-${LIQUIBASE_VERSION}.tar.gz
```

This downloads and extracts the tarball, and then leaves it in the image, so that's some really low hanging fruit for image size improvement (if you want to go deep down this rabbit hole, you could move the `gnupg`+`wget` installation down to where they're used so they can be removed in the same layer afterwards, but that's a much larger refactor).

r2liquibase:
Thank you @tianon . We got wrapped up into EOY business. We'll turn around an update shortly. Thanks again!!!

r2liquibase:
@tianon Thanks a ton for the feedback. We have incorporated it all except `RUN chmod 0755 /usr/local/bin/docker-entrypoint.sh`.

We have a number of developers on Windows that this broke for them. Setting the +x bit for them was not working on their machines. That is the only thing we did not incorporate. From your note, that sounded like a nice to have, not a must have. Cool?

tianon:
> We have a number of developers on Windows that this broke for them. Setting the +x bit for them was not working on their machines. That is the only thing we did not incorporate. From your note, that sounded like a nice to have, not a must have. Cool?

See https://github.com/docker/cli/blob/70a00157f161b109be77cd4f30ce0662bfe8cc32/cli/command/image/build.go#L418-L426 (which is the warning that would be displayed while doing `docker build` on a Windows box) -- the reason for this behavior in `docker build` is exactly for that use case of copying a script that's supposed to be executable already. :confused:

I think this is also missing some GitHub Actions (esp. the one that posts the updated `diff`), so if you can rebase your branch, that would be helpful in our re-review.

r2liquibase:
On it. Rebase was successful. 

Working on the docker-entrypoint.sh exec bit. 

Thanks!

Robert

r2liquibase:
Hi, @tianon ! Are we good for a merge? Looks like we've met all of Docker's requirements. Happy to fix any that we have not!!!

jandroav:
Hi @tianon ! 

We would to continue the process. I updated [library/liquibase](https://github.com/docker-library/official-images/pull/8409/files#diff-b0215954f46dbc6280ab24f06a60ee412ba186f0d2f030dfa3a68158ffa6ad4c) to match our current tags.

Is there any other work to be done here?
Thanks!

jandroav:
Hi @tianon , just in case you missed this I am adding @yosifkit

jnewton03:
ping @tianon @yosifkit .  Anything else required on our end?  Thanks!

jnewton03:
ping @tianon @yosifkit 😄

jnewton03:
another ping @tianon @yosifkit .  Anything I need to do here?

jnewton03:
@tianon @yosifkit please advise

jnewton03:
@mcred @r2-lf anything you can help with here?

mcred:
@deitch Any chance you know someone we can talk to about this? We've been trying for a long time to talk to the right people. There is a Sponsored OSS liquibase image, but it's very old, a fork from the main repo, and not supported by us. We haven't been able to get any help to make the official image, you know, official. 

Screenshot 2023-08-17 at 9 16 12 AM

deitch:
Wish I knew. Did you hit people up on Slack?

deitch:
Often @thaJeztah has ideas. He seems to know _everyone_.

mcred:
@deitch Is there a public slack group we can join? I'm not connected to any docker related slack group, yet.

deitch:
The docker community slack? I think it's here https://www.docker.com/community/

yosifkit:
Apologies for the delay. :bow:

Here are the initial items that I see in the [most recent diff](https://github.com/docker-library/official-images/pull/8409#issuecomment-1540303066).

```bash
apt-get upgrade -y
```

We recommend against using blanket package upgrades (`apt-get upgrade`/`apk upgrade`/`yum upgrade`/`yum update`) for official-images. When package upgrades are applied in a dependent image, it duplicates content of the base image, making the image larger than necessary. It also only delays the inevitable "there are outdated packages". The Official Images build pipeline only rebuilds on a update to the Dockerfile or a base image update, so we make periodic base image updates to then fully rebuild all dependent images (e.g., the [Debian](https://github.com/docker-library/official-images/pulls?q=is%3Apr+label%3Alibrary%2Fdebian) and [Ubuntu](https://github.com/docker-library/official-images/pulls?q=is%3Apr+label%3Alibrary%2Fubuntu) images are updated a least every 30 day).

---

Combine lines that affect the same files to minimize duplication in the container:

```dockerfile
RUN addgroup --gid 1001 liquibase \
	&& adduser --disabled-password --uid 1001 --ingroup liquibase liquibase
```

---

Since this is dealing with downloading, installing, and cleaning up a zip, it needs to be a single `RUN` in order for the cleanup to have an effect in final image size.

```diff
-RUN mkdir /liquibase/bin 
-RUN wget -q -O lpm.zip "https://github.com/liquibase/liquibase-package-manager/releases/download/v${LPM_VERSION}/lpm-${LPM_VERSION}-linux.zip"
-RUN unzip lpm.zip -d bin/
-RUN rm lpm.zip
+RUN mkdir /liquibase/bin \
+	&& wget -q -O lpm.zip "https://github.com/liquibase/liquibase-package-manager/releases/download/v${LPM_VERSION}/lpm-${LPM_VERSION}-linux.zip" \
+	&& unzip lpm.zip -d bin/ \
+	&& rm lpm.zip
```

---

This `RUN` line has no effect on the image and should be removed.

```dockerfile
RUN export LIQUIBASE_HOME=/liquibase
```

If `LIQUIBASE_HOME` is needed at runtime, then an `ENV` is the solution.

```dockerfile
ENV LIQUIBASE_HOME=/liquibase
```

---

```dockerfile
RUN lpm update
RUN /liquibase/liquibase --version
```

I'd recommend attaching these to the `RUN` lines that download the software so that if there is an incorrect download, then the docker layer doesn't get committed with a broken binary.

```dockerfile
RUN ... \
	&& rm liquibase-${LIQUIBASE_VERSION}.tar.gz \
	&& liquibase --version
```
```dockerfile
RUN ... \
	&& rm lpm.zip \
	&& lpm update
```

---

`lpm update`

> This mirrors `apt-get update` functionality.
>
> \-https://www.liquibase.com/blog/new-package-manager-utility

This populates a package cache in the image? Will this quickly get out of date just like `apt` package cache does? Would this be better for users to do when they need to use it?

---

```dockerfile
VOLUME /liquibase/classpath
VOLUME /liquibase/changelog
```

Does the image/application expect to modify these in a persistent way that needs to stick around from invocation to invocation? (Given the "permanence" of a `VOLUME` declaration, we recommend keeping them to just the directories the application stores active state in, and even more specifically active state that's actually necessary to keep if the container is completely destroyed and recreated, such as MySQL's `/var/lib/mysql` directory.)

---

These lines in the entrypoint should use `exec` so that the shell process doesn't stick around unnecessarily and get in the way of signals.

```diff
-    liquibase "$@"
+    exec liquibase "$@"

-    liquibase "--defaultsFile=/liquibase/liquibase.docker.properties" "$@"
+    exec liquibase "--defaultsFile=/liquibase/liquibase.docker.properties" "$@"
```

jandroav:
Hey, @yosifkit thanks a lot for reviewing this!

Since May we added some improvements and several recommendations were already applied. We have just updated `official-images/library/liquibase` to liquibase `4.23.1`. You can check the latest diff.

> `lpm update`
> This mirrors apt-get update functionality.
> -https://www.liquibase.com/blog/new-package-manager-utility
> This populates a package cache in the image? Will this quickly get out of date just like apt package cache does? Would this be better for users to do when they need to use it?

@jnewton03 and @mcred any thoughts on this?

> ```
> VOLUME /liquibase/classpath
> VOLUME /liquibase/changelog
> ```
> Does the image/application expect to modify these in a persistent way that needs to stick around from invocation to invocation? (Given the "permanence" of a VOLUME declaration, we recommend keeping them to just the directories the application stores active state in, and even more specifically active state that's actually necessary to keep if the container is completely destroyed and recreated, such as MySQL's /var/lib/mysql directory.)

- `/liquibase/classpath`: If drivers are not included or if using liquibase extensions, you can mount a local directory that contains JAR files to `/liquibase/classpath` and add JAR files to the classpath setting.

- ` /liquibase/changelog`: can be used for commands that write output like `liquibase generateChangeLog`

We added an `alpine` flavor. Can this be certified in this PR or does it need to be tracked in another one?

mcred:
@jandroav 

> lpm update
This mirrors apt-get update functionality.
-https://www.liquibase.com/blog/new-package-manager-utility
This populates a package cache in the image? Will this quickly get out of date just like apt package cache does? Would this be better for users to do when they need to use it?

We were using lpm to install some packages at one point in time, so we needed to update the package manifest. It's no longer needed, so I think it's find to remove the `lpm update` from the dockerfile.

jandroav:
Done @mcred 
cc: @yosifkit @jnewton03

jnewton03:
@yosifkit ⬆️

jnewton03:
@yosifkit @thaJeztah please advise.  thanks!

jnewton03:
@thaJeztah @deitch @yosifkit please advise.  Thank you.

jnewton03:
@thaJeztah @deitch @yosifkit ping

jnewton03:
@thaJeztah @deitch @yosifkit ping

thaJeztah:
Oh! Should've replied; I'm not a maintainer for the official images, so don't know if there's things missing to be addressed. @yosifkit and @tianon (currently on leave) may know

library/liquibase
```diff
@@ -0,0 +1,12 @@
...
+
+Tags: latest, latest-alpine
+GitFetch: refs/heads/main
+GitCommit: 838c2b21806320fbb0118b21a34f140064301d10
+Directory: .
+
+Tags: 4.23, 4.23.1, 4.23-alpine, 4.23.1-alpine
+GitFetch: refs/heads/main
+GitCommit: 838c2b21806320fbb0118b21a34f140064301d10
+Directory: .
```
LaurentGoderre, line 12: All of these tags point to the same thing. It should only be one entry like

```
Tags: 4.23, 4.23.1, 4.23-alpine, 4.23.1-alpine latest, latest-alpine
GitFetch: refs/heads/main
GitCommit: 838c2b21806320fbb0118b21a34f140064301d10
Directory: .
```

jnewton03: Done @LaurentGoderre and thank you for the review!

library/liquibase
```diff
@@ -0,0 +1,7 @@
+Maintainers: Jake Newton <docker@liquibase.com> (@jnewton03)
+GitRepo: https://github.com/liquibase/docker.git
+
+Tags: 4.23, 4.23.1, 4.23-alpine, 4.23.1-alpine, latest, latest-alpine
```
LaurentGoderre, line 4: Can you remove the `latest-alpine`. We usually only have one latest for the latest version of the default OS

LaurentGoderre:
Can you also squash into a single commit?

jnewton03:
> Can you also squash into a single commit?

Done.  Thanks, @mcred !

LaurentGoderre:
The build is now failing because the commit referenced is not in the ref branch. This commit would be better: https://github.com/liquibase/docker/commit/ea764a08f079fa2ffe24cc8d517bacf439adb670

jnewton03:
hi @LaurentGoderre !  Anything else you need here?

jnewton03:
ping @LaurentGoderre  😄

jnewton03:
@LaurentGoderre when do you think we could get this merged in?  Thanks!

jnewton03:
ping @LaurentGoderre

jnewton03:
ping @LaurentGoderre @thaJeztah @deitch @yosifkit @tianon

whalelines:
@jnewton03 thank you for your continued interest in providing a liquibase DOI. There appear to be a number of previously voiced concerns that have yet to be addresses in this pull request. Having enumerated them a few different times, we would ask that you please go through all of the above comments in detail and make another attempt to address those concerns. We will be notified when a new push is made to this PR and will re-review at that time.

jandroav:
Hey @whalelines we were taking a look at all the previous comments and we don't see any pending concerns to be resolved. Could you please be more specific so we can address them quickly? Thanks a lot!

whalelines:
Just looking quickly through https://github.com/docker-library/official-images/pull/8409#issuecomment-1769193041 I noticed:

-   `lpm update` is still run
-   the `VOLUME` statements are still there but your response about them indicates they are not necessary, see https://docs.docker.com/engine/reference/builder/#volume
-   The `--chown` still operates on the whole `/liquibase` directory

I would guess there are others. So please carefully go through the requested changes and ensure all of them are addressed. Once you push those changes, we will review again.

whalelines:
Thank you for making many of the requested changes. It still seems as though a few are still outstanding and, with some of the other changes you've made, some additional changes should be considered.

1.   Remove `TARGETARCH` as a build argument and use the output of `dpkg --print-architecture`.
2.   I would recommend moving the remaining `ARG` statements closer to their usage so you are not busting the build cache every time there is an update. In other words, move `LIQUIBASE_VERSION` and `LB_SHA256` after the `WORKDIR` statement and `LPM_VERSION` before the `RUN` statement that downloads and unpacks it.
3.   Add a checksum and check it for the lpm zip.
4.   Remove the `mkdir /liquibase && chown root:root /liquibase` from the `RUN` adding the user and group. It is not necessary (the `WORKDIR` statement creates the directory with the appropriate ownership).
5.   There is no need to symlink the `docker-entrypoint.sh` script to `/docker-entrypoint.sh`, remove command from the `RUN ln -s` commands.
6.   Set the group in the `USER` statement, `USER liquibase:liquibase`, to be more clear. (optional)
7.   Remove the `--chown` options from the `COPY` commands (see below).
8.   The order of the production stage has some issues. Since you set `USER` prior the `COPY` instructions, the contents of the `/liquibase` directory are still owned by the `liquibase` user. Similarly, the `--chown` options when `COPY`ing the entrypoint and properties files are unnecessary, and incorrect. As mentioned in a previous comment, the source code files should not be writable by the process using them. If there are locations that the process needs to write to, e.g., a log directory, then only those locations should be changed to be owned by the `liquibase` user.
9.   Another order issue is that the entrypoint and properties files will likely change less frequently than the files copied from the build stage, so it would be better to have the `COPY` statements for the entrypoint and properties files before the `COPY --from=builder`.

jandroav:
Thanks a lot @whalelines ! We went through all your comments and recommendations and updated this PR.

whalelines:
Thanks for the updates.

1.  You should put double quotes around the command substitution in the case statement, `"$(dpkg --print-architecture)".
2.  You have not moved the `USER` directive _after_ the `COPY` directives (item 8 in my [previous comment](https://github.com/docker-library/official-images/pull/8409#issuecomment-1885809753). If the liquibase process requires write access to paths under `/liquibase`, only those paths should be owned by the `liquibase` user.
3.  You have `-alpine` tags but the image appears to be Debian based. Please remove the `-alpine` tags.

jnewton03:
@whalelines Thanks so much!  I've got a PR open on our side to address items 1 and 2.  On item 3, we do have an alpine [Dockerfile](https://github.com/liquibase/docker/blob/main/Dockerfile.alpine) as well.  How do we get that to be an official image as well?

library/liquibase
```diff
@@ -0,0 +1,7 @@
+Maintainers: Jake Newton <docker@liquibase.com> (@jnewton03)
+GitRepo: https://github.com/liquibase/docker.git
+
+Tags: 4.25, 4.25.1, 4.25-alpine, 4.25.1-alpine, latest
+GitFetch: refs/heads/main
+GitCommit: 5e9438c2fdfabbce058946229e3df2bce958aa32
+Directory: .
```
whalelines, line 7: Presumably the `-alpine` tags would have a different branch, commit, and/or directory.

jandroav: @whalelines We addressed the latest items and we also added a new entry for `alpine` tags.

whalelines:
Thanks for the updates. There are a couple items with the latest version that should be addressed.

1.  I would recommend basing your Alpine image on `3.19` rather than `3.19.0` so the image gets updated when there are new patch releases in the 3.19 branch.
2.  `dpkg` will not be available in an Alpine-based image. You can use `"$(apk --print-arch)"`. Note the output for arm64 is aarch64, so you will need to update the case statement match pattern as well.

whalelines:
And please build and test the images to make sure they are working properly.

jandroav:
@whalelines done! Thanks for your latest recommendations.

jnewton03:
hi @whalelines , thanks so much for all of your guidance.  Is there anything else you need us to do?

whalelines:
Thanks for the latest round of changes. I think we are getting close.

My main request this time is that you build and run the images locally using the Dockerfiles and verify the build works and liquibase runs successfully. It seems to me that the build will fail on either amd64 or arm64 because a different lpm zip is downloaded for each architecture yet only one checksum is provided in the Dockerfile. So please, please, please build and run these four images (amd64/jammy, arm64/jammy, amd64/alpine, arm64/alpine) locally to make sure everything is working as you would expect.

After that, there are some minor things:

1.  The alpine image builder is based off of alpine and looking at the code there does not seem to be anything done in that stage that would require the JDK. Would it make sense to base the jammy builder off ubuntu:jammy?
2.  Since you set `WORKDIR /liquibase` before your `COPY` directives in the second stage, the copy destination could just be `./`. It is more clear as it currently is, but less future proof. If you don't anticipate the location of the `WORKDIR` changing, you could leave as is.

jandroav:
Hey @whalelines thanks for the new advice. We were able to make the changes and built all images locally:

**amd64/jammy**
image

**arm64/jammy**
image

**amd64/alpine**
image

**arm64/alpine**
image

We also tested that `liquibase` works by running some commands with all generated images i.e:

image

We had some issues changing the build to `ubuntu:jammy` while executing:

```
RUN wget -q -O liquibase-${LIQUIBASE_VERSION}.tar.gz "https://github.com/liquibase/liquibase/releases/download/v${LIQUIBASE_VERSION}/liquibase-${LIQUIBASE_VERSION}.tar.gz" && \
    echo "$LB_SHA256 liquibase-${LIQUIBASE_VERSION}.tar.gz" | sha256sum -c - && \
    tar -xzf liquibase-${LIQUIBASE_VERSION}.tar.gz && \
    rm liquibase-${LIQUIBASE_VERSION}.tar.gz
```

`wget`, `tar`, and `sha256sum` are correctly installed before running the command and we also tested the commands in a fresh `ubuntu:jammy` VM. Could we leave the builder as it is while we work on it?

yosifkit:
In Docker Official Images, we try hard to avoid multi-stage builds in almost every instance. Since this does a multi-stage build to basically prevent an package cleanup like `apt-get purge --auto-remove ...`, we would rather that it be just a regular non-multistage build. It will help make it clearer for users viewing docker history where things came from. See also the FAQ section on [multi-stage builds](https://github.com/docker-library/faq/tree/76862506fde541f964df742db3b5abe549d9beb6#multi-stage-builds).

There seems to be a couple dependencies installed in the build stage of both the Alpine-based and Debian-based images, but they don't seem to be used in the Dockerfile and are not installed in the final image (`gnupg`, `openrc`, `krb5-user`, `libpam-krb5`). Perhaps a single stage image could solve this for the ones needed at runtime, but anything unused should not be installed.

The `sha256sum` should either have a `*` or keep two spaces; see the `sha25sum` manpage:

> The default mode is to print a line with: checksum, a space, a character  indicating  input mode  ('\*'  for binary, ' ' for text or where binary is insignificant), and name for each FILE.
>
> Note: There is no difference between binary mode and text mode  on  GNU systems.

---

I've updated the Dockerfiles in a fork of `liquibase/docker` to show the changes I suggest: https://github.com/yosifkit/liquibase-docker/commit/b537a60cfa7bdd644f4e936b1c2944c6dbc9653e. I am happy to make it a PR; you are also welcome to just take the changes as well. Most of the changes are just to convert it to a single stage build (like moving the items that change less often to before the liquibase install). As far as the krb5 packages or `openrc`, I left those uninstalled like the current final stage images.

---

One more thing, to tell the build system that the Dockerfiles support amd64 and arm64v8, the `library/liquibase` needs an `Architectures: arm64v8, amd64` line added to the global section (up with the `GitRepo`) or can be added to both the other blocks so that the Alpine-based image architectures can be chosen independently from the temurin-based image. (Or can be added to the global section and later overridden for an image).

jandroav:
Hey @yosifkit !

Thanks a lot for the fork and the changes there.  We appreciate your work here and we are very happy to learn and apply best practices to our Dockerfiles.

We have pushed the latest changes.

whalelines:
It looks like maybe you missed the asterisk/extra space in the `sha256sum` command lines in the alpine variant.

jandroav:
Yes, @whalelines ! Done.

jnewton03:
hi @whalelines @yosifkit.  Thanks again for all of the help on this.  Anything else we should do?

yosifkit:
The only thing to change/remove is the `latest-alpine` tag (probably to just `alpine`).

> Incorrectly formatted 'latest' tags detected:
>  liquibase:latest-alpine
>
> Read https://github.com/docker-library/official-images#tags-and-aliases for more details.

LGTM otherwise

jandroav:
Hi, @yosifkit @whalelines we did the change, and `liquibase/liquibase:alpine` has been published.

Thanks!

jnewton03:
hi @yosifkit @whalelines.  Thanks for all of the assistance here.  Anything else we can do to get this merged?

jnewton03:
hi @yosifkit @whalelines.  Just checking in to see what else you need from us. Thanks!

jandroav:
hi @tianon @yosifkit just pinging you in case we need to make more changes to get this done. Thanks!

tianon:
I promise you're in the queue.  Please be patient.

There are a few really minor things, but overall this looks reasonable.

Here are the important ones (in order of importance); I've pushed them all as a commit at https://github.com/liquibase/docker/commit/15b6b4dda3aadc688f7bfc798a10f4a2649abf0d if that's easier to integrate (`git fetch origin 15b6b4dda3aadc688f7bfc798a10f4a2649abf0d && git cherry-pick 15b6b4dda3aadc688f7bfc798a10f4a2649abf0d`):

---

`apk --print-architecture` is not a valid flag (https://github.com/liquibase/docker/actions/runs/7918608937/job/21617433482#step:4:105 -> "apk: unrecognized option: print-architecture"), so this just happens to complete "successfully" because the `case` statement isn't coded very defensively and as a result falls through to downloading the amd64 artifact, and you might notice that the `lpm` binary in your Alpine-based arm64 images is actually an amd64 binary as a result.  Additionally, the strings being matched are not correct in either case for Alpine (should be `x86_64` and `aarch64` for APK architectures strings).

Here's what I'd suggest instead (explicit `apk --print-arch` as a separate command so its failure is captured correctly, fixed `case` strings, explicit failure case for unknown/unsupported architectures):

```diff
@@ -29,19 +32,18 @@ ARG LPM_SHA256_ARM=375acfa1e12aa0e11c4af65e231e6471ea8d5eea465fb58b516ea2ffbd18f
 # Download and Install lpm
 RUN mkdir /liquibase/bin && \
     apk add --no-cache --virtual .fetch-deps wget unzip && \
-    case "$(apk --print-architecture)" in \
-      *amd64*)  DOWNLOAD_ARCH=""  ;; \
-      *arm64*)  DOWNLOAD_ARCH="-arm64" && LPM_SHA256=$LPM_SHA256_ARM  ;; \
+    arch="$(apk --print-arch)" && \
+    case "$arch" in \
+      x86_64)   DOWNLOAD_ARCH=""  ;; \
+      aarch64)  DOWNLOAD_ARCH="-arm64" && LPM_SHA256=$LPM_SHA256_ARM  ;; \
+      *) echo >&2 "error: unsupported architecture '$arch'" && exit 1 ;; \
     esac && wget -q -O lpm-${LPM_VERSION}-linux${DOWNLOAD_ARCH}.zip "https://github.com/liquibase/liquibase-package-manager/releases/download/v${LPM_VERSION}/lpm-${LPM_VERSION}-linux${DOWNLOAD_ARCH}.zip" && \
```

```diff
@@ -30,10 +33,12 @@
     apt-get -yqq install unzip --no-install-recommends && \
     rm -rf /var/lib/apt/lists/* && \
     mkdir /liquibase/bin && \
-    case "$(dpkg --print-architecture)" in \
-      *amd64*)  DOWNLOAD_ARCH=""  ;; \
-      *arm64*)  DOWNLOAD_ARCH="-arm64" && LPM_SHA256=$LPM_SHA256_ARM ;; \
+    arch="$(dpkg --print-architecture)" && \
+    case "$arch" in \
+      amd64)  DOWNLOAD_ARCH=""  ;; \
+      arm64)  DOWNLOAD_ARCH="-arm64" && LPM_SHA256=$LPM_SHA256_ARM ;; \
+      *) echo >&2 "error: unsupported architecture '$arch'" && exit 1 ;; \
     esac && wget -q -O lpm-${LPM_VERSION}-linux${DOWNLOAD_ARCH}.zip "https://github.com/liquibase/liquibase-package-manager/releases/download/v${LPM_VERSION}/lpm-${LPM_VERSION}-linux${DOWNLOAD_ARCH}.zip" && \
```

(I'd even suggest going further and not using `LPM_SHA256` as a fallback at all so that every part of this becomes explicit, but that's not necessary -- however, adding `set -e` and/or `set -u` or just `set -eux` to all these `RUN` lines would be strongly recommended, especially as it would've caught the missing `DOWNLOAD_ARCH` fallback)

---

Adding validation that the commands were not downloaded successfully, but _work_ would also help here; I'd suggest something like this:

```diff
@@ -20,7 +20,10 @@ RUN set -x && \
     echo "$LB_SHA256 *liquibase-${LIQUIBASE_VERSION}.tar.gz" | sha256sum -c - && \
     tar -xzf liquibase-${LIQUIBASE_VERSION}.tar.gz && \
     rm liquibase-${LIQUIBASE_VERSION}.tar.gz && \
-    apk del --no-network .fetch-deps
+    apk del --no-network .fetch-deps && \
+    ln -s /liquibase/liquibase /usr/local/bin/liquibase && \
+    ln -s /liquibase/docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh && \
+    liquibase --version
 
@@ -38,12 +41,9 @@ RUN mkdir /liquibase/bin && \
     echo "$LPM_SHA256 *lpm-${LPM_VERSION}-linux${DOWNLOAD_ARCH}.zip" | sha256sum -c - && \
     unzip lpm-${LPM_VERSION}-linux${DOWNLOAD_ARCH}.zip -d bin/ && \
     rm lpm-${LPM_VERSION}-linux${DOWNLOAD_ARCH}.zip && \
-    apk del --no-network .fetch-deps
-
-# Setup symbolic links
-RUN ln -s /liquibase/docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh && \
-    ln -s /liquibase/liquibase /usr/local/bin/liquibase && \
-    ln -s /liquibase/bin/lpm /usr/local/bin/lpm
+    apk del --no-network .fetch-deps && \
+    ln -s /liquibase/bin/lpm /usr/local/bin/lpm && \
+    lpm --version
 
```

```diff
@@ -19,7 +19,10 @@ ARG LB_SHA256=46850b5fd21c548f969253cbbc97dc6c846198a8225581e3af5346ac8aa7dbf2
 RUN wget -q -O liquibase-${LIQUIBASE_VERSION}.tar.gz "https://github.com/liquibase/liquibase/releases/download/v${LIQUIBASE_VERSION}/liquibase-${LIQUIBASE_VERSION}.tar.gz" && \
     echo "$LB_SHA256 *liquibase-${LIQUIBASE_VERSION}.tar.gz" | sha256sum -c - && \
     tar -xzf liquibase-${LIQUIBASE_VERSION}.tar.gz && \
-    rm liquibase-${LIQUIBASE_VERSION}.tar.gz
+    rm liquibase-${LIQUIBASE_VERSION}.tar.gz && \
+    ln -s /liquibase/liquibase /usr/local/bin/liquibase && \
+    ln -s /liquibase/docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh && \
+    liquibase --version
 
@@ -39,12 +42,9 @@ RUN apt-get update && \
     echo "$LPM_SHA256 *lpm-${LPM_VERSION}-linux${DOWNLOAD_ARCH}.zip" | sha256sum -c - && \
     unzip lpm-${LPM_VERSION}-linux${DOWNLOAD_ARCH}.zip -d bin/ && \
     rm lpm-${LPM_VERSION}-linux${DOWNLOAD_ARCH}.zip && \
-    apt-get purge -y --auto-remove unzip
-
-# Setup symbolic links
-RUN ln -s /liquibase/docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh && \
-    ln -s /liquibase/liquibase /usr/local/bin/liquibase && \
-    ln -s /liquibase/bin/lpm /usr/local/bin/lpm
+    apt-get purge -y --auto-remove unzip && \
+    ln -s /liquibase/bin/lpm /usr/local/bin/lpm && \
+    lpm --version
 
```

---

Switching the Ubuntu-based variant from the `adduser`/`addgroup` wrappers to the lower level `useradd`/`groupadd` commands will clean up some errorneous output (https://github.com/liquibase/docker/actions/runs/7918608937/job/21617432684#step:4:63; yes, confusingly Alpine is the other way around because `adduser` and `addgroup` there come from BusyBox, so should be left as-is):

```diff
@@ -2,8 +2,8 @@
 FROM eclipse-temurin:17-jre-jammy
 
 # Create liquibase user
-RUN addgroup --gid 1001 liquibase && \
-    adduser --disabled-password --uid 1001 --ingroup liquibase liquibase
+RUN groupadd --gid 1001 liquibase && \
+    useradd --uid 1001 --gid liquibase liquibase
 
```

tianon:

jandroav:
Hey @tianon, we added your recommendations on Feb 26. We also updated this PR with our latest `4.27.0` release.

tianon (approved):
Thanks, looks good after the latest update :+1:

I left some documentation notes at https://github.com/docker-library/docs/pull/1803#pullrequestreview-1987336609, which I think is all that's blocking this now :+1:

jandroav:
Thanks @tianon , I updated the docs PR with the fixes.

jandroav:
Hi @tianon thanks for all your work here. Now that we are tagged as an official image we noticed that there are two separate repositories for Liquibase Docker images: https://hub.docker.com/_/liquibase and https://hub.docker.com/r/liquibase/liquibase 

Could you please provide some clarification on the difference between these two repositories? Can they be unified? Is there a specific reason for having two separate repositories for Liquibase images?

Additionally, we are wondering if there are any implications in our CICD pipeline. Do we need to choose one repository over the other for specific reasons? Any good practices here? We also would like not to lose the download count metric (`10M+`) shown here https://hub.docker.com/r/liquibase/liquibase

Thanks!

whalelines:
@jandroav , there are actually [quite a few more than two liquibase related images](https://hub.docker.com/search?q=liquibase&source=) and we do not control namespaces on Docker Hub other than the library namespace.

Can I deduce from your question that [Liquibase](https://www.liquibase.com/) does not control the [liquibase organization on Docker Hub](https://hub.docker.com/u/liquibase)?

jandroav:
@whalelines we own the Liquibase org. We are publishing our liquibase/liquibase image to https://hub.docker.com/r/liquibase/liquibase using this user https://hub.docker.com/u/liquibase in https://github.com/liquibase/docker/blob/main/.github/workflows/create-release.yml

Our latest release matches this:

image

image

whalelines:
> So, from now instead of doing our CI, we only need to update this https://github.com/docker-library/official-images/blob/master/library/liquibase and https://doi-janky.infosiftr.net/ will push it, right?

Correct, you will open a PR to update the library/liquibase file. Once it is reviewed, approved, and merged, the DOI build system will create the images and push them to Docker Hub.

>It seems we need to update our docs to reflect that the correct pull command is `docker pull liquibase` instead of the old `docker pull liquibase/liquibase`, correct?

That is correct. You can also update the documentation on https://hub.docker.com/r/liquibase/liquibase to point to https://hub.docker.com/_/liquibase .

It is probably a good idea to maintain the `liquibase/liquibase` image to ensure users who have not updated their Dockerfiles and scripts continue to get an up-to-date image.

jandroav:
Thanks for your comments @whalelines. We will do it.

jnewton03:
> > So, from now instead of doing our CI, we only need to update this https://github.com/docker-library/official-images/blob/master/library/liquibase and https://doi-janky.infosiftr.net/ will push it, right?
> 
> Correct, you will open a PR to update the library/liquibase file. Once it is reviewed, approved, and merged, the DOI build system will create the images and push them to Docker Hub.
> 
> > It seems we need to update our docs to reflect that the correct pull command is `docker pull liquibase` instead of the old `docker pull liquibase/liquibase`, correct?
> 
> That is correct. You can also update the documentation on https://hub.docker.com/r/liquibase/liquibase to point to https://hub.docker.com/_/liquibase .
> 
> It is probably a good idea to maintain the `liquibase/liquibase` image to ensure users who have not updated their Dockerfiles and scripts continue to get an up-to-date image.

hi @whalelines We're trying to determine if we should continue maintaining and updating both the liquibase/liquibase image AND the official image or if we should stop updating the liquibase/liquibase image and _only_ update the official image.  Do you have any examples of how other companies have handled this transition in the past?  We'd like to follow best practices.

One thought we had was to have a cutover version where for versions 4.27.0 and earlier you can use liquibase/liquibase but for anything 4.28.0 and newer, to use the official image.

tianon:
Yeah, I think you've got the right idea that it's really a communication problem -- having a specific version at which you'll stop updating the other image is a very sane way to handle that (straightforward for users to understand), although our experience is that getting them to _notice_ the communication can sometimes be a challenge.  An approach we've used for that in the past has been deleting generic tags like `latest` so that users who aren't pinning to an explicit older version get a 404 as a signal that they should check the repository and hopefully see the notice that the images are in a new place now.  (I don't think there's "one" standard flow that's best because each company's situation and userbase is inevitably a little bit different.)

LaurentGoderre:
From a Hub perspective you could do something like this: https://hub.docker.com/r/mattermost/mattermost-prod-db

PR #8411, opened 2020-07-22
Telegraf 1.15.0

reimda:
Updating telegraf version, debian base image, alpine base image.


reimda:
Found an issue with packaging

PR #8417, opened 2020-07-23
Update buildpack-deps

tianon:
Changes:

- https://github.com/docker-library/buildpack-deps/commit/3229203: Remove Ubuntu Eoan (EOL)


tianon:
(Deletions only, merging)

PR #2097, opened 2016-08-27
Improve redis-basics-persistent by actually checking

TimWolla:
I'm not entirely sure what redis-basics-persistent should check, but it currently does not check whether redis persists something. So I thought I'd recycle that test.

I removed the building of the new image, as setting the `appendonly --yes` parameter no `dump.rdb` is created on exit.

If you want I can create an entirely new test for this as well.

see docker-library/redis#4, https://github.com/docker-library/redis/pull/71#issuecomment-241876353

PR #8439, opened 2020-07-27
Update spiped:alpine to alpine:3.12

TimWolla:


TimWolla:
Somehow I must have lost track of Alpine releases for that one :pensive:

TimWolla:
Apparently GH Actions did not run the spiped-basics test. Did I do anything wrong / did I miss a location to adjust?

tianon:
Yeah, it needs to be added to the `spiped` image in `test/config.sh`:

https://github.com/docker-library/official-images/blob/530bbdd66c3c44cd6fc0a172877f5c2c86364f35/test/config.sh#L232-L238

TimWolla:
Meh. I made the change locally, but apparently forgot to `git add` the `config.sh` :facepalm:

test/tests/spiped-basics/run.sh
```diff
@@ -0,0 +1,59 @@
...
+# ensure the clientImage is ready and available
+if ! docker image inspect "$clientImage" &> /dev/null; then
+	docker pull "$clientImage" > /dev/null
+fi
+
+# Create an instance of the container-under-test
+serverImage="$("$dir/../image-name.sh" librarytest/spiped "$image")"
+"$dir/../docker-build.sh" "$dir" "$serverImage" <<EOD
+FROM $image
+COPY dir/keyfile /spiped/key
```
yosifkit, line 21: Why commit an opaque keyfile when we could just generate one when running the test?

ala `RUN dd if=/dev/urandom bs=32 count=1 of=keyfile`

---

or maybe just `docker run ... sh -c "spiped-generate-key.sh && exec docker-entrypoint.sh -d -s ..."` with share for the key?

TimWolla: I implemented it like that, because I remember from here that using the filesystem / bind mounts is unsafe in context of the tests: https://github.com/docker-library/official-images/pull/2097#issuecomment-244195736

But of course I should be able to replace the `COPY` with a `RUN` that statically bakes the keyfile into the created image.

yosifkit: Yeah, we can't mount from the host, but `volumes-from` or a named volume with a `$RANDOM` element that gets cleaned up on `EXIT` should work fine.

test/tests/spiped-basics/run.sh
```diff
@@ -0,0 +1,59 @@
...
+EOD
+cid_d="$(docker run -d "$serverImage" -d -s '[0.0.0.0]:8080' -t 'example.com:80')"
+cid_e="$(docker run --link "$cid_d":spiped_d -d "$serverImage" -e -s '[0.0.0.0]:80' -t 'spiped_d:8080')"
+trap "docker rm -vf $cid_d > /dev/null" EXIT
+trap "docker rm -vf $cid_e > /dev/null" EXIT
```
yosifkit, line 26: Traps on the same signal replace the old `trap`, so the later have to include everything any previous `trap` does.

TimWolla: Thanks, even without newer traps overriding older ones not doing this in a single call is dumb. I've fixed that.

test/tests/spiped-basics/run.sh
TimWolla: I've opted for `--volumes-from`.

test/tests/spiped-basics/run.sh
```diff
@@ -0,0 +1,56 @@
...
+image="$1"
+
+clientImage='buildpack-deps:buster-curl'
+# ensure the clientImage is ready and available
+if ! docker image inspect "$clientImage" &> /dev/null; then
+	docker pull "$clientImage" > /dev/null
+fi
+
+# The keygen below is a bit ugly, because spiped-generate-key.sh expects /spiped/key to be a directory (you can't bind mount a non-existing file),
+# but the entrypoint expects /spiped/key to be the actual keyfile.
```
yosifkit, line 18: It seems like the extra complexity described here is because either `spiped-generate-key.sh` or `docker-entrypoint.sh` should be adjusted, but :shrug: that's your choice.

TimWolla: Yeah, it should. But I don't want to break compatibility here and I'm not sure what the correct solution would be. I certainly don't want to expect the users bind mounting a full folder containing a single keyfile vs. just the keyfile.

test/tests/spiped-basics/run.sh
```diff
@@ -0,0 +1,56 @@
...
+# So we first symlink /spiped/key to some directory, then generate the keyfile and then replace the symlink by the generated keyfile.
+cid_keygen="$(docker run -d "$image" sh -c 'ln -s /tmp /spiped/key && spiped-generate-key.sh && rm /spiped/key && mv /tmp/spiped-keyfile /spiped/key')"
+cid_d="$(docker run --volumes-from="$cid_keygen" -d "$image" -d -s '[0.0.0.0]:8080' -t 'example.com:80')"
+cid_e="$(docker run --volumes-from="$cid_keygen" --link "$cid_d":spiped_d -d "$image" -e -s '[0.0.0.0]:80' -t 'spiped_d:8080')"
+trap "docker rm -vf $cid_keygen $cid_d $cid_e > /dev/null" EXIT
```
tianon, line 23: If any of these `docker run` lines fail, this will leave dangling containers, so it would probably be good to (later) update this to do a `trap` after each one:

```suggestion
cid_keygen="$(docker run -d "$image" sh -c 'ln -s /tmp /spiped/key && spiped-generate-key.sh && rm /spiped/key && mv /tmp/spiped-keyfile /spiped/key')"
trap "docker rm -vf $cid_keygen > /dev/null" EXIT
cid_d="$(docker run --volumes-from="$cid_keygen" -d "$image" -d -s '[0.0.0.0]:8080' -t 'example.com:80')"
trap "docker rm -vf $cid_keygen $cid_d > /dev/null" EXIT
cid_e="$(docker run --volumes-from="$cid_keygen" --link "$cid_d":spiped_d -d "$image" -e -s '[0.0.0.0]:80' -t 'spiped_d:8080')"
trap "docker rm -vf $cid_keygen $cid_d $cid_e > /dev/null" EXIT
```

PR #8440, opened 2020-07-27
Add spiped-basics test

TimWolla:
This test is based on the existing haproxy-basics test and [the Travis configuration in the origin repository](https://github.com/TimWolla/docker-spiped/blob/master/.travis.yml).

On my local machine it worked on the first attempt. Something's fishy there. Please carefully review for dumbness.


TimWolla:
It appears that GH Actions is not terribly happy with a test only PR: https://github.com/docker-library/official-images/actions/runs/184869600 and I also received an email about that.

tianon:
The easy fix would be to make this part of #8439 :sweat_smile:

tianon:
(Then it'll actually test those updates)

TimWolla:
> The easy fix would be to make this part of #8439

I did this now, but I like creating a single PR per distinct change :innocent:

PR #8448, opened 2020-07-28
Update rabbitmq

tianon:
Changes:

- https://github.com/docker-library/rabbitmq/commit/a999b78: Update to 3.8.6-rc.1, Erlang/OTP 23.0.3, OpenSSL 1.1.1g


itmecho:
Is it normal for these to get hung up on a pending github action? I'm waiting on this image for some bug fixes

yosifkit:
> Is it normal for these to get hung up on a pending github action?

Yes and no. Hung github actions do not happen that often and for this one, it is stuck on a version that is not being updated, so I feel fine in merging it.

PR #8459, opened 2020-07-30
Add Java 11 release for Flink 1.11.1

rmetzger:
This effectively adds 2 Java11-based images.
Since we've changed the tool that generates the the stackbrew library file, the order of the entries has changed. Future releases of Flink should not create such noisy pull requests.


rmetzger:
I will fix the issue reported in the check...

rmetzger:
Sorry for the delay. The pull request is now ready to review.

PR #8469, opened 2020-07-30
Update Consul to 1.8.1/1.7.5/1.6.7

kyhavlov:


tianon:

mikemorris:
@tianon We migrated from what appeared to be a [deprecated](https://docs.docker.com/engine/reference/builder/#maintainer-deprecated) `MAINTAINER` in https://github.com/hashicorp/docker-consul/pull/149, should this is reverted or is there a preferred alternative?

tianon:
From https://github.com/opencontainers/image-spec/blob/v1.0.1/annotations.md#pre-defined-annotation-keys:

> Keys SHOULD be named using a reverse domain notation - e.g. `com.example.myKey`.

If you want to use a standard well-defined label for this same bit of metadata, I'd suggest `org.opencontainers.image.authors` (but honestly, if it were me, I'd just remove the label entirely since I don't see the added value, and since it's auto-inherited by any image that's `FROM` this one and doesn't overwrite the value).

mikemorris:
@tianon How's this look now?

tianon:
Much better :+1:

PR #8493, opened 2020-08-04
[ros] bump package version for foxy and rolling ros1-bridge image

mikaelarguedas:


yosifkit:

PR #8497, opened 2020-08-04
Update docker

tianon:
Changes:

- https://github.com/docker-library/docker/commit/184d0e7: Update to 19.03.13-beta2


yosifkit:

PR #8498, opened 2020-08-04
Update rabbitmq

tianon:
Changes:

- https://github.com/docker-library/rabbitmq/commit/0c7faaf: Update to 3.8.6, Erlang/OTP 23.0.3, OpenSSL 1.1.1g


yosifkit:

PR #8499, opened 2020-08-04
Update ghost

tianon:
Changes:

- https://github.com/docker-library/ghost/commit/437aff1: Update to 3.28.0, ghost-cli 1.14.1


yosifkit:

PR #8500, opened 2020-08-04
Update pypy

tianon:
Changes:

- https://github.com/docker-library/pypy/commit/7823435: Update to 7.3.1, pip 20.2.1
- https://github.com/docker-library/pypy/commit/dbe6fb9: Handle compressed responses from downloads.python.org when scraping versions


yosifkit:

PR #8505, opened 2020-08-06
Fixed permissions bug in Neo4j 4.0.7 and 4.1.1

jennyowen:
I fixed a file permissions bug that I'd created in Neo4j 4.0.7 and 4.1.1 caused by different ownership defaults between host file mounting, named volumes and whatever docker-compose does by default.

Oh, and thanks for the `gosu` tool tianon, it's super useful!


tianon:
Glad you're seeing usefulness out of it! :+1:

PR #8509, opened 2020-08-06
[Swift] Update Docker images to Swift 5.2.5 and add CentOS 7 support

shahmishal:


tianon:
I'm not sure how we missed it previously, but `yum -y update` shouldn't be used in `Dockerfile`s -- we recommend instead asking the base image maintainers to update their images if there are packages in the base image that need to be updated (both because often the packages from the base image will fail to update, but more importantly because it severely bloats the size of the resulting images by including many of the same files twice).

(This is discussed in https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#apt-get, https://github.com/docker-library/faq#why-does-my-security-scanner-show-that-an-image-has-cves, and somewhat in https://github.com/docker-library/official-images#library-definition-files.)

shahmishal:
@tianon Thanks for the feedback, I updated the Dockerfiles.

tianon:
LGTM, thanks :+1:

PR #8517, opened 2020-08-09
Update the CentOS images for August 2020

bstinsonmhk:
Respin CentOS container images for updated content as of August 9th, 2020


bstinsonmhk:
FYI: I'll have a PR soon to clean up some of these old tags

PR #8518, opened 2020-08-09
jobber: Update from 1.4.3 to 1.4.4

dshearer:


tianon:
@dshearer I'm sorry to find https://github.com/dshearer/jobber/commit/d226a5ece9a199b95bc2b7695cfb99edb34f5ee5, but I'm wondering if it makes sense to put something similar on the image description?  It's not *quite* the same as the `deprecated.md` that we've added on several other images (since it's not deprecated), but it *is* unmaintained, so maybe "deprecated" is appropriate?

PR #8530, opened 2020-08-11
Update Solr to 8.6.0

dsmiley:
See [Announcement](https://lists.apache.org/thread.html/rd0bf6ba079d86bb7af5f42ebb5990cceef769b61272b929aa0b289cc%40%3Csolr-user.lucene.apache.org%3E) which links to the release notes which further links to [upgrade considerations](https://github.com/apache/lucene-solr/blob/master/solr/solr-ref-guide/src/solr-upgrade-notes.adoc).

One the docker-solr side, we now include [jattach](https://github.com/apangin/jattach) to get some basic tooling equivalents to what the JDK provides.

### DIFF
--- a/library/solr
+++ b/library/solr
@@ -6,12 +6,22 @@ Maintainers: The Apache Lucene/Solr Project <solr-user@lucene.apache.org> (@asfb
              Jan Høydahl (@janhoy)
 GitRepo: https://github.com/docker-solr/docker-solr.git
 
-Tags: 8.5.2, 8.5, 8, latest
+Tags: 8.6.0, 8.6, 8, latest
+Architectures: amd64, arm64v8
+GitCommit: d86046a9ea94e91ee494c86a9bae4b6b5a44c64d
+Directory: 8.6
+
+Tags: 8.6.0-slim, 8.6-slim, 8-slim, slim
+Architectures: amd64, arm64v8
+GitCommit: d86046a9ea94e91ee494c86a9bae4b6b5a44c64d
+Directory: 8.6/slim
+
+Tags: 8.5.2, 8.5
 Architectures: amd64, arm64v8
 GitCommit: 86159575c4b6da2010ae9ee003bf5a6de31157cb
 Directory: 8.5
 
-Tags: 8.5.2-slim, 8.5-slim, 8-slim, slim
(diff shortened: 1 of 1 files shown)


library/solr
```diff
@@ -6,12 +6,22 @@ Maintainers: The Apache Lucene/Solr Project <solr-user@lucene.apache.org> (@asfb
...
+GitCommit: d86046a9ea94e91ee494c86a9bae4b6b5a44c64d
+Directory: 8.6/slim
+
+Tags: 8.5.2, 8.5
 Architectures: amd64, arm64v8
 GitCommit: 86159575c4b6da2010ae9ee003bf5a6de31157cb
 Directory: 8.5
 
-Tags: 8.5.2-slim, 8.5-slim, 8-slim, slim
+Tags: 8.5.2-slim, 8.5-slim
```
tianon, line 24: Any particular reason to keep this version (and 8.4.1, 8.3.1, 8.2.0, 8.1.1, 8.0.0, 7.7.3, 6.6.6, and 5.5.5) listed here?  Removing tags here will remove them from the "Supported" section on the Hub readme (and will prevent us from spending cycles rebuilding them on the official build servers), but the tags will still be available to users who want them.  (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)

janhoy: Are you saying that the ones we don't list in this file will not be re-built, but the (stale) existing image will still be available, but remain on old Linux and Java versions, which may have vulnerabilities? @dsmiley please chime in here, but to me it makes sense to at least list the versions that are still supported by the project, i.e. 8.x and 7.7.x. If we cut out anything older than 7.7.x, we should add a warning notice somewhere that these versions are stale, unsupported and likely not secure.

Can we push the 8.6 release as-is for now, and then we re-consider for our next release which to "support"? I can create a github issue for it.

tianon: Yes, I'm saying if 8.x and 7.7.x and what's officially supported by the project, what I'm saying is that those should be the only versions listed here (since they're the only ones users are "recommended" to use as they're actually actively supported), but some users still need the older versions/tags, so we don't actively delete them unless the image maintainer explicitly requests that we do so.

We'll merge this for now and let you discuss separately. :+1:

PR #8548, opened 2020-08-13
eclipse-mosquitto: Don't abort on chmod error.

ralight:
If the filesystem is read only this shouldn't cause a quit, for example.


tianon:
Instead of `set +e`, why not just add `|| true` (or the equivalent `|| :`) to the end of the `chown` line?

```bash
[ -d "/mosquitto" ] && chown -R mosquitto:mosquitto /mosquitto || :
```

ralight:
If you think that's clearer, I'm reasonable indifferent :)

PR #8565, opened 2020-08-17
Aerospike 5.1.0.5

volmarl:


yosifkit:
I am a little confused about the `-dev` package, since they are usually just header files. It seems that `libcurl4-openssl-dev` also [contains a `libcurl.so`](https://packages.debian.org/stretch/amd64/libcurl4-openssl-dev/filelist), but there is also `libcurl.so.3` and `libcurl.so.4` in the [`libcurl3` package](https://packages.debian.org/stretch/amd64/libcurl3/filelist).

PR #8585, opened 2020-08-20
Fix python installation in CrateDB 3.3.5.

kovrus:
The 3.3.5 image was at some point rebuilt with the newer base image
which broke the python installation. Consequently, CrateDB crash CLI
was also not possible to use.


yosifkit:
> Tags in the library file are only built through an update to that library file or as a result of its base image being updated (ie, an image `FROM debian:jessie` would be rebuilt when `debian:jessie` is built).

Or in your case, `centos:7` (updated in https://github.com/docker-library/official-images/pull/8517)

LGTM

PR #8616, opened 2020-08-27
maven: install find and add openjdk-16

carlossg:
fixes https://github.com/carlossg/docker-maven/issues/165

### DIFF
--- a/library/maven
+++ b/library/maven
@@ -18,7 +18,7 @@ Directory: openjdk-11-slim
 
 Tags: 3.6.3-jdk-14, 3.6-jdk-14, 3-jdk-14
 Architectures: amd64
-GitCommit: 26ba49149787c85b9c51222b47c00879b2a0afde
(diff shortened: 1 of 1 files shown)


tianon:
Overall this looks fine, but I have a couple minor recommendations:

- `FROM openjdk:16-jdk` should probably be more specific like `FROM openjdk:16-jdk-oracle` or even `FROM openjdk:16-jdk-oraclelinux8` (to help you avoid surprises later, like `oraclelinux7` being updated to `oraclelinux8` just recently, which would've affected you)

- `RUN microdnf install findutils` should probably move up above the `ARG MAVEN_VERSION` block, since it's not version dependent

tianon:
(But to be clear, if you want to merge this as-is for now, I'm OK with that -- these aren't blockers)

carlossg:
> which would've affected you

it did :)

ok, I'll update the tags in a sec, thanks!

PR #8626, opened 2020-08-28
Adjusted the Kong containers to infer their dependencies from the packages

hutchic:
We'd like to de-dup our dependency list and trust `fpm` to specify the requirements


yosifkit:
Just a heads up on using `apt` instead of `apt-get` in Dockerfiles.

```console
WARNING: apt does not have a stable CLI interface. Use with caution in scripts.
```

PR #8652, opened 2020-09-02
Update CrateDB to 4.2.4

mkleen:


yosifkit:
Sorry, was in a grove and forgot to double check maintainer. I see that @mkleen has approval/commit access to `crate/crate` so this seems fine. Let me know if we need to change anything. cc @christianbader @mfussenegger @chaudum @autophagy

mkleen:
Thanks a lot. I am ok from my side with the current cc.

PR #8658, opened 2020-09-03
Mediawiki: Add fpm and fpm-alpine variants

agp8x:
Relating to https://github.com/wikimedia/mediawiki-docker/pull/81


yosifkit:
GPG signatures! :100::+1:

We've adjusted our documentation to recommend both checksum and detached gpg signatures since they verify different things; see the [Image Build](https://github.com/docker-library/official-images/tree/a0ae21a233bc74700e7f2474091ca707592e6a77#image-build) section for our reasoning. While not strictly necessary for merge, we do recommend keeping the checksum.

PR #8664, opened 2020-09-03
[nats] Release v2.1.8

kozlovic:
Details can be found [here](https://github.com/nats-io/nats-server/releases/tag/v2.1.8)

### DIFF
--- a/library/nats
+++ b/library/nats
@@ -3,31 +3,31 @@ Maintainers: Derek Collison <derek@synadia.com> (@derekcollison),
              Waldemar Salinas <wally@synadia.com> (@wallyqs),
              Jaime Piña <jaime@synadia.com> (@variadico)
 GitRepo: https://github.com/nats-io/nats-docker.git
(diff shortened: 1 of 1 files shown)


kozlovic:
@yosifkit I am just checking if I messed-up something or what. It seems that the arm32v6 image is not being built:
```
docker run --rm -ti mplatform/mquery nats:2.1.8
Image: nats:2.1.8
 * Manifest List: Yes
 * Supported platforms:
   - linux/amd64
   - linux/arm/v7
   - linux/arm64
   - windows/amd64:10.0.17763.139
```
Last build was more than a month ago: https://doi-janky.infosiftr.net/job/multiarch/job/arm32v6/
Looks like there are errors with other projects, but not sure if related or not. Thanks!

yosifkit:
Ah, the generator was stuck and so the triggers were not going.

kozlovic:
Thanks! I see it is now running. Will check if it errors out.

PR #8673, opened 2020-09-04
Remove Sentry image

BYK:
Sentry now requires multiple services to be set up and having a single official image, especially with the lack of cooperation from the Docker team for updates (see getsentry/docker-sentry#211) does not make sense. This page and image is now dragging us back as we want people to use the newer versions of Sentry.

Ideally, the old page would redirect to https://develop.sentry.dev/self-hosted/ or at least https://hub.docker.com/r/getsentry/sentry but it is okay if this is not possible.

/cc @mattrobenolt @benvinegar @dcramer

### DIFF
--- a/library/sentry
+++ b/library/sentry
@@ -1,12 +1,2 @@
-# this file is generated via https://github.com/getsentry/docker-sentry/blob/f58f91fe5dc31bfe77af277dae7002a5542326a9/generate-stackbrew-library.sh
-
-Maintainers: Matt Robenolt <matt@sentry.io> (@mattrobenolt)
+Maintainers: Sentry Open Source <oss@sentry.io> (@getsentry)
 GitRepo: https://github.com/getsentry/docker-sentry.git
-
-Tags: 9.1.2, 9.1, 9, latest
-GitCommit: 09a7761e841eee7fab758526b14d46ae56134952
-Directory: 9.1
-
-Tags: 9.1.2-onbuild, 9.1-onbuild, 9-onbuild, onbuild
-GitCommit: f58f91fe5dc31bfe77af277dae7002a5542326a9
-Directory: 9.1/onbuild


BYK:
Ping @yosifkit

tianon:
This is unfortunate, but understandable.

We don't have any means of doing a redirect, so this should start with a "deprecation" PR over at https://github.com/docker-library/docs (adding `sentry/deprecated.md`); you can see a few examples in `crux`, `rapidoid`, `swarm`, `opensuse`, `sourcemage`, and `thrift`.

BYK:
@tianon I am still okay working with you and maintaining the image with monthly releases we do now. That said this image alone does not offer much value anymore (compared to Sentry 9.x times) so I think best path forward is to remove it and save resources from your side too 🙂

BYK:
@tianon what's a typical timeline for the deprecation notice to be up before removing the code/image?

BYK:
Ping @tianon

Steg17 (approved):

BYK:
@tianon @yosifkit are you okay merging this next week or latest the week of October 19th since now the deprecation notice is up for a few weeks?

yosifkit:
This should be a mostly empty file, rather than deleted (like https://github.com/docker-library/official-images/commit/12edf617e277db4d95cbd43e122f99cdc7ad7757). This will get the "Supported Tags" section on the Hub page to be empty.

It is up to you how long you would like the deprecation notice before removal. Once this merges, the Docker Hub page will continue to be up, so there won't really be any change. The `sentry` image is `FROM` an image that is no longer updated and so it doesn't get periodic rebuilds. If a build does run, the build cache on the server would prevent any change to the published image.

To be clear, the historical images will still be available via `docker pull`.

BYK:
>To be clear, the historical images will still be available via docker pull.

@yosifkit I'd rather this no to be the case to give people a clearer heads up that this is stale. Is this possible in any way?

BYK:
@yosifkit also, updated the PR per your suggestion. Merge at will 🙂

PR #8674, opened 2020-09-04
introduce hitch

gquintard:
Hi,

Here's a new version of https://github.com/docker-library/official-images/pull/7611, hopefully it'll be more aligned with the guidelines. It reuses a lot of the Varnish image experience and tries to fix the problems pointed out by #7611, notably, the command line is much easier to override and we are using a `gpg` fingerprint and the "right" keyservers. Oh, and we are not using the new project-built packages instead of the distro ones, so we can keep up with the releases.

# Checklist for Review
-	[x] associated with or contacted upstream? (yes, I'm part of Varnish Software who maintains `hitch`)
-	[x] does it fit into one of the common categories? ("service", "language stack", "base distribution") (yes, "service")
-	[x] is it reasonably popular, or does it solve a particular use case well? (yes, provides up-to-date versions of the de facto TLS terminator used with Varnish)
-	[x] does a [documentation](https://github.com/docker-library/docs/blob/master/README.md) PR exist? (should be reviewed and merged at roughly the same time so that we don't have an empty image page on the Hub for very long) (docker-library/docs#1675)
-	[x] existing official images have been considered as a base? (ie, if `foobar` needs Node.js, has `FROM node:...` instead of grabbing `node` via other means been considered?) (no, no relevant base image here)
-	[x] ~~if `FROM scratch`, tarballs only exist in a single commit within the associated history?~~ (not `FROM scratch`)

### DIFF
--- a/library/hitch
+++ b/library/hitch
@@ -1,0 +1,8 @@
+# this file was generated using https://github.com/varnish/docker-hitch/blob/d2feb9f1a1a3426da633383c2bac4a31559248bd/populate.sh
+Maintainers: Thijs Feryn <thijs@varni.sh> (@thijsferyn),
+             Guillaume Quintard <guillaume@varni.sh> (@gquintard)
+GitRepo: https://github.com/varnish/docker-hitch.git	
+
+Tags: 1, 1.7, 1.7.0, 1.7.0-1, latest
+Architectures: amd64
+GitCommit: d2feb9f1a1a3426da633383c2bac4a31559248bd


ThijsFeryn:
The updated docs PR can be found here: https://github.com/docker-library/docs/pull/1675

gquintard:
Happy Thanksgiving!

I just pushed an update to this PR as we have released `hitch` 1.7.0 recently. Let me know if you want me to squash the commits

tianon:
FWIW, you didn't do yourself any favors by opening a new PR -- the conversation is now split (having to refer back to that previous thread during my review), and you put yourself back at the bottom of the new-image queue. :sweat_smile: :heart:

I think some of my comments from https://github.com/docker-library/official-images/pull/7611#issuecomment-625636574 are still relevant, but I'll try to recapture the salient bits here.

---

```diff
+	cp /etc/hitch/testcert.pem /etc/hitch/certs/default
```

> Having this example certificate pair pre-baked into the image feels a little strange IMO -- what're the details of that certificate? How does it get regenerated? What's it valid for, how long, etc? Is it signed by any other party? (These are all questions that are hard to answer from just an opaque certificate.)

(The docs mention this is a self-signed certificate, but I think more details would probably still be helpful, especially since some systems like Mac OS are picky about the validity period of even self-signed certs.)

---

What user/UID does this end up running as?  Does it matter for that to be consistent?  (Is there any persistent storage for things like session/cache that might be affected if it were to change underneath a user?)

> > ```ini
> > user = "_hitch"
> > ```
> 
> If users accidentally omit this bit from their configuration files, will Hitch run as root instead, or does it have some kind of protection against that?
> 
> Is it worth considering using `USER _hitch` or similar instead to enforce this at a higher level?  (Usually filesystem permissions and/or "privileged" ports are the reasons setting `USER` becomes a hassle.)
> 
> Did you want to create a dedicated `hitch` user separate from the one auto-created by the Debian package? (Think about whether that might be something you'll want to consider doing later too, because it's *much* easier on you to make that sort of user-breaking change now than after we merge this, *especially* if it ends up getting embedded in user's configuration files too.)

gquintard:
> FWIW, you didn't do yourself any favors by opening a new PR -- the conversation is now split (having to refer back to that previous thread during my review), and you put yourself back at the bottom of the new-image queue. sweat_smile heart

Dagnabit! Sorry, I didn't realize that.

> (The docs mention this is a self-signed certificate, but I think more details would probably still be helpful, especially since some systems like Mac OS are picky about the validity period of even self-signed certs.)

@ThijsFeryn, would you mind updating the docs to match that we now use the packaged default file, and tell us a bit more about it?

> What user/UID does this end up running as? Does it matter for that to be consistent? (Is there any persistent storage for things like session/cache that might be affected if it were to change underneath a user?)

Without user configuration, the `hitch` user is used (https://github.com/varnish/pkg-hitch/blob/master/systemd/hitch.conf#L97) and is the same as the one created by the `debian` package, and the only directory we need for persistence will be `/var/lib/hitch-ocsp`. I think that is what we want.

My instinct here is that we don't need the `USER hitch` bit here as the process will drop permissions itself and this is a no-home user anyway, so we don't really want to `exec` as it either. But maybe I'm missing something here.

gquintard:
@tianon, @ThijsFeryn updated the docs PR to explain the test certificate: https://github.com/docker-library/docs/pull/1675/files#diff-001e280f2a3c2c5174bd4d733268cc2f2b4478f3baa6c5fc09ed6c5b7ee3d20cR43-R53

That should address all your concerns, right?

gquintard:
Hi team, anything you need us to do here?

gquintard:
bump @tianon, @yosifkit

tianon:
Generally, I think this looks OK.  Thank you for your patience. :pray:

A few minor (non-binding) notes/suggestions:

- `apt-transport-https` is a transitional package in Buster+, and thus not necessary :smile:

- if `hitch` creates any files on-disk (cache, etc -- `/var/lib/hitch-ocsp`?), you might still want to consider pre-creating the package's `_hitch` user so it can have a known, fixed UID/GID (because that will cause potential upgrade headaches later)

I'm a little confused at the entrypoint control flow -- if `HITCH_CONFIG_FILE` has a non-empty value, and there are no arguments (or the first argument starts with `-`, it'll run `hitch --config=$HITCH_CONFIG_FILE ...`, but if `HITCH_CONFIG_FILE` is empty in the same case, it'll run `hitch /etc/hitch/hitch.conf ...` (without `--config=`); is that the intentional usage?

It seems like maybe something like the following was intended?

```sh
#!/bin/sh
set -eu

# this will check if the first argument is a flag
# but only works if all arguments require a hyphenated flag
# -v; -SL; -f arg; etc will work, but not arg1 arg2
if [ "$#" -eq 0 ] || [ "${1#-}" != "$1" ]; then
	if [ -n "${HITCH_CONFIG_FILE:-}" ]; then
		# only add --config=... if HITCH_CONFIG_FILE was set and not empty
		set -- --config="$HITCH_CONFIG_FILE" "$@"
	fi

	set -- hitch "$@"
fi

exec "$@"
```

(Also, a rebase/squash of the final proposed commit would indeed be appreciated. :+1:)

gquintard:
Hi!

So, the logic behind the entrypoint:
- if you specify a command, we use it verbatim, no question asked
- otherwise, we inject a `--config` argument, either using `$HITCH_CONFIG_FILE` or `/etc/hitch/hitch.conf`

For example, replacing `exec` with `echo`:
```
$ docker run hitch
hitch --config=/etc/hitch/hitch.conf

$ docker run hitch --some-arg
hitch --config=/etc/hitch/hitch.conf --some-arg

$ docker run hitch some-command
some-command

$ docker run -e HITCH_CONFIG_FILE=/some/other/config.conf hitch
hitch --config=/some/other/config.conf

$ docker run -e HITCH_CONFIG_FILE=/some/other/config.conf hitch --some-arg
hitch --config=/some/other/config.conf --some-arg

$ docker run -e HITCH_CONFIG_FILE=/some/other/config.conf hitch some-command
some-command
```

The goal was to save the user from setting `HTCH_CONFIG_FILE` in most cases, but maybe that's confusing?

Thanks for the `apt-transport-https`, I wasn't aware of it, I'm fixing it right away.

The UID/GID makes sense, I will change it too, is there a rule for picking these IDs, or will any random values do the trick?

tianon:
> * if you specify a command, we use it verbatim, no question asked
> * otherwise, we inject a `--config` argument, either using `$HITCH_CONFIG_FILE` or `/etc/hitch/hitch.conf`

Ah, so what you're looking for is something even simpler, like this:

```sh
#!/bin/sh
set -eu

# this will check if the first argument is a flag
# but only works if all arguments require a hyphenated flag
# -v; -SL; -f arg; etc will work, but not arg1 arg2
if [ "$#" -eq 0 ] || [ "${1#-}" != "$1" ]; then
	set -- hitch --config="${HITCH_CONFIG_FILE:-/etc/hitch/hitch.conf}" "$@"
fi

exec "$@"
```

> The UID/GID makes sense, I will change it too, is there a rule for picking these IDs, or will any random values do the trick?

We typically choose something in the high 900s because `useradd --system` will typically pick whatever number comes next in the low range, and non-system will typically start at 1000 and go up from there (so 999 is a common choice).

Another interesting convention we've seen is choosing the "default port" of the service, if it's a reasonable UID/GID that doesn't have a lot of potential for overlap, but given the nature of `hitch` I don't think that really works here. :smile:

Given this image hasn't been published yet, you could also just choose 1000 directly, if you think that's reasonable -- the main goal is to avoid Debian UID/GID changes (however unlikely) from changing your Hitch user's UID/GID and breaking users unexpectedly. :smile:

gquintard:
yeah, your code is indeed way more straightforward, I'll go with that.

I did think about going with 443 and expected some resitance, 999 has a nice ring to it though.

I'll push something this afternoon

tianon:
I mean, 443 is high enough that it shouldn't have conflicts, and it's kind of cute :smirk: -- your call :smile:

gquintard:
fixed, rebased and force-pushed!

tianon:
If you're using `debhelper` to build that `hitch.deb`, you should be able to pre-create the `hitch` user instead of modifying it post-install (so it's created with the expected UID from the very start), but either way you'll want to ensure the permissions on any `hitch`-owned directories are correct (notably `/var/lib/hitch-ocsp`).

I also just left a review with some more notes/discussion on the docs PR: https://github.com/docker-library/docs/pull/1675#pullrequestreview-622558037

yosifkit (approved):

gquintard:
almost there! The packages are the generic ones, so we don't get to tinker with their creation. I didn't realize the OCSP directory was created by the package (I thought it was done at runtime), so I flipped things around and created the user before the package is installed.

Thank you for the docs review, it'll probably wait until @ThijsFeryn is back, but that is great to have your feedback on it

PR #5958, opened 2019-05-25
Update Zookeeper to 3.5.5

31z4:
* Use openjdk:8-jre-slim as a base image
* Drop architectures to amd64

PR #8676, opened 2020-09-05
Add arm64v8 architecture support to zookeeper

ertanden:


yosifkit:
This and other architectures were dropped by the maintainer in https://github.com/docker-library/official-images/pull/5958. So unless @31z4 wants `arm64v8` back, this PR will not be acceptable.

ertanden:
@31z4 any specific reasons not to have arm64 support for 3.5 and 3.6 branches? OpenJDK images have official support for arm64. I see no reason not to support this.

31z4:
Indeed, as OpenJDK images (except 8) now have official support for `arm64v8` we can have it back in Zookeeper `3.5` and `3.6`. See https://github.com/docker-library/official-images/pull/8752

PR #8691, opened 2020-09-09
Friendica Hotfix 2020.07-1

nupplaphil:
https://github.com/friendica/friendica/releases/tag/2020.07-1

### DIFF
--- a/library/friendica
+++ b/library/friendica
@@ -1,63 +1,63 @@
-# This file is generated via https://github.com/friendica/docker/blob/ade37e5af6488f520bca1a0d8222ad31b0fd8063/generate-stackbrew-library.sh
+# This file is generated via https://github.com/friendica/docker/blob/bb91661bd2753d706449e3a56e47d271c4accbff/generate-stackbrew-library.sh
 Maintainers: Friendica <info@friendi.ca> (@friendica), Philipp Holzer <admin@philipp.info> (@nupplaphil)
 GitRepo: https://github.com/friendica/docker.git
 
-Tags: 2020.03-apache, 2020.03
+Tags: 2020.07-apache, 2020.07
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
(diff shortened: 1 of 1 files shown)


library/friendica
```diff
@@ -1,63 +1,63 @@
-# This file is generated via https://github.com/friendica/docker/blob/ade37e5af6488f520bca1a0d8222ad31b0fd8063/generate-stackbrew-library.sh
+# This file is generated via https://github.com/friendica/docker/blob/bb91661bd2753d706449e3a56e47d271c4accbff/generate-stackbrew-library.sh
 Maintainers: Friendica <info@friendi.ca> (@friendica), Philipp Holzer <admin@philipp.info> (@nupplaphil)
 GitRepo: https://github.com/friendica/docker.git
 
-Tags: 2020.03-apache, 2020.03
+Tags: 2020.07-apache, 2020.07
```
yosifkit, line 5: Any particular reason to keep the `2020.07` versions here?   Removing it here will remove it from the "Supported" section on the Hub readme (and will prevent us from spending cycles rebuilding it on the official build server), but the tag will still be available to users who want it, assuming you don't tag `2020.07-1` as `2020.07`.  (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)

yosifkit (commented):
See comment about older versions below.

yosifkit:
There might be an issue on the `-dev` version:
```diff
-ENV FRIENDICA_VERSION 2020.09-rc
-ENV FRIENDICA_ADDONS 2020.09-rc
+ENV FRIENDICA_VERSION ""
+ENV FRIENDICA_ADDONS ""
```

library/friendica
nupplaphil: I was too fast. I reverted the additional, explicit `2020.07-1` image and map the `2020.07-1` archive onto the existing `2020.07` branch. This now reflects the upstream branch better as before (because there isn't an explicit `2020.07-1` branch as well). Thanks for your input!

PR #8704, opened 2020-09-10
Updating Oracle Linux 8 images for ELBA-2020-3653 / bind

Djelibeybi:
Details: https://linux.oracle.com/errata/ELBA-2020-3653.html


yosifkit:
Build queue is still quite full from Debian and child image rebuilds (https://github.com/docker-library/official-images/pull/8696), so it may be a while before the updated images show on the Docker Hub.

Djelibeybi:
Ok, thanks for letting me know.

PR #8714, opened 2020-09-12
Usability change for oraclelinux:8-slim image

Djelibeybi:
We added the `shadow-utils` package to the `oraclelinux:8-slim` image. Apparently creating non-root users is a thing people want to do.

### DIFF
--- a/library/oraclelinux
+++ b/library/oraclelinux
@@ -4,10 +4,10 @@ GitCommit: 01a15ec99c7470a3391c691509db1759b41eaf66
 GitFetch: refs/heads/master
 # https://github.com/oracle/container-images/tree/dist-amd64
 amd64-GitFetch: refs/heads/dist-amd64
-amd64-GitCommit: 71082ca8be173ae4a73c7187eb7c3403bbd9c641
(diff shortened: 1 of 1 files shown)


Djelibeybi:
@yosifkit @tianon I see a build that occurred two hours ago: https://doi-janky.infosiftr.net/job/multiarch/view/images/view/oraclelinux/ but the image I'm pulling from Docker Hub isn't the one I'm expecting. Am I misinterpreting the Jenkins output or has something broken somewhere? :)

yosifkit:
Because registries are dumb, there is the second step to push the "manifest list" that combines the multiple architectures: https://doi-janky.infosiftr.net/job/put-shared/job/light/job/oraclelinux/.  Looks like it finally finished a few minutes ago.

Djelibeybi:
Ah, lovely. Also, I'll bookmark _that_ page too, so I can get both the build and push jobs in future.

yosifkit:
The regular jobs will at least let you know when https://hub.docker.com/r/amd64/oraclelinux or https://hub.docker.com/r/arm64v8/oraclelinux is ready.

PR #8722, opened 2020-09-14
Added OpenContainers image-spec labels to the currently updated tags.

Djelibeybi:
Yet another packaging update.


tianon:

(Speaking specifically of that bare `maintainer` label.)

tianon:
(You likely intended to use `org.opencontainers.image.authors` instead?)

Djelibeybi:
Well, no. I was maintaining (see what I did there?) the deprecated `MAINTAINER` directive and using `org.opencontainers.image.vendor` as the OpenContainers label that references the source organisation. I can switch the maintainer to author instead, if you prefer?

Djelibeybi:
(Slightly more minutes are required to force push the images, though)

PR #8731, opened 2020-09-16
Release of GeoNetwork 4.0.0-alpha.2

juanluisrp:
Release of GeoNetwork 4.0.0-alpha.2
https://github.com/geonetwork/docker-geonetwork/pull/53

### DIFF
--- a/library/geonetwork
+++ b/library/geonetwork
@@ -1,4 +1,4 @@
-# this file is generated via https://github.com/geonetwork/docker-geonetwork/blob/7259ce435432fe1814da0e444328f450b8b6654b/generate-stackbrew-library.sh
+# this file is generated via https://github.com/geonetwork/docker-geonetwork/blob/9d7d9cffd5746a263599dd69e84a00e0405e6b9d/generate-stackbrew-library.sh
 
 Maintainers: Joana Simoes <jo@doublebyte.net> (@doublebyte1),
      Juan Luis Rodriguez <juanluisrp@geocat.net> (@juanluisrp)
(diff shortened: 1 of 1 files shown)


yosifkit:
You definitely still need `exec` to apply to the resident server process so that the shell doesn't stay resident.

```diff
-  java $JAVA_OPTIONS -jar /usr/local/jetty/start.jar
+  exec java $JAVA_OPTIONS -jar /usr/local/jetty/start.jar
```

Also, the change from `JAVA_OPTS` to `JAVA_OPTIONS` might cause some user problems on upgrade if they are overriding it?

juanluisrp:
For these new versions we have switched from Tomcat to Jetty. The option for settings JVM options in Jetty based images is to use `JAVA_OPTIONS`. 
What have other projects in this case?
Should we keep `JAVA_OPTS` and assign its value to `JAVA_OPTIONS` in the entrypoint?

yosifkit:
> For these new versions we have switched from Tomcat to Jetty. The option for settings JVM options in Jetty based images is to use `JAVA_OPTIONS`.
> What have other projects in this case?
> Should we keep `JAVA_OPTS` and assign its value to `JAVA_OPTIONS` in the entrypoint?

You can use whichever variable you want, I was just concerned for breaking users.  On the other hand, I don't see a need to use/support the one from jetty, since you aren't using their scripts.

tianon:
(IMO, it seems a little odd to be `FROM jetty` at all if you're not going to use their scripts :sweat_smile:)

juanluisrp:
> (IMO, it seems a little odd to be `FROM jetty` at all if you're not going to use their scripts 😅)

We could install Jetty ourselves but I would like to get the image rebuilt when base Jetty image is updated.

yosifkit:
closing in favor of https://github.com/docker-library/official-images/pull/8969

PR #8735, opened 2020-09-17
caddy: Updated :builder and v2.2.0 final release

hairyhenderson:
- Adds images for https://github.com/caddyserver/caddy/releases/tag/v2.2.0
- Updates the builder image to use [xcaddy](https://github.com/caddyserver/xcaddy), the official Caddy plugin-bundling tool
- Adds a Windows-based builder image

### DIFF
--- a/library/caddy
+++ b/library/caddy
@@ -1,97 +1,97 @@
 # this file is generated with gomplate:
 # template: https://github.com/caddyserver/caddy-docker/blob/c74bdca53a1efd5195a9c35005e5515afb5feeff/stackbrew.tmpl
-# config context: https://github.com/caddyserver/caddy-docker/blob/98b7497fbd428e50040ef34f7a4ba2833c22780f/stackbrew-config.yaml
+# config context: https://github.com/caddyserver/caddy-docker/blob/153e0aaa8e8f9f7cc2879c4ffd1016dec368cd9e/stackbrew-config.yaml
 Maintainers: Dave Henderson (@hairyhenderson)
 
-Tags: 2.1.1-alpine, 2-alpine, alpine
-SharedTags: 2.1.1, 2, latest
+Tags: 2.1.1-alpine
+SharedTags: 2.1.1
 GitRepo: https://github.com/caddyserver/caddy-docker.git
 Directory: 2.1/alpine
 GitCommit: 98b7497fbd428e50040ef34f7a4ba2833c22780f
 Architectures: amd64, arm64v8, arm32v6, arm32v7, ppc64le, s390x
 
-Tags: 2.1.1-builder-alpine, 2-builder-alpine, builder-alpine
-SharedTags: 2.1.1-builder, 2-builder, builder
+Tags: 2.1.1-builder-alpine
+SharedTags: 2.1.1-builder
 GitRepo: https://github.com/caddyserver/caddy-docker.git
 Directory: 2.1/builder
 GitCommit: 98b7497fbd428e50040ef34f7a4ba2833c22780f
 Architectures: amd64, arm64v8, arm32v6, arm32v7, ppc64le, s390x
 
-Tags: 2.1.1-windowsservercore-1809, 2-windowsservercore-1809, windowsservercore-1809
-SharedTags: 2.1.1-windowsservercore, 2-windowsservercore, windowsservercore, 2.1.1, 2, latest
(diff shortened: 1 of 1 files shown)


hairyhenderson:
Uhhh....

```

 - caddy:2.1.1-builder-windowsservercore-1809 (FROM golang:1.14) -- completely unsupported base!
 - caddy:2.1.1-builder-windowsservercore-ltsc2016 (FROM golang:1.14) -- completely unsupported base!
 - caddy:2.2.0-rc.1-builder-windowsservercore-1809 (FROM golang:1.15) -- completely unsupported base!
 - caddy:2.2.0-rc.1-builder-windowsservercore-ltsc2016 (FROM golang:1.15) -- completely unsupported base!
```

Guess I'll need to go figure this out first 😅

tianon:
That looks like a bug in that script 😅😬

hairyhenderson:
Oh thank goodness. Thank you @tianon! 🎉

hairyhenderson:
The CI for at least one of the "naughty" Dockerfiles is passing: https://github.com/docker-library/official-images/pull/8735/checks?check_run_id=1126266914... though I'm not sure I quite follow what "naughty" actually means in this context 😉

yosifkit:
While there is a bug in our detection script, it is also a problem with the Dockerfile. `FROM golang:1.15` isn't specific enough for windows hosts. In order to properly target `ltsc2016` vs `1809`, that needs to be in the `FROM` line (we build both on the same Windows Server).

francislavoie:
So we should use these instead?

```
FROM golang:1.15-windowsservercore-ltsc2016
FROM golang:1.15-windowsservercore-1809
```

tianon:
Yes, if you want `caddy:xxx-windowsservercore-ltsc2016` to actually work on ltsc2016, it has to be the more specific `FROM` (otherwise Docker will use whatever the latest our host supports is, which is 1809 ATM).

hairyhenderson:
Ok, sorry for the delay. @francislavoie got a fix in _ages_ ago and I've only just bumped the PR to reflect it 😅

Also the 2.2.0 final release is probably happening tomorrow, so it _may_ make sense to hold off. 2.2.0-rc.1 isn't current anymore anyway (it's at rc.3, and there's a bug in rc.3 😅).

hairyhenderson:
Ok, v2.2.0 is released now - I've updated the manifest.

PTAL @tianon @yosifkit 🙏 😬

yosifkit:
```diff
-WORKDIR /src/custom-caddy/cmd/caddy
+WORKDIR /usr/bin
```
This directory seems like a strange choice. Won't this cause weird issues for users that `COPY` into `.`?

---

> For backwards compatibility it'd be nice to at least wrap `xcaddy` so existing users aren't broken, but it would be good to log a warning in stderr to let users know they should switch to calling `xcaddy` directly.

What happened to the planned warning and why do the new 2.2 releases and windows versions even have the `caddy-builder.sh` script? Since they didn't have it before, would it be sane to not even add them in the first place?

hairyhenderson:
> This directory seems like a strange choice. Won't this cause weird issues for users that `COPY` into `.`?

That was an optimization so that when `xcaddy` runs it spits the binary out into the expected location (`/usr/bin`). There isn't really an expectation that users will `COPY` into `.` for the builder image...

> What happened to the planned warning and why do the new 2.2 releases and windows versions even have the `caddy-builder.sh` script? Since they didn't have it before, would it be sane to not even add them in the first place?

Yeah, that's a good point. I'd totally forgotten about that in the rush of preparing for the release 😅

I added a `caddy-builder.ps1` script to the Windows images more for compatibility than anything, but in retrospect that was unnecessary. For now I think we'll keep it and have it print out a deprecation message (so it's consistent with Linux).

I'm also realizing that there are docs changes we were planning to go along with this, so I'll issue a PR for that soon.

francislavoie:
I wrote a commit to update the docs here https://github.com/francislavoie/docs-1/commit/78086a55990f1403dca149475b190dd6337b3bd7?short_path=790afa8#diff-790afa88c0ba417c78ade9a50f69ea9c as a preview.

hairyhenderson:
> For now I think we'll keep it

I take it back. Powershell is not my favourite so I'll just ditch it instead 😝

hairyhenderson:
Ok - I've updated _again_... PTAL @yosifkit 😬

PR #8738, opened 2020-09-17
Matomo 3.14.1

J0WI:


library/matomo
```diff
@@ -2,17 +2,17 @@
 Maintainers: Pierre Ozoux <pierre@piwik.org> (@pierreozoux)
 GitRepo: https://github.com/matomo-org/docker.git
 
-Tags: 3.14.0-apache, 3.14-apache, 3-apache, apache, 3.14.0, 3.14, 3, latest
+Tags: 3.14.1-apache, 3.14-apache, 3-apache, 3.14.1, 3.14, 3
```
tianon, line 5: Was the loss of `apache` and `latest` here intentional?

J0WI: No
https://github.com/docker-library/official-images/pull/8935

PR #8745, opened 2020-09-17
CouchDB 3.1.1 release

wohali:
Title says it all.

### DIFF
--- a/library/couchdb
+++ b/library/couchdb
@@ -3,18 +3,18 @@
 
 Maintainers: Joan Touzet <wohali@apache.org> (@wohali)
 GitRepo: https://github.com/apache/couchdb-docker
-GitCommit: 01e1c16bb405c1174393c76a7ebc3792b0c5e2c9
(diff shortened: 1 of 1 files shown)


wohali:
@tianon Not sure what we need to do for the "Naughty" check failure, this seems new since our last PR. Obviously 2.3.1 builds against `ppc64le` since we've not touched that image in years. Please advise if action is required on our part.

wohali:
We'll go ahead and remove `ppc64le` from `2.3.1` just as a prophylactic here. Supporting that version on that platform isn't a priority for us.

tianon:
Yep, that "naughty" is part of what I was trying to communicate with https://github.com/docker-library/official-images/pull/8405#issuecomment-662167454 -- namely, Stretch has moved from being supported by the Security Team (who officially support all release architectures of Debian) to being supported by the LTS Team (who only support the subset of architectures that the LTS sponsors have explicitly requested). :+1:

actionmanmbc (approved):

PR #8746, opened 2020-09-18
Plone: Release 5.2.2

avoinea:
Refs https://github.com/plone/plone.docker/issues/138


actionmanmbc:
🤘

yosifkit:
While we don't actively recommend using labels, we do recommend that Image label names should adhere to the image spec: https://github.com/opencontainers/image-spec/blob/v1.0.1/annotations.md.

This can definitely be addressed later so that this update doesn't get delayed through a weekend.

PR #8749, opened 2020-09-18
bump Kong to 2.1.4

hutchic:


yosifkit:
Using `uname -m` for platform detection is very strongly discouraged, as it tests the host kernel's architecture, when what you really wanted to test was the image's userspace architecture, so `apk --print-arch` or `dpkg --print-architecture` are going to be safer/more reliable options.

sanjog152 (approved):

PR #8765, opened 2020-09-24
erlang/otp 23.1, and adding immutable versions

getong:
close erlang/docker-erlang-otp#289

### DIFF
--- a/library/erlang
+++ b/library/erlang
@@ -1,21 +1,21 @@
-# this file is generated via https://github.com/erlang/docker-erlang-otp/blob/7ad81df8943ef6a5b15b8e4839ef45474f8760b0/generate-stackbrew-library.sh
+# this file is generated via https://github.com/erlang/docker-erlang-otp/blob/c0f9d0d3cb0e86581269ec68b12a0ccf5c30276e/generate-stackbrew-library.sh
 
 Maintainers: Mr C0B <denc716@gmail.com> (@c0b)
 GitRepo: https://github.com/erlang/docker-erlang-otp.git
 
-Tags: 23.0.4, 23.0, 23, latest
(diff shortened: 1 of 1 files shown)


tianon:
IMO, `immutable-18.3.4.11` seems like an odd tag name, especially since "Erlang Immutable" isn't a thing, and also because the image will be anything *but* immutable (see for example https://github.com/docker-library/official-images#repeatability, https://github.com/docker-library/faq#an-images-source-changed-in-git-now-what, https://github.com/docker-library/faq#why-does-my-security-scanner-show-that-an-image-has-cves for where we talk about some of the reasons the image will be rebuilt).  If users want something that will never change, what they're looking for is pull-by-digest (`erlang@sha256:xxxx`), which is immutable by definition.

It seems like what's being proposed here is in response to the fact that Erlang doesn't use version numbers in the way that the rest of the industry does, is that accurate?  So "Erlang 21.3" is a thing, and "Erlang 21.3.8.17" is not necessarily the same thing as "the latest "Erlang 21.3 release" but is more accurately "the 17th subrelease of the 8th subrelease of the 3rd subrelease of the 21st major release of Erlang" ?

WordPress has a similar problem, where they have an X.Y release and then an X.Y.Z later, and we resolve that by padding with zeros so that users can still pin to the "initial release" with X.Y.0 if they so desire to do so.  Would an approach like that work here?

An alternative solution would be to just drop the shortened tag aliases, but that makes the image significantly more difficult to consume, especially since it will very quickly no longer receive security updates / image rebuilds, and requires the user to stay up to date on what the "latest Erlang 21" release version is.

getong:
Thanks, the solution pr is made by erlang/docker-erlang-otp#308

getong:
@tianon 
The pr is updated.

PR #8786, opened 2020-09-26
[ros] add labels with sha of ros packages to invalidate cache

mikaelarguedas:
Update all ros images to use the new way of invalidating cache for each new sync of ROS packages.
This adds labels with the sha of the ros packages installed. relates to https://github.com/osrf/docker_images/issues/112 and https://github.com/osrf/docker_images/issues/376

### DIFF
--- a/library/ros
+++ b/library/ros
@@ -9,22 +9,22 @@ GitRepo: https://github.com/osrf/docker_images.git
 
 Tags: kinetic-ros-core, kinetic-ros-core-xenial
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 1c7a5895537e7909774bb8e634d549acd2d71cc9
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/kinetic/ubuntu/xenial/ros-core
 
 Tags: kinetic-ros-base, kinetic-ros-base-xenial, kinetic
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 1c7a5895537e7909774bb8e634d549acd2d71cc9
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/kinetic/ubuntu/xenial/ros-base
 
 Tags: kinetic-robot, kinetic-robot-xenial
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 1c7a5895537e7909774bb8e634d549acd2d71cc9
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/kinetic/ubuntu/xenial/robot
 
 Tags: kinetic-perception, kinetic-perception-xenial
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 1c7a5895537e7909774bb8e634d549acd2d71cc9
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/kinetic/ubuntu/xenial/perception
 
 
@@ -36,22 +36,22 @@ Directory: ros/kinetic/ubuntu/xenial/perception
 
 Tags: melodic-ros-core, melodic-ros-core-bionic
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 421722cdc4b6b3fa5ec3697f851130435454a059
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/melodic/ubuntu/bionic/ros-core
 
 Tags: melodic-ros-base, melodic-ros-base-bionic, melodic
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 421722cdc4b6b3fa5ec3697f851130435454a059
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/melodic/ubuntu/bionic/ros-base
 
 Tags: melodic-robot, melodic-robot-bionic
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 421722cdc4b6b3fa5ec3697f851130435454a059
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/melodic/ubuntu/bionic/robot
 
 Tags: melodic-perception, melodic-perception-bionic
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 421722cdc4b6b3fa5ec3697f851130435454a059
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/melodic/ubuntu/bionic/perception
 
 ########################################
@@ -59,22 +59,22 @@ Directory: ros/melodic/ubuntu/bionic/perception
 
 Tags: melodic-ros-core-stretch
 Architectures: amd64, arm64v8
-GitCommit: 421722cdc4b6b3fa5ec3697f851130435454a059
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/melodic/debian/stretch/ros-core
 
 Tags: melodic-ros-base-stretch
 Architectures: amd64, arm64v8
-GitCommit: 421722cdc4b6b3fa5ec3697f851130435454a059
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/melodic/debian/stretch/ros-base
 
 Tags: melodic-robot-stretch
 Architectures: amd64, arm64v8
-GitCommit: 421722cdc4b6b3fa5ec3697f851130435454a059
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/melodic/debian/stretch/robot
 
 Tags: melodic-perception-stretch
 Architectures: amd64, arm64v8
-GitCommit: 421722cdc4b6b3fa5ec3697f851130435454a059
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/melodic/debian/stretch/perception
 
 
@@ -86,22 +86,22 @@ Directory: ros/melodic/debian/stretch/perception
 
 Tags: noetic-ros-core, noetic-ros-core-focal
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 421722cdc4b6b3fa5ec3697f851130435454a059
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/noetic/ubuntu/focal/ros-core
 
 Tags: noetic-ros-base, noetic-ros-base-focal, noetic
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 421722cdc4b6b3fa5ec3697f851130435454a059
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/noetic/ubuntu/focal/ros-base
 
 Tags: noetic-robot, noetic-robot-focal
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 421722cdc4b6b3fa5ec3697f851130435454a059
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/noetic/ubuntu/focal/robot
 
 Tags: noetic-perception, noetic-perception-focal
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 421722cdc4b6b3fa5ec3697f851130435454a059
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/noetic/ubuntu/focal/perception
 
 ########################################
@@ -109,22 +109,22 @@ Directory: ros/noetic/ubuntu/focal/perception
 
 Tags: noetic-ros-core-buster
 Architectures: amd64, arm64v8
-GitCommit: 421722cdc4b6b3fa5ec3697f851130435454a059
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/noetic/debian/buster/ros-core
 
 Tags: noetic-ros-base-buster
 Architectures: amd64, arm64v8
-GitCommit: 421722cdc4b6b3fa5ec3697f851130435454a059
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/noetic/debian/buster/ros-base
 
 Tags: noetic-robot-buster
 Architectures: amd64, arm64v8
-GitCommit: 421722cdc4b6b3fa5ec3697f851130435454a059
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/noetic/debian/buster/robot
 
 Tags: noetic-perception-buster
 Architectures: amd64, arm64v8
-GitCommit: 421722cdc4b6b3fa5ec3697f851130435454a059
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/noetic/debian/buster/perception
 
 
@@ -136,17 +136,17 @@ Directory: ros/noetic/debian/buster/perception
 
 Tags: dashing-ros-core, dashing-ros-core-bionic
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 04ec552de26d6cc17823ff6e1de0191c136d94f8
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/dashing/ubuntu/bionic/ros-core
 
 Tags: dashing-ros-base, dashing-ros-base-bionic, dashing
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 04ec552de26d6cc17823ff6e1de0191c136d94f8
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/dashing/ubuntu/bionic/ros-base
 
 Tags: dashing-ros1-bridge, dashing-ros1-bridge-bionic
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 04ec552de26d6cc17823ff6e1de0191c136d94f8
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/dashing/ubuntu/bionic/ros1-bridge
 
 
@@ -158,17 +158,17 @@ Directory: ros/dashing/ubuntu/bionic/ros1-bridge
 
 Tags: eloquent-ros-core, eloquent-ros-core-bionic
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 421722cdc4b6b3fa5ec3697f851130435454a059
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/eloquent/ubuntu/bionic/ros-core
 
 Tags: eloquent-ros-base, eloquent-ros-base-bionic, eloquent
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 421722cdc4b6b3fa5ec3697f851130435454a059
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/eloquent/ubuntu/bionic/ros-base
 
 Tags: eloquent-ros1-bridge, eloquent-ros1-bridge-bionic
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 0885dd6fd1a8c465828a76c99b348826a62205be
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/eloquent/ubuntu/bionic/ros1-bridge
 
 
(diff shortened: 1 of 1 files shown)


tianon:
This seems like kind of a funky solution that's functionally not really any different/better than just embedding a timestamp that gets updated -- can't the packages themselves be verified by these checksums somehow so that this is functional data instead of just informational?

---

mikaelarguedas:
Thanks @tianon for the feedback

> This seems like kind of a funky solution that's functionally not really any different/better than just embedding a timestamp that gets updated

Indeed, in both approaches (timestamp or sha) we're restricted by the fact that they will be different depending on the architecture, packages for different arches being built at different times and not having the same sha while the Dockerfile is common to all architectures.

> can't the packages themselves be verified by these checksums somehow so that this is functional data instead of just informational?

This would require to get the different shas for each arch and then conditionnaly check them adding extra logic in the dockerfile.

> If an image maintainer wants to have labels, that is fine, but label names should adhere to the image spec: https://github.com/opencontainers/image-spec/blob/v1.0.1/annotations.md

I thought the labels generated in these dockerfiles were complying with the rules defined in that document. Could you point to which rule is violated by the current approach ?

I'm happy to consider another solution allowing to break cache when the packages change without adding arch specific logic in the dockerfiles. We converged on the sha one as it seemed to tick all the boxes on our side. Would you prefer for the labels to use the timestamps instead of the sha ?

ruffsl:
> Could you point to which rule is violated by the current approach ?

Perhaps @tianon is alluding to this point?

> Keys SHOULD be named using a reverse domain notation - e.g. com.example.myKey.

We could change the labels to something like:

```
LABEL org.ros.sha256.ros-melodic-ros-comm=...
LABEL org.osrfoundation.sha256.gazebo10=...
```

> Would you prefer for the labels to use the timestamps instead of the sha ?

Agreed, this would be an Ok  alternative, as we'd like to avoid any architecture specific embedding in the Dockerfiles. One issue with using timestamps however is that there doesn't seem to be a way to query the timestamp on a per package basis from the apt source list repo. That's a point I think is important, as we don't want to break the build cache if merely an unrelated package in the same package list file is changed, and not all packages include build time stamps in there package version ID. 

http://packages.ros.org/ros2/ubuntu/dists/focal/main/binary-amd64/Packages

yosifkit:
Yes, any labels do need to be reverse DNS style (and conform to any other requirements in the image-spec).

---

The other part he was trying to point out is that these sha256 `LABEL`s are just as useless as sticking a timestamp there since its only purpose is to break the build cache at that point, but it doesn't guarantee anything about what gets installed.

It comes down to the [repeatability section](https://github.com/docker-library/official-images/tree/4824d81111ea74658937e5319c5528594201b9a3#repeatability); any image listed in a `library/` file will be rebuilt if the dependent parent image is rebuilt and we regenerate all Debian and Ubuntu images at least monthly.

ruffsl:
> Yes, any labels do need to be reverse DNS style (and conform to any other requirements in the image-spec).

Done

---

> its only purpose is to break the build cache at that point

Correct, but this is a strict monotonic improvement over what exists currently. 

> it doesn't guarantee anything about what gets installed.

We are limited by the package versioning and distribution software that package.ros.org is currently reliant upon, as well as by the templating CI for our Dockerfiles. We can revisit this infrastructure with alternatives, but in the interim, we have a number of releases/updates backing up, that should be upstreamed to the registry.

yosifkit:
> improvement over what exists currently

Improvement in what way? Breaking the build cache with unused data will result in the data becoming stale and irrelevant to the image.

Let's try a scenario. The Dockerfile "pins" to a specific version like `0.9.1-2*` along with a corresponding `ENV ...sha256=zzz`. Say that Dockerfile is built and installs a version like `0.9.1-2~0` which probably corresponds to the sha256 in some way. Then a few days/weeks later, the base Ubuntu image is rebuilt (as we do every \~30 days) and so the same Dockerfile for `ros` is rebuilt. This time it installs something like version `0.9.1-2~1`.  This updated version has nothing to do with the corresponding `ENV ...sha256` that is part of the image.

ruffsl:
> Let's try a scenario.

Yet, the scenario you described is already analogous to the current status quo... :(

And without us having any means to break the build cache, the ROS packages become stale and irrelevant to the current release of the ROS distro, unless we wait ~30 days for the base Ubuntu image to rebuild. This is just an internal `LABEL` to ratchet up, as package.ros.org has no support for ratcheting down anyhow. Timestamps could work just as well, but wouldn't be as deterministically derivable. The benefit of triggering image rebuilds for package updates in a timely and controlled manner seems to outway the cons of an otherwise unused LABEL becoming stale before our CI can open a PR minutes later.

We elected to not use the `ENV` directive, as to avoid clobberin the environment with values irrelevant to the end user. Using build `ARG`s could be a suitable alternative, but I don't think that's available to us here.

ruffsl:
Could we please get these updates? It seems the ros images are growing quite stale as these release syns have been backing up.

tianon:
We're happy to merge the version bumps, but the added `LABEL` values need more work.  If you would like to remove the `LABEL` updates for now to get the update in, we'd love to move forward on that.

Regarding the issue the `LABEL` values are trying to solve as I understand it, the biggest problem is that each architecture has a different version number for packages like `ros-rolling-ros-base`?  If you've got something that can generate these `sha256` values from relevant data, I don't understand why it can't be used for the version numbers directly (since all the `Dockerfile`s are generated already using data from the repo), something like the following:

```dockerfile
# install ros2 packages
RUN set -eux; \
	arch="$(dpkg --print-architecture)"; \
	case "$arch" in \
		amd64) version='0.9.2-1focal.20201007.223858' ;; \
		arm64) version='0.9.2-1focal.20201008.005415' ;; \
		*) echo >&2 "error: unsupported arch: $arch"; exit 1 ;; \
	esac; \
	apt-get update; \
	apt-get install -y --no-install-recommends ros-rolling-ros-base="$version"; \
	rm -rf /var/lib/apt/lists/*
```

ruffsl:
> We're happy to merge the version bumps, but the added LABEL values need more work. If you would like to remove the LABEL updates for now to get the update in, we'd love to move forward on that.

Could we compromise with using the default field of an `ARG` directive instead? This would have the benefit over both `LABEL` and `ENV`, in that there would be no additional data becoming stale and irrelevant to the image, while still giving us a simple way to ratchet up with release syncs.

``` diff
- LABEL @(' \\\n      '.join('org.osrfoundation.{name}.sha256={sha256}'.format(**p) for p in packages))@
+ ARG @(' \\\n      '.join('org.osrfoundation.{name}.sha256={sha256}'.format(**p) for p in packages))@
```
> https://docs.docker.com/engine/reference/builder/#default-values
> https://github.com/osrf/docker_templates/blob/adad8152da75670c9a157102b322770a931c6b2a/docker_templates/templates/snippet/label_and_install_package_list.Dockerfile.em#L3

> If you've got something that can generate these sha256 values from relevant data, I don't understand why it can't be used for the version numbers directly 

Switching to set default `ARG` value could also be implemented in a few minutes today, where as we'd otherwise have to overhaul our template infrastructure and CI backend, given it was never designed for generating Dockerfiles that support multiple CPU architectures simultaneously; as this all predates multiarch support for DockerHub.

Regardless, let us know. It would take a while if we have to heavily refactor our template, as we are mainly volunteer maintainers.

tianon:
What I'm suggesting is that you simply remove that line as a temporary change for now (since `ubuntu` has had a recent update, and the version bumps you need will cause a rebuild anyhow, so the problem it is solving is not relevant right this second).

ruffsl:
> What I'm suggesting is that you simply remove that line as a temporary change for now

Done now in 09e7488c7e010f307d4fd7a5cbafb3112cdcb3e4

> (since ubuntu has had a recent update, and the version bumps you need will cause a rebuild anyhow, so the problem it is solving is not relevant right this second)

The underlying issue will persists though, regardless if ubuntu gets an update once in a while. I think `ARG` would be a nice compromise here, allowing our CI to submit PRs to break the build cache, without the potential of baking stale info into the image binaries.

yosifkit:
<details>
<summary>Diff:</summary>

```diff
diff --git a/_bashbrew-cat b/_bashbrew-cat
index a7672a4..27adbd7 100644
--- a/_bashbrew-cat
+++ b/_bashbrew-cat
@@ -3,160 +3,160 @@ GitRepo: https://github.com/osrf/docker_images.git
 
 Tags: dashing-ros-core, dashing-ros-core-bionic
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 0eae97e0fa2b34e99d684d88565553b1d67cc590
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/dashing/ubuntu/bionic/ros-core
 
 Tags: eloquent-ros-core, eloquent-ros-core-bionic
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 0b33e61b5bbed5b93b9fba2d5bae5db604ff9b58
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/eloquent/ubuntu/bionic/ros-core
 
 Tags: foxy-ros-core, foxy-ros-core-focal
 Architectures: amd64, arm64v8
-GitCommit: 30496b03a10f37d3ee54d5df672a0c27e0ab3952
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/foxy/ubuntu/focal/ros-core
 
 Tags: kinetic-ros-core, kinetic-ros-core-xenial
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 0b33e61b5bbed5b93b9fba2d5bae5db604ff9b58
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/kinetic/ubuntu/xenial/ros-core
 
 Tags: melodic-ros-core, melodic-ros-core-bionic
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 0b33e61b5bbed5b93b9fba2d5bae5db604ff9b58
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/melodic/ubuntu/bionic/ros-core
 
 Tags: melodic-ros-core-stretch
 Architectures: amd64, arm64v8
-GitCommit: 0b33e61b5bbed5b93b9fba2d5bae5db604ff9b58
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/melodic/debian/stretch/ros-core
 
 Tags: noetic-ros-core, noetic-ros-core-focal
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: bba0ef85a9b417f3815d2cba3b36e7754611a428
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/noetic/ubuntu/focal/ros-core
 
 Tags: noetic-ros-core-buster
 Architectures: amd64, arm64v8
-GitCommit: bba0ef85a9b417f3815d2cba3b36e7754611a428
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/noetic/debian/buster/ros-core
 
 Tags: rolling-ros-core, rolling-ros-core-focal
 Architectures: amd64, arm64v8
-GitCommit: f8cfc3fa7ee88d6b59ae18deedd179b8de4c3eb2
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/rolling/ubuntu/focal/ros-core
 
 Tags: dashing-ros-base, dashing-ros-base-bionic, dashing
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 0eae97e0fa2b34e99d684d88565553b1d67cc590
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/dashing/ubuntu/bionic/ros-base
 
 Tags: eloquent-ros-base, eloquent-ros-base-bionic, eloquent
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 0b33e61b5bbed5b93b9fba2d5bae5db604ff9b58
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/eloquent/ubuntu/bionic/ros-base
 
 Tags: foxy-ros-base, foxy-ros-base-focal, foxy, latest
 Architectures: amd64, arm64v8
-GitCommit: 30496b03a10f37d3ee54d5df672a0c27e0ab3952
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/foxy/ubuntu/focal/ros-base
 
 Tags: kinetic-ros-base, kinetic-ros-base-xenial, kinetic
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 0b33e61b5bbed5b93b9fba2d5bae5db604ff9b58
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/kinetic/ubuntu/xenial/ros-base
 
 Tags: melodic-ros-base, melodic-ros-base-bionic, melodic
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 0b33e61b5bbed5b93b9fba2d5bae5db604ff9b58
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/melodic/ubuntu/bionic/ros-base
 
 Tags: melodic-ros-base-stretch
 Architectures: amd64, arm64v8
-GitCommit: 0b33e61b5bbed5b93b9fba2d5bae5db604ff9b58
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/melodic/debian/stretch/ros-base
 
 Tags: noetic-ros-base, noetic-ros-base-focal, noetic
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: cc4e832f92bcac995b9a2c9ca8a7b8fcf85c5c28
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/noetic/ubuntu/focal/ros-base
 
 Tags: noetic-ros-base-buster
 Architectures: amd64, arm64v8
-GitCommit: cc4e832f92bcac995b9a2c9ca8a7b8fcf85c5c28
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/noetic/debian/buster/ros-base
 
 Tags: rolling-ros-base, rolling-ros-base-focal, rolling
 Architectures: amd64, arm64v8
-GitCommit: f8cfc3fa7ee88d6b59ae18deedd179b8de4c3eb2
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/rolling/ubuntu/focal/ros-base
 
 Tags: dashing-ros1-bridge, dashing-ros1-bridge-bionic
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 92b0c6f7753fc2d7bb7697a57f2871cf3895180e
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/dashing/ubuntu/bionic/ros1-bridge
 
 Tags: eloquent-ros1-bridge, eloquent-ros1-bridge-bionic
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 576e10eb03567a24596c88d515026c84a2ea0dec
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/eloquent/ubuntu/bionic/ros1-bridge
 
 Tags: foxy-ros1-bridge, foxy-ros1-bridge-focal
 Architectures: amd64, arm64v8
-GitCommit: 947dddf543e6e788fd58dabc38ea4371091787e3
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/foxy/ubuntu/focal/ros1-bridge
 
 Tags: kinetic-perception, kinetic-perception-xenial
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 7ff09c2a75e902bc2bb25a1f1ae748ec4e9c7a4b
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/kinetic/ubuntu/xenial/perception
 
 Tags: kinetic-robot, kinetic-robot-xenial
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 7ff09c2a75e902bc2bb25a1f1ae748ec4e9c7a4b
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/kinetic/ubuntu/xenial/robot
 
 Tags: melodic-perception, melodic-perception-bionic
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 7ff09c2a75e902bc2bb25a1f1ae748ec4e9c7a4b
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/melodic/ubuntu/bionic/perception
 
 Tags: melodic-perception-stretch
 Architectures: amd64, arm64v8
-GitCommit: 7ff09c2a75e902bc2bb25a1f1ae748ec4e9c7a4b
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/melodic/debian/stretch/perception
 
 Tags: melodic-robot, melodic-robot-bionic
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 7ff09c2a75e902bc2bb25a1f1ae748ec4e9c7a4b
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/melodic/ubuntu/bionic/robot
 
 Tags: melodic-robot-stretch
 Architectures: amd64, arm64v8
-GitCommit: 7ff09c2a75e902bc2bb25a1f1ae748ec4e9c7a4b
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/melodic/debian/stretch/robot
 
 Tags: noetic-perception, noetic-perception-focal
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: bba0ef85a9b417f3815d2cba3b36e7754611a428
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/noetic/ubuntu/focal/perception
 
 Tags: noetic-perception-buster
 Architectures: amd64, arm64v8
-GitCommit: bba0ef85a9b417f3815d2cba3b36e7754611a428
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/noetic/debian/buster/perception
 
 Tags: noetic-robot, noetic-robot-focal
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: bba0ef85a9b417f3815d2cba3b36e7754611a428
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/noetic/ubuntu/focal/robot
 
 Tags: noetic-robot-buster
 Architectures: amd64, arm64v8
-GitCommit: bba0ef85a9b417f3815d2cba3b36e7754611a428
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/noetic/debian/buster/robot
 
 Tags: rolling-ros1-bridge, rolling-ros1-bridge-focal
 Architectures: amd64, arm64v8
-GitCommit: c098e08c9e9a077516dbba9648893e89aecf1d63
+GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
 Directory: ros/rolling/ubuntu/focal/ros1-bridge
diff --git a/ros_dashing-ros1-bridge-bionic/Dockerfile b/ros_dashing-ros1-bridge-bionic/Dockerfile
index fc18a8a..7c453fe 100644
--- a/ros_dashing-ros1-bridge-bionic/Dockerfile
+++ b/ros_dashing-ros1-bridge-bionic/Dockerfile
@@ -10,16 +10,17 @@ RUN echo "deb http://packages.ros.org/ros/ubuntu bionic main" > /etc/apt/sources
 
 ENV ROS1_DISTRO melodic
 ENV ROS2_DISTRO dashing
+
 # install ros packages
 RUN apt-get update && apt-get install -y --no-install-recommends \
-    ros-melodic-ros-comm=1.14.9-1* \
+    ros-melodic-ros-comm=1.14.10-1* \
     ros-melodic-roscpp-tutorials=0.9.3-1* \
     ros-melodic-rospy-tutorials=0.9.3-1* \
     && rm -rf /var/lib/apt/lists/*
 
 # install ros2 packages
 RUN apt-get update && apt-get install -y --no-install-recommends \
-    ros-dashing-ros1-bridge=0.7.6-1* \
+    ros-dashing-ros1-bridge=0.7.6-3* \
     ros-dashing-demo-nodes-cpp=0.7.9-1* \
     ros-dashing-demo-nodes-py=0.7.9-1* \
     && rm -rf /var/lib/apt/lists/*
diff --git a/ros_eloquent-ros1-bridge-bionic/Dockerfile b/ros_eloquent-ros1-bridge-bionic/Dockerfile
index 2f831bc..852f51e 100644
--- a/ros_eloquent-ros1-bridge-bionic/Dockerfile
+++ b/ros_eloquent-ros1-bridge-bionic/Dockerfile
@@ -10,16 +10,17 @@ RUN echo "deb http://packages.ros.org/ros/ubuntu bionic main" > /etc/apt/sources
 
 ENV ROS1_DISTRO melodic
 ENV ROS2_DISTRO eloquent
+
 # install ros packages
 RUN apt-get update && apt-get install -y --no-install-recommends \
-    ros-melodic-ros-comm=1.14.9-1* \
+    ros-melodic-ros-comm=1.14.10-1* \
     ros-melodic-roscpp-tutorials=0.9.3-1* \
     ros-melodic-rospy-tutorials=0.9.3-1* \
     && rm -rf /var/lib/apt/lists/*
 
 # install ros2 packages
 RUN apt-get update && apt-get install -y --no-install-recommends \
-    ros-eloquent-ros1-bridge=0.8.2-1* \
+    ros-eloquent-ros1-bridge=0.8.2-3* \
     ros-eloquent-demo-nodes-cpp=0.8.4-1* \
     ros-eloquent-demo-nodes-py=0.8.4-1* \
     && rm -rf /var/lib/apt/lists/*
diff --git a/ros_rolling-ros-core-focal/Dockerfile b/ros_rolling-ros-core-focal/Dockerfile
index 52972d7..a099ebc 100644
--- a/ros_rolling-ros-core-focal/Dockerfile
+++ b/ros_rolling-ros-core-focal/Dockerfile
@@ -29,7 +29,7 @@ ENV ROS_DISTRO rolling
 
 # install ros2 packages
 RUN apt-get update && apt-get install -y --no-install-recommends \
-    ros-rolling-ros-core=0.9.1-2* \
+    ros-rolling-ros-core=0.9.2-1* \
     && rm -rf /var/lib/apt/lists/*
 
 # setup entrypoint
diff --git a/ros_rolling-ros1-bridge-focal/Dockerfile b/ros_rolling-ros1-bridge-focal/Dockerfile
index 43e9416..26fbc7d 100644
--- a/ros_rolling-ros1-bridge-focal/Dockerfile
+++ b/ros_rolling-ros1-bridge-focal/Dockerfile
@@ -19,9 +20,9 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 
 # install ros2 packages
 RUN apt-get update && apt-get install -y --no-install-recommends \
-    ros-rolling-ros1-bridge=0.9.2-2* \
-    ros-rolling-demo-nodes-cpp=0.10.0-1* \
-    ros-rolling-demo-nodes-py=0.10.0-1* \
+    ros-rolling-ros1-bridge=0.9.4-2* \
+    ros-rolling-demo-nodes-cpp=0.10.1-1* \
+    ros-rolling-demo-nodes-py=0.10.1-1* \
     && rm -rf /var/lib/apt/lists/*
 
 # setup entrypoint
diff --git a/ros_rolling/Dockerfile b/ros_rolling/Dockerfile
index c37f255..60ce2c0 100644
--- a/ros_rolling/Dockerfile
+++ b/ros_rolling/Dockerfile
@@ -26,6 +26,6 @@ RUN colcon mixin add default \
 
 # install ros2 packages
 RUN apt-get update && apt-get install -y --no-install-recommends \
-    ros-rolling-ros-base=0.9.1-2* \
+    ros-rolling-ros-base=0.9.2-1* \
     && rm -rf /var/lib/apt/lists/*
 
```

</details>

PR #8793, opened 2020-09-28
XWiki 12.8 released

tmortagne:


tmortagne:
Fixing the connector's checksum, sorry about that.

PR #8810, opened 2020-09-30
Update nextcloud

tilosp:
Manually for now since @tilosp-bot is retired.

Is it possible to use doi-janky only for `generate-stackbrew-library.sh` and not `update.sh`?
Since `update.sh ` is now handled by GitHub Actions.

### DIFF
--- a/library/nextcloud
+++ b/library/nextcloud
@@ -1,4 +1,4 @@
-# This file is generated via https://github.com/nextcloud/docker/blob/323015718be9a2abacf9fb6213ddcf082db435a0/generate-stackbrew-library.sh
+# This file is generated via https://github.com/nextcloud/docker/blob/9c9e8154198131aa283ad478034eb22abe92bdfc/generate-stackbrew-library.sh
 
 Maintainers: Nextcloud <docker@nextcloud.com> (@nextcloud)
 GitRepo: https://github.com/nextcloud/docker.git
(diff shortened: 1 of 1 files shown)


tianon:
> Is it possible to use doi-janky only for `generate-stackbrew-library.sh` and not `update.sh`?

Yes!  I've done a quick test with this of doing exactly that via https://github.com/docker-library/oi-janky-groovy/commit/35f8f5ca5a93a885eb69ea7c85645369f4339df9, and you can see the result via https://github.com/docker-library/official-images/compare/master...docker-library-bot:nextcloud (looks like we have the same odd entry reordering we see with J0WI's PRs :confused:).

tilosp:
> Yes! I've done a quick test with this of doing exactly that via [docker-library/oi-janky-groovy@35f8f5c](https://github.com/docker-library/oi-janky-groovy/commit/35f8f5ca5a93a885eb69ea7c85645369f4339df9), and you can see the result via [master...docker-library-bot:nextcloud](https://github.com/docker-library/official-images/compare/master...docker-library-bot:nextcloud) (looks like we have the same odd entry reordering we see with J0WI's PRs confused).

Great, this saves me the work of figuring out how to do this with github actions.

Feel free to close this pr and use the `docker-library-bot:nextcloud` branch to get the reordering done right away.

tianon:
I did some testing locally, and found that if I set `LANG=C`, I end up getting a result similar to what you're getting, so it's probably worth adding something like `export LANG=C` early in your script to just force that deterministic locale setting for more deterministic ordering.

Also, I think we should probably have @docker-library-bot commit to a fork of https://github.com/docker-library/official-images that you can actually open a PR from (unless you want to monitor/copy https://github.com/docker-library-bot/official-images/tree/nextcloud anytime the bot pushes to it with something like https://github.com/docker-library-bot/official-images/commits/nextcloud.atom).

yosifkit:
LGTM, Just the regular test failures:

```console
testing nextcloud:19.0.3-apache
	'nextcloud-cli-postgres' [6/8]...failed
```

PR #8837, opened 2020-10-05
Upgrade to 8.5

philippe-perrin-sonarsource:
This upgrades to the newest SonarQube version.

### DIFF
--- a/library/sonarqube
+++ b/library/sonarqube
@@ -4,16 +4,16 @@ Maintainers: Janos Gyerik <janos.gyerik@sonarsource.com> (@janos-ss),
              Pierre Guillot <pierre.guillot@sonarsource.com> (@pierre-guillot-sonarsource),
              Michal Duda <michal.duda@sonarsource.com> (@michal-duda-sonarsource)
 GitRepo: https://github.com/SonarSource/docker-sonarqube.git
-GitCommit: 2908701e165041cd2ffc9b618effb3345d4c60e9
+GitCommit: e612cf000910bfec4cf0d3d08b89d25c61cfcbf7
 
 Tags: 7.9.4-community, 7.9-community, lts
 Directory: 7/community
 
-Tags: 8.4.2-community, 8.4-community, 8-community, community, latest
+Tags: 8.5.0-community, 8.5-community, 8-community, community, latest
 Directory: 8/community
 
-Tags: 8.4.2-developer, 8.4-developer, 8-developer, developer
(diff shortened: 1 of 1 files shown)


library/sonarqube
```diff
@@ -4,16 +4,16 @@ Maintainers: Janos Gyerik <janos.gyerik@sonarsource.com> (@janos-ss),
...
              Michal Duda <michal.duda@sonarsource.com> (@michal-duda-sonarsource)
 GitRepo: https://github.com/SonarSource/docker-sonarqube.git
-GitCommit: 2908701e165041cd2ffc9b618effb3345d4c60e9
+GitCommit: f5cdd43e0d6d931ca12becbac9a495536a3869dd
 
 Tags: 7.9.4-community, 7.9-community, lts
 Directory: 7/community
 
-Tags: 8.4.2-community, 8.4-community, 8-community, community, latest
+Tags: 8.5-community, 8-community, community, latest
```
yosifkit, line 12: Did you still want to have the more specific tag for `8.5.0`?

philippe-perrin-sonarsource: Yes we do. Thank you.

philippe-perrin-sonarsource: Indeed. I've just pushed the suggested changes. Thanks

tianon:
This bit seems really weird to me -- it installs a number of packages, then purges them immediately (and purges a cache which should already be empty):

```diff
+    export PATH="/opt/java/openjdk/bin:$PATH"; \
+    apk add --no-cache --virtual .build-deps bash binutils; \
+    apk del --purge .build-deps; \
+    rm -rf /var/cache/apk/*; \
```

Is this intentional?  What's the goal of it?

philippe-perrin-sonarsource:
> This bit seems really weird to me -- it installs a number of packages, then purges them immediately (and purges a cache which should already be empty):
> 
> ```diff
> +    export PATH="/opt/java/openjdk/bin:$PATH"; \
> +    apk add --no-cache --virtual .build-deps bash binutils; \
> +    apk del --purge .build-deps; \
> +    rm -rf /var/cache/apk/*; \
> ```
> 
> Is this intentional? What's the goal of it?

Good catch, it wasn't intentional. I've just made the modification.
Thanks for the feedback.

tianon:
FWIW, I think the following is also still not actually doing anything (since the `RUN` line it is in ends shortly after it is set):

```diff
+    export PATH="/opt/java/openjdk/bin:$PATH"; \
```

PR #8848, opened 2020-10-06
Update openjdk

tianon:
Changes:

- https://github.com/docker-library/openjdk/commit/2d4433f: Update to 16-ea+14


yosifkit:
GitHub Actions is having a bit of a flake (https://www.githubstatus.com/incidents/t34640ccrmfs), but the `openjdk:16-ea-14-jdk-alpine3.12` build is green :innocent:

PR #8854, opened 2020-10-07
Update for the release of websphre-liberty 20.0.0.10

crpotter:
Release updates for websphere-liberty version 20.0.0.10

### DIFF
--- a/library/websphere-liberty
+++ b/library/websphere-liberty
@@ -2,7 +2,7 @@ Maintainers: Chris Potter <crpotter@us.ibm.com> (@crpotter),
              Wendy Raschke <wraschke@us.ibm.com> (@wraschke),
              Arthur De Magalhaes <arthurdm@ca.ibm.com> (@arthurdm)
 GitRepo: https://github.com/WASdev/ci.docker.git
-GitCommit: 259400d0cfe000c3741919640983c4e9a4fca3b8
+GitCommit: 79c6e4c3f4a0f6b08f399ce4781176877b5b420d
 Architectures: amd64, i386, ppc64le, s390x
 
 Tags: beta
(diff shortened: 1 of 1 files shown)


tianon:
Can we please figure out some way to get those massive constantly-changing license files out of the build context so that the diffs here are reasonable?

yosifkit:
The only other image that almost had a large licenses context was https://github.com/docker-library/official-images/pull/8366 and they decided to download and verify them during the build to get them out of the build context.

I guess my question is that if they are provided for open source license compliance of software included with the websphere liberty bundle, then why is it not just part of the `wlp.zip`?  But if it is for software installed via `apt`, then surely `/usr/share/doc/[package]/copyright` is sufficient.

tianon:
I think the misunderstanding here is on me -- I was under the impression that this was more akin to a license enabling file (since it's been here but not in `open-liberty`), not just a raw dump of open source license text.

I fully agree with @yosifkit here -- if this is just the licenses of software contained in the image and is something you want to keep in this aggregated form, it should really be generated during the build of the image itself or else it will almost always suffer bitrot (as discussed in https://github.com/docker-library/official-images/pull/8366#issuecomment-662138721 and the following comments).

arthurdm:
thanks guys - we are looking into updating this PR with a change that pulls in the licenses via URL.

arthurdm:
I've created [this PR](https://github.com/WASdev/ci.docker/pull/363) into the WebSphere Liberty repo.

@crpotter - once that is merged, can you please update the GitCommit for this PR?

arthurdm:
hi @tianon / @yosifkit - I've created this new PR with the proposed license changes: https://github.com/docker-library/official-images/pull/8930

Since @crpotter is away this week, can you please review that one, and close this one?

PR #8858, opened 2020-10-08
Release Oracle Linux 7 Update 9

Djelibeybi:
Release of [Oracle Linux 7 Update 9](https://blogs.oracle.com/linux/announcing-the-release-of-oracle-linux-7-update-9) for the `amd64` and `arm64v8` architectures.

### DIFF
--- a/library/oraclelinux
+++ b/library/oraclelinux
@@ -4,10 +4,10 @@ GitCommit: 01a15ec99c7470a3391c691509db1759b41eaf66
 GitFetch: refs/heads/master
 # https://github.com/oracle/container-images/tree/dist-amd64
 amd64-GitFetch: refs/heads/dist-amd64
-amd64-GitCommit: 283700269a887af4a45ba9bc614d8ad777cf702c
+amd64-GitCommit: 52339856c3cf67775e37b128952ba43c19507ab2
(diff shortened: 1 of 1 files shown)


library/oraclelinux
```diff
@@ -25,7 +25,11 @@ Tags: 8-slim
...
 Directory: 8-slim
 
-Tags: 7.8, 7
+Tags: 7.9, 7
+Architectures: amd64, arm64v8
+Directory: 7.9
+
+Tags: 7.8
 Architectures: amd64, arm64v8
 Directory: 7.8
```
tianon, line 34: I think you probably know what I'm going to say here... :smirk:

Any particular reason to keep this version listed here?  Removing tags here will remove them from the "Supported" section on the Hub readme (and will prevent us from spending cycles rebuilding them on the official build servers), but the tags will still be available to users who want them.  (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)

tianon: (I suppose the same goes for 8.0 and 8.1, which we've likely talked about previously?)

tianon:
LGTM (so I'm merging), but let's continue discussing the above :smile:

PR #8864, opened 2020-10-08
Remove older oraclelinux tags.

Djelibeybi:
As pointed out by @tianon in #8858, these older tags are no longer required.


tianon:
:heart::exclamation:

PR #8866, opened 2020-10-09
[percona] update psmdb-36

EvgeniyPatlan:


tianon:
Is there a specific reason we have four PRs for what appears to be mostly the same change across different versions instead of just one PR that ecompasses all four? (#8866, #8867, #8868, #8869)

EvgeniyPatlan:
Hi @tianon 
Well it is just because of our procedure how we issue releases.
It is very rare case once we update all products together. So we updated them one by one and for each we have to create PR to update dockers.
If it is needed I can rework and combine these PRs.
Should I do so?

tianon:
It would definitely be much easier to review, and reduce the chance of
merge conflicts. 😅

EvgeniyPatlan:
@tianon here it is https://github.com/docker-library/official-images/pull/8878

PR #8881, opened 2020-10-11
Update Git

keeganwitt:
Temporarily using PPA for Git until https://github.com/AdoptOpenJDK/openjdk-docker/pull/397 gets merged.

### DIFF
--- a/library/gradle
+++ b/library/gradle
@@ -1,6 +1,6 @@
 Maintainers: Keegan Witt <keeganwitt@gmail.com> (@keeganwitt)
 GitRepo: https://github.com/keeganwitt/docker-gradle.git
-GitCommit: f6e27a40b3c8b6dcc8dc07a278f7a9f0c2a808cb
+GitCommit: 391db05e9a4c9aad68979d7df2eb42f2ada5e303
 
 Tags: 6.6.1-jdk8, 6.6-jdk8, jdk8, 6.6.1-jdk, 6.6-jdk, jdk, 6.6.1, 6.6, latest
 Architectures: amd64, arm32v7, arm64v8, ppc64le, s390x


tianon:
Somehow, I don't think updating `library/amazoncorretto` is what you intended here :smile:

keeganwitt:
> 
> 
> Somehow, I don't think updating `library/amazoncorretto` is what you intended here 😄

Oops.  It certainly wasn't.

tianon:
(force push due to https://github.com/actions/runner/issues/751 so we can hopefully get some test builds)

tianon:
Arg, looks like `add-apt-repository` defaults to using `https://keyserver.ubuntu.com`, which our mitigations for https://github.com/docker-library/faq#openpgp--gnupg-keys-and-verification end up breaking.

It looks like (https://manpages.ubuntu.com/manpages/bionic/en/man1/add-apt-repository.1.html) it's supposed to support a `--keyserver` flag to explicitly overwrite that, which would be an easy solution here, but it appears at some point the `ppa:` functionality of it was rewritten to *only* support `https://keyserver.ubuntu.com` (hard-coded; https://salsa.debian.org/pkgutopia-team/software-properties/-/blob/bb1cf92d19d48aa0779f6a93e15dbb360fcb6dee/softwareproperties/ppa.py#L49), so I'm not sure what to do here. :disappointed:

keeganwitt:
I'm thinking maybe
`apt-key adv --keyserver ha.pool.sks-keyservers.net --recv E1DD270288B4E6030699E45FA1715D88E1DF1F24` (that seems to be the key mentioned on the PPA page: https://launchpad.net/~git-core/+archive/ubuntu/ppa).

https://github.com/keeganwitt/docker-gradle/pull/158

yosifkit:
We recommend using gpg directly (as recommended by the [`apt-key add` manpage](https://manpages.debian.org/buster/apt/apt-key.8.en.html#COMMANDS)).

https://github.com/docker-library/official-images/tree/548d851413ad8c3682e6514b7f0f58fd47305852#image-build:
```Dockerfile
RUN set -eux; \
    key='A4A9406876FCBD3C456770C88C718D3B5072E1F5'; \
    export GNUPGHOME="$(mktemp -d)"; \
    gpg --batch --keyserver ha.pool.sks-keyservers.net --recv-keys "$key"; \
    gpg --batch --armor --export "$key" > /etc/apt/trusted.gpg.d/mysql.gpg.asc; \
    gpgconf --kill all; \
    rm -rf "$GNUPGHOME"; \
    apt-key list
```

keeganwitt:
What's the `apt-key list` for?

keeganwitt:
Well shoot, it looks like it runs the key import even if its already in the keyring.  Unless I imported it wrong?  I didn't notice this in my test, but I must not have had logging turned up enough to see it.

tianon:
At this point, you're now importing the key manually and all `add-apt-repository` is going to do for you is set up `sources.list` and do `apt-get update`, so you might as well just do `echo 'deb http://ppa.launchpad.net/git-core/ppa/ubuntu bionic main' > /etc/apt/sources.list.d/git-core-ppa.list` (or similar) yourself. :sweat_smile:

> What's the `apt-key list` for?

We use `apt-key list` as a basic smoke test to make sure the format of the resulting file is something APT is willing to ingest.  In some places, we get slightly more specific (`apt-key list | grep -i mysql`, for example), but we've found that just invoking `apt-key list` is usually enough to catch the most common errors.  However, since you're then immediately doing `apt-get update` in the same layer, it's likely a moot point since APT will then validate it anyhow. :smile:

keeganwitt:
Well, it's less flakey than before, but 1 of the pulls from PPA timed out.  Hopefully AdoptOpenJDK base will be updated soon, so this won't be very long-lived anyway...

PR #8896, opened 2020-10-15
[percona] Update PS-8.0 version

Sudokamikaze:


library/percona
```diff
@@ -7,8 +7,8 @@ GitRepo: https://github.com/percona/percona-docker.git
 GitFetch: refs/heads/master
 Architectures: amd64
 
-Tags: 8.0.20-11-centos, 8.0-centos, 8-centos, 8.0.20-11, 8.0, 8, ps-8.0.20-11, ps-8.0, ps-8
-GitCommit: 65855912b9946093cee42f987336440972405391
+Tags: 8.0.21-12-centos, 8.0-centos, 8-centos, 8.0.21-12, 8.0, 8, ps-8.0.21-12, ps-8.0, ps-8
+GitCommit: 387a5ec03894c49e89cb992611fa365dcaf40ff1
 Directory: percona-server.80
```
yosifkit, line 12:
```suggestion
Directory: percona-server-8.0
```

Sudokamikaze: Gah, I thought we had a switch to new scheme for 8.0..
Thanks, added this

yosifkit:
We don't actively recommend using labels. If an image maintainer wants to have labels, that is fine, but Image label names should only use the pre-defined annotation keys of the [image spec](https://github.com/opencontainers/image-spec/blob/v1.0.1/annotations.md).

See also https://github.com/docker-library/official-images/issues/3540

Sudokamikaze:
> We don't actively recommend using labels. If an image maintainer wants to have labels, that is fine, but Image label names should only use the pre-defined annotation keys of the [image spec](https://github.com/opencontainers/image-spec/blob/v1.0.1/annotations.md).
> 
> See also #3540

Thank you, it'll be reworked. For now, moving this PR into draft state

Sudokamikaze:
@yosifkit Could you please take a look? We've removed conflicting labels from our images

tianon:
```diff
+ENV PS_VERSION 8.0.21-12.2
+ENV OS_VER el8
+ENV FULL_PERCONA_VERSION "$PS_VERSION.$OS_VER"
```

I'm a little confused as to why this block was moved up so far?  It seems unnecessary to bust the build cache so far up when these variables aren't used until much lower.  Thoughts?

tianon:
```diff
 # purge and re-create /var/lib/mysql with appropriate ownership
-RUN /usr/bin/install -m 0775 -o mysql -g root -d /var/lib/mysql /var/run/mysqld /docker-entrypoint-initdb.d \
+RUN /usr/bin/install -m 0775 -o mysql -g root -d /var/lib/mysql /var/run/mysqld /docker-entrypoint-initdb.d; \
 # comment out a few problematic configuration values
```

I'm 100% all for switching from `&&` to `;` but this block will also need a `set -e` (and I'd suggest also at least `-x` but `-u` is similarly useful, so we tend to use `set -eux;` in our repos).

Sudokamikaze:
> ```diff
> +ENV PS_VERSION 8.0.21-12.2
> +ENV OS_VER el8
> +ENV FULL_PERCONA_VERSION "$PS_VERSION.$OS_VER"
> ```
> 
> I'm a little confused as to why this block was moved up so far? It seems unnecessary to bust the build cache so far up when these variables aren't used until much lower. Thoughts?

Thank you, created a PR in our repository.

> ```diff
>  # purge and re-create /var/lib/mysql with appropriate ownership
> -RUN /usr/bin/install -m 0775 -o mysql -g root -d /var/lib/mysql /var/run/mysqld /docker-entrypoint-initdb.d \
> +RUN /usr/bin/install -m 0775 -o mysql -g root -d /var/lib/mysql /var/run/mysqld /docker-entrypoint-initdb.d; \
>  # comment out a few problematic configuration values
> ```
> 
> I'm 100% all for switching from `&&` to `;` but this block will also need a `set -e` (and I'd suggest also at least `-x` but `-u` is similarly useful, so we tend to use `set -eux;` in our repos).

This will be included as well ;)

Sudokamikaze:
@tianon Could you please review? I've pushed needed changes.
Thanks in advance :)

tianon:
Looks good, thanks! :+1:

PR #8916, opened 2020-10-20
Replace Java 14 with Java 15

keeganwitt:
Given the current base tags for Java 15 images are now 20.04, but will be flipped to 18.04, then back to 20.04 (https://github.com/AdoptOpenJDK/openjdk-docker/issues/445), I'm not sure when it'd be best to merge this.


keeganwitt:
Now that the rollback has merged (https://github.com/docker-library/official-images/pull/8918), I think we can merge this, and I'll revert my manual Git upgrade when the tag with 20.04 is available.

JohannesLichtenberger:
:+1:

PR #8932, opened 2020-10-22
Telegraf v1.15.4, v1.16.0

ssoroka:


yosifkit:
`alpine:latest` and `buildpacks-deps:stable` (i.e. `debian:stable`) will keep the image more up-to-date, then say `buster`, and `3.12`, but the warning is that they are a square wheel of change and might cause build issues in the future when they move to the next version (likely December 2020 for the next [Alpine release](https://wiki.alpinelinux.org/wiki/Alpine_Linux:Releases), and possibly mid 2021 for the next Debian stable [release](https://wiki.debian.org/DebianReleases)).

PR #8933, opened 2020-10-22
node: update to v15.0.0

SimenB:
https://nodejs.org/en/blog/release/v15.0.0/

### DIFF
--- a/library/node
+++ b/library/node
@@ -1,39 +1,74 @@
-# this file is generated via https://github.com/nodejs/docker-node/blob/94a4913649bf6b3d13da30c4cb480770a86b3877/generate-stackbrew-library.sh
+# this file is generated via https://github.com/nodejs/docker-node/blob/9a8ea84f6258c8ad9edbb010484a52016df66e8d/generate-stackbrew-library.sh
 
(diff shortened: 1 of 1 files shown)


nschonni:
Should we try and figure out why `arm64v8` and `ppcle64` are having issues before landing the 15.0.1?

SimenB:
no, we shouldn't block a bug fix for regressions in v15 on something like that

yosifkit:
From the logs, it looks like the pp64le build worked fine but must've accidentally got deleted before the job was done (server was completely full, so "ripe" images were being cleaned); a new build is in the queue.  As for `arm64v8`, we're looking into it (`node` isn't the only failure :scream:), but it seems to be host related.

nschonni:
✨ awesome. If we land the 15.0.1 upstream, you can make the call on whether the PR here lands here anyway

tianon:
I've retriggered arm64v8, and the only failures are `node:15.0.0-stretch` and `node:15.0.0-stretch-slim` (not `buster` or `alpine3.12`), so IMO this is not worth holding anything up for.

PR #8940, opened 2020-10-22
Update Ubuntu: Release Ubuntu 20.10 (Groovy Gorilla)

powersj:


powersj:
@philroche can you please review as well to ensure I did this correctly? It does looks like groovy picked up the rolling tag.

philroche (approved):

PR #8949, opened 2020-10-24
Composer 1.10.16 & 2.0.2

alcohol:


alcohol:
2 is not tagged as latest yet on purpose (if you were wondering)

naderman:
Hey, is there anything we can do on our side to speed up the publication of these new tags? I'm getting contacted by tons of people who are asking why they cannot install the new 2.0 major release from dockerhub yet.

Seldaek (approved):

tianon:
> Hey, is there anything we can do on our side to speed up the publication of these new tags? I'm getting contacted by tons of people who are asking why they cannot install the new 2.0 major release from dockerhub yet.

Not publish them on a Saturday? :smile:

Seldaek:
@tianon that's completely fair of course for a manual process, but our question was more on the line of.. isn't there a way to get simple version bumps merged faster/automatically somehow? If the dockerfile does not change at all, it's kind of a waste of time for you to review, and it always means a delay for our users no matter how fast you react.

PR #8960, opened 2020-10-26
Fedora Update 2020-10-26

cverna:
Fedora 33 is now the latest Fedora release.

Signed-off-by: Clement Verna <cverna@tutanota.com>


tianon:
> Fedora 31 will be maintained until four weeks after the release of Fedora 33.

:eyes:

PR #8969, opened 2020-10-27
Release of GeoNetwork 4.0.0

juanluisrp:
Release of GeoNetwork 4.0.0

### DIFF
--- a/library/geonetwork
+++ b/library/geonetwork
@@ -26,5 +26,5 @@ Directory: 3.8.3/postgres
 
 Tags: 4.0.0, 4.0, 4
 Architectures: amd64
-GitCommit: 0f8341a07639191ad221b7b16a0757393b475444
+GitCommit: 7676a4c49dfdbb896b8a14b810b3dcbf51d56390
 Directory: 4.0.0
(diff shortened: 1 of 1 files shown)


yosifkit:
With the new `VOLUME`, you'll likely need to update the Hub docs page since, as far as I can tell, this new image doesn't support changing the `DATA_DIR` like they [currently suggest](https://github.com/docker-library/docs/tree/6682d2bd72929a3573a9dae8293bfd1347dbb787/geonetwork#set-the-data-directory-and-h2-db-file).

LGTM

tianon:
Generally LGTM, just a couple minor comments that can be addressed later (if at all):

---

```diff
+    echo "$@"
```

This one looks like it was probably left in on accident. :sweat_smile:

---

```diff
+ENV GN_FILE GeoNetwork-4.0.0-0.war
+ENV GN_VERSION 4.0.0
+ENV GN_DOWNLOAD_MD5 6cf3524a0e3e51e713402b97e086ba20
```

I'd suggest moving these down closer to where they're used for better caching.

PR #8972, opened 2020-10-27
bump kong to 2.2.0

gszr:


hutchic:
an explanation as to why we're doing `useradd kong || true` is we're migrating the `useradd` for non-alpine packages to `after-install.sh` but need a version of compatibility. So subsequent releases we should be able to safely remove the `useradd` lines entirely

yosifkit:
Similar comment to https://github.com/docker-library/official-images/pull/8211#issuecomment-646239481; specifically and (`&&`) binds tighter than or (`||`) so it would need to switch all `&&` to semicolons like the if statement or use curly braces for grouping.

gszr:
@yosifkit We've implemented the change. Thanks!

yosifkit:
This bracket enclosure won't quite work. The `useradd` never runs since the shell tries to run `{` with `useradd` as its arguments:
```console
$ docker run -it --rm debian sh -xc 'echo hi && \{ useradd kong || true \} && echo done'+ echo hi
hi
+ { useradd kong
sh: 1: {: not found
+ true }
+ echo done
done
```
 What you really wanted was like this (https://www.gnu.org/software/bash/manual/html_node/Command-Grouping.html):
```console
docker run -it --rm debian sh -xc 'echo hi && { useradd kong || true; } && echo done'
+ echo hi
hi
+ useradd kong
+ echo done
done
```

```diff
-&& \{ useradd -ms /bin/bash kong || true \}
+&& { useradd -ms /bin/bash kong || true; }
```

gszr:
Done, @yosifkit! Thanks!

yosifkit:
> WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

It was not new to this change, but you should likely swap the `apt install` to be `apt-get install`.

PR #8979, opened 2020-10-28
elixir 1.11.1, and change to the new repo url

getong:
As the c0b/docker-elixir#145 does not merge for a long time, we decide to move to new repo, https://github.com/erlef/docker-elixir/

### DIFF
--- a/library/elixir
+++ b/library/elixir
@@ -1,19 +1,35 @@
-# this file is generated via https://github.com/c0b/docker-elixir/blob/e2d20313fefcd13b8d3b1bcd94df60bd4f02a0b3/generate-stackbrew-library.sh
+# this file is generated via https://github.com/erlef/docker-elixir/blob/4d889a2a5bc2611ac6b354ec4a9392c085cd7f0c/generate-stackbrew-library.sh
 
-Maintainers: . <c0b@users.noreply.github.com> (@c0b)
-GitRepo: https://github.com/c0b/docker-elixir.git
+Maintainers: . <c0b@users.noreply.github.com> (@c0b),
+             Tristan Sloughter <t@crashfast.com> (@tsloughter)
+GitRepo: https://github.com/erlef/docker-elixir.git
 
-Tags: 1.10.4, 1.10, latest
+Tags: 1.11.1, 1.11, latest
+Architectures: amd64, arm32v7, arm64v8, i386, s390x, ppc64le
+GitCommit: 5d25131693118e23795b4605f5473cb548311403
+Directory: 1.11
+
+Tags: 1.11.1-slim, 1.11-slim, slim
+Architectures: amd64, arm32v7, arm64v8, i386, s390x, ppc64le
+GitCommit: 5d25131693118e23795b4605f5473cb548311403
+Directory: 1.11/slim
+
+Tags: 1.11.1-alpine, 1.11-alpine, alpine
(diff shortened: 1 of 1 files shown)


library/elixir
```diff
@@ -1,19 +1,34 @@
-# this file is generated via https://github.com/c0b/docker-elixir/blob/e2d20313fefcd13b8d3b1bcd94df60bd4f02a0b3/generate-stackbrew-library.sh
+# this file is generated via https://github.com/erlef/docker-elixir/blob/a6086e61c4d193a569e046d57f212872345839e3/generate-stackbrew-library.sh
 
 Maintainers: . <c0b@users.noreply.github.com> (@c0b)
```
vrcca, line 3: Should this be updated as well?

getong: I think we should respect him and his work.

vrcca: Of course. I meant, add a new maintainer to the list.

getong: I don't want to add myself.

getong:
With erlef/docker-elixir#3 , erlef/docker-elixir#4 , all the github action error has been fixed. I think this pr is ok to be merged.

library/elixir
tianon: If @c0b is no longer maintaining this image, it is not appropriate for @c0b to be the only human listed (or even to be listed at all, really) -- we use this list for pinging maintainers on related issues (docs changes, base image changes, etc), so we really need some humans here.  The format is either `Name <email> (@github)` or `Name (@github)` so it doesn't really need to be anything elaborate.

I don't have a problem with leaving @c0b listed, but I really wish we could get some kind of reply that this is OK with them. :disappointed:  (I hope they're doing ok!)

tsloughter: @getong since you don't want to be added I can be used:`Tristan Sloughter t@crashfast.com (@tsloughter)` -- we should also put `Erlang Ecosystem Foundation Build and Packaging Working Group https://github.com/erlef/eef-build-and-packaging-wg` since that working group is really who will be maintaining it.

getong: Add Tristan Sloughter as the maintainer, the pr now is ready to be merged.

tsloughter (approved):

library/elixir
```diff
@@ -34,91 +50,91 @@ GitCommit: 0d9f47458468a8bf1407374731cbec077ab6f895
...
 Directory: 1.6/otp-21
 
 Tags: 1.6.6-otp-21-alpine, 1.6-otp-21-alpine
-Architectures: amd64, arm32v7, arm64v8, i386, s390x, ppc64le
-GitCommit: 084efbdda747411b3a41c231deaff03c437e5aad
+Architectures: amd64, arm32v7, arm64v8, i386
+GitCommit: 373b3a8017bd774b7d193818a326de0fde7f6733
 Directory: 1.6/otp-21-alpine
 
 Tags: 1.5.3, 1.5
```
yosifkit, line 117: Any particular reason to keep these older versions listed here? It looks like 1.5 and 1.4 aren't even getting security patches any more (https://hexdocs.pm/elixir/compatibility-and-deprecations.html)  Removing it here will remove it from the "Supported" section on the Hub readme (and will prevent us from spending cycles rebuilding it on the official build server), but the tag will still be available to users who want it.  (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)

PR #8999, opened 2020-11-02
archlinux: Release 20201102.0.8190

archlinux-github:

Maintainers: @SantiagoTorres @shibumi @pierres @hashworks

### DIFF
--- a/library/archlinux
+++ b/library/archlinux
@@ -1,12 +1,18 @@
-# tarballs built by soyuz on a rootless environment
-# see https://github.com/archlinux/archlinux-docker/tree/no-root-build
+# https://gitlab.archlinux.org/archlinux/archlinux-docker
 
-Maintainers: Santiago Torres-Arias <santiago@archlinux.org> (@SantiagoTores),
+Maintainers: Santiago Torres-Arias <santiago@archlinux.org> (@SantiagoTorres),
              Christian Rebischke <Chris.Rebischke@archlinux.org> (@shibumi),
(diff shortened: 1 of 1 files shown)


hashworks:
Just wanted to chime in and confirm that this is a valid PR. We're now switching to weekly automated releases (every Sunday at 00:00 UTC) as noted [here](https://github.com/docker-library/docs/pull/1825).

Not sure why the "Generate Jobs" CI failes:
```
bashbrew version v0.1.2
Processing archlinux:latest ...
/home/runner/bashbrew/scripts/github-actions/generate.sh: line 105: metas["$parent"]: unbound variable
```

tianon:
Conceptually this seems OK (and in line with what we've discussed), but in practice this means that `archlinux:latest` is now `FROM archlinux:latest`, and I think that circular dependency is what's causing the CI to fail, and is likely to cause a lot more weird bugs/failures in other bits of our tooling (because it's a difficult to resolve circular dependency that technically requires a double-build each time in order to "settle" properly, assuming the build is fully deterministic/reproducible).

hashworks:
> that circular dependency is what's causing the CI to fail

If it changes anything, that is just the first stage of the build to verify the rootfs download. If needed we could switch to a different official image here or use a previous tag (not `latest`).

hashworks:
This will be fixed in https://gitlab.archlinux.org/archlinux/archlinux-docker/-/merge_requests/47

PR #9001, opened 2020-11-02
Update bash

tianon:
Changes:

- https://github.com/tianon/docker-bash/commit/755b314: Update devel to 20201026, commit 22b8a76dd6269038576e0370074ff46c7b38cb32


yosifkit:
over9000

PR #9034, opened 2020-11-06
Aerospike 5.2.0.7

volmarl:

### DIFF
--- a/library/aerospike
+++ b/library/aerospike
@@ -1,16 +1,16 @@
 Maintainers: Lucien Volmar <lucien@aerospike.com> (@volmarl)
 
-Tags: 5.0.0.15
+Tags: 5.0.0.17
 Architectures: amd64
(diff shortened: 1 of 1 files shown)


yosifkit:
The change is fine, but did you want to keep the entrypoint script in the entrypoint command so that users could still do `docker run aerospike --flag`?

For example, like this:
```diff
+ENTRYPOINT ["/usr/bin/dumb-init", "--", "/entrypoint.sh"]
 # Execute the run script in foreground mode
-ENTRYPOINT ["/entrypoint.sh"]
 CMD ["asd"]
```

tianon:
(friendly ping :heart:)

volmarl:
@tianon I think you have a good point. I'll update the next build with the changes. 5.2.0.7 should be released tomorrow. I'll update docker file with ENTRYPOINT ["/usr/bin/dumb-init", "--", "/entrypoint.sh"]

PR #9046, opened 2020-11-08
Add tests for phpMyAdmin

J0WI:
Merging the tests from https://github.com/phpmyadmin/docker/tree/master/testing


tianon:
I'm a bit confused -- this (especially `test/tests/phpmyadmin-apache-run/world.sql` :scream:) seems kind of large/intensive for a basic smoke test?

If phpmyadmin/docker#313 is the motivation for this, I'd suggest something like https://github.com/tianon/dockerfiles/blob/bb3600a7b4027e9c74f63215a4d85398c657b73f/.github/workflows/generate.sh#L42-L48 in order to provide an additional `config.sh`, which should allow additional tests to live in that repo directly. (Alternatively, have that CI add to the prepare step to inject some contents into `~/oi/test/tests/` and append to `~/oi/test/config.sh` directly, if that feels easier to maintain than an admittedly unsightly regex.)

williamdes (approved):

J0WI:
> I'm a bit confused -- this (especially test/tests/phpmyadmin-apache-run/world.sql scream) seems kind of large/intensive for a basic smoke test?

True, I just moved the existing tests. I guess I could easily reduce it.

Is there any documentation for `config.sh`?

williamdes:
> True, I just moved the existing tests. I guess I could easily reduce it.

You can go ahead :+1: , I will be ready to merge your work

test/tests/phpmyadmin-apache-run/phpmyadmin_test.py
```diff
@@ -0,0 +1,77 @@
...
+
+import mechanize
+import tempfile
+import pytest
+
+
+def do_test_content(match, content):
+    assert(match in content)
+
+def test_phpmyadmin(url, username, password, server, sqlfile):
```
williamdes, line 16: I recently pushed more changes upstream
And started to try to plug them in on our CI
But the having to start a mysql server blocked me

I think you can pick my changes on the py test script :)

PR #9055, opened 2020-11-10
Switch from PPA Git to Focal Git and temporarily comment out OpenJ9 Java 15 images

keeganwitt:
to avoid an OpenJ9 [bug](https://github.com/eclipse/openj9/issues/11013).


keeganwitt:
It looks like d06e5d9e886645ae34254740b8b9d76b709b9d57 removed arm32v7 from the Java 8 hotspot images. I'm asking for clarification (if this is temporary or what), but feel free to chime in here if you know offhand.

PR #9056, opened 2020-11-10
[varnish] update lts image to 6.0.7

gquintard:
This one is a bit more than the usual switch to a more recent package, it also trade `stretch` for `buster`, since the former is EOL.

I don't know if there are special announcements (I'll post on the Vanrish mailing list), tagging or something like that to do here, if so, please let me know!


tianon:
> I don't know if there are special announcements (I'll post on the Vanrish mailing list), tagging or something like that to do here, if so, please let me know!

We typically only worry about it in cases where there might be breaking changes (like https://github.com/docker-library/postgres/issues/582) and/or it's common for users to be `FROM` the image (and install more stuff), which is why images like `python` and `golang` often have multiple variants like `x.y-buster` vs `x.y-stretch`.  I'm guessing neither of those apply here?

gquintard:
We should be fine, there's a `jemalloc` upgrade, but it that had been an issue, I'm sure somebody would have reported it on the `fresh` variants of the image. Go ahead!

gquintard:
@tianon , can I attract your attention to https://github.com/docker-library/official-images/pull/8674 while I have you?

PR #9063, opened 2020-11-11
Add back hotspot v8 arm32 builds, remove hotspot v8 s390x temporarily.

dinogun:
hotspot v8 on s390x on Ubuntu 20.04 currently fails with an error as documented [here](https://github.com/AdoptOpenJDK/openjdk-docker/issues/458).


dinogun:
@tianon Looks like the jre failures are fixed ! (The only failure seems to be a network issue ?)

tianon:
Well, kinda -- the JRE failures only happen for new major versions (otherwise it automatically prefers to `javac` against the older patch release, as happened here). :sweat_smile:

dinogun:
haha 🤣

PR #9072, opened 2020-11-12
add kapacitor 1.5.7

docmerlin:
This is to update kapacitor to 1.5.7

### DIFF
--- a/library/kapacitor
+++ b/library/kapacitor
@@ -10,9 +10,9 @@ Directory: kapacitor/1.4
 Tags: 1.4-alpine, 1.4.1-alpine
 Directory: kapacitor/1.4/alpine
 
-Tags: 1.5, 1.5.6, latest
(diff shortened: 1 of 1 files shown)


library/kapacitor
```diff
@@ -10,9 +10,9 @@ Directory: kapacitor/1.4
 Tags: 1.4-alpine, 1.4.1-alpine
 Directory: kapacitor/1.4/alpine
 
-Tags: 1.5, 1.5.6, latest
+Tags: 1.5, 1.5.6, 1.5.7, latest
```
yosifkit, line 13: You probably don't want to also tag `1.5.7` as `1.5.6`

There also isn't any change to the `GitCommit` on line 4, so there aren't any new Dockerfiles.

docmerlin:
Thanks am closing this in favor of https://github.com/docker-library/official-images/pull/9092

PR #9097, opened 2020-11-16
Add new "Munge PR" workflow using "pull_request_target"

tianon:
This deprecates our (silly) "Periodic" job in favor of the pull request itself triggering the labels/diff (which is what we wanted originally, as seen by the comments in `periodic.yml`, but we had to wait for `pull_request_target` to come along and implement what we actually need! :smile:)


tianon:
You can see the results of my testing in https://github.com/docker-library/official-images/pull/9085.

yosifkit (approved):

PR #9102, opened 2020-11-16
Re-target the CentOS 7.9.2009 ppc64le Image

bstinsonmhk:
and clean up older tags so that the tests pass more cleanly.


tianon:
Welp, GitHub is doing something weird and won't let me test the CI fix from https://github.com/docker-library/official-images/commit/752c8cde5402c689b626c7cc77736830f53db789, so I've just labelled this manually and we'll hope the next PR that gets opened works properly. :disappointed: :heart:

Thanks for pruning these old versions! :+1:

PR #9104, opened 2020-11-17
Gradle 6.7.1

keeganwitt:


tianon:
Not urgent, but you probably want to drop `s390x` from your `adoptopenjdk:8-*-hotspot` builds (https://github.com/AdoptOpenJDK/openjdk-docker/issues/458).

keeganwitt:
> Not urgent, but you probably want to drop `s390x` from your `adoptopenjdk:8-*-hotspot` builds ([AdoptOpenJDK/openjdk-docker#458](https://github.com/AdoptOpenJDK/openjdk-docker/issues/458)).

Thanks. Sounds like that's another temporary issue as well.

PR #9130, opened 2020-11-19
PHP 8.0 RC5.

desrosj:
Release notes: https://www.php.net/archive/2020.php#2020-11-19-1.


tianon:
Niiiice, thank you for the assist! :metal:

PR #8089, opened 2020-05-29
alpine: add 3.12.0 and remove EOL 3.8

ncopa:

PR #9135, opened 2020-11-19
Update Consul to 1.8.6, 1.7.10, 1.6.10

freddygv:


yosifkit:
CI tests are complaining that `alpine:3.8` is being used (it was EOL back in May in https://github.com/docker-library/official-images/pull/8089). It looks like https://github.com/hashicorp/docker-consul/commit/4d5b197d230a11731c4f90640194465173b482e6 inadvertently downgraded the Alpine version on the `1.6.10` Dockerfile.

See also https://github.com/hashicorp/docker-consul/compare/7650dbe7ef6dd06a67fd202bd0c8e847dd6a4798...4d5b197d230a11731c4f90640194465173b482e6 (the two commits for `1.6.9` and `1.6.10`)

tianon:
```diff
+    apk -X https://dl-cdn.alpinelinux.org/alpine/v3.8/main add tzdata=2020a-r0 && \
```

I'm guessing https://gitlab.alpinelinux.org/alpine/aports/-/issues/12057 is probably the underlying motivation for this, but generally mixing packages from Alpine releases doesn't end very well, so I'd suggest including at the very least a comment as to why this (otherwise eyebrow-raising) line is included such as a link to https://gitlab.alpinelinux.org/alpine/aports/-/issues/12057.

tianon:
Very nice, thank you!  LGTM :bow: :heart:

PR #9139, opened 2020-11-20
Update photon image

gerrit-photon:
Update photon image


gerrit-photon:
Create a new one soon

PR #9141, opened 2020-11-20
Update photon image

gerrit-photon:
Update photon image

### DIFF
--- a/library/photon
+++ b/library/photon
@@ -4,19 +4,19 @@ Maintainers: Fabio Rapposelli <fabio@vmware.com> (@frapposelli),
 GitRepo: https://github.com/vmware/photon-docker-image.git
 Directory: docker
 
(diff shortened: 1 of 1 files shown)


gerrit-photon:
Hi,

Could you please show me how to fix the error 'message: 'Body is too long (maximum is 65536 characters)'' complained by Munge PR step?

Is the photon file itself too long or any other message is too long?

Thanks,
Michelle

yosifkit:
> Could you please show me how to fix the error 'message: 'Body is too long (maximum is 65536 characters)'' complained by Munge PR step?

After it generates a diff of the current image contexts to the new image contexts (including the file list in base image `tar.gz` files), it tries to make a comment to github on the issue (since that is much nicer to view than the build logs). Sometimes the comment is too long and so the failure is expected.

PR #9172, opened 2020-11-25
Update Ubuntu November 2020

philroche:
Update Ubuntu November 2020


philroche:
@yosifkit These Munge PR errors are new to me. Any guidance on how to replicate and fix?

tianon:
I get the same error when I try to run `diff-pr.sh` locally (`tar: Unexpected EOF in archive`), and looking at the "Test PR" failure it seems `ubuntu:20.10` is failing, so I think what's happening is that `ubuntu-groovy-core-cloudimg-amd64-root.tar.gz` must be broken in some way?

philroche:
@tianon Thanks. This is the first release of ubuntu:20.10 so I will investigate

philroche:
There were build issues for the Ubuntu 20.10 image. I will close this PR until the rebuild is complete

PR #9173, opened 2020-11-25
Update Ubuntu November 2020

philroche:
Also adds first release of Ubuntu 21.04 which is now the devel release


philroche:
@tianon New PR with passing checks

tianon:
Great work, thanks! :+1:

PR #9177, opened 2020-11-26
[percona] Fix UID for Percona-Server-57

EvgeniyPatlan:


EvgeniyPatlan:
@tianon I also added fixes for 5.6 tests (I have found that version wasn't updated)

das-peter:
Thanks for the quick reaction! Let's hope this can be merged before to many people run into a failing image.
That said, is there a way to show some sort of notification on the official docker image page while this is pending?

EvgeniyPatlan:
hi @yosifkit could you please take a look on this PR as this is critical fix

tianon:
Ouch, looks like we missed that this `RUN` block is also missing `set -e` (like the one we caught further down in a different version).

(Also, kind of an unfortunate day to be having it -- today is a very popular US holiday, second most commonly celebrated next to Christmas.)

tianon:
(Also, this specific image tag hasn't functionally changed since 87a73c767cf6eb29d036602a1614fbd39b974d9f back in July, so I'm not sure I understand the urgency?)

EvgeniyPatlan:
@tianon we found that in previous version we changed UID from 999 to 1001 so it can break permissions and can lead to issues with it

EvgeniyPatlan:
@tianon  I have added set -ex

PR #9188, opened 2020-11-30
Update photon image

gerrit-photon:
Update photon image

### DIFF
--- a/library/photon
+++ b/library/photon
@@ -4,19 +4,19 @@ Maintainers: Fabio Rapposelli <fabio@vmware.com> (@frapposelli),
 GitRepo: https://github.com/vmware/photon-docker-image.git
 Directory: docker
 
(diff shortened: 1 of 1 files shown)


tianon:
I think this is struggling because https://github.com/docker-library/official-images/pull/9143 had so many commits that it got a squash-and-merge instead of just a merge, and your bot appears to be using an older "master" to make PRs with.

What I would suggest is that you have your automation check out the master branch from https://github.com/docker-library/official-images, add your commits, then push *that* to https://github.com/gerrit-photon/official-images.

gerrit-photon:
Hi,

I try to do that - like https://github.com/gerrit-photon/official-images/commit/3bcddea1364a0c901950b903562e4bfb318dc61a
But it is still failed.

Thanks,
Michelle

PR #9189, opened 2020-11-30
Photon image update

gerrit-photon:
Photon image update


tianon:
Please, stop opening new PRs.  Just force push to the existing PR's branch and it will update the PR.  See also https://github.com/docker-library/official-images/pull/9188#issuecomment-735951377.

PR #9209, opened 2020-12-03
eclipse-mosquitto 2.0.0

ralight:
This is a new version of Mosquitto with breaking behaviour changes, hence the major version number change.

I've left the `latest` tag where it is right now on purpose and will move it at some point in the future. I presume there are people out there using `latest` when they really should be using a fixed tag and want to give them the chance to migrate.

### DIFF
--- a/test/tests/eclipse-mosquitto-basics/run.sh
+++ b/test/tests/eclipse-mosquitto-basics/run.sh
@@ -15,11 +15,18 @@ dir="$(dirname "$(readlink -f "$BASH_SOURCE")")"
 
 image="$1"
 
+# Create an instance of the container-under-test
+serverImage="$("$dir/../image-name.sh" librarytest/eclipse-mosquitto-basics "$image")"
+"$dir/../docker-build.sh" "$dir" "$serverImage" <<EOD
+FROM $image
+COPY dir/mosquitto.conf /mosquitto/config/
+EOD
+
 cname="eclipse-mosquitto-container-$RANDOM-$RANDOM"
 cid="$(docker run -d \
 	--name "$cname" \
-	"$image"
-)"
+	"$serverImage")"
+
 trap "docker rm -vf $cid > /dev/null" EXIT
 
 _publish() {
@@ -30,8 +37,8 @@ _publish() {
 	shift
 
 	docker run --rm \
-		--link "$cname":eclipse-mosquitto \
-		"$image" \
+		--link "$cid":eclipse-mosquitto \
+		"$serverImage" \
 		mosquitto_pub \
 			-t $topic \
 			-m ${payload} \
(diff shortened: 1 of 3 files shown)


test/tests/eclipse-mosquitto-basics/run.sh
```diff
@@ -15,11 +15,13 @@ dir="$(dirname "$(readlink -f "$BASH_SOURCE")")"
...
-	--name "$cname" \
-	"$image"
-)"
+# Create an instance of the container-under-test
+cname="$("$dir/../image-name.sh" librarytest/eclipse-mosquitto-basics "$image")"
+"$dir/../docker-build.sh" "$dir" "$cname" <<EOD
+FROM $image
+COPY dir/mosquitto.conf /mosquitto/config/
+EOD
+cid="$(docker run -d "$cname")"
```
tianon, line 24: Try something like this:

```suggestion
serverImage="$("$dir/../image-name.sh" librarytest/eclipse-mosquitto-basics "$image")"
"$dir/../docker-build.sh" "$dir" "$serverImage" <<EOD
FROM $image
COPY dir/mosquitto.conf /mosquitto/config/
EOD
cname="eclipse-mosquitto-container-$RANDOM-$RANDOM"
cid="$(docker run -d \
	--name "$cname" \
	"$serverImage"
)"
```

ralight: Thank you, I think I got there in the end. Sorry for the noise, I should've left it for the morning after the first fail.

ralight:
Is there any way around the `unable to prepare context: path "/tmp/docker-library-test-build-EPR2ucKryn" not found` type errors when trying to run these type of tests locally?

tianon:
> Is there any way around the `unable to prepare context: path "/tmp/docker-library-test-build-EPR2ucKryn" not found` type errors when trying to run these type of tests locally?

I've just pushed b3e25495, which changes this `mktemp` to be more conservative -- want to rebase and give it a try? :pray:

ralight:
That hasn't helped I'm afraid.

I did this:
```
 ...
 cp -RL "$dir" "$tmp/dir"

+ls $tmp > /tmp/rrr
 command docker build -t "$imageTag" "$tmp" > /dev/null
 ...
```
And after run.sh failed, /tmp/rrr showed that the $tmp existed and had the right files in.

This is on Ubuntu 20.04. I'm happy to do more debugging (but perhaps it would be better in a different place than this PR).

tianon:
Oh interesting - are you running Docker via Snap, by chance?  (It's super
buggy, and restrictive on what things Docker can access by design.)

tianon:
I would love to see some verification on what gets downloaded from `https://github.com/DaveGamble/cJSON/archive/v${CJSON_VERSION}.tar.gz`, but I don't think that's necessarily a blocker here.

LGTM

ralight:
Yes I am running as a snap.

I've added sha256sum checks to all of the cJSON and libwebsockets downloads.

PR #9219, opened 2020-12-04
rebar3 3.14.3

getong:
erlang/docker-erlang-otp#315


tsloughter:
Needs 22.3.4.12 rerun because it got a 500 from github when downloading rebar3. Should be a temporary failure, I can fetch it fine locally.

yosifkit:
GitHub actions only support restarting the whole workflow. :disappointed: Since `22.3.4.12` didn't change, I'll just merge.

PR #9230, opened 2020-12-06
Update rocket.chat versions 2 and 3

sampaiodiego:


yosifkit:
It seems the signing key changed for `2.4.13`?

```console
+ gpg --batch --verify rocket.chat.tgz.asc rocket.chat.tgz
gpg: Signature made Mon Dec  7 21:49:00 2020 UTC
gpg:                using RSA key E060152B30502562
gpg: Can't check signature: No public key
```

sampaiodiego:
I messed it up.. since it is an old release, I probably got the wrong key.. I'll take a look an fix it.. thx @yosifkit

sampaiodiego:
@yosifkit I need help now.. The signature is now using the right sign key, but the build keeps failing.. I can now build this locally, so I wonder if there is some cache going on.

tianon:
I tried to pull the key that `gpg` is claiming that file is signed with (`E060152B30502562`), and `gpg` is telling me it can't find it?

```console
$ gpg --keyserver ha.pool.sks-keyservers.net --recv-keys E060152B30502562
gpg: keyserver receive failed: No data
```

Has it been pushed to the keyserver network?

(Per https://github.com/docker-library/faq#openpgp--gnupg-keys-and-verification, I'd also highly recommend checking out https://keys.openpgp.org/about, which we've found to be much more reliable than the SKS network in general, both in availability and in usability.)

tianon:
Ahh, disregard; I restarted the build and now I see what you meant - the key was right, the signature file needed to be fixed. :+1:

LGTM

sampaiodiego:
> Ahh, disregard; I restarted the build and now I see what you meant - the key was right, the signature file needed to be fixed. 👍

exactly.. thanks again @tianon .. will take a look link you mentioned. 👍

PR #9243, opened 2020-12-08
Update to Flink 1.12.0

dianfu:

PR #7281, opened 2020-01-15
Convertigo official 7.7.0 release.

nicolas-albert:
Adds a AKS version with specific Kubernetes tools installed.

Still maintains the 7.6.x versions.

Happy new year and thx !

PR #9249, opened 2020-12-08
[flink] add 1.12 ; remove 1.10

rmetzger:
This might be a replacement for https://github.com/docker-library/official-images/pull/9243

### DIFF
--- a/library/flink
+++ b/library/flink
@@ -1,4 +1,4 @@
-# this file is generated via https://github.com/apache/flink-docker/blob/11221b1fc63ccacff701ad5b86be7f9063aacee7/generate-stackbrew-library.sh
+# this file is generated via https://github.com/apache/flink-docker/blob/f47f4a2e257279c4c28667e3ceff12bc5eb70afb/generate-stackbrew-library.sh
 
 Maintainers: Patrick Lucas <me@patricklucas.com> (@patricklucas),
              Ismaël Mejía <iemejia@gmail.com> (@iemejia)
@@ -24,12 +24,22 @@ Architectures: amd64
 GitCommit: 7035f03679b11352f2fdecd9f6a9bb0ec8bc2022
 Directory: ./1.11/scala_2.12-java8-debian
 
-Tags: 1.10.2-scala_2.12, 1.10-scala_2.12, 1.10.2, 1.10
+Tags: 1.12.0-scala_2.11-java11, 1.12-scala_2.11-java11, scala_2.11-java11
 Architectures: amd64
-GitCommit: 58a29fca7c6ff05999fad4371638d16335f7e93e
-Directory: ./1.10/scala_2.12-debian
+GitCommit: 4d7cb3068d670bef274975c9412d5e40317fba3f
+Directory: ./1.12/scala_2.11-java11-debian
 
-Tags: 1.10.2-scala_2.11, 1.10-scala_2.11
+Tags: 1.12.0-scala_2.11-java8, 1.12-scala_2.11-java8, scala_2.11-java8, 1.12.0-scala_2.11, 1.12-scala_2.11, scala_2.11
 Architectures: amd64
-GitCommit: 58a29fca7c6ff05999fad4371638d16335f7e93e
-Directory: ./1.10/scala_2.11-debian
+GitCommit: 4d7cb3068d670bef274975c9412d5e40317fba3f
+Directory: ./1.12/scala_2.11-java8-debian
+
+Tags: 1.12.0-scala_2.12-java11, 1.12-scala_2.12-java11, scala_2.12-java11, 1.12.0-java11, 1.12-java11, java11
+Architectures: amd64
+GitCommit: 4d7cb3068d670bef274975c9412d5e40317fba3f
+Directory: ./1.12/scala_2.12-java11-debian
+
+Tags: 1.12.0-scala_2.12-java8, 1.12-scala_2.12-java8, scala_2.12-java8, 1.12.0-scala_2.12, 1.12-scala_2.12, scala_2.12, 1.12.0-java8, 1.12-java8, java8, 1.12.0, 1.12, latest
+Architectures: amd64
+GitCommit: 4d7cb3068d670bef274975c9412d5e40317fba3f
+Directory: ./1.12/scala_2.12-java8-debian


dianfu:
cc @yosifkit @tianon

yosifkit:
`args` by reference seems a bit weird, but is harmless. (since it is a global value, it could be accessed by the function if declared earlier.) No change necessary.

---

This inserted `bash -c` may be a non-intuitive interface. I'm not entirely certain it needs to change, just ensuring that it makes sense for users.

```diff
+    exec $(drop_privs_cmd) bash -c "${args[@]}"
```

```console
$ # current
$ docker run -it flink native-k8s 'some-command --with --its arguments && other stuff'
$ # missing quotes would mean the arguments to "some-command" would be interpreted as arguments to "bash"
$ docker run -it flink native-k8s some-command --with --its arguments

$ # vs if you drop the bash -c
$ docker run -it flink native-k8s some-command --with --its arguments
$ docker run -it flink native-k8s bash -c 'some-command --with --its arguments && other stuff'
```

rmetzger:
Thank you for your review! I agree that the `bash -c` invocation is potentially confusing. However, as far as I understand, the `native-k8s` command is not meant to by used directly by humans. It's used by Flink's Kubernetes integration, where a java process generates the K8s deployments for starting Flink. This Java process also generates the arguments for launching the docker containers. @wangyang0918 correct me if I'm wrong here.

wangyang0918:
@tianon @rmetzger Yes, the `native-k8s` is not meant to be used directly. The Flink client will generate the full command and then it is executed in the `docker-entrypoint.sh`. The command and arguments may be a little complex and looks like followings.

```
  Containers:
  flink-job-manager:
    Container ID:  docker://b02897603fdf7e74d94ffa928727ea594e6247e6cdf9adf9a393c957350fcda2
    Image:         registry.cn-beijing.aliyuncs.com/streamcompute/flink:flink-1.12.0
    Image ID:      docker-pullable://registry.cn-beijing.aliyuncs.com/streamcompute/flink@sha256:7e8985b0c1250f62c5b3e5fb9d13c5ef287241fc24af2a84902e37ebafd018ea
    Ports:         8081/TCP, 6123/TCP, 6124/TCP
    Host Ports:    0/TCP, 0/TCP, 0/TCP
    Command:
      /docker-entrypoint.sh
    Args:
      native-k8s
      $JAVA_HOME/bin/java -classpath $FLINK_CLASSPATH -Xmx654311424 -Xms654311424 -XX:MaxMetaspaceSize=268435456 -Dlog.file=/opt/flink/log/jobmanager.log -Dlogback.configurationFile=file:/opt/flink/conf/logback-console.xml -Dlog4j.configuration=file:/opt/flink/conf/log4j-console.properties -Dlog4j.configurationFile=file:/opt/flink/conf/log4j-console.properties org.apache.flink.kubernetes.entrypoint.KubernetesApplicationClusterEntrypoint -D jobmanager.memory.off-heap.size=134217728b -D jobmanager.memory.jvm-overhead.min=201326592b -D jobmanager.memory.jvm-metaspace.size=268435456b -D jobmanager.memory.heap.size=654311424b -D jobmanager.memory.jvm-overhead.max=201326592b
```

yosifkit:
> Extra [script functions] that are only for a specific vendor platform (Amazon, Azure, Google Cloud, OpenShift, OpenStack, etc) breaks the whole "run anywhere" Docker experience. Images in Official-Images should be platform independent. It is up to image maintainers whether or not to adjust their images to better support the specific needs of a platform in a **general** way (e.g. https://github.com/docker-library/postgres/issues/359 + https://github.com/docker-library/postgres/pull/448).
>
>\ -https://github.com/docker-library/official-images/pull/7281#issuecomment-574909447

Seems a little strange for the Flink client to generate a command reliant on interpreted env vars, shouldn't it just generate some args for `java` and let the entrypoint just run `java` with the correct classpath?  This would be especially relevant if any arguments ever have a path that includes spaces.  And what about non-kubernetes users, would they not also want to use these calculated memory values and other custom flags?  In other words, why have a just-for-kubernetes part of the entrypoint?

wangyang0918:
@yosifkit Thanks for your comments and suggestions.

We could move the `$JAVA_HOME/bin/java -classpath $FLINK_CLASSPATH` into the `docker-entrypoint.sh` and only generate the jvm args in Flink. Then `exec $(drop_privs_cmd) bash -c "${args[@]}"` could be `exec $(drop_privs_cmd) $JAVA_HOME/bin/java -classpath $FLINK_CLASSPATH "${args[@]}"`. And the `native-k8s` command should be renamed to `java-run-with-classpath`. This is a possible solution to make things clearer.

However, I prefer another one. Currently, the `native-k8s` only works for native K8s integration for Flink. It is also why we introduce this command. Maybe we could completely remove it and let the flink generate a `launch_container.sh` and mount into the JobManager and TaskManager pod. The `launch_container.sh` will be executed in the `docker-entrypoint.sh` directly.

BTW, could we do this improvement in following release version(Flink 1.12.1)? Both the two solutions require us to update the Flink source codes, but we have released the Flink 1.12.0 and only one step left(publishing the image to dockerhub).

rogaha:
thanks @wangyang0918 for submitting the `PR`. @tianon thanks for keeping up with `official` images support.

wangyang0918:
Hi @rogaha @tianon @yosifkit, do you think the new `docker-entrypoint.sh` in https://github.com/apache/flink-docker/pull/49 make sense? We will use `run` command in Flink and remove the deprecated `native-k8s` in next Flink major release(1.13).

xintongsong:
Thanks all for the efforts so far.
Since @rmetzger is out-of-office, I opened #9345 as a replacement of this one.

baobeidaodao:
When will it be released ?

tianon:
(Commenting here instead of on #9345 because this is where the previous discussion is and I'd really appreciate if we didn't splinter it.)

I've given this more thought, and figured out what's got me so confused/conflicted.

I see that there exist a lot of (useful looking) scripts in https://github.com/apache/flink/tree/2eaee8344bdcab961799127c7e72d3f4abe75726/flink-dist/src/main/flink-bin that I can only assume exist in this directory because they are part of Flink releases, and are how users are expected to use Flink directly (outside containers), correct?

My main concern (and confusion) here is that we have a separate custom script that's specific to the Dockerization, and it keeps growing, and it feels like it's really duplicating the efforts that exist there.  If this had been implemented as part of those scripts and been part of the Flink release artifacts, we wouldn't have even noticed their addition, their usage would still fall under the "consistent interface" guidelines we are looking for, and we wouldn't still be sitting here with a PR that both adds *and* deprecates a feature in the same new release, which feels really off.

zentol:
@tianon Your assessment is somewhat correct.

All deployments of Flink makes use of the the flink-bin scripts in some form; for YARN or non-containerized setups users are [directly using the contained scripts](https://ci.apache.org/projects/flink/flink-docs-release-1.12/deployment/resource-providers/standalone/#starting-a-standalone-cluster-session-mode) to create a cluster.

For [docker deployments](https://ci.apache.org/projects/flink/flink-docs-release-1.12/deployment/resource-providers/standalone/docker.html#starting-a-session-cluster-on-docker), users work against the interface provided by the docker entrypoint, which ultimately just wraps the flink-bin scripts which just add some additional docker-specific logic.

Finally, the native kubernetes integration has users interact with the flink-bin scripts to issue the [creation of a cluster](https://ci.apache.org/projects/flink/flink-docs-release-1.12/deployment/resource-providers/native_kubernetes.html#session-mode), where Flink requests containers and generates a command to be executed within the container through the docker-entrypoint.sh.

In other words the Flink docker image is used in 2 ways; directly by users and in some essentially-internal fashion where we're just passing a command through.

Re deprecating a new feature: The annoying reality is that the 1.12.0 code expects the (now deprecated) command to exists, and we would like 1.12.0 clients to continue working with 1.12.1 images. (The used images are not hard-wired to a specific release)

If I understand you correctly, then it would be fine for us to move everything in the docker-entrypoint.sh into a new scripts within distribution, and just call that scripts from docker-entrypoint.sh with all passed arguments. We had the same idea on our side, and would probably be fine with doing that starting from 1.13.

tianon:
> Re deprecating a new feature: The annoying reality is that the 1.12.0 code expects the (now deprecated) command to exists, and we would like 1.12.0 clients to continue working with 1.12.1 images. (The used images are not hard-wired to a specific release)
> 
> If I understand you correctly, then it would be fine for us to move everything in the docker-entrypoint.sh into a new scripts within distribution, and just call that scripts from docker-entrypoint.sh with all passed arguments. We had the same idea on our side, and would probably be fine with doing that starting from 1.13.

Yes, I'd love to see this (anything that decreases the delta between a "standard" Flink install and a "Docker" Flink install is a big win IMO) -- the part about this particular instance that feels so off is that this feature appears to have been worked on for months with the intention of adding this additional code to the bits of the image we review (which is one of the core purposes of this program), but we weren't looped into the conversation until it was already too late to discuss it, which is why we're having this conversation. :grimacing: :smile:

wangyang0918:
@tianon 
We could make the `docker-entrypoint.sh` in flink-docker project as thin as possible and move the most functionalities into flink distribution. Of cause, the native K8s mode will have a dedicated entrypoint `kubernetes-entrypoint.sh` in flink distribution, which will export some environment variables for internal use. Then we will not need to do it in the common `docker-entrypoint.sh`. But I think we could only do this for next major release(1.13).

How could we move the Flink 1.12.* official images publishing forward? Is marking the `native-k8s` command deprecated enough? We have already done this[1].


[1]. https://github.com/apache/flink-docker/blob/master/1.12/scala_2.11-java8-debian/docker-entrypoint.sh#L22

zentol:
> Of cause, the native K8s mode will have a dedicated entrypoint kubernetes-entrypoint.sh in flink distribution.

This is TBD. There may be other ways to solve this, but let's not have this discussion here.

library/flink
```diff
@@ -4,32 +4,42 @@ Maintainers: Patrick Lucas <me@patricklucas.com> (@patricklucas),
...
-Tags: 1.10.2-scala_2.12, 1.10-scala_2.12, 1.10.2, 1.10
+Tags: 1.12.0-scala_2.12-java11, 1.12-scala_2.12-java11, scala_2.12-java11, 1.12.0-java11, 1.12-java11, java11
 Architectures: amd64
-GitCommit: 58a29fca7c6ff05999fad4371638d16335f7e93e
-Directory: ./1.10/scala_2.12-debian
+GitCommit: 041ac78d9264cc596d9007eb524d24c2b65ca830
+Directory: ./1.12/scala_2.12-java11-debian
 
-Tags: 1.10.2-scala_2.11, 1.10-scala_2.11
+Tags: 1.12.0-scala_2.12-java8, 1.12-scala_2.12-java8, scala_2.12-java8, 1.12.0-scala_2.12, 1.12-scala_2.12, scala_2.12, 1.12.0-java8, 1.12-java8, java8, 1.12.0, 1.12, latest
```
heroldus, line 42: What is the reason to allow java8 images?

xintongsong: Java 8 is the lowest version required and the maven target version for Flink, and is AFAIK used by the majority of Flink developers and users.

tianon:
(Reviewing again given this was updated and I'm assuming it's ready? :sweat_smile:)

Per https://github.com/apache/flink/pull/14630#issuecomment-763180710, this jemalloc usage looks much better; thank you.  Looking forward to improved/extracted k8s integration for 1.13 (https://issues.apache.org/jira/browse/FLINK-21128). :+1:

(Happy to merge now following confirmation that this is indeed "ready" from your perspective)

zentol:
@tianon Yes, from our side it is ready to go 👍

xintongsong:
@tianon @yosifkit @rogaha 
Thank you all for helping us work this out. :)

PR #9257, opened 2020-12-10
Update Debian, especially for 10.7, CVE-2020-1971, CVE-2020-27350

tianon:
Both CVEs have pretty minor impact, but we're due for an update anyhow (thanks to 10.7) so it made sense to wait for them to be included.

### DIFF
--- a/library/debian
+++ b/library/debian
@@ -7,31 +7,31 @@ GitRepo: https://github.com/debuerreotype/docker-debian-artifacts.git
 GitCommit: 90a83b6b730031399d45a49809364b07df0d0087
 # https://github.com/debuerreotype/docker-debian-artifacts/tree/dist-amd64
 amd64-GitFetch: refs/heads/dist-amd64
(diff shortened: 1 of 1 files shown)


tianon:
I've verified that every arch/suite combo has one of the following `apt` versions:

```
apt	1.0.9.8.6
apt	1.4.11
apt	1.8.2.2
apt	2.1.12
```

1.4.11 is stretch's fixed version, 1.8.2.2 is buster's fixed version, 1.0.9.8.6 is jessie's version (EOL), and 2.1.12 is the testing/unstable version (unfixed)

This morning, unstable got 2.1.13 (with the fix), but given the low severity I'm thinking it's probably not worth rebuilding for right this minute.

tianon:
Talked to APT maintainers: "you don't care about the security updates for apt and python-apt unless you run a service exposing the archive code to untrusted content ... And even then, I can still DoS you anyway" :laughing: :heart:

(Which was my understanding, but it's nice to have it confirmed - so updating unstable/sid isn't all that urgent.)

PR #9260, opened 2020-12-10
Bump for Mono 6.12 stable

directhex:


directhex:
Well that's interesting. Wonder why it hasn't repro'd locally

directhex:
Fix has finished publishing

PR #9266, opened 2020-12-10
Update JRuby release to 9.2.14.0

reidmorrison:


reidmorrison:
Naughty is complaining about:
` - jruby:9.2-jdk14 (FROM openjdk:14-jdk-buster) -- completely unsupported base!`

It appears the tag still exists: https://hub.docker.com/layers/openjdk/library/openjdk/14-jdk-buster/images/sha256-12fa83956c7ba2423622e293eced64b475e6e3bb151c06c54812d5d8c39f4bc0?context=explore

tianon:
The tag still exists yes, but given that JDK 14 is EOL (https://jdk.java.net/14/), it will not be updated anymore.

cpuguy83:
Didn't realize this was open, I made #9318 which is all green. Just removed the jdk14 tags.

PR #9268, opened 2020-12-11
consul: 1.9.1, 1.8.7 and 1.7.11

mikemorris:
Removes 1.8.7-beta1 and 1.6 series


yosifkit:
LGTM.

PR #9269, opened 2020-12-11
Update ghost

tianon:
Changes:

- https://github.com/docker-library/ghost/commit/25c66b1: Update to 3.40.1, ghost-cli 1.15.3


yosifkit:
LGTM.

PR #9270, opened 2020-12-11
Update haproxy

tianon:
Changes:

- https://github.com/docker-library/haproxy/commit/71238cd: Update to 2.4-dev3


yosifkit:
LGTM.

PR #9271, opened 2020-12-12
Update docker

tianon:
Changes:

- https://github.com/docker-library/docker/commit/811d896: Merge pull request https://github.com/docker-library/docker/pull/280 from infosiftr/ip6tables
- https://github.com/docker-library/docker/commit/094faa8: Add ip6tables in 20.10+


yosifkit:
LGTM.

With both Debian (https://github.com/docker-library/official-images/pull/9257) and Alpine (https://github.com/docker-library/official-images/pull/9262) merged yesterday, the [build queue](https://doi-janky.infosiftr.net/) is still quite full so it may be a bit before this new version is built and available on Docker Hub.

PR #9318, opened 2020-12-18
Bump jruby to 9.2.14.0

cpuguy83:
Closes #9266


yosifkit:
https://github.com/docker-library/official-images/pull/9266#issuecomment-742861994

cpuguy83:
Updated to remove that tags since it's no longer supported.

PR #9319, opened 2020-12-18
Transfer jruby library maintenance to JRuby team

headius:
The JRuby team will be taking over maintenance of the official `jruby` image as per https://github.com/jruby/jruby/issues/5598

The admin@jruby.org email goes to the two co-leads of the project responsible for releases and administration, me (@headius) and Tom Enebo (@enebo).

This does not reflect the pending update of the images to JRuby 9.2.14.0 by the previous maintainer @cpuguy83: https://github.com/docker-library/official-images/pull/9318

### DIFF
--- a/library/jruby
+++ b/library/jruby
@@ -1,5 +1,7 @@
-Maintainers: Brian Goff <cpuguy83@gmail.com> (@cpuguy83)
-GitRepo: https://github.com/cpuguy83/docker-jruby.git
+Maintainers: JRuby Admin <admin@jruby.org> (@jruby),
+             Charles Oliver Nutter <headius@headius.com> (@headius),
+             Thomas E Enebo <tom.enebo@gmail.com> (@enebo)
+GitRepo: https://github.com/jruby/docker-jruby.git
 GitFetch: refs/heads/master
 GitCommit: 510284ff5ffbb257937d755a9d351e9df50ba31b
 


headius:
Failure is due to the EOL of JDK14, also fixed in the update PR #9318.

cpuguy83 (approved):
LGTM

headius:
I rebased this on master, post #9318.

library/jruby
```diff
@@ -1,5 +1,5 @@
-Maintainers: Brian Goff <cpuguy83@gmail.com> (@cpuguy83)
-GitRepo: https://github.com/cpuguy83/docker-jruby.git
+Maintainers: JRuby Admin <admin@jruby.org> (@jruby)
```
tianon, line 1: This is used for when we need to find the "humans" who are responsible for this image (to inform you of major changes that might cause issues, etc), and "@jruby" isn't something we can usefully mention on an issue/PR, so I'd appreciate if we could instead or also list both @headius and @enebo (if email is the hesitation, you can skip that -- the parser will accept `Name (@github)` just as well as `Name <email> (@github)`).

library/jruby
```diff
@@ -1,5 +1,5 @@
...
+GitRepo: https://github.com/jruby/docker-jruby.git
```
tianon, line 2: This being a new repo instead of a moved repo means we lose issues/PRs, which isn't great -- is there a particular reason moving @cpuguy83's existing repo didn't work out?

headius: Only that I did not want to move it before this transfer is official... chicken and egg. If it is fine to move it directly I am sure @cpuguy83 can help us make that change. The JRuby repo can be dumped at any time.

tianon: It's looking pretty official to me :eyes: :wink: :heart:

library/jruby
headius: Ok, for whatever reason I thought the @jruby referred to Twitter, but if it refers to Github that makes sense. There is no JRuby user.

I can list also @headius and @enebo in this, but we would prefer the emails go to admin@jruby.org for general support. Should we just remove the @github part from the `JRuby Admin` entry?

headius: I updated the commit to include @headius and @enebo contact information.

It seems that @jruby can stay based on the contact info for Amazon Corretto, which also includes a "team" entry pointing at the @corretto organization, yes?

https://github.com/docker-library/official-images/blob/master/library/amazoncorretto

library/jruby
headius: Ok, I will work with @cpuguy83 to transfer the repo directly to @jruby organization. Nothing will visibly change related to this PR.

library/jruby
headius: D'oh... fixed.

headius:
The `docker-jruby` repository has been **transferred** to the `jruby` organization. I believe we are ready to merge this.

library/jruby
headius: Transfer complete. https://github.com/jruby/jruby/issues/5598#issuecomment-749180697

tianon: Can confirm!  Clicking https://github.com/cpuguy83/docker-jruby/blob/510284ff5ffbb257937d755a9d351e9df50ba31b/9000/jre8/Dockerfile (from https://hub.docker.com/_/jruby) takes me to https://github.com/jruby/docker-jruby/blob/510284ff5ffbb257937d755a9d351e9df50ba31b/9000/jre8/Dockerfile properly! :tada: :heart:

tianon:
Awesome, welcome to our newest maintainers!

(And thank you again, @cpuguy83 for your years of loyal service :heart:)

PR #9322, opened 2020-12-18
Telegraf v1.17.0

ssoroka:


library/telegraf
```diff
@@ -17,9 +10,16 @@ Directory: telegraf/1.15
...
 Architectures: amd64, arm32v7, arm64v8
 Directory: telegraf/1.16
 
-Tags: 1.16-alpine, 1.16.3-alpine, alpine
+Tags: 1.16-alpine, 1.16.3-alpine
 Directory: telegraf/1.16/alpine
+
+Tags: 1.17, 1.17.0, latest
+Architectures: amd64, arm32v7, arm64v8
+Directory: telegraf/1.17
```
tianon, line 22: Looking at https://github.com/influxdata/influxdata-docker/tree/22528387700ac8e119269db2adf5ad26a3e3efee/telegraf, either this `Directory:` or the `GitCommit:` above doesn't seem to be quite right? :grimacing:

ssoroka: fixed

tianon: Was that a commit you meant to push here? :innocent: :heart:

ssoroka: Updated, thanks for assist :)

PR #9381, opened 2021-01-06
Add arm64v8 support to clojure:latest

cap10morgan:
This should support arm64v8 now that it's based on openjdk-11 upstream. I think only the openjdk-8 images don't at this point.


cap10morgan:
I noticed some other tags that should support arm64v8 weren't being built for it, so I added those too.

PR #9382, opened 2021-01-07
[security] Update php

tianon:
Changes:

- https://github.com/docker-library/php/commit/4d752fa: Update 7.4 to 7.4.14
- https://github.com/docker-library/php/commit/a310c86: Update 7.3 to 7.3.26


tianon:
(There was an 8.0.1 just now, but it's not labelled as a security fix, so I'm planning to merge this without it if the tests can finish in a reasonable timeframe.)

PR #9417, opened 2021-01-13
Release XWiki 12.6.7

surli:


library/xwiki
```diff
@@ -25,15 +25,31 @@ Directory: 12/postgres-tomcat
...
-Tags: 12.6-postgres-tomcat, 12.6.6-postgres-tomcat
+Tags: 12.6-postgres-tomcat, 12.6.7-postgres-tomcat
 Architectures: amd64, arm64v8
 Directory: 12/postgres-tomcat
-GitCommit: 3f09b1d7e286b4c2ff611742c417a118111f5e84
+GitCommit: b0cf482e10e8b0507122ce654f5a7d660063620f
+GitFetch: refs/heads/stable-12.6.x
+
+# The following tags are provided to fix a mistake done in recent release
+# and to restore back the right version for XWiki 12.6.6.
```
yosifkit, line 42: :heart: Thanks for this note! I was confused that our diff was showing that the `12.6.6` images were changing from `12.10.2` to `12.6.6` and thought our diff script was being silly.

library/xwiki
```diff
@@ -25,15 +25,31 @@ Directory: 12/postgres-tomcat
 GitCommit: 3f09b1d7e286b4c2ff611742c417a118111f5e84
```
yosifkit, line 25: Doh! it looks like `12.10.1` tagged images actually contain `12.10.2` (https://github.com/xwiki/xwiki-docker/commit/3f09b1d7e286b4c2ff611742c417a118111f5e84)

surli: Outch you're right apparently https://github.com/docker-library/official-images/pull/9353 was really wrong we forgot to update the tags for 12.10... I'll provide a new PR for fixing it.

library/xwiki
surli: No your script was clearly ok :)

surli: Just a question: I guess the change has already been taken into account to fix back the revision for tag 12.6.6, meaning I can delete those lines in my next PR, right?

PR #9422, opened 2021-01-14
Fix tags for XWiki 12.10.x

surli:


yosifkit:
(Double checked that the `12.6.6*` tags were all updated on Docker Hub yesterday)

PR #9425, opened 2021-01-14
Update Postfixadmin

J0WI:


library/postfixadmin
```diff
@@ -2,17 +2,17 @@
 Maintainers: David Goodwin <david@codepoets.co.uk> (@DavidGoodwin)
 GitRepo: https://github.com/postfixadmin/docker.git
 
-Tags: 3.2.4-apache, 3.2-apache, 3-apache, apache, 3.2.4, 3.2, 3, latest
+Tags: 3.3.3-apache, 3.3-apache, 3-apache, 3.3.3, 3.3, 3
```
yosifkit, line 5: Some tags like `latest` and `fpm-alpine` are being removed. Is that on purpose?

J0WI: Nope, thanks for the hint!

PR #9426, opened 2021-01-14
Update PhpMyAdmin

J0WI:


tianon:
```diff
+if [ ! -z "${MAX_EXECUTION_TIME}" ]; then
+    echo "Adjusting the maximum execution time."
+    sed -i "s/^max_execution_time = .*$/max_execution_time = ${MAX_EXECUTION_TIME}/g" $PHP_INI_DIR/conf.d/phpmyadmin-misc.ini
+fi
```

Given [that `php.ini` supports environment variables](https://www.php.net/manual/en/configuration.file.php#example-36), wouldn't it be better to embed this directly in the generated `php.ini` snippet (with a default set via environment variable in the image) and let it just DTRT in all instances?

tianon:
(Also, shouldn't it be `PMA_MAX_EXECUTION_TIME` for consistency?)

williamdes:
Hi :wave: 

I would just post here what I said upstream: we will not rename variables because our users would hate us.
And second point is that `MAX_EXECUTION_TIME` sets the phpMyAdmin and the PHP value, like the 2 other variables in this script.

Hope it clarifies your question ;)

yosifkit:
This one should be `if (...MAX)`, not `DB` ?
```php
if (isset($_ENV['PMA_QUERYHISTORYDB'])) {
    $cfg['QueryHistoryMax'] = intval($_ENV['PMA_QUERYHISTORYMAX']);
}
```

As for the the `MEMORY_LIMIT`, `UPLOAD_LIMIT`, and `MAX_EXECUTION_TIME` variables, the names are really generic, and could potentially have more overlap with other software (limited impact since this is a container, but something to consider). They haven't been in an image pushed to the official-images so renaming shouldn't affect any users.

williamdes:
> They haven't been in an image pushed to the official-images so renaming shouldn't affect any users.

They are on pushed images since a long time and our users use them :/

Checked and UPLOAD_LIMIT is on the official image and for MAX_EXECUTION_TIME it was added last year (11/2020) and our users started using it.

`MEMORY_LIMIT` could be removed because it was just added but now that it is added it would like not to.

Thank you for `This one should be if (...MAX), not DB ?` :+1: 
I fixed it and added back to the documentation the `HIDE_PHP_VERSION` ENV

yosifkit:
No worries on the generic variables, the diff made them look like they were new, my bad :disappointed:.

I think we can merge once the `PMA_QUERYHISTORYMAX` if statement fix shows up here. @J0WI, can you update the PR to reference the new commit?

J0WI:
>  I fixed it and added back to the documentation the HIDE_PHP_VERSION ENV

@williamdes I removed that (together with the [related code](https://github.com/phpmyadmin/docker/pull/320/files#diff-79738685a656fe6b25061bb14181442210b599f746faeaba408a2401de45038aL16-L20)) on purpose because there is no common use case to set this explicitly. Removing the ENV should not "break" anything.

williamdes:
> > I fixed it and added back to the documentation the HIDE_PHP_VERSION ENV
> 
> @williamdes I removed that (together with the [related code](https://github.com/phpmyadmin/docker/pull/320/files#diff-79738685a656fe6b25061bb14181442210b599f746faeaba408a2401de45038aL16-L20)) on purpose because there is no common use case to set this explicitly. Removing the ENV should not "break" anything.

The ENV still works and should be documented (IMO). Our users requested to have this security option to hide their PHP versions. I added a test for that :)

J0WI:
It's still hidden (by default in the image), I just removed the env to unhide it.

williamdes:
> It's still hidden (by default in the image), I just removed the env to unhide it.

Hmm, let me know if you need some action from my side :)

J0WI:
If you agree with the removal of the env to *un*hide it the documentation change should be reverted. Otherwise the [hardcoded config](https://github.com/phpmyadmin/docker/blob/bd7de176315049e4b226b377a361f4e05b2e4cd9/Dockerfile-alpine.template#L61) should be changed to a env. The current readme documents a non-functional env. I should have separated those changes into multiple pull requests.

williamdes:
> If you agree with the removal of the env to _un_hide it the documentation change should be reverted. Otherwise the [hardcoded config](https://github.com/phpmyadmin/docker/blob/bd7de176315049e4b226b377a361f4e05b2e4cd9/Dockerfile-alpine.template#L61) should be changed to a env. The current readme documents a non-functional env. I should have separated those changes into multiple pull requests.

I would say that if the ENV exists it can not be removed, I would like if you could make a fix to make the ENV work again.

> I should have separated those changes into multiple pull requests.

I would say that the test suite was not very good and still could be better even now, that would have pointed out that the changes where not working

williamdes:
Hi @J0WI 
Could you update this PR please ?

yosifkit (approved):

PR #9429, opened 2021-01-14
[oraclelinux] Republish all Oracle Linux images without LABEL directive

Djelibeybi:
Because LABELs are automatically inherited and cannot be removed
by downstream images, we are removing them from our base images
instead. This is to avoid a downstream image misidentifying itself
as an Oracle produced artefact.

### DIFF
--- a/library/oraclelinux
+++ b/library/oraclelinux
@@ -4,10 +4,10 @@ GitCommit: 01a15ec99c7470a3391c691509db1759b41eaf66
 GitFetch: refs/heads/master
 # https://github.com/oracle/container-images/tree/dist-amd64
 amd64-GitFetch: refs/heads/dist-amd64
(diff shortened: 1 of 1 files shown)


tianon:
For (reminder of) context on why I love this PR so much: https://github.com/docker-library/official-images/issues/3540#issuecomment-335229052

> Honestly, I find labels on base images to be irritating, especially since they're all inherited by any image `FROM` the base ...

:smile: :heart:

Djelibeybi:
> For (reminder of) context on why I love this PR so much: [#3540 (comment)](https://github.com/docker-library/official-images/issues/3540#issuecomment-335229052)

Yeah, the ramifications of that weren't obvious to me back then. Hindsight and all.

PR #9456, opened 2021-01-19
Update Amazon Corretto (8.282.08.1 and 11.0.10.9.1)

jguo11:
Update Amazon Corretto 8 to 8.282.08.1 and Amazon Corretto 11 to 11.0.10.9.1.


tianon:
Multi-stage build here is kind of strange -- if "Amazon Corretto 11 'JRE' on Alpine" is something officially supported, shouldn't it be officially built and released as part of the upstream artifacts?

(Also, why only for Alpine?)

jguo11:
@tianon Thanks for the catch. The Amazon Corretto 11 JRE on Alpine was a mistake generated by our script and slipped away from my check. I've removed it and updated the PR.

PR #9464, opened 2021-01-21
Add version 1.9.2

freddygv:


yosifkit:
Image label names should adhere to the image spec: https://github.com/opencontainers/image-spec/blob/v1.0.1/annotations.md.

tianon:
(See also https://github.com/docker-library/official-images/issues/3540#issuecomment-530925319)

PR #9474, opened 2021-01-22
Fix that 6.8 still had some 6.7 tags (re-release 6.8)

keeganwitt:
Merge after releasing changes from #9473.


keeganwitt:
Gradle 6.8.1 is out now. The 6.8.0 tags still exist, so we can merge both these commits at once without a build in between.

PR #9488, opened 2021-01-25
Abcl

daewok:
Add an official image for Armed Bear Common Lisp (ABCL).

### DIFF
--- a/library/abcl
+++ b/library/abcl
@@ -1,123 +1,159 @@
-# this file is generated via https://github.com/cl-docker-images/abcl/blob/e6c45d7980eae07ace9a7a77476516507bfd8b56/generate-stackbrew-library.sh
-Maintainers: Eric Timmons <nasafreak@gmail.com> (@etimmons)
+# this file is generated via https://github.com/cl-docker-images/abcl/blob/9f5e8dba91b92e7751417a563761c2192536e1d2/generate-stackbrew-library.sh
+Maintainers: Eric Timmons <nasafreak@gmail.com> (@daewok)
 GitRepo: https://github.com/cl-docker-images/abcl.git
 
 Tags: 1.8.0-jdk15-buster, 1-jdk15-buster, 1.8-jdk15-buster, jdk15-buster, 1.8.0-jdk-buster, 1-jdk-buster, 1.8-jdk-buster, jdk-buster, 1.8.0-buster, 1-buster, 1.8-buster, buster
 SharedTags: 1.8.0-jdk15, 1-jdk15, 1.8-jdk15, jdk15, 1.8.0-jdk, 1-jdk, 1.8-jdk, jdk, 1.8.0, 1, 1.8, latest
 Architectures: amd64, arm64v8
-GitCommit: 90356f1334c6f714845f8df80823c8cdee9ff8cc
-Directory: 1.8.0/buster/jdk-15
+GitCommit: f3e8a2a6da84fa0a372e46bb7e8d5a6b696bec24
+Directory: 1.8.0/buster/jdk-15/
+
+Tags: 1.8.0-jdk15-slim-buster, 1-jdk15-slim-buster, 1.8-jdk15-slim-buster, jdk15-slim-buster, 1.8.0-jdk-slim-buster, 1-jdk-slim-buster, 1.8-jdk-slim-buster, jdk-slim-buster, 1.8.0-slim-buster, 1-slim-buster, 1.8-slim-buster, slim-buster
+SharedTags: 1.8.0-jdk15-slim, 1-jdk15-slim, 1.8-jdk15-slim, jdk15-slim, 1.8.0-jdk-slim, 1-jdk-slim, 1.8-jdk-slim, jdk-slim, 1.8.0-slim, 1-slim, 1.8-slim, slim
+Architectures: amd64, arm64v8
+GitCommit: f3e8a2a6da84fa0a372e46bb7e8d5a6b696bec24
+Directory: 1.8.0/buster/jdk-15/slim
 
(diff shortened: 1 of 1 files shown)


daewok:
@easye is the ABCL maintainer and has blessed this image.

Documentation PR at https://github.com/docker-library/docs/pull/1864

tianon:
Thanks for the proposal! :stars:

I wasn't familiar with ABCL, but I'm impressed that it's been around and active since 2003! :smile:

However, I'm having a hard time verifying the popularity of ABCL since the source/history of the project is a bit spread out (with what appears to be several "official" mirrors), so the usual metrics we look for to help give us clues don't appear to be very strong (GitHub Stars, Docker Hub Stars/Pulls, Twitter followers, etc).  Is there an additional source of validation we could look to for the (potential) size of the interested community for this image?  It seems like a niche userbase. :sweat_smile:

(I'm also honestly somewhat concerned looking at https://github.com/cl-docker-images and that there are several _other_ "Common Lisp" variants there. :thinking:)

For context, the reviewers on this program are a small team, so we have to try to ensure our (limited) time is spent in such a way to benefit as many users as possible.

Perhaps this would make sense to publish to Docker Hub under a general "Common Lisp" organization to get interested users testing/validating it, and to help gauge the relative popularity such a thing might receive if it were to become a top-level image?

easye:
> However, I'm having a hard time verifying the popularity of ABCL since the source/history of the project is a bit spread out (with what appears to be several "official" mirrors), so the usual metrics we look for to help give us clues don't appear to be very strong (GitHub Stars, Docker Hub Stars/Pulls, Twitter followers, etc). Is there an additional source of validation we could look to for the (potential) size of the interested community for this image? It seems like a niche userbase. 😅

We deliberately maintain the synchronicity of ABCL across svn bridged to git to maintain the reach-ability or decades-old URI references contained in the [mailing lists](https://mailman.common-lisp.net/pipermail/armedbear-devel/), the [Trac issue wiki](https://abcl.org/trac/report/1), and as embedded in discussion platforms available via the World Wide Web.   Due to tooling deficiencies that I have never had the time to properly overcome, the git mirrors are missing history prior to 2009 which the svn repository alone retains.  

> (I'm also honestly somewhat concerned looking at https://github.com/cl-docker-images and that there are several _other_ "Common Lisp" variants there. 🤔)

One of the strengths of Common Lisp *is* the multiple implementations, as they each have different strengths in various deployment scenarios.  ECL is based around a C transpiler, so is especially interesting for use embedded in a shared object, SBCL has an extremely optimized compiler,  CCL has a long heritage of debugging facilities, and ABCL's hosting on the Java Virtual Machine (JVM)allows easy interop with the Java ecosystem.  
 
> For context, the reviewers on this program are a small team, so we have to try to ensure our (limited) time is spent in such a way to benefit as many users as possible.

ABCL is perhaps a square peg in a round hole for usual metrics.  To be brutally honest, when newcomers think they want to run a Lisp on the JVM these days, they are going to find and use Clojure.  But those who know something about what a conforming ANSI Common Lisp implementation provides, usually because they have existing software, are happy to discover ABCL exists when their runtime requirements suggest that they need to run on the JVM.  Many times have I gotten a request out of the blue on the mailing list for help with a problem running a large, mature, unusual software system like a theorem prover, or a symbolic algebra system on ABCL.  Since most of these software systems are decades old, they sort of predate the "social mediafication" of software development.  

I *do* think that more widespread, trusted Docker images will get more people to try ABCL a shot.

daewok:
As you've probably guessed from looking at the cl-docker-images project, our end goal is to get official Docker images for every widely used Common Lisp implementation. As easye said, while they all implement the same language each has its own extensions and claim to fame. But because they are all developed by different teams, we don't have an idea at the moment of how many would *want* images in the program. We started with ABCL because easye was enthusiastic about getting an image released and we would use it as a trial run before reaching out to the other maintainers.

> Perhaps this would make sense to publish to Docker Hub under a general "Common Lisp" organization to get interested users testing/validating it, and to help gauge the relative popularity such a thing might receive if it were to become a top-level image?

We have recently made the [clfoundation org](https://hub.docker.com/u/clfoundation) to host these images (better org names were sadly illegal or already squatted). It also hosts other images which are outside the scope of the Official Images program. Many of the images are derived from images on my [personal account](https://hub.docker.com/u/daewok).

It is rather hard to get a feel for the popularity of an individual implementation. Both because the CL community is very decentralized (even on Docker Hub there are a smattering of images) and because the implementations mostly do predate the "social mediafication" of software (many implementations have roots in the 1980s) that information simply isn't tracked or treated as important.

I would say that within the CL community [SBCL](http://sbcl.org/) is perceived as the most popular open source implementation. [CLASP](https://github.com/clasp-developers/clasp) is probably the youngest widely used implementation (claim to fame: C++ interop) and is one of the only implementations I can think of that grew up in the age of Github (the other being [SICL](https://github.com/robert-strandh/SICL)). I'd be curious if any of those implementations meet your typical standards for popularity.

yosifkit:
Apologies for the long silence. :bowing_man:

Finally coming back to review and if you are still interested, we'd love to move forward. :heart: 

From a first pass review, I think the main feedback I have is about the constant changing of directory via `WORKDIR`; since this often creates a layer along with modifying container metadata, this Dockerfile command is not as cheap as just `cd` in the `RUN` line. This will also allow you to control the permissions or ownership, if desired, of any new directories; whereas docker creates them as `root` owned. We normally only recommend using `WORKDIR` for a change that should persist in the resulting image.

The other feedback is just to update the branch here with the most recent versions that are based on newer/supported releases of OpenJDK since some OpenJDK releases went end of life (15 and 16 had quite short life cycles, but 17 is an LTS release).

tianon:

yosifkit:

PR #9490, opened 2021-01-25
AdoptOpenJDK update: 8u282-b08, 11.0.10+9 and 15.0.2+7 releases

dinogun:


dinogun:
Hotspot 8 missing for arches arm64v8 and arm32v7 as on date as the tags are not available upstream as yet. They are expected to be available end of this week tentatively.

yosifkit:
In my experience, using `kill -9` is usually very suspect. Was `shutdown.sh` not working is some situation? Does it need a longer or smarter `sleep` to wait for tomcat to stop?

bharathappali:
> Was shutdown.sh not working is some situation?

As in some situations tomcat is taking more time to come up. And in some situations `shutdown.sh` is not able to kill the tomcat on time so in the next cycle tomcat is not showing up due to port already in use error.

> Does it need a longer or smarter sleep to wait for tomcat to stop?

Yes

dinogun:
The latest changes remove the loop to reduce the overall time to build and use `shutdown -force` which will attempt a regular stop of the server and only use kill by pid if shutdown does not happen even after 5 seconds.

PR #9491, opened 2021-01-25
Add "cassandra.ring_delay_ms" to "cassandra-basics" test for 4.0+

tianon:
This is a new feature in 4.0 that delays the startup for 30 seconds (by default; https://github.com/apache/cassandra/blob/5e8f7f591dfec5a61d8eb2e9e977ec29f3a2bbe4/src/java/org/apache/cassandra/service/StorageService.java#L150-L162).

See https://github.com/docker-library/cassandra/issues/221, especially https://github.com/docker-library/cassandra/issues/221#issuecomment-764939173. :+1:


tianon:
I've tested this against a local build of `cassandra:4.0-beta4` and it (as expected) halves the time the test takes to complete, thus allowing it to complete successfully in a reasonable amount of time. :+1:

PR #9511, opened 2021-01-27
Silverpeas 6.2

mmoqui:
Release of both SIlverpeas 6.2 and 6.1.2

### DIFF
--- a/library/silverpeas
+++ b/library/silverpeas
@@ -1,11 +1,21 @@
 # This file is generated via https://github.com/Silverpeas/docker-silverpeas-prod/blob/master/generate-docker-library.sh
-
 Maintainers: Miguel Moquillon <miguel.moquillon@silverpeas.org> (@mmoqui)
 GitRepo: https://github.com/Silverpeas/docker-silverpeas-prod.git
 
-Tags: 6.1.1, latest
+Tags: 6.2, latest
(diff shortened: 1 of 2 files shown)


yosifkit:
After some local build tests, I found that the build is hanging on configuring `tzdata`. Underlying bug is https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=929417 and there is a related discussion in https://github.com/docker-library/buildpack-deps/pull/117. One fix is to install `tzdata` first: an `apt-get install -y tzdata` just before the large `apt-get install`.

tianon:
`ENV DEBIAN_FRONTEND` has unintended consequences for downstream images / users -- I would recommend instead either putting that in the `RUN` line directly (`RUN ... DEBIAN_FRONTEND=noninteractive apt-get install ...`) or more preferred, just do `apt-get install -y tzdata` in a command by itself before the `apt-get install` of the rest of the packages.

yosifkit:
I don't foresee a problem since you already have config files using `COPY` that haven't changed much. Just a recommendation to keep changes to copied files (`COPY`) minimal to help keep our build context diffs* easy to read so that we can continue to review quickly.  If it becomes a large file that changes every release, it may make sense to include it in your default silverpeas release bundles (since it could likely be useful to those outside the Docker context).

\* the [comment](https://github.com/docker-library/official-images/pull/9511#issuecomment-769890545) just above

PR #9526, opened 2021-01-29
[Flink] add 1.10.3 ; update to 1.12.1

xintongsong:
During the review process of the 1.12.0 official image, Apache Flink published 1.10.3 and 1.12.1 bugfix releases.
This PR is to add official images for these two released.


xintongsong:
Thanks for the quick response, @yosifkit

PR #9540, opened 2021-02-01
New mga 8 image and Rebuilt mga 7 and caludron images

juanluisbaptiste:
Hi @tianon, we are pushing new images for the upcoming mageia 8 that entered release candidate status last week.

Related Issue juanluisbaptiste/docker-brew-mageia#18


juanluisbaptiste:
TY

PR #9570, opened 2021-02-04
[security] Update php

tianon:
Changes:

- https://github.com/docker-library/php/commit/1ad1881: Update 7.4 to 7.4.15
- https://github.com/docker-library/php/commit/29b7e57: Update 7.3 to 7.3.27

There's also an 8.0.2 release, but according to https://www.php.net/ only these two are security releases and 8.0.2 failed in `update.sh` for reasons I need to dig further into (although no reason to hold up the things noted to be security updates on that).


tianon:
Someone already reported the 8.0.2 issue! https://bugs.php.net/bug.php?id=80711 :partying_face:

PR #9578, opened 2021-02-05
Gradle 6.8.2

keeganwitt:


keeganwitt:
I saw the build failed with

```

 - gradle:6.8.1-jdk8 (FROM adoptopenjdk:8-jdk-hotspot) [arm32v7, arm64v8]
 - gradle:6.8.1-jre8 (FROM adoptopenjdk:8-jre-hotspot) [arm32v7, arm64v8]
 ```
 
It seems like architectures periodically get removed then added back in AdoptOpenJDK, which makes this frustrating to maintain. It looks like that's happened again (https://github.com/docker-library/official-images/pull/9490#issuecomment-767010171). I guess I'll wait a couple days then remove it if it's not available still.

tianon:
> It seems like architectures periodically get removed then added back in AdoptOpenJDK, which makes this frustrating to maintain. It looks like that's happened again ([#9490 (comment)](https://github.com/docker-library/official-images/pull/9490#issuecomment-767010171)). I guess I'll wait a couple days then remove it if it's not available still.

Yeah, fully agree (ours are even automated, but it ends up creating quite a bit of churn). :disappointed:

We'll merge this as-is for now and hope the Adopt folks get this resolved quickly.

PR #9594, opened 2021-02-08
Add InfluxDB OSS 2.0

danxmoran:
Updated documentation staged at docker-library/docs#1871

### DIFF
--- a/library/influxdb
+++ b/library/influxdb
@@ -1,6 +1,6 @@
 Maintainers: Daniel Moran <dmoran@influxdata.com> (@danxmoran)
 GitRepo: git://github.com/influxdata/influxdata-docker
-GitCommit: cfdf53f5e9ab71671ebca007b7d9e46f8603493e
+GitCommit: c3c73281e6c0dfa14383c1cf5cd4fb6a7521872a
 
 Tags: 1.7, 1.7.10
 Architectures: amd64, arm32v7, arm64v8
--- a/library/influxdb
+++ b/library/influxdb
@@ -1,6 +1,6 @@
 Maintainers: Daniel Moran <dmoran@influxdata.com> (@danxmoran)
 GitRepo: git://github.com/influxdata/influxdata-docker
-GitCommit: 21b42bc2042d806d6175285646cf20f1f0a80057
+GitCommit: 2c1aa43aac238e3439866663ab4c6810b397ac32
 
 Tags: 1.7, 1.7.10
 Architectures: amd64, arm32v7, arm64v8
--- a/library/influxdb
+++ b/library/influxdb
@@ -1,6 +1,6 @@
 Maintainers: Daniel Moran <dmoran@influxdata.com> (@danxmoran)
 GitRepo: git://github.com/influxdata/influxdata-docker
-GitCommit: 2c1aa43aac238e3439866663ab4c6810b397ac32
+GitCommit: cfdf53f5e9ab71671ebca007b7d9e46f8603493e
 
 Tags: 1.7, 1.7.10
 Architectures: amd64, arm32v7, arm64v8
@@ -45,4 +45,5 @@ Architectures: amd64, arm64v8
 Directory: influxdb/2.0
 
 Tags: 2.0-alpine, 2.0.4-alpine, alpine
+Architectures: amd64, arm64v8
 Directory: influxdb/2.0/alpine
--- a/library/influxdb
+++ b/library/influxdb
@@ -1,6 +1,6 @@
-Maintainers: Jonathan A. Sternberg <jonathan@influxdata.com> (@jsternberg)
+Maintainers: Daniel Moran <dmoran@influxdata.com> (@danxmoran)
 GitRepo: git://github.com/influxdata/influxdata-docker
-GitCommit: 4d733eabcc1e3691e462a702db8eede9d19a9c0a
+GitCommit: 21b42bc2042d806d6175285646cf20f1f0a80057
 
 Tags: 1.7, 1.7.10
 Architectures: amd64, arm32v7, arm64v8
@@ -21,11 +21,11 @@ Directory: influxdb/1.7/meta
 Tags: 1.7-meta-alpine, 1.7.10-meta-alpine
 Directory: influxdb/1.7/meta/alpine
 
-Tags: 1.8, 1.8.4, latest
+Tags: 1.8, 1.8.4
 Architectures: amd64, arm32v7, arm64v8
 Directory: influxdb/1.8
 
-Tags: 1.8-alpine, 1.8.4-alpine, alpine
+Tags: 1.8-alpine, 1.8.4-alpine
 Directory: influxdb/1.8/alpine
 
 Tags: 1.8-data, 1.8.3-data, data
@@ -39,3 +39,10 @@ Directory: influxdb/1.8/meta
 
 Tags: 1.8-meta-alpine, 1.8.3-meta-alpine, meta-alpine
 Directory: influxdb/1.8/meta/alpine
+
+Tags: 2.0, 2.0.4, latest
+Architectures: amd64, arm64v8
+Directory: influxdb/2.0
+
+Tags: 2.0-alpine, 2.0.4-alpine, alpine
+Directory: influxdb/2.0/alpine


yosifkit:
Along the lines of https://github.com/docker-library/official-images/pull/9249#issuecomment-756496043, our concern is that some of the functionality of this script is likely duplicated in your packaging for distributions (any debs and rpms) and would be useful to users outside docker (like database migration when installing version 2 from version 1). So we like to make sure that the Dockerization makes use of these types of scripts so that deployments in and out of docker can benefit from fixes (and minimize the added maintenance of the Docker images).

One way we try to emphasize this is by recommending that changes to files in the build context should be kept minimal to help keep our build context diffs easy to read so that we can continue to review quickly. If it becomes a large file that changes every release, it may make sense to include it in the upstream release (or released as a fully maintained project/example that can be downloaded by the image).

------

-  `VOLUME /etc/influxdb2`

	Does the image/application expect to modify it in a persistent way that needs to stick around from invocation to invocation? (Given the "permanence" of a `VOLUME` declaration, we recommend keeping them to just the directories the application stores active state in, and even more specifically active state that's actually necessary to keep if the container is completely destroyed and recreated, such as MySQL's `/var/lib/mysql` directory.)

- `ARCH=$(arch) && \`

	Using `uname -m` or `arch` for platform detection is very strongly discouraged, as it tests the host kernel's architecture, when what you really wanted to test was the image's userspace architecture, so `apk --print-arch` or `dpkg --print-architecture` are going to be safer/more reliable options.

------

A few edge cases that could cause issues for users:

- unquoted strings and arrays like `${INFLUXDB_INIT_PASSWORD}`, `${BOLT_PATH}`, and `${@}`

- unremoved `trap` calls for `TERM` and `INT`

- `exec gosu` doesn't use the `influxd_main` function

danxmoran:
Thanks @yosifkit! Responding to some of your points:

> our concern is that some of the functionality of this script is likely duplicated in your packaging for distributions (any debs and rpms) and would be useful to users outside docker (like database migration when installing version 2 from version 1). So we like to make sure that the Dockerization makes use of these types of scripts so that deployments in and out of docker can benefit from fixes (and minimize the added maintenance of the Docker images)

There are some scripts, but they have minimal overlap. Core functionality like running the DB upgrade is baked into the `influxd` binary, and we invoke it in different ways for different packaging scenarios. IMO the Docker scripts are much better-tested than what's packaged with our deb/rpm packages, but any backport-ing would have to wait for our 2.0.5 release.

> recommending that changes to files in the build context should be kept minimal to help keep our build context diffs easy to read so that we can continue to review quickly. If it becomes a large file that changes every release, it may make sense to include it in the upstream release

I expect that the big diff in this PR will be a one-time thing, forced by all the breaking changes of the major version bump

>  `VOLUME /etc/influxdb2`
>
> Does the image/application expect to modify it in a persistent way that needs to stick around from invocation to invocation? (Given the "permanence" of a VOLUME declaration, we recommend keeping them to just the directories the application stores active state in, and even more specifically active state that's actually necessary to keep if the container is completely destroyed and recreated, such as MySQL's /var/lib/mysql directory.)

Yes, both of the auto-initialization modes for the DB will generate configs stored under `/etc/influxdb2`
* "setup" mode will generate CLI configs
* "upgrade" mode will generate both CLI configs and an upgraded server config

> unremoved `trap` calls for `TERM` and `INT`

If I understand correctly, the handlers are removed [here](https://github.com/influxdata/influxdata-docker/blob/master/influxdb/2.0/entrypoint.sh#L268), after the server instance used to run initialization logic is shut down.

> `exec gosu` doesn't use the `influxd_main` function

We have `exec gosu` re-call the entire entrypoint script so users can run commands other than `influxd` (i.e. run the `influx` CLI) as a non-root user

---

I'll work on fixing these points tomorrow:
* Using `apk --print-arch` and `dpkg --print-arch` instead of `arch`
* Unquoted strings and arrays like `${INFLUXDB_INIT_PASSWORD}`, `${BOLT_PATH}`, and `${@}`

yosifkit:
> I expect that the big diff in this PR will be a one-time thing, forced by all the breaking changes of the major version bump

:+1: :heart: 

> If I understand correctly, the handlers are removed here, after the server instance used to run initialization logic is shut down.

Oh, it looks like the Alpine version didn't get the other trap removals (https://github.com/influxdata/influxdata-docker/blob/21b42bc2042d806d6175285646cf20f1f0a80057/influxdb/2.0/alpine/entrypoint.sh#L276).

> We have `exec gosu` re-call the entire entrypoint script so users can run commands other than `influxd` (i.e. run the `influx` CLI) as a non-root user

Doh, I must've missed the `$0` there :disappointed:. That's how many of our own entrypoint scripts work :+1: 

Thanks!

danxmoran:
@yosifkit I think I've covered everything

yosifkit:
Alpine Dockerfile has one lingering `arch` invocation.
```Dockerfile
ENV INFLUXDB_VERSION 2.0.4
RUN set -eux && \
    ARCH=$(arch) && \
```

The alpine Dockerfile for `2.0` is written to support `arm64v8`, but the `library/influxdb` file only has the default `amd64` architecture. Did you want this built on `arm64v8` too?

```diff
 Tags: 2.0-alpine, 2.0.4-alpine, alpine
+Architectures: amd64, arm64v8
 Directory: influxdb/2.0/alpine
```

Without overriding the entrypoint, non-influxdb commands like `bash` cannot be run as root, since it always steps down. This could be unexpected and confusing for users. 

```console
$ docker run -it --rm influxdb:2.0 bash
influxdb@cfb552cd8cce:/$ exit
$ docker run -it --rm --user root influxdb:2.0 bash
influxdb@c577c7115858:/$ exit
```

danxmoran:
> The alpine Dockerfile for 2.0 is written to support arm64v8, but the library/influxdb file only has the default amd64 architecture. Did you want this built on arm64v8 too?

Yes, will add! I was following the pattern used elsewhere in the file... I guess that means all the other alpine/meta/data variants of influxdb have only been bulding for `amd64` as well 🤦

danxmoran:
I've set myself up for merge conflicts with #9614... if I get a vote on the merge order, this one is higher priority from our end 😄

danxmoran:
@yosifkit @tianon all comments addressed. I'm going to be OOO tomorrow and Monday, so I'll be a bit delayed in addressing any additional feedback. Thanks for the thorough review 😄

danxmoran:
Ping @yosifkit @tianon, is there anything I can do to help move this along?

yosifkit:
Will the differing prefixes on some variables confuse users? `INFLUX_`,  `INFLUXD_`, and  `INFLUXDB_`. (If I had a `yml` defining all three I would think I misspelled some)

It seems like there still could be discrepancies between the two `entrypoint.sh` scripts since one is `#!/bin/bash` and one `#!/bin/sh`; `bash` is installed in the alpine image ( `RUN apk add --no-cache tzdata bash ca-certificates gnupg run-parts &&` ), why not keep them the same?

danxmoran:
> Will the differing prefixes on some variables confuse users? INFLUX_, INFLUXD_, and INFLUXDB_. (If I had a yml defining all three I would think I misspelled some)

The `INFLUX_` and `INFLUXD_` prefixes are known and expected by InfluxDB users. The docs linked in my original comment explain the naming convention (`<uppercase-binary-name>_<upper-snake-case-setting-name>`). I didn't want to reuse either of those for the env vars that only affect the Docker entrypoint to avoid confusion over real/assumed collisions, but I see your point... I could expand the new prefix to `INFLUXDB_DOCKER_`? I'll check with our product and dev. advocate teams as well.

> It seems like there still could be discrepancies between the two entrypoint.sh scripts since one is #!/bin/bash and one #!/bin/sh; bash is installed in the alpine image ( RUN apk add --no-cache tzdata bash ca-certificates gnupg run-parts && ), why not keep them the same?

Honestly, I copy-pasted the first few lines for the alpine `Dockerfile` from an earlier version and completely missed that `bash` was in there. I'll test using the same entrypoint script in both, and delete the `/bin/sh` variant if everything works.

danxmoran:
@yosifkit comments addressed:
1. "normal" and alpine entrypoints are now identical
2. Updated the env vars that are specific to the DockerHub image to start with `DOCKER_INFLUXDB_`

yosifkit (approved):

tianon:
Notes for the next update (not going to block this on them):

```diff
+    gpg --keyserver keys.openpgp.org --recv-keys 8C2D403D3C3BDB81A4C27C883C3E4B7317FFE40A && \
+    gpg --batch --verify influxdb2-${INFLUXDB_VERSION}-linux-${ARCH}.tar.gz.asc influxdb2-${INFLUXDB_VERSION}-linux-${ARCH}.tar.gz && \
```

Both of these should include `--batch` :smile:

```diff
+    if [ ${ARCH} = x86_64 ]; then \
+        ARCH=amd64; \
+    elif [ ${ARCH} = aarch64 ]; then \
+        ARCH=arm64; \
+    else \
+        echo "Unsupported architecture: ${ARCH}" && exit 1; \
+    fi && \
```

This would probably be easier to update/maintain as a `case` statement instead of an `if`/`elif` chain (I think there's a couple instances of a similar pattern). :+1:

```diff
+    export BOLT_PATH="$(influxd print-config --key-name bolt-path "${@}")"
+    export ENGINE_PATH="$(influxd print-config --key-name engine-path "${@}")"
```

I would suggest splitting these into separate set/export so that `set -e` can fail appropriately if `influxd` fails for some unexpected reason, like:

```bash
    BOLT_PATH="$(influxd print-config --key-name bolt-path "${@}")"
    ENGINE_PATH="$(influxd print-config --key-name engine-path "${@}")"
    export BOLT_PATH ENGINE_PATH
```

(There are a few other similar cases of `<local|declare|export|etc> var="$(...)"` that would benefit from a similar change like `local foo; foo="$(...)"`, although things like `dirname` are probably even less likely than `influxd` to fail :sweat_smile:)

danxmoran:
Thanks @yosifkit @tianon!

Will the linked [docs PR](https://github.com/docker-library/docs/pull/1871) also merge today?

tianon:
Oh, I suppose I can go look at that right now :sweat_smile: :heart:

PR #9609, opened 2021-02-10
Update to Adminer 4.8.0

TimWolla:


library/adminer
```diff
@@ -1,14 +1,14 @@
-# this file is generated via https://github.com/TimWolla/docker-adminer/blob/7730bed48bb21686486346ea70c75b423a5d3755/generate-stackbrew-library.sh
+# this file is generated via https://github.com/TimWolla/docker-adminer/blob/a46dbffde41a1f48f61e5881033a21011f472de1/generate-stackbrew-library.sh
 
 Maintainers: Tim Düsterhus <tim@bastelstu.be> (@TimWolla)
 GitRepo: https://github.com/TimWolla/docker-adminer.git
 
-Tags: 4.7.9-standalone, 4.7-standalone, 4-standalone, standalone, 4.7.9, 4.7, 4, latest
+Tags: 4.8.0-standalone, 4-standalone, standalone, 4.8.0, 4, latest
```
yosifkit, line 6: Did you want a plain `4.8` (and `4.8-fastcgi`) tag?

TimWolla: No, I specifically dropped them here: https://github.com/docker-library/official-images/pull/9609/commits/44173ef29f3e83c21208d13497ce22517d262332.

From the looks it does not appear like Adminer is using a versioning system where branches will get releases after the release of a new second digit version and I am also not able to see a meaningful difference between the changelog of a second digit update and a third digit update.

So in the interest of simplicity I dropped the x.y tag.

TimWolla: See also: https://github.com/vrana/adminer/blob/master/SECURITY.md#supported-versions (the SECURITY.md was published just a few minutes ago).

PR #9631, opened 2021-02-15
Update Sonarqube to 8.7.0

zipeng-wu-sonarsource:
- Add Sonarqube datacenter edition images. 
- Documentation is updated in the PR: https://github.com/docker-library/docs/pull/1877

### DIFF
--- a/library/sonarqube
+++ b/library/sonarqube
@@ -4,16 +4,22 @@ Maintainers: Janos Gyerik <janos.gyerik@sonarsource.com> (@janos-ss),
              Pierre Guillot <pierre.guillot@sonarsource.com> (@pierre-guillot-sonarsource),
              Michal Duda <michal.duda@sonarsource.com> (@michal-duda-sonarsource)
 GitRepo: https://github.com/SonarSource/docker-sonarqube.git
-GitCommit: 73a85e01c1f5acfab08026318cfa9da598b9bfb1
+GitCommit: 7e2c6a2bf55a250c332f8e4bacf0a1fe87264ebf
 
 Tags: 7.9.5-community, 7.9-community, lts
 Directory: 7/community
 
-Tags: 8.6.1-community, 8.6-community, 8-community, community, latest
+Tags: 8.7.0-community, 8.7-community, 8-community, community, latest
 Directory: 8/community
 
-Tags: 8.6.1-developer, 8.6-developer, 8-developer, developer
+Tags: 8.7.0-developer, 8.7-developer, 8-developer, developer
 Directory: 8/developer
 
-Tags: 8.6.1-enterprise, 8.6-enterprise, 8-enterprise, enterprise
+Tags: 8.7.0-enterprise, 8.7-enterprise, 8-enterprise, enterprise
 Directory: 8/enterprise
+
+Tags: 8.7.0-datacenter-app, 8.7-datacenter-app, 8-datacenter-app, datacenter-app
+Directory: 8/datacenter/app
+
+Tags: 8.7.0-datacenter-search, 8.7-datacenter-search, 8-datacenter-search, datacenter-search
+Directory: 8/datacenter/search


zipeng-wu-sonarsource:
Hi @tianon , commit hash has been updated to reflect changes in https://github.com/SonarSource/docker-sonarqube/pull/468

yosifkit:
```bash
IP=$(ifconfig | sed -En 's/127.0.0.1//;s/.*inet (addr:)?(([0-9]*\.){3}[0-9]*).*/\2/p')
```

This is going to be a multi-line string if the container is on more than one network. From the [sonarqube docs](https://docs.sonarqube.org/latest/setup/operate-cluster/), it looks like sonarqube already has this behavior as its default.

>sonar.cluster.node.host
>
>IP address of the network card that will be used by Hazelcast to communicate with the members of the cluster. If not specified, the first interface will be chosen (note that loopback interfaces won't be selected).

Assuming the first non-local IP address is still needed for other config values, you might need to watch out for issues like https://github.com/docker-library/cassandra/issues/223.

zipeng-wu-sonarsource:
Hi @yosifkit, thanks a lot for your feedback.
> From the sonarqube docs, it looks like sonarqube already has this behavior as its default.
In fact, thanks to you, we just discovered that it's a mistake in our documentation, and we have created a [PR](https://github.com/SonarSource/sonar-enterprise/pull/3787) to fix this issue on ourside.

> Assuming the first non-local IP address is still needed for other config values, you might need to watch out for issues like docker-library/cassandra#223.
Indeed, @tobias-trabelsi-sonarsource has update docker-sonarqube repo by this PR https://github.com/SonarSource/docker-sonarqube/pull/469 .  I have just updated the commit hash to reflect this change.

ivan-uqido:
I need this docker release version ASAP because of an important bug fixing from the previous one...

tianon:
```diff
-    GLIBC_VER="2.32-r0"; \
+    GLIBC_VER="2.31-r0"; \
```

Is downgrading the glibc version intentional?  It really makes me nervous to learn these are just being copied cargo-cult style from `Dockerfile`s elsewhere - is that the reason for this downgrade?

---

```diff
-set -euo pipefail
+set -eo pipefail
```

Dropping `-u` here seems strange -- is it because of `if [ -z "${SONAR_PATH_LOGS}" ]` (which could be `if [ -z "${SONAR_PATH_LOGS:-}" ]` instead and keep `-u`)?

---

```diff
+    if [ -z "${SONAR_PATH_LOGS}" ]
+    then
+        SONAR_CLUSTER_PATH_LOGS="logs/${HOSTNAME}"
+        mkdir -p ${SONARQUBE_HOME}/${SONAR_CLUSTER_PATH_LOGS}
+        chown -R sonarqube:sonarqube ${SONARQUBE_HOME}/${SONAR_CLUSTER_PATH_LOGS}
+    else
+        SONAR_CLUSTER_PATH_LOGS="${SONAR_PATH_LOGS}/${HOSTNAME}"
+        mkdir -p ${SONAR_CLUSTER_PATH_LOGS}
+        chown -R sonarqube:sonarqube ${SONAR_CLUSTER_PATH_LOGS}
+    fi
```

This seems like kind of a strange workaround for "cluttered" logs -- how common is it for users to scale up with `docker-compose` on a single host?  Isn't the service kind of heavy, making that a bad idea generally?

---

```diff
-exec java -jar lib/sonar-application-"${SONAR_VERSION}".jar -Dsonar.log.console=true "$@"
+exec java -jar lib/sonar-application-"${SONAR_VERSION}".jar -Dsonar.log.console=true -Dsonar.log.console=true "$@"
```

Is the duplicated `-Dsonar.log.console=true` intentional here?

tobias-trabelsi-sonarsource:
Hi @tianon ,

good catch on the glibc version. i updated it in September last year but forgot about it. i also re added the missing u option in the run script and removed the duplication of the console property. you can see the changes in this pr: https://github.com/SonarSource/docker-sonarqube/pull/471

zipeng-wu-sonarsource:
Commit hash has been updated accordingly to reflect changes in https://github.com/SonarSource/docker-sonarqube/pull/471

yosifkit (approved):

zipeng-wu-sonarsource:
Hi @tianon, @yosifkit , thanks a lot for merging this PR and publishing Sonarqube 8.7.0 official docker image. Meanwhile, we have also a open PR https://github.com/docker-library/docs/pull/1877 to update Docker documentation correspondingly.

PR #9635, opened 2021-02-16
Updated nginx to 1.19.7

thresheek:


tianon:
This looks generally fine, but IMO it seems like `30-tune-worker-processes.sh` should probably be part of `worker_processes auto` eventually, right?  Especially given that `systemd` might be applying cgroup limits to NGINX outside Docker, so this isn't really a container-specific problem anymore. :grimacing:

(Is there maybe a wishlist request somewhere this could link to?)

Along a similar vein, perhaps instead of (or in addition to?) the explicit `NGINX_ENTRYPOINT_WORKER_PROCESSES_AUTOTUNE` variable, it would make sense to only replace `worker_processes auto` instead of replacing any value (and adjust the default to `auto`, I suppose)?

thresheek:
>  it seems like 30-tune-worker-processes.sh should probably be part of worker_processes auto eventually, right?

Yes, but getting patches in nginx is a challenge. I don't see it hitting upstream any time soon, especially with current projects like http3 that take an enormous amount of dev time (although admittedly I'm not a PM for nginx).

> instead of replacing any value (and adjust the default to auto, I suppose)?

That sounds like something that we can indeed do, e.g. in a 1.20/1.21 change coming in April for nginx packages. We're generally quite conservative with default configuration changes, but `auto` has been there for the commercial version since its inception and I dont think we ever seen complaints.  If we do that, we can change the script to only care about `auto` in the configuration, and leave predefined values as is.

PR #9651, opened 2021-02-18
Sunset Classic Swarm image

thaJeztah:
Further deprecate the classic Swarm image, as classic (standalone, not swarmkit) is deprecated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,7 +168,7 @@ All official images should provide a consistent interface. A beginning user shou
 3.	If the image only contains the main executable and its linked libraries (ie no shell) then it is fine to use the executable as the `ENTRYPOINT`, since that is the only thing that can run:
 
 	```Dockerfile
-	ENTRYPOINT ["swarm"]
+	ENTRYPOINT ["fully-static-binary"]
 	CMD ["--help"]
 	```
 
--- a/library/swarm
+++ b/library/swarm
@@ -1,9 +1,5 @@
-Maintainers: Alexandre Beslic <abronan@docker.com> (@abronan),
-             Andrea Luzzardi <al@docker.com> (@aluzzardi),
-             Dong Chen <dong@docker.com> (@dongluochen),
-             Nishant Totla <nishanttotla@gmail.com> (@nishanttotla),
-             Victor Vieux <vieux@docker.com> (@vieux)
-GitRepo: https://github.com/docker/swarm-library-image.git
+# Classic Swarm has been archived and is no longer actively developed.
+# You may want to use the Swarm mode built into the Docker Engine instead, or another orchestration system.
 
-Tags: 1.2.9, latest
-GitCommit: a2a08d733833327354b5d3336ffbd19e4f369a2d
+Maintainers: Docker <info@docker.com> (@docker)
+GitRepo: https://github.com/docker-archive/swarm-library-image.git
--- a/test/config.sh
+++ b/test/config.sh
@@ -284,8 +284,6 @@ globalExcludeTests+=(
 	[nats-streaming_utc]=1
 	[nats_no-hard-coded-passwords]=1
 	[nats_utc]=1
-	[swarm_no-hard-coded-passwords]=1
-	[swarm_utc]=1
 	[traefik_no-hard-coded-passwords]=1
 	[traefik_utc]=1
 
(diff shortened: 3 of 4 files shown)


library/swarm
```diff
@@ -1,9 +1,2 @@
-Maintainers: Alexandre Beslic <abronan@docker.com> (@abronan),
-             Andrea Luzzardi <al@docker.com> (@aluzzardi),
-             Dong Chen <dong@docker.com> (@dongluochen),
-             Nishant Totla <nishanttotla@gmail.com> (@nishanttotla),
-             Victor Vieux <vieux@docker.com> (@vieux)
-GitRepo: https://github.com/docker/swarm-library-image.git
-
-Tags: 1.2.9, latest
-GitCommit: a2a08d733833327354b5d3336ffbd19e4f369a2d
+Maintainers: Docker <info@docker.com> (@docker)
```
thaJeztah, line 1: Opted to just put some "boilerplate" information here, as I wasn't sure if this could be removed altogether

test/config.sh
```diff
@@ -284,7 +284,9 @@ globalExcludeTests+=(
 	[nats-streaming_utc]=1
 	[nats_no-hard-coded-passwords]=1
 	[nats_utc]=1
+	[swarm_cve-2014--shellshock]=1
```
thaJeztah, line 287: I didn't find a mention of "how to exclude an image for everything"; perhaps it's there but I didn't find it

README.md
```diff
@@ -168,7 +168,7 @@ All official images should provide a consistent interface. A beginning user shou
 3.	If the image only contains the main executable and its linked libraries (ie no shell) then it is fine to use the executable as the `ENTRYPOINT`, since that is the only thing that can run:
 
 	```Dockerfile
-	ENTRYPOINT ["swarm"]
+	ENTRYPOINT ["mybinary"]
```
justincormack, line 171: why??

library/swarm
justincormack: I would say this image is no longer maintained since (date)

README.md
thaJeztah: Thought it didn't make sense to use `swarm` as an example; this is just for illustration?

library/swarm
thaJeztah: I wasn't sure if there's scripts relying on the format (and if removing/changing to something else would cause breakage)

thaJeztah: If not, then I'll update (I'll wait for @tianon to check)

test/config.sh
yosifkit: With no tags in the library file, there is nothing to test so any mentions of swarm can just be removed.

(The failed GitHub actions email you are getting is because it is generating an empty matrix to run tests on and that is fine)

README.md
yosifkit: I think we chose `swarm` since it was one of the only single binary images we had at the time. I guess `hello` from the hello-world image would work.

library/swarm
yosifkit: Yeah, format of the library file is important. Comment (`#`) lines are fine though.

yosifkit:
Just a reminder that there is already a deprecation notice in the Docs (and so on the Docker Hub page): https://github.com/docker-library/docs/pull/1738.

README.md
tianon: It might be more clear if we choose something more self-descriptive like `fully-static-binary`

library/swarm
tianon: Yeah, maybe we should duplicate the deprecation notice in a comment?

```suggestion
# Classic Swarm has been archived and is no longer actively developed.
# You may want to use the Swarm mode built into the Docker Engine instead, or another orchestration system.

Maintainers: Docker <info@docker.com> (@docker)
```

test/config.sh
thaJeztah: okay; I'll remove these lines: thanks!

library/swarm
thaJeztah: I'll do that 👍

thaJeztah:
> Just a reminder that there is already a deprecation notice in the Docs (and so on the Docker Hub page): docker-library/docs#1738.

Thanks! Yes, I had that one on my list 🤗 

Screenshot 2021-02-18 at 22 43 41

thaJeztah:
(oh, and updated with the suggestions 👍)

yosifkit (approved):

tianon (approved):
:metal:

thaJeztah:
Thanks ✅

PR #9693, opened 2021-02-25
Composer 2.0.11

alcohol:
Side note, is it possible to automate part of this process @yosifkit / @tianon ?


yosifkit:
>  is it possible to automate part of this process?

If you are talking about making PRs to us automated, then that is acceptable and we have some minimal guidelines: 
https://github.com/docker-library/faq/tree/0fd4aeb047fc37ed37bd9991cae479140450ae65#can-i-use-a-bot-to-make-my-image-update-prs

alcohol:
Thanks, will look into that!

PR #9695, opened 2021-02-25
Gradle 6.8.3

keeganwitt:


yosifkit:
I guess we are again waiting on adoptopenjdk for arm builds: https://github.com/docker-library/official-images/pull/9578#issuecomment-774179861. Merging anyway.

PR #9696, opened 2021-02-25
eclipse-mosquitto 2.0.8

ralight:
As well as the version bump, the 2.0 images now have the option to
change runtime behaviour using an environment variable.


yosifkit:
Matching on `$*` doesn't seem very flexible. Why a whole environment variable when they could just be told to change the command?

 ```console
$ docker run -d eclipse-mosquitto:2.0.8 mosquitto -c /mosquitto-no-auth.conf
```

ralight:
Yes, that's a much better idea. I had users repeating "environment variable" in my ear so that's why I went down that route.

PR #9705, opened 2021-02-26
Update photon image

gerrit-photon:
Update photon image


michellew-vmware:
Hi docker team,

We deploy 4.0GA docker image today in the docker file. And latest tag will snapped at 4.0 instead of 3.0 from now on.

Thanks,
Michelle

PR #9720, opened 2021-03-01
add Kong 2.2.2

gszr:


library/kong
```diff
@@ -20,21 +20,21 @@ GitFetch: refs/tags/2.3.2
 Directory: centos
 Architectures: amd64
 
-Tags: 2.2.1-alpine, 2.2-alpine, 2.2.0, 2.2
```
yosifkit, line 23: If you need to have us rebuild `2.2.0` since it seems to contain `2.2.1` (and maybe the plain `2.2.1` tag?), feel free to make a new PR that adds those tags. We are open to temporarily back-fill old versions if you think it is important to fix the old tags.

PR #9747, opened 2021-03-04
Update php

tianon:
Changes:

- https://github.com/docker-library/php/commit/6481179: Update 8.0 to 8.0.3
- https://github.com/docker-library/php/commit/1bc63c1: Update 7.4 to 7.4.16

### DIFF
--- a/library/php
+++ b/library/php
@@ -4,94 +4,94 @@ Maintainers: Tianon Gravi <admwiggin@gmail.com> (@tianon),
              Joseph Ferguson <yosifkit@gmail.com> (@yosifkit)
 GitRepo: https://github.com/docker-library/php.git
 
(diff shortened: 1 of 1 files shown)


zoispag:
Getting: `no matching manifest for linux/amd64 in the manifest list entries`.
Something wrong with the build?

tianon:
Indeed, unfortunately the official images amd64 build server stopped responding and we're still trying to investigate 😱

Details/tracking in https://github.com/docker-library/official-images/issues/9749 😬❤️

zoispag:
> Indeed, unfortunately the official images amd64 build server stopped responding and we're still trying to investigate 😱
> 
> Details/tracking in #9749 😬❤️

Thanks for the heads up @tianon 🙏🏻

PR #9758, opened 2021-03-06
Update ELK images

tianon:
elasticsearch:
- https://github.com/docker-library/elasticsearch/commit/4ac154b: Update to 7.11.1
- https://github.com/docker-library/elasticsearch/commit/29bc2ab: Update to 6.8.14
- https://github.com/docker-library/elasticsearch/commit/db67263: Update to 7.11.0
- https://github.com/docker-library/elasticsearch/commit/9043b04: Update to 7.10.2

logstash:
- https://github.com/docker-library/logstash/commit/2b17ad4: Update to 7.11.1
- https://github.com/docker-library/logstash/commit/4a715b5: Update to 6.8.14
- https://github.com/docker-library/logstash/commit/c4aa9b9: Update to 7.11.0
- https://github.com/docker-library/logstash/commit/5ae8fc1: Update to 7.10.2

kibana:
- https://github.com/docker-library/kibana/commit/877c966: Update to 7.11.1
- https://github.com/docker-library/kibana/commit/a813600: Update to 6.8.14
- https://github.com/docker-library/kibana/commit/89b7f10: Update to 7.11.0
- https://github.com/docker-library/kibana/commit/2282b15: Update to 7.10.2

### DIFF
--- a/library/elasticsearch
+++ b/library/elasticsearch
@@ -4,12 +4,12 @@ Maintainers: Tianon Gravi <admwiggin@gmail.com> (@tianon),
              Joseph Ferguson <yosifkit@gmail.com> (@yosifkit)
 GitRepo: https://github.com/docker-library/elasticsearch.git
 
-Tags: 7.10.1
+Tags: 7.11.1
 Architectures: amd64, arm64v8
-GitCommit: 38f3a55628528582a5a951e2fab6c01878af163c
+GitCommit: 4ac154b859d306286cf240607fe2bdefa0a1caab
 Directory: 7
 
-Tags: 6.8.13
+Tags: 6.8.14
 Architectures: amd64
-GitCommit: 008dd9ddffa1881494e32f6c09763e5097293964
+GitCommit: 29bc2aba27cf15e72e0fd05fbfca2f6ac6d07606
 Directory: 6
--- a/library/kibana
+++ b/library/kibana
@@ -4,10 +4,10 @@ Maintainers: Tianon Gravi <admwiggin@gmail.com> (@tianon),
              Joseph Ferguson <yosifkit@gmail.com> (@yosifkit)
 GitRepo: https://github.com/docker-library/kibana.git
 
-Tags: 7.10.1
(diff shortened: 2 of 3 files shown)


tianon:
Upstream Diffs:

- https://github.com/elastic/dockerfiles/compare/v6.8.13...v6.8.14
- https://github.com/elastic/dockerfiles/compare/v7.10.1...v7.11.1

(I have not fully reviewed these yet)

yosifkit:
> COPY bin/transform-log4j-config-7.11.1.jar /tmp/

It seems strange to commit a compiled file to the git repository, even if it is only used in an earlier stage. :man_shrugging:

> RUN curl -Lo - https://artifacts.elastic.co/downloads/logstash/logstash-7.11.1-linux-$(arch).tar.gz | \

- Using `uname -m` or `arch` for platform detection is very strongly discouraged, as it tests the host kernel's architecture, when what you really wanted to test was the image's userspace architecture, so `apk --print-arch` or `dpkg --print-architecture` are going to be safer/more reliable options.

Not good, but for now, I'll assume that upstream Elastic Co are not going to build `i386` or `arm32x` Logstash on `amd64` or `arm64v8`, respectively, and the fact that rpm-based distros have no such thing as `dpkg --print-architecture`. As far as current non-amd64 architecture support, there is only `arm64v8` for `elasticsearch:7.+`; I find it highly unlikely that they will even have 32-bit architectures.

tianon:
cc @mgreau @pugnascotia

pugnascotia:
> It seems strange to commit a compiled file to the git repository, even if it is only used in an earlier stage. 🤷‍♂️

🤦 that was me. We build a tool from source when we build the Docker image in our Gradle build, but I forgot that when create the image via this repo we need to fetch or compile it instead. I'll file an issue to fix this in a subsequent release, if that's OK?

Thanks also for your comments re: architecture. We intend to move away from a CentOS base, but haven't yet decided what we'll move to. We have no plans to support 32-bit architectures right now.

PR #9770, opened 2021-03-09
Plone 5.2.4 release

avoinea:
Refs https://github.com/plone/plone.docker/issues/155


avoinea:
I think the failure within `5.2.4-alpine` test was a temporary network glitch :smile:

yosifkit:
LGTM

---

Just a note that `buildout.cfg` and `docker-initialize.py` seem like quite a lot to have outside the standard Plone packaging. Would it benefit users outside of Docker images to have all/part of these files in the regular `Plone-*-UnifiedInstaller-1.0.tgz` release?

avoinea:
@yosifkit Thank you!

With Plone 6 we plan to move away from zc.buildout and UnifiedInstaller and just use something like: 

```
pip install Plone
```

Until then, indeed some parts from buildout.cfg can be moved to the common Plone buildout (versions, cors, plonesite, etc.)

PR #9772, opened 2021-03-10
Update wordpress

tianon:
Changes:

- https://github.com/docker-library/wordpress/commit/fb920c6: Merge pull request https://github.com/docker-library/wordpress/pull/572 from infosiftr/un-beta
- https://github.com/docker-library/wordpress/commit/891b710: Move "wp-config-docker.php" to non-beta (5.7 GA)
- https://github.com/docker-library/wordpress/commit/24c4469: Update latest to 5.7.0
- https://github.com/docker-library/wordpress/commit/59404ef: Update beta to 5.7.0

**Important Note:** now that 5.7 is GA, this includes https://github.com/docker-library/wordpress/pull/557 in all tags (no longer `beta`-only).

### DIFF
--- a/library/wordpress
+++ b/library/wordpress
@@ -1,52 +1,52 @@
-# this file is generated via https://github.com/docker-library/wordpress/blob/8f5bcc1c980fce198bd191cfb7f51a3e76bb59cd/generate-stackbrew-library.sh
+# this file is generated via https://github.com/docker-library/wordpress/blob/891b7108294a629e6cc16c2f4bf643d9475894cc/generate-stackbrew-library.sh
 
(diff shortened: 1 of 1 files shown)


tianon:
Just to make sure the important bits / migration path for users are in more places they might look when things don't seem to be working the same way they used to, see https://github.com/docker-library/wordpress/pull/557#issuecomment-756453476:

> For users who are changing `WORDPRESS_*` variables and expecting our `docker-entrypoint.sh` to update them, the "migration" path is to remove your `wp-config.php` file and let the new entrypoint copy in the updated version, and everything will "just work" from then on.

PR #9828, opened 2021-03-19
Update yourls

LeoColomb:
This change won't affect published images, but reflect a refactor of https://github.com/YOURLS/docker-yourls repo.
* The main branch is now named `main`
* The built Dockerfiles and relative files are moved to a separate branch (`dist`)
  * They don't live in main branch anymore, only templates there
* CI/CD changes (we won't use @docker-library-bot help anymore. It was nice, thanks @tianon, now even better!)

I'm making this PR to make sure everything follow correctly, and to have a clean base for next concrete changes.
- [x] It seems to to be OK

### DIFF
--- a/library/yourls
+++ b/library/yourls
@@ -1,4 +1,4 @@
-# this file is generated via https://github.com/YOURLS/docker-yourls/blob/0b8dc066336508aeafae001698f970af7ca1ec27/bin/generate-stackbrew-library.sh
+# this file is generated via https://github.com/YOURLS/docker-yourls/blob/f1756127e0b8eebad4f301d4fe6790f6d9794ce2/bin/generate-stackbrew-library.sh
 
 Maintainers: YOURLS <yourls@yourls.org> (@YOURLS),
              Léo Colombaro <git@colombaro.fr> (@LeoColomb)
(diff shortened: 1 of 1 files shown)


library/yourls
```diff
@@ -1,20 +1,20 @@
-# this file is generated via https://github.com/YOURLS/docker-yourls/blob/eded14078d93e7c85b9be1cdc5f522f2d669019c/generate-stackbrew-library.sh
+# this file is generated via https://github.com/YOURLS/docker-yourls/blob/a358b4f51a61732a89347ba35d59c7bc4c65f6e6/bin/generate-stackbrew-library.sh
 
 Maintainers: YOURLS <yourls@yourls.org> (@YOURLS),
              Léo Colombaro <git@colombaro.fr> (@LeoColomb)
 GitRepo: https://github.com/YOURLS/docker-yourls.git
```
tianon, line 6: You'll want to also include `GitFetch:` in here:

```suggestion
GitRepo: https://github.com/YOURLS/docker-yourls.git
GitFetch: refs/heads/main

```

tianon: (Sorry, updated my suggestion to use `refs/heads/dist` appropriately.)

It's also kind of strange to have `dist` so divorced from `main`, since if users browse around the repository from links on https://hub.docker.com/_/yourls, they'll not have any of the other bits from the repository.

LeoColomb: Ah, thanks for the suggestion!

Well, this is my point of view on this divorce:
* Each time a new external contribution is made, the contributor does not understand the way variant images have to be updated. No blame on them, because have distribution files and source templates in the same branch is 
* This makes local maintenance much more straightforward: I don't need a complete bash each time I what to fix a typo.
* `dist` branch does not need to host all the file. I use the assumption if someone want to see any of the other bits, they'll need to go on repo home.
* And last but not least: when accessing dist Dockerfiles from Hub page, the page loaded is the exact commit page and the not the head. So the way for other bits differs very little.

LeoColomb: For example, this is a very usual contribution: https://github.com/YOURLS/docker-yourls/pull/68/files
Only the template and one of the 3 variant has been updates.
Then:
* Not sure which one the contributor used as a reference
* The CI is less relevant, only 1 variant is fully tested under the PR change
* We need to ask the contributor to run the `update.sh` script or do it ourselves.

PR #9838, opened 2021-03-23
Revert "Update redmine"

tianon:
Reverts docker-library/official-images#9837

See https://github.com/docker-library/redmine/pull/230


tianon:
(thanks @J0WI 🤦)

J0WI:
@tianon the update is still important due https://redmine.org/projects/redmine/wiki/Security_Advisories

PR #9848, opened 2021-03-24
Update httpd

tianon:
Changes:

- https://github.com/docker-library/httpd/commit/98fa1e7: Merge pull request https://github.com/docker-library/httpd/pull/173 from J0WI/alpine-3.13
- https://github.com/docker-library/httpd/commit/d36dfa5: Alpine 3.13

### DIFF
--- a/library/httpd
+++ b/library/httpd
@@ -11,5 +11,5 @@ Directory: 2.4
 
 Tags: 2.4.46-alpine, 2.4-alpine, 2-alpine, alpine
 Architectures: amd64, arm32v6, arm32v7, arm64v8, i386, ppc64le, s390x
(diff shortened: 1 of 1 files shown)


library/httpd
```diff
@@ -11,5 +11,5 @@ Directory: 2.4
 
 Tags: 2.4.46-alpine, 2.4-alpine, 2-alpine, alpine
```
tianon, line 12: I wonder if we should add some more specific aliases (even though we only support one version)?

yosifkit: :man_shrugging: we've really only done that for languages and none of our "service" images have Alpine versions in their tags:  (`bash`, `docker` , `haproxy`, `memcached`, `postgres`, `redis`).

(Maybe `bash` needs it?)

PR #9855, opened 2021-03-25
Update CrateDB to 4.4.3

seut:


seut:
@yosifkit Looks like something went wrong with this version on publishing to docker hub.

```
docker pull crate:4.4.3
4.4.3: Pulling from library/crate
no matching manifest for linux/amd64 in the manifest list entries
```

The tag description looks bogus and also `LATEST` doesn't point to `4.4.3`, https://hub.docker.com/_/crate?tab=tags&page=1&ordering=last_updated.

Could someone please have a look?
Thanks!

yosifkit:
The build queue has just been very full with the merging of updates to Alpine, Ubuntu, and Debian for an OpenSSL CVE yesterday, which causes all dependent images to be rebuilt as well.  It seems to be updated now.

seut:
I see, thanks for the update!

PR #9862, opened 2021-03-25
Update Debian, esp. for CVE-2021-3449

tianon:
We'll have another update next week for Debian 10.9. :disappointed:

### DIFF
--- a/library/debian
+++ b/library/debian
@@ -7,31 +7,31 @@ GitRepo: https://github.com/debuerreotype/docker-debian-artifacts.git
 GitCommit: 90a83b6b730031399d45a49809364b07df0d0087
 # https://github.com/debuerreotype/docker-debian-artifacts/tree/dist-amd64
 amd64-GitFetch: refs/heads/dist-amd64
-amd64-GitCommit: 819ea4f3f6feeb849244ca1fa17b6105a7742daa
+amd64-GitCommit: 7c8c3447048f6b2866c2c55311606449ea088630
 # https://github.com/debuerreotype/docker-debian-artifacts/tree/dist-arm32v5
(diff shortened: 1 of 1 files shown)


tianon:
Gah, `sid` includes `libssl` now:

```
 libssl1.1:amd64	1.1.1j-1
```

I've got to regenerate these with https://tracker.debian.org/news/1237247/accepted-openssl-111k-1-source-into-unstable/, which might not have hit snapshot for all the arches we need yet.

tianon:
Confirmed, I need to rebuild and the necessary update hasn't hit snapshot yet.  Converted to Draft for now while we wait. :weary:

tianon:
To clarify: I do intend to get this updated and merged ASAP, but Debian Unstable *is* officially security supported.

- `src:openssl` version `1.1.1k-1` has been uploaded fixing this (https://tracker.debian.org/news/1237247/accepted-openssl-111k-1-source-into-unstable/)
- ... _and_ has built successfully across all the architectures we need (https://buildd.debian.org/status/package.php?p=openssl)
- "all" we need is for a new timestamp on https://snapshot.debian.org/ to appear which includes those builds and I can trigger rebuilds of all these artifacts to pull that update in (which I'm actively monitoring for)

I have converted the pull request to a Draft to prevent it being merged prematurely because the downstream effect of updating Debian (and Ubuntu, and Alpine) is **extremely** massive (which is why I'm sad about the timing of this with respect to the upcoming 10.9), and I thus want to prevent it happening *three* times within 7 days if we can get away with only two.

tianon:
```diff
-libssl1.1:amd64	1.1.1j-1
+libssl1.1:amd64	1.1.1k-1
```

*Now* we're cooking with gasoline. :metal:

PR #9878, opened 2021-03-30
Update Debian, esp. for 10.9

tianon:
This also finally removes the EOL Debian 8 tags.


tianon:
As far as I can tell, here's the full list of images/maintainers still on `debian:jessie`(`-slim`):

- `fsharp`: @ninjarobot @stevedesmond-ca

stevedesmond-ca:
@ninjarobot My assumption is we should just ditch 4.x at this point?

PR #9895, opened 2021-03-31
Haskell: GHC 9.0.1 and sync up on releases.

psftw:
Releasing GHC 8.8.4 to pick up a newer `cabal` tool version before
closing out support of 8.8.x. We also missed the GHC 8.10.3 release, but
would still like to ship it for completeness. A follow-up change will
remove both of these to get us back in line with our two release support
strategy.


psftw:
cc @alistairb thank you! This is the way we get new images published to DockerHub (docs in a separate repo). Feel free to add yourself as a Maintainer to this file as well if you'd like to get involved in this side of things!

tianon:
Fair warning, thanks to recent Debian and Alpine updates, it's gonna be a bit before these are built :sweat_smile:

PR #9896, opened 2021-04-01
Haskell: drop old releases.

psftw:
Follow-up to https://github.com/docker-library/official-images/pull/9895 :tada:


tianon:
LGTM, but needs to wait until https://doi-janky.infosiftr.net/job/multiarch/view/images/view/haskell/job/amd64/job/haskell/74/console is finished before it merges :sweat_smile:

tianon:
Nice, no conflict with #9897 :sunglasses: 

Thanks for following through on the cleanup! :+1:

PR #9901, opened 2021-04-01
Add bullseye images for Rust

sfackler:
We'll switch the default from buster to bullseye in the 1.52.0 release.


tianon:
Putting my Debian hat on, I'd add a warning here that Bullseye is _not_ officially security-supported (yet -- until release), and isn't quite in [full freeze](https://release.debian.org/bullseye/freeze_policy.html#full) yet, so while it's probably a pretty stable base (and I actively run it on several of my host systems), it's still technically a little bit of a moving target. :smile:

sfackler:
Ah thanks - we'll keep the default on buster then until bullseye becomes the new stable.

PR #9914, opened 2021-04-02
[varnish] force workdir

gquintard:
Hi,

This one is probably more of an RFC than a PR as I'm not sure this is the best approach.

The gist of the problem is that `varnishd` uses `/varlib/varnish/$HOSTNAME` as workdir to write logs, metrics and whatnot. So if you have sidecars want to fetch logs or metrics, you need to pass `-n /common/volume` to everyone, in addition of mounting said volume everywhere already (but that can't be helped).

The proposed situation is to detect if the user is running a `varnish*` tool and set `-n /var/lib/varnish` so everyone agrees on the workdir, while still being user-settable.

Question is: is that too much magic?

### DIFF
--- a/library/varnish
+++ b/library/varnish
@@ -1,18 +1,13 @@
-# this file was generated using https://github.com/varnish/docker-varnish/blob/82d7e7a3e0d39bed9f799b9a9455335b27ee0e83/populate.sh
+# this file was generated using https://github.com/varnish/docker-varnish/blob/3b38b0e1c38c1f6324ac8afa156ff73cfbbffcf4/populate.sh
 Maintainers: Guillaume Quintard <guillaume@varni.sh> (@gquintard)
 GitRepo: https://github.com/varnish/docker-varnish.git
 
 Tags: 6.0, 6.0.7-1, 6.0.7, stable
 Architectures: amd64
 Directory: stable/debian
-GitCommit: c0846f15b04cd499a1df1fb62f0693b41a84e636
+GitCommit: 3b38b0e1c38c1f6324ac8afa156ff73cfbbffcf4
 
 Tags: 6.6, 6.6.0-1, 6.6.0, 6, latest, fresh
 Architectures: amd64
 Directory: fresh/debian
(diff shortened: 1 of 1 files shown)


yosifkit:
Ah, this is similar to RabbitMQ that also uses the hostname in their data path.  I don't think any of the other tools needs to know the path, so we had only recommended that users set `--hostname` on their containers.

I think it makes sense to set a default path that doesn't use the hostname.  I guess the drawback is that a user can't get back to path with hostname unless they skip the entrypoint, but that's not new for `varnishd`.

Maybe one of these is simpler/clearer?

```bash
#if echo "$1" | grep -qE '^varnish(d|adm|hist|log|ncsa|reload|stat|top)$'; then

case "$1" in
	varnish*)
		...
		;;
esac

# switch to /bin/bash instead of /bin/sh
if [[ "$1" == varnish* ]]; then
```

tianon:
Also `/varlib/varnish/` looks like a typo and `shift > /dev/null` is odd -- does `shift` actually ever have any output? :sweat_smile:

gquintard:
> Also /varlib/varnish/ looks like a typo shift > /dev/null is odd -- does shift actually ever have any output? sweat_smile

oopsie, good catch. The `shift` thing maaaaaaaaay have been me trying a couple of things and...meh, it didn't hurt.

> I think it makes sense to set a default path that doesn't use the hostname. I guess the drawback is that a user can't get back to path with hostname unless they skip the entrypoint, but that's not new for varnishd.

you actually can with `-n ""`

> Maybe one of these is simpler/clearer?

It is clearer, but I wanted to be very explicit to avoid collateral effect (e.g. catching `varnish_custom_script`). I did have a stab at `switch/case` but that line became stupidly long.

Thanks a lot to you two, you are awesome (it's not said enough)

tianon:
> This one is probably more of an RFC than a PR as I'm not sure this is the best approach.

(Converted this to a draft so we don't accidentally merge it before you're ready for us to.)

gquintard:
all good from me :-)

PR #9922, opened 2021-04-04
Add Zookeeper 3.7

31z4:


tianon:
This looks good, and I'm going to merge it, but I think you should probably also consider removing 3.4 (since it appears to be EOL now). :smile: :+1:

31z4:
Thanks @tianon! I'll probably remove 3.4 soon.

PR #9943, opened 2021-04-08
Updated mageia images

juanluisbaptiste:


yosifkit:
Diff: https://gist.github.com/docker-library-bot/959446b693a86734656c1385a7af76a5

juanluisbaptiste:
Hi @tianon , yes, you are right, I will try to push updated images this week.

juanluisbaptiste:
@tianon We are having issues building the aarch64 images, so this update will take longer than expected.

tianon:
Ouch, thanks for the update! :heart:

I hope you get it figured out. :crossed_fingers:

PR #9961, opened 2021-04-10
Replace Java 15 with 16

keeganwitt:
Merge #9963 first, need to fix 3.0.6 tags then I'll rebase this MR on top of that.


yosifkit:
The builds for https://github.com/docker-library/official-images/pull/9963 have finished and it looks like all the multi-arch tags have updated: https://hub.docker.com/_/groovy?tab=tags&page=1&ordering=last_updated&name=3.0.6.

LGTM

PR #9962, opened 2021-04-10
Gradle 7.0 and replace Java 15 with 16

keeganwitt:


library/gradle
```diff
@@ -21,13 +21,13 @@ Tags: 6.8.3-jre11, 6.8.3-jre11-hotspot, 6.8-jre11, 6.8-jre11-hotspot, jre11, jre
 Architectures: amd64, arm32v7, arm64v8, ppc64le, s390x
 Directory: hotspot/jre11
 
-Tags: 6.8.3-jdk15, 6.8.3-jdk15-hotspot, 6.8-jdk15, 6.8-jdk15-hotspot, jdk15, jdk15-hotspot
+Tags: 6.8.3-jdk16, 6.8.3-jdk16-hotspot, 6.8-jdk16, 6.8-jdk16-hotspot, jdk16, jdk16-hotspot
```
tbroyer, line 24: Shouldn't that read 7.0.0 rather than 6.8.3?

keeganwitt: Oops, yes. Thank you.

library/gradle
```diff
@@ -1,57 +1,57 @@
...
 GitRepo: https://github.com/keeganwitt/docker-gradle.git
-GitCommit: ef27237d21e823cc0e075fff946322b5e551e8c4
+GitCommit: 605a08fb025ad34e8e9d5d2391427886e64fc9f7
 
 
 # Hotspot
 
-Tags: 6.8.3-jdk8, 6.8.3-jdk8-hotspot, 6.8-jdk8, 6.8-jdk8-hotspot, jdk8, jdk8-hotspot, 6.8.3-jdk, 6.8.3-jdk-hotspot, 6.8-jdk, 6.8-jdk-hotspot, jdk, jdk-hotspot, 6.8.3, 6.8.3-hotspot, 6.8, 6.8-hotspot, latest, hotspot
+Tags: 7.0.0-jdk8, 7.0.0-jdk8-hotspot, 7.0-jdk8, 7.0-jdk8-hotspot, jdk8, jdk8-hotspot, 7.0.0-jdk, 7.0.0-jdk-hotspot, 7.0-jdk, 7.0-jdk-hotspot, jdk, jdk-hotspot, 7.0.0, 7.0.0-hotspot, 7.0, 7.0-hotspot, latest, hotspot
 Architectures: amd64, arm32v7, arm64v8, ppc64le, s390x
```
tbroyer, line 9: "Naughty" says you should remove the two ARM architectures from here (and indeed [the base image doesn't support those architectures](https://hub.docker.com/layers/adoptopenjdk/library/adoptopenjdk/8-jdk-hotspot/images/sha256-b0123b7758d74f5ff3b22c4d4fe27df2dac92d02d99af9d3c2461bea9490ebf2?context=explore))
```suggestion
Architectures: amd64, ppc64le, s390x
```

library/gradle
```diff
@@ -1,57 +1,57 @@
...
 Directory: hotspot/jdk8
 
-Tags: 6.8.3-jre8, 6.8.3-jre8-hotspot, 6.8-jre8, 6.8-jre8-hotspot, jre8, jre8-hotspot, 6.8.3-jre, 6.8.3-jre-hotspot, 6.8-jre, 6.8-jre-hotspot, jre, jre-hotspot
+Tags: 7.0.0-jre8, 7.0.0-jre8-hotspot, 7.0-jre8, 7.0-jre8-hotspot, jre8, jre8-hotspot, 7.0.0-jre, 7.0.0-jre-hotspot, 7.0-jre, 7.0-jre-hotspot, jre, jre-hotspot
 Architectures: amd64, arm32v7, arm64v8, ppc64le, s390x
```
tbroyer, line 13: …and [same here](https://hub.docker.com/layers/adoptopenjdk/library/adoptopenjdk/8-jre-hotspot/images/sha256-8702f57b8baae2118780264531051033b971e3c3277779cb7bf75061a813af41?context=explore).
```suggestion
Architectures: amd64, ppc64le, s390x
```

library/gradle
```diff
@@ -1,57 +1,57 @@
...
-Directory: hotspot/jre15
+Directory: hotspot/jre16
 
 
 # OpenJ9
 
-Tags: 6.8.3-jdk8-openj9, 6.8-jdk8-openj9, jdk8-openj9, 6.8.3-jdk-openj9, 6.8-jdk-openj9, jdk-openj9, 6.8.3-openj9, 6.8-openj9, openj9
+Tags: 7.0.0-jdk8-openj9, 7.0-jdk8-openj9, jdk8-openj9, 7.0.0-jdk-openj9, 7.0-jdk-openj9, jdk-openj9, 7.0.0-openj9, 7.0-openj9, openj9
 Architectures: amd64, ppc64le, s390x
```
tbroyer, line 36: Apparently, you could add arm64v8 for all openj9 variants: https://hub.docker.com/layers/adoptopenjdk/library/adoptopenjdk/8-jdk-openj9/images/sha256-605695fedbf637d3425a4b8623a9c4554a73aa9d6c50a311fcef5b350e7aa330?context=explore

tbroyer (changes requested):

library/gradle
keeganwitt: This is a constant struggle, and I don't know what to do about it.  AdoptOpenJDK folks keep adding and removing those architectures, depending on the timing of those binaries being released in relation to when they update the Docker images. So it's always a coin toss as to whether my PR will fail the check or not.

library/gradle
keeganwitt: Ah, OK might as well.

PR #9974, opened 2021-04-12
[varnish] overload the binaries with scripts

gquintard:
This is a follow-up of #9914. The now current approach works well EXCEPT
when people exec into the container with `bash` or whatever since the
entrypoint can't work it magic.

So, the blunt approach is to create a series of trampoline scripts that
will just add `-n /var/lib/varnish` before all arguments. It's not super
elegant, and I'm not a fan of overloading `PATH` but:
- it works transparently
- you can easily bypass the trampoline by using the binary's absolute
  path
- the logic is pretty obvious with `/varnish_cmds/` sitting proudly at
  the root level of the file-system.

I decided against generating the scripts from the `Dockerfile`s to avoid
escape hell, but I can be convinced otherwise if it makes reviews
easier.

### DIFF
--- a/test/tests/varnish/run.sh
+++ b/test/tests/varnish/run.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+set -eo pipefail
+
+dir="$(dirname "$(readlink -f "$BASH_SOURCE")")"
+
+image="$1"
+
+vname="varnish-volume-$RANDOM-$RANDOM"
+trap "docker volume rm $vname > /dev/null" EXIT
+docker volume create --driver local \
+	--opt type=tmpfs \
+	--opt device=tmpfs \
+	--opt o=size=100m \
+	$vname
+
+
+cname="varnish-container-$RANDOM-$RANDOM"
+cid="$(
+	docker run -d \
+		-v $vname:/var/lib/varnish \
+		--name "$cname" \
+		"$image" \
+		varnishd -F -a 0:0 -f /etc/varnish/default.vcl
+)"
+trap "docker rm -vf $cid > /dev/null; docker volume rm $vname > /dev/null" EXIT
+
+sidecar() {
+	docker run --rm -i \
+		--network container:"$cid" \
+		-v $vname:/var/lib/varnish \
+		"$image" \
+		"$@" > /dev/null
+}
+
+sidecar varnishlog -d
+sidecar varnishncsa -d
+sidecar varnishstat -1
+sidecar varnishreload
+sidecar varnishadm ping
--- a/library/varnish
+++ b/library/varnish
@@ -1,13 +1,13 @@
-# this file was generated using https://github.com/varnish/docker-varnish/blob/3b38b0e1c38c1f6324ac8afa156ff73cfbbffcf4/populate.sh
+# this file was generated using https://github.com/varnish/docker-varnish/blob/51a66e74c3a56e4def767d3ad3f6c8ad1a4addaa/populate.sh
 Maintainers: Guillaume Quintard <guillaume@varni.sh> (@gquintard)
 GitRepo: https://github.com/varnish/docker-varnish.git
 
 Tags: 6.0, 6.0.7-1, 6.0.7, stable
 Architectures: amd64
 Directory: stable/debian
-GitCommit: 3b38b0e1c38c1f6324ac8afa156ff73cfbbffcf4
+GitCommit: 51a66e74c3a56e4def767d3ad3f6c8ad1a4addaa
 
 Tags: 6.6, 6.6.0-1, 6.6.0, 6, latest, fresh
 Architectures: amd64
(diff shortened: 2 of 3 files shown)


yosifkit:
I think you don't need the `if` and can just `exec`. Even `/bin/dash` in Debian doesn't fail when `"$@"` is empty.  With something like the following, they could all just be symlinks to a single script

```bash
#!/bin/bash
set -e

script="$(basename $0)"
exec "/usr/bin/$script" -n /var/lib/varnish "$@"
```

tianon:
Also, `/usr/local/bin` is in `PATH` ahead of `/usr/bin`, so you could even do this without the `PATH` addition if you wanted to. :smirk:

gquintard:
indeed, thanks for the reviews!
`varnishd` is in `/usr/sbin/` so needs to be treated on its own, but the symlinks clean the PR nicely (I just generated the scripts and that never crossed my mind...)

force-pushed with all that goodness in

tianon:
:trollface:

```bash
#!/bin/sh -e
bin="$(basename $0)"
[ "$bin" = 'varnishd' ] && bin="/usr/sbin/$bin" || bin="/usr/bin/$bin"
exec "$bin" -n /var/lib/varnish "$@"
```

Or if you wanted to get more generic:

```bash
#!/bin/sh -e
bin="$(basename $0)"
[ -x "/usr/sbin/$bin" ] && bin="/usr/sbin/$bin" || bin="/usr/bin/$bin"
exec "$bin" -n /var/lib/varnish "$@"
```

(Also, `/usr/bin/local` looks like a typo of `/usr/local/bin` :sweat_smile:)

gquintard:
erf, "don't do two things at once" they said, I should have listened.

I'll probably get some flak for the "cryptic" shell scripts, but now I can blame you :-D

gquintard:
ready to go on my side

tianon:
```diff
...
diff --git a/varnish_fresh/scripts/varnishd b/varnish_fresh/scripts/varnishd
new file mode 120000
index 0000000..56da77f
--- /dev/null
+++ b/varnish_fresh/scripts/varnishd
@@ -0,0 +1 @@
+varnishlog
\ No newline at end of file
...
diff --git a/varnish_fresh/scripts/varnishlog b/varnish_fresh/scripts/varnishlog
new file mode 120000
index 0000000..6b2a530
--- /dev/null
+++ b/varnish_fresh/scripts/varnishlog
@@ -0,0 +1 @@
+varnishd
\ No newline at end of file
...
```

spiderman pointing fingers meme

I think something's missing here :thinking:

gquintard:
I'm a bit confused as my tests ran, but something is obviously wrong...

tianon:
How hard would it be to write up a quick/short test for the thing this is fixing, just to be sure it's right? :innocent:

gquintard:
found it, the `populate.sh` script didn't do its job, and we basically were pointing at the wrong commit, which is why my tests came back clean.
I now realize that I can add tests to this repo, so we can test exactly what I'm trying to push, instead of the local stuff.

I'll write something up

gquintard:
ok, I did something, using `mysql-basics` as template, and it seems the tests are being run by github, so it's not completely wrong

test/tests/varnish/run.sh
```diff
@@ -0,0 +1,41 @@
...
+		-v $vname:/var/lib/varnish \
+		--name "$cname" \
+		"$image" \
+		varnishd -F -a 0:0 -f /etc/varnish/default.vcl
+)"
+trap "docker rm -vf $cid > /dev/null; docker volume rm $vname > /dev/null" EXIT
+
+sidecar() {
+	docker run --rm -i \
+		--network host \
```
yosifkit, line 31: To keep things isolated and ensure we don't ever run into port conflicts, we should drop `host` networking and just use `--network container:"$cid"` for the sidecar.

gquintard: nice, I wasn't aware that it was an option! Fixed

test/tests/varnish/run.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+set -eo pipefail
+
+set -x
```
yosifkit, line 4: Some debugging got left in. :eyes: 

Everything else looks good to me :+1:   Thanks!

gquintard: that one was actually intentional, I swear! I figured it wouldn't harm to have verbose/noisy that would save time if something went wrong in the future. Is there a rule about this? If so, I might have to silence the `sidecar` commands as they are bound to be chatty

gquintard: in doubt, I silenced everyone

yosifkit: Sorry for the delay. Yeah, the tests are designed to only be noisy on failure. Thanks for updating!

PR #9979, opened 2021-04-14
bump Kong to 2.4.0

locao:


locao:
There is an outage happening, some requirements could not be retrieved. We will reopen this when everything is back to normal state.

PR #9980, opened 2021-04-14
AdoptOpenJDK: Add back arm arches for builds that are now available.

dinogun:
Remove `arm64v8` for OpenJ9 for now as it is still EA and it is not properly tagged.

### DIFF
--- a/library/adoptopenjdk
+++ b/library/adoptopenjdk
@@ -6,15 +6,15 @@ GitRepo: https://github.com/AdoptOpenJDK/openjdk-docker.git
 #-----------------------------hotspot v8 images---------------------------------
 Tags: 8u282-b08-jdk-hotspot-focal, 8-jdk-hotspot-focal, 8-hotspot-focal
 SharedTags: 8u282-b08-jdk-hotspot, 8-jdk-hotspot, 8-hotspot, 8-jdk, 8
-Architectures: amd64, ppc64le, s390x
-GitCommit: 820829025ee3eb66b5dc9e5fed0ee3171aababe9
+Architectures: amd64, arm32v7, arm64v8, ppc64le, s390x
+GitCommit: 4f1c550c7c1a47a3bab419dcaf8892bdf03c8ad5
 Directory: 8/jdk/ubuntu
 File: Dockerfile.hotspot.releases.full
 
 Tags: 8u282-b08-jdk-hotspot-windowsservercore-1809, 8-jdk-hotspot-windowsservercore-1809, 8-hotspot-windowsservercore-1809
 SharedTags: 8u282-b08-jdk-hotspot-windowsservercore, 8-jdk-hotspot-windowsservercore, 8-hotspot-windowsservercore, 8u282-b08-jdk-hotspot, 8-jdk-hotspot, 8-hotspot, 8-jdk, 8
 Architectures: windows-amd64
-GitCommit: 820829025ee3eb66b5dc9e5fed0ee3171aababe9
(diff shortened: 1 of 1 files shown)


yosifkit:
Looks like the `ltsc2016` images are all failing

```console
Step 4/4 : RUN Write-Host ('Downloading https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u282-b08/OpenJDK8U-jdk_x64_windows_hotspot_8u282b08.msi ...');     curl.exe -LfsSo openjdk.msi https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u282-b08/OpenJDK8U-jdk_x64_windows_hotspot_8u282b08.msi;     Write-Host ('Verifying sha256 (fe137353ffa9f5b02c7783737e73ddf7668a3222b02c5d91abb8b8a2e55871ff) ...');     if ((Get-FileHash openjdk.msi -Algorithm sha256).Hash -ne 'fe137353ffa9f5b02c7783737e73ddf7668a3222b02c5d91abb8b8a2e55871ff') {             Write-Host 'FAILED!';             exit 1;     };         New-Item -ItemType Directory -Path C:\temp | Out-Null;         Write-Host 'Installing using MSI ...';     Start-Process -FilePath "msiexec.exe" -ArgumentList '/i', 'openjdk.msi', '/L*V', 'C:\temp\OpenJDK.log',     '/quiet', 'ADDLOCAL=FeatureEnvironment,FeatureJarFileRunWith,FeatureJavaHome' -Wait -Passthru;     Remove-Item -Path C:\temp -Recurse | Out-Null;     Write-Host 'Removing openjdk.msi ...';     Remove-Item openjdk.msi -Force
 ---> Running in 3bc4d7d666f5
Downloading https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u282-b08/OpenJDK8U-jdk_x64_windows_hotspot_8u282b08.msi ...
curl.exe : The term 'curl.exe' is not recognized as the name of a cmdlet, 
function, script file, or operable program. Check the spelling of the name, or 
if a path was included, verify that the path is correct and try again.
```

tianon:
A few thoughts:

- Instead of using `Invoke-WebRequest` on LTSC 2016 and `curl.exe` on 1809, why not use the built-in `Invoke-WebRequest` on both?  Is there a benefit to using `curl.exe` instead, especially now that it introduces an unnecessary delta?

- I believe with `Start-Process -Wait`, you probably don't need `$proc.WaitForExit`, right?

- In the "FAILED installing MSI!" case, you probably want to include `Get-Content C:\temp\OpenJDK.log` so the build logs include the failure. :sweat_smile:

slide:
> A few thoughts:
> 
> * Instead of using `Invoke-WebRequest` on LTSC 2016 and `curl.exe` on 1809, why not use the built-in `Invoke-WebRequest` on both?  Is there a benefit to using `curl.exe` instead, especially now that it introduces an unnecessary delta?

Invoke-WebRequest can be brittle, curl is much more robust, it has better error handling, better support for more protocols, etc.  

> * I believe with `Start-Process -Wait`, you probably don't need `$proc.WaitForExit`, right?

You would think so, but sometimes the `-Wait` is flaky and doesn't actually work in practice all the time. The `.WaitForExit` is just to make sure the installation actually does finish.

> * In the "FAILED installing MSI!" case, you probably want to include `Get-Content C:\temp\OpenJDK.log` so the build logs include the failure. 😅

That's a good idea, I can add that.

tianon:
Alright, fair enough -- I'll merge this for now then. :+1:

PR #9993, opened 2021-04-16
consul: 1.9.5, 1.8.10, and 1.7.14

rboyer:


tianon:
`iptables` is an interesting addition (that normally won't work inside the container without additional privileges) -- can you elaborate a little on it?  (It's not a blocking concern so I've merged anyhow, but I'm definitely curious. :smile:)

PR #10014, opened 2021-04-20
Update stable nginx to 1.20.0

thresheek:


library/nginx
```diff
@@ -23,22 +23,22 @@ Architectures: arm64v8, arm32v6, arm32v7, ppc64le, s390x, i386, amd64
 GitCommit: 464886ab21ebe4b036ceb36d7557bf491f6d9320
 Directory: mainline/alpine-perl
 
-Tags: 1.18.0, stable, 1.18
+Tags: 1.20.0, stable, 1.18
```
tianon, line 26: I think you meant to make these 1.18 -> 1.20? (in all `Tags:` entries)

tianon: I think https://github.com/nginxinc/docker-nginx/blob/ab8516ed3a212d8a03392567c8f55c570c839e59/generate-stackbrew-library.sh#L7 is probably the culprit :eyes:

thresheek: Arf, mea culpa

thresheek: Fixed in the force-push.

PR #10022, opened 2021-04-21
21.0.0.4 release

crpotter:
Updates to `websphere-liberty` and `open-liberty` for the release of 21.0.0.4.

### DIFF
--- a/library/open-liberty
+++ b/library/open-liberty
@@ -2,7 +2,7 @@ Maintainers: Chris Potter <crpotter@us.ibm.com> (@crpotter),
              Wendy Raschke <wraschke@us.ibm.com> (@wraschke),
              Arthur De Magalhaes <arthurdm@ca.ibm.com> (@arthurdm)
 GitRepo: https://github.com/OpenLiberty/ci.docker.git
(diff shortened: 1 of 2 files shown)


library/open-liberty
```diff
@@ -35,6 +35,26 @@ Directory: releases/latest/full
...
+Directory: releases/21.0.0.4/full
+Architectures: amd64, ppc64le, s390x
+File: Dockerfile.ubuntu.adoptopenjdk8
+
+Tags: 21.0.0.4-full-java11-openj9
+Directory: releases/21.0.0.4/full
+Architectures: amd64, ppc64le, s390x
+File: Dockerfile.ubuntu.adoptopenjdk11
+
 Tags: 21.0.0.3-kernel-slim-java8-openj9
```
tianon, line 58: Shouldn't 21.0.0.3 be removed in both of these since it's superseded by 21.0.0.4?

crpotter: Our quarterly releases are meant to be longer term support releases. Our policy is to maintain N-2 releases actively, so we currently support 20.0.0.12, 21.0.0.3, and the current release of 21.0.0.4.

PR #10060, opened 2021-04-28
Update base image for all 6.x version of Couchbase Server

ceejatec:
We have updated the base image for all 6.x versions (except 6.0.0) to Ubuntu 18.04, and 6.6.2 to Ubuntu 20.04, as those are the newest versions of Ubuntu supported by the corresponding Server versions. This change updates the active 6.x images as well as a one-time update of all earlier 6.x versions. Those earlier versions will be removed from the next PR for library/couchbase.


ceejatec:
I've found some problems with the older versions; retracting this change for now. I'll re-propose soon.

PR #10068, opened 2021-04-28
Friendica Release 2021.04

valvin1:
as @nupplaphil did on each friendica release


yosifkit:
It looks like https://github.com/docker-library/official-images/pull/10067 is more complete. Thanks for the initiative.

PR #10070, opened 2021-04-28
MariaDB 10.6.0 alpha release, 10.1 is EOL

grooverdan:
Hi,

We've done our 10.6.0 alpha release.

I've tested this against some updated tests https://github.com/grooverdan/official-images/blob/mariadb-tests/test/tests/mariadb-extras/run.sh that passes, (though it depends on local mysql for connections, so I do need to clean that up before submitting).

Next major release will be coming soon.

### DIFF
--- a/library/mariadb
+++ b/library/mariadb
@@ -1,8 +1,13 @@
-# this file is generated via https://github.com/docker-library/mariadb/blob/c969655b3c92a1ae3f066e506ab7b32ff4ba4582/generate-stackbrew-library.sh
+# this file is generated via https://github.com/MariaDB/mariadb-docker/blob/b180797cece7d65e23156fd872368e34d3d2a757/generate-stackbrew-library.sh
 
-Maintainers: Tianon Gravi <admwiggin@gmail.com> (@tianon),
-             Joseph Ferguson <yosifkit@gmail.com> (@yosifkit)
-GitRepo: https://github.com/docker-library/mariadb.git
+Maintainers: Daniel Black <daniel@mariadb.org> (@grooverdan),
+             Daniel Bartholomew <dbart@mariadb.com> (@dbart)
+GitRepo: https://github.com/MariaDB/mariadb-docker.git
+
+Tags: 10.6.0-focal, 10.6-focal, alpha-focal, 10.6.0, 10.6, alpha
+Architectures: amd64, arm64v8, ppc64le
+GitCommit: 40c61c4726ab063708a7f4137b81b39773e7dbd1
+Directory: 10.6
 
 Tags: 10.5.9-focal, 10.5-focal, 10-focal, focal, 10.5.9, 10.5, 10, latest
 Architectures: amd64, arm64v8, ppc64le
@@ -23,8 +28,3 @@ Tags: 10.2.37-bionic, 10.2-bionic, 10.2.37, 10.2
 Architectures: amd64, arm64v8, ppc64le
 GitCommit: f0e04a7e0e4b7d83691efc889865da8d6ae5f00e
 Directory: 10.2
-
-Tags: 10.1.48-bionic, 10.1-bionic, 10.1.48, 10.1
-Architectures: amd64, arm64v8, ppc64le
-GitCommit: 0d197a924990af521cded73d04f113b06488844f
-Directory: 10.1


tianon:
```diff
 docker_temp_server_stop() {
-	if ! mysqladmin --defaults-extra-file=<( _mysql_passfile ) shutdown -uroot --socket="${SOCKET}"; then
+	export MYSQL_PWD=$MARIADB_ROOT_PASSWORD
+	if ! mysqladmin shutdown -uroot --socket="${SOCKET}"; then
 		mysql_error "Unable to shut down server."
 	fi
+	unset MYSQL_PWD
 }
```

This could probably be simplified to:

```bash
docker_temp_server_stop() {
	if ! MYSQL_PWD="$MARIADB_ROOT_PASSWORD" mysqladmin shutdown -uroot --socket="${SOCKET}"; then
 		mysql_error "Unable to shut down server."
 	fi
}
```

(Then it's not touching the global environment.)

There's a similar issue with the changes to `docker_process_sql` but unfortunately the fix is less obvious. :sweat_smile:

---

```diff
+		export MARIADB_ROOT_PASSWORD="$(pwgen --numerals --capitalize --symbols --remove-chars="'\\" -1 32)"
```

If `pwgen` fails here for whatever reason, `set -e` won't catch it and `MARIADB_ROOT_PASSWORD` will be set to the empty string, which is very strongly the opposite of your goal (which to be fair, is a problem the old code suffered from as well :see_no_evil:); easy fix:

```bash
		MARIADB_ROOT_PASSWORD="$(pwgen --numerals --capitalize --symbols --remove-chars="'\\" -1 32)"; export MARIADB_ROOT_PASSWORD
```

or separate lines joining with the next statement, perhaps:

```bash
		MARIADB_ROOT_PASSWORD="$(pwgen --numerals --capitalize --symbols --remove-chars="'\\" -1 32)"
		export MARIADB_ROOT_PASSWORD MYSQL_ROOT_PASSWORD=$MARIADB_ROOT_PASSWORD
```

---

```diff
+	local rootPasswordEscaped=$( docker_sql_escape_string_literal "${MARIADB_ROOT_PASSWORD}" )
```

similar problem/solution:

```bash
	local rootPasswordEscaped
	rootPasswordEscaped=$( docker_sql_escape_string_literal "${MARIADB_ROOT_PASSWORD}" )
```

(and again for `userPasswordEscaped`)

I usually style those as `local foo; foo="$(bar)"` :sweat_smile:

tianon:
(You probably also want to update all the other images with the new script/changes, right?)

grooverdan:
> (You probably also want to update all the other images with the new script/changes, right?)

The other images are getting released in maybe the next 1-2 weeks so was going to incorporate those changes then.

The idea that a released specific version tag could change is a little foreign/frighting to me.

> Overall though, LGTM and it's great to have this file updated to match the new maintainership smile +1
> (welcome to you too, @dbart, if you aren't already watching wave smile)
> 
> Do you want us to merge this now and take those as async advice

A merge sounds good to me.

Thanks for the review. I'll look at getting those fixes in soon.

tianon:
> The idea that a released specific version tag could change is a little foreign/frighting to me.

This is an unfortunate limitation of the way containers get consumed such that it's actually a core "feature" of the official images program: https://github.com/docker-library/repo-info/commits/master/repos/mariadb/remote/10.1.48.md (this is the history of times `mariadb:10.1.48` has changed since it was first pushed); see also https://github.com/docker-library/official-images#library-definition-files and/or https://github.com/docker-library/faq#why-does-my-security-scanner-show-that-an-image-has-cves (which touches on the update cadence for several popular base images a bit).

For users who want to make sure what they're consuming "never changes", that's what the content-addressable digest is for (ie, the current `mariadb:10.1.48` can also be consumed as `mariadb@sha256:1205b21b713812a6ba9fb59df206017dc06976fe68e172fbc6d153d7a4e13dfe` or even `mariadb:10.1.48@sha256:1205b21b713812a6ba9fb59df206017dc06976fe68e172fbc6d153d7a4e13dfe` if users want a reminder of where they got that digest from, and they're guaranteed to get the exact same bytes from that every single time they pull it, and that's also why the https://github.com/docker-library/repo-info repository exists in the first place :smile:).

All that being said, we try to keep major/breaking changes to a minimum, but in our experience users will hit those regardless of whether they're in a specific version or whether you do it on a major/minor bump (and will complain loudly), so :shrug:.

PR #10074, opened 2021-04-29
Update ghost

acburdine:


acburdine:
there's two updates here:
1 - update Ghost itself to v4.3.3
2 - revert 4-alpine back to the node 12 base image to fix https://github.com/docker-library/ghost/issues/256

tianon:
I've updated the title and OP to give a little more detail / pointers to the upstream details (hopefully giving users a few more places to find the information about why all the other 4.x tags are gone, because I'm about to go delete them).

acburdine:
Thank you for the quick turnaround on this @tianon !

tianon:
Alright, all the 4.x tags except those pointing to the new (security fixed) 4.3.3 have been removed from https://hub.docker.com/_/ghost.  If you're running Ghost 4.x, please update to 4.3.3 ASAP (and go read [the security advisory](https://github.com/TryGhost/Ghost/security/advisories/GHSA-9fgx-q25h-jxrg) linked above).

PR #10105, opened 2021-05-03
Update EG to 1.16.11

XVincentX:
It also changes the base imge to node 12﻿

### DIFF
--- a/library/express-gateway
+++ b/library/express-gateway
@@ -1,7 +1,7 @@
 Maintainers: The Express-Gateway Team <info@express-gateway.io> (@express_gateway)
 GitRepo: https://github.com/ExpressGateway/docker-express-gateway.git
 
-Tags: 1.x, 1.16.x, 1.16.10, latest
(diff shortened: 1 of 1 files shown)


library/express-gateway
```diff
@@ -1,7 +1,7 @@
 Maintainers: The Express-Gateway Team <info@express-gateway.io> (@express_gateway)
 GitRepo: https://github.com/ExpressGateway/docker-express-gateway.git
 
-Tags: 1.x, 1.16.x, 1.16.10, latest
+Tags: 1.x, 1.16.x, 1.16.11, latest
 Architectures: arm64v8, amd64, i386, ppc64le, s390x
```
yosifkit, line 5: It seems that `node` images don't have `i386` support anymore. (cause of GitHub Actions failure)

XVincentX: Whoa, ok!

yosifkit:
Thanks for the quick responses!

PR #10132, opened 2021-05-06
consul: add 1.10.0-beta2

mikemorris:


yosifkit:
:confused: The older (unchanged) versions are failing to build with `Can't check signature: No public key`; did their signature files get updated and they also need the new key that the `1.10-beta2` image now has?

mikemorris:
Yep, that's exactly it @yosifkit - sorry for not catching that before opening the PR, just pushed a commit that should fix those now.

PR #10144, opened 2021-05-10
Add dart

athomas:
See also https://github.com/docker-library/docs/pull/1944 for the docs.

---

# Checklist for Review

    - https://github.com/dart-lang
    - language stack
    - https://github.com/docker-library/docs/pull/1944


yosifkit:
Is there a way the list of hard-linked libraries could be generated from `ldd` or something? I'd hate to miss one on a new release of dart and it will definitely be necessary once another architecture like `arm64v8` is added.  We don't allow separate Dockerfiles per architecture except for base images like Debian where they are `FROM scratch` and add a compressed `tar` file.

The `ARCH` and `PLATFORM` environment variables are not acceptable. These are both inherent to an image (part of the platform object in the OCI image index). To get the architecture within the image, `dpkg --print-architecture` is the most accurate. OS platform can just be hard-coded, since scraping "linux" from `/etc/os-release` would be longer than just using it and Linux-specific code anyway.

`CHANNEL` is rather generic for an environment variable name. This will leak down to users of the image, so something more specific like `DART_CHANNEL` might be desirable.

tianon:
To clarify, we're not suggesting you need to go multi-architecture right off the bat (although I think it would be really low effort), but that embedding `ARCH` and `PLATFORM` directly in the one place they're each used for now is a more acceptable intermediate step towards that, for very similar reasons to why we're suggesting that `CHANNEL` is a bit generic and you probably want to either embed that one too or choose a more descriptive name like `DART_CHANNEL`.

(I can't say for sure, but I imagine @yosifkit is also having the same :raised_eyebrow: moment I did after reading https://dart.dev/get-dart#linux which clearly claims "Supported architectures: x64, ia32, arm, arm64" but the documented channels for consumption and even this PR are still amd64-only.)

subfuzion:
Thanks @tianon and @yosifkit, I've got a draft PR that we're iterating on [here](https://github.com/dart-lang/dart-docker/pull/21) to address your concerns other than multi-arch support. Will ping when we're ready for another look. Given our timeline for next week, we're currently focused on getting the Debian arch pushed for now and plan to iterate on other architectures immediately after.

subfuzion:
@tianon @yosifkit PTAL now, thanks!

yosifkit (approved):

subfuzion:
Speaking on behalf of our team, thank you for your guidance and helping land this so quickly! You're all awesome and we're grateful. ❤️

PR #10154, opened 2021-05-10
Mariadb release 20210510

dbart:
We've done our MariaDB 2021 Q2 releases: MariaDB 10.5.10, 10.4.19, 10.3.29, and 10.2.38.

### DIFF
--- a/library/mariadb
+++ b/library/mariadb
@@ -6,25 +6,25 @@ GitRepo: https://github.com/MariaDB/mariadb-docker.git
 
 Tags: 10.6.0-focal, 10.6-focal, alpha-focal, 10.6.0, 10.6, alpha
 Architectures: amd64, arm64v8, ppc64le
-GitCommit: 40c61c4726ab063708a7f4137b81b39773e7dbd1
+GitCommit: 8be7abacca5f19d487b137c7d8f9516d07116109
 Directory: 10.6
 
-Tags: 10.5.9-focal, 10.5-focal, 10-focal, focal, 10.5.9, 10.5, 10, latest
+Tags: 10.5.10-focal, 10.5-focal, 10-focal, focal, 10.5.10, 10.5, 10, latest
 Architectures: amd64, arm64v8, ppc64le
-GitCommit: 5d57b119775458cd37994a0f313ea3a29603efbb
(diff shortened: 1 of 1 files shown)


grooverdan:
> I'm a little confused by [this wait loop addition](https://github.com/MariaDB/mariadb-docker/blob/8295244635e584bc1e0fa226e409ab94090d9ffe/docker-entrypoint.sh#L225-L230) in the entrytpoint. We've not seen an issue before, especially since the [`docker_temp_server_start`](https://github.com/MariaDB/mariadb-docker/blob/8295244635e584bc1e0fa226e409ab94090d9ffe/docker-entrypoint.sh#L118-L137) is already looping until the server is ready.

I did hit it once that there where two `--dont-use-mysql-root-password` passed. Also thought the shell checking for the socket was a bit more ruggard with less verbose log generated (as the server side bind is straight before the `accept` connection loop), but I'm changed this to a more minimal approach. I'll look at it again later with more consideration to loops within loops.

> Is there a possible issue with doing `unset` and `export` on `MYSQL_PWD` int he `docker_process_sql` function since it could have been set by the user on their `docker run` or in their `/docker-entrypoint-initdb.d/` scripts?

Ah, missed that https://github.com/docker-library/official-images/pull/10070#issuecomment-828854682 sorry. Does this look better?

grooverdan:
Also reminder - with this merged, if acceptable, https://github.com/docker-library/docs/pull/1889 is ready to go live.

yosifkit:
It all seems fine to me. :+1: 

Just a note that the most recent design of the entrypoint was to allow users to use the provided functions from their `initdb.d` scripts or to source the whole file to customize the startup flow. We've not been able to document the interface but have encouraged users that function that begin with an underscore are "private" but any other would be a stable interface. So any functions added could be used by users and subject to upsetting them if changed in the future.

(see https://github.com/docker-library/mysql/pull/471 and  https://github.com/MariaDB/mariadb-docker/pull/271 and https://docs.google.com/spreadsheets/d/1yYlDjwHw58ng_9eZntI2PAuGGn_WgdatqUZQp5Ra-_s/edit#gid=0)

grooverdan:
Thank you.

PR #10182, opened 2021-05-17
Update clojure tools-deps to 1.10.3.833

cap10morgan:
Upstream release

### DIFF
--- a/library/clojure
+++ b/library/clojure
@@ -2,7 +2,7 @@ Maintainers: Paul Lam <paul@quantisan.com> (@Quantisan),
              Wes Morgan <wesmorgan@icloud.com> (@cap10morgan)
 Architectures: amd64, arm64v8
 GitRepo: https://github.com/Quantisan/docker-clojure.git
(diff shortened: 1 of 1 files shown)


tekacs:
After this update, it appears that `clj` no longer executes in the container:

```clojure
~> docker run --rm -it clojure:openjdk-11-tools-deps bash
root@31e78425c7f3:/tmp# clj
rlwrap: error: Cannot execute /usr/local/lib/clojure/bin/clojure: No such file or directory 
```

I've found this to be the case on multiple systems.

yosifkit:
Turns out the installer has a bug:
>1.10.3.839 (May 12, 2021)
> - Fix Linux installer breakage in 1.10.3.833
>
> \- https://clojure.org/releases/tools#v1.10.3.839

cap10morgan:
I’ll get an update PR up shortly

cap10morgan:
New PR: https://github.com/docker-library/official-images/pull/10186

PR #10183, opened 2021-05-17
Add JDK16

headius:


headius:
https://github.com/jruby/docker-jruby/issues/73

PR #10191, opened 2021-05-18
Update dart (stable) to 2.13.0

athomas:


library/dart
```diff
@@ -1,9 +1,9 @@
...
 GitRepo: https://github.com/dart-lang/dart-docker.git
 GitFetch: refs/heads/main
-GitCommit: 202eef5e34bc0b034766c9e37c5ecb44fd72f9ff
+GitCommit: ee4ca6649d4c4f184a356bcd4c9c1648b6674806
 
-Tags: 2.12.4-sdk, 2.12-sdk, 2-sdk, stable-sdk, sdk, 2.12.4, 2.12, 2, stable, latest
+Tags: 2.13.0-sdk, 2.13-sdk, 2-sdk, stable-sdk, sdk, 2.13.0, 2.13, 2, stable, latest
 Directory: stable/buster
 
 Tags: 2.13.0-211.14.beta-sdk, beta-sdk, 2.13.0-211.14.beta, beta
```
yosifkit, line 9: Is this `beta` still relevant with the release of `2.13.0`?

athomas: Fair point, the head of the beta channel _is_ `2.13.0` right now. That will change in a few weeks when the first 2.14 beta gets created. 

So perhaps this would be more appropriate:
```
Maintainers: Alexander Thomas <athom@google.com> (@athomas), Tony Pujals <tonypujals@google.com> (@tonypujals)
GitRepo: https://github.com/dart-lang/dart-docker.git
GitFetch: refs/heads/main
GitCommit: ee4ca6649d4c4f184a356bcd4c9c1648b6674806

Tags: 2.13.0-sdk, 2.13-sdk, 2-sdk, stable-sdk, sdk, 2.13.0, 2.13, 2, stable, latest, beta-sdk, beta
Directory: stable/buster
```

And then, I'd re-add beta when 2.14 beta becomes available.

For users on `dart:beta`, this means that they don't have to edit their dockerfiles to get the latest stable. This is useful, because packages with "2.13.0 or later" constraints would fail to resolve otherwise.

athomas: I've updated the PR and removed the beta image.

PR #10193, opened 2021-05-18
Fix bugs in influxdb 2.0 entrypoint scripts.

danxmoran:
* Remove arbitrary timeout from initial DB boot.
* Skip booting temporary DB instance if not needed.
* Detect and handle when --help is passed.
* Detect and handle when TLS is enabled in DB config.


yosifkit:
> ```bash
> while kill -0 "${influxd_pid}"; do
> ```

Are there instances of `influxd` that take longer than the current 60 second timeout? I am wary of adding possibly infinite loops to startup scripts, especially db initialization. Maybe it would be better for users if it were configurable so that they can explicitly choose to change the timeout or to opt-in to wait indefinitely.

danxmoran:
@yosifkit unfortunately yes, we've seen a handful of reports from users trying to auto-upgrade a lot of data from 1.x->2.x and getting blocked by the old timeout.

I can add a configurable timeout back into the image.

PR #10205, opened 2021-05-21
mariadb: use own test suite in addition to mysql

grooverdan:
Hi,

I've started a more detailed test suite for MariaDB.

Its significantly more verbose that other tests and I've justified this by making it such that failures are hopefully observable enough to discover the case.

Its undoubtedly not up to the higher level of scripting standard that you expect within the entrypoints.

I've also not used the `. "$dir/../../retry.sh" --tries 30 "echo 'SELECT 1' | mysql"` as my current testing is on podman without the `--link` feature.

Also notably it will fail on 10.6 due to reverting https://github.com/MariaDB/mariadb-docker/commit/6f5d272ca053105901c23fcdc44884907aa4d11d, the next 10.6 release fixes this and will be out in the next day or so.

Any feedback would be appreciated.

### DIFF
--- a/test/config.sh
+++ b/test/config.sh
@@ -26,7 +26,6 @@ testAlias+=(
 	[ubuntu]='debian'
 	[ubuntu-debootstrap]='debian'
 
-	[mariadb]='mysql'
 	[percona]='mysql'
 	[percona:psmdb]='mongo'
 
@@ -115,6 +114,11 @@ imageTests+=(
 	'
 	[logstash]='
 		logstash-basics
+	'
+	[mariadb]='
+		mysql-basics
+		mysql-initdb
+		mariadb-extras
 	'
 	[memcached]='
 		memcached-basics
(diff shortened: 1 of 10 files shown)


grooverdan:
@tianon is this acceptable? This is a moderately comprehensive test case of entrypoint and container behaviour. Passes tests on all 10.X latest tags.

yosifkit:
The number of times this creates a new MariaDB server container in a single test is too much. Our tests usually only test a single bit of functionality and, even altogether, are not meant to be an exhaustive test suite. We'd rather not add ~20 new 
MariaDB specific tests.

---

If you do want to utilize our test runner, it's possible to extend it with a local `config.sh` like Tianon [does](https://github.com/tianon/dockerfiles/blob/ce109068e7e6085861c17c8e63edc7e3800b3e7e/.github/workflows/generate.sh#L46-L52) for his own repo of Dockerfiles. Adding something like that directly to [`ci.yaml`](https://github.com/MariaDB/mariadb-docker/blob/94e05986d626a473f83ccc5e8a3ac2dd54403b8c/.github/workflows/ci.yml#L27) should allow extra tests to be run on PRs to the `MariaDB/mariadb-docker` repo.

grooverdan:
Thanks for the hints. Its probably better to stay in the repository anyway. WIP: https://github.com/MariaDB/mariadb-docker/tree/test

grooverdan:
For the next person that asks https://github.com/MariaDB/mariadb-docker/commit/9b432ef7ae1364e8586c613af7a822c5a5b5d321 is the simplest working form I've come up with so far. Learning jq was a fun exercise.

PR #10217, opened 2021-05-24
library/perl: remove one-time backfill for EOL Perl versions

zakame:
Images from Perl 5.8 to 5.28 are done rebuilding for
https://github.com/Perl/docker-perl/issues/100.


tianon:
We really appreciate you filing this proactively! :smile:

The only snag is that our arm32 builder is currently in the shop for a tune-up, so all the arm32v7 images haven't rebuilt yet -- do you think that's important enough to hold up on this, or do you think the demand for arm32v7 images of these EOL releases is low enough that we can proceed without them?

zakame:
We can probably wait for the arm32 builder to come back :+1:

tianon:
It looks like 5.12 and 5.8 fail (tests) on arm32v7, but otherwise we're done! :+1:

zakame:
5.8 and 5.12 for https://hub.docker.com/r/arm32v7/perl seems good to go now :+1:

zakame:
Thanks @tianon ! :tada:

PR #10220, opened 2021-05-24
MariaDB 10.6.1 beta release 20210524

dbart:
We've done our 10.6.1 beta release.


yosifkit:
These seem like a bug in templating that could just drop in the correct `libjemalloc`, but I don't see it as blocking to this PR.

```bash
if [ bionic = focal ]; ...
if [ focal = focal ]; ...
```

PR #10233, opened 2021-05-25
[ros] add galactic distribution images

ruffsl:
and update ros1 bridge images

### DIFF
--- a/library/ros
+++ b/library/ros
@@ -190,8 +190,30 @@ Directory: ros/foxy/ubuntu/focal/ros-base
 
 Tags: foxy-ros1-bridge, foxy-ros1-bridge-focal
 Architectures: amd64, arm64v8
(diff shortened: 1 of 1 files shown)


yosifkit:
Did you want to also update `dashing` and `eloquent` to address the build failures?

```
E: Version '1.14.10-1*' for 'ros-melodic-ros-comm' was not found

E: Version '1.14.10-1*' for 'ros-melodic-ros-comm' was not found
```

ruffsl:
@yosifkit , we plan to as they will be EOL'd, but we are waiting on final snapshots to be released before archiving the tags:

> Eloquent and Dashing CI failures are expected and currently held by a snapshots.ros.org issue.
> https://github.com/osrf/docker_images/pull/532#issuecomment-848045397

PR #10264, opened 2021-05-29
AlmaLinux: release 8.4 x86_64 version

ezamriy:
Hello!

There is a difference on a package level between AlmaLinux 8.3 and 8.4 images:

```diff
+cracklib-dicts
-ima-evm-utils
+hwdata
+ima-evm-utils0
+libibverbs
+libnl3
+pciutils
+pciutils-libs
+rdma-core
```

Unfortunately, in EL 8.4 those packages became systemd dependencies and removing of any of them leads to something like this:
```
Error:
 Problem: The operation would result in removing the following protected packages: systemd
(try to add '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)
```

I've checked it with RHEL 8.4 and CentOS 8 Stream with the same result.

### DIFF
--- a/library/almalinux
+++ b/library/almalinux
@@ -3,4 +3,4 @@ GitRepo: https://github.com/AlmaLinux/docker-images.git
 
 Tags: latest, 8
 GitFetch: refs/heads/almalinux-8-x86_64
-GitCommit: 1dc91dd3b9f69fd2f570c94441104004a9ef9811
+GitCommit: 3317c1f96704c1cbbab9257ac51a791fb7e28987


ezamriy:
Hello @yosifkit @tianon! Sorry for bothering you, but I have a question: recently we released an AlmaLinux OS 8.4 beta for the aarch64 platform and going to start building Docker images for it.
Is such kind of content acceptable for the official library? Or it would be better to distribute it from our user account until we release a stable version?
I believe it stable enough, but you know, we need some feedback from the community before calling it stable.

yosifkit:
It would be fine from our perspective; it is up to you if you want to expose it to more users through the official-images. You could add a note to the Docker Hub page description to let users know that the arm64v8 images are beta.

I'll get this version merged, feel free to make another PR to add arm64v8.

ezamriy:
Thanks! I think a beta image is much better than no image at all. We tested it on our workflows and it works pretty well, so let's do it.

PR #10270, opened 2021-05-31
[ros] eloquent and melodic/debian stretch now use snapshots repository

mikaelarguedas:
This is to trigger a final build of ROS eloquent images and ROS melodic on debian strretch before retiring them.

Related to https://github.com/docker-library/official-images/pull/8405, https://github.com/osrf/docker_images/pull/529 and https://github.com/osrf/docker_images/pull/430

### DIFF
--- a/library/ros
+++ b/library/ros
@@ -59,7 +59,7 @@ Directory: ros/melodic/ubuntu/bionic/perception
 
 Tags: melodic-ros-core-stretch
 Architectures: amd64, arm64v8
-GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
The build failure on dashing are expected on this PR and are addressed by https://github.com/docker-library/official-images/pull/10272

yosifkit:
Thanks for the clarification. There was some overlap between the two PRs and so I wasn't sure if one replaced the other.  I think they'll merge without a conflict. Next time I'd suggest just doing one PR and then you wouldn't have build failures.

mikaelarguedas:
Thanks, as I wasn't sure the approach in the other PR would need iterations, I went with 2 PRs addressing 2 different issues (one for the key change and one for EOLing the old images)

PR #10272, opened 2021-05-31
[ros] swap key and list file to trigger rebuild

mikaelarguedas:
The ROS [apt repo key expired](https://status.ros.org/incidents/mhqz8yyhrtsg), a new key has been pushed, for apt to work in the docker images, they need to be rebuilt without cache.
To do so we modified our images to swap the creation of the source.list file and the retrieval of the key from the keyserver.

Another more futureprool approach would be to [ADD the key using a fixed URL.](https://github.com/osrf/docker_images/issues/535#issuecomment-850893664) If the key changes the cache would be burst and the images rebuilt, otherwise the cache would be kept. Is this an approach that would be acceptable in official images ?

Relates to https://github.com/osrf/docker_images/issues/535

### DIFF
--- a/library/ros
+++ b/library/ros
@@ -9,7 +9,7 @@ GitRepo: https://github.com/osrf/docker_images.git
 
 Tags: kinetic-ros-core, kinetic-ros-core-xenial
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: df19ab7d5993d3b78a908362cdcd1479a8e78b35
+GitCommit: 11c613986e35a1f36fd0fa18b49173e0c564cf1d
 Directory: ros/kinetic/ubuntu/xenial/ros-core
 
 Tags: kinetic-ros-base, kinetic-ros-base-xenial, kinetic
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
The build failure on eloquent is expected and addressed by https://github.com/docker-library/official-images/pull/10270

yosifkit:
As for `ADD` to download the key, we don't recommend `ADD` for any remote URL. We only recommend using `ADD` for a base image to extract the tar contents.  An explicit PR when keys change or are updated is the expected flow so that everything is transparent to users.

Related: https://github.com/docker-library/faq#openpgp--gnupg-keys-and-verification

mikaelarguedas:
Thanks for the clarification and the link.
Our issue was that the key was updated but had the same fingerprint so we were not sure how to ensure a rebuild (as the dockerfile could stay identical but just needed a rebuild without cache). Is there a way to request such a rebuild ? or do we need to submit a PR with a change to the Dockerfiles such as this one ?

yosifkit:
> Is there a way to request such a rebuild ?

Not really, we make heavy use of Docker build cache.

> or do we need to submit a PR with a change to the Dockerfiles such as this one ?

This type of PR is the one option. These scenarios are quite rare; most software goes out of support before a key is updated. What I'd suggest is ensuring that the key is updated more than 30 days before expiration and then the natural rebuilds that we do *every* 30 days when Debian and Ubuntu images are updated will just pull in the updated key automatically.

PR #10276, opened 2021-06-01
Add Zend Server 2021.0 as latest version

rainviigipuu:


tianon:
Any particular reason you're using `COPY` instead of just `RUN echo 'deb http://repos.zend.com/zend-server/2021.0/deb_ssl1.1 server non-free' > /etc/apt/sources.list.d/zend-server.list`? (The cache semantics of `COPY` are still a bit ... complicated on the edges.)

Also, are any of the older tags now EOL and thus safe to think about removing?  (8.5, 9.1, 2019.0)

(Neither of these are blockers, so I'm going to go ahead and merge.)

rainviigipuu:
> Any particular reason you're using `COPY` instead of just `RUN echo 'deb http://repos.zend.com/zend-server/2021.0/deb_ssl1.1 server non-free' > /etc/apt/sources.list.d/zend-server.list`? (The cache semantics of `COPY` are still a bit ... complicated on the edges.)
>

No particular reason. Just it works and haven't have reason to change that. Is "RUN echo ..." preferred way of doing something like that? 
 
> Also, are any of the older tags now EOL and thus safe to think about removing? (8.5, 9.1, 2019.0)

No. They are all supported as LTS versions.

tianon:
Yeah, doing `RUN` is going to cache more reliably than `COPY` does, but it's fine (just expect us to forget we discussed it and bring it up again next time if you plan to keep it :sweat_smile: )

PR #10283, opened 2021-06-01
Update Corretto-8 for arm64v8 to 8.292.10.2

cliveverghese:
Update Corretto-8 for arm64v8 to 8.292.10.2. 

This addresses issues : https://github.com/corretto/corretto-8/issues/302

### DIFF
--- a/library/amazoncorretto
+++ b/library/amazoncorretto
@@ -4,7 +4,7 @@ Maintainers: Amazon Corretto Team <corretto-team@amazon.com> (@corretto),
              Clive Verghese <verghese@amazon.com> (@cliveverghese)
 GitRepo: https://github.com/corretto/corretto-docker.git
 GitFetch: refs/heads/master
-GitCommit: 4e2c302e3ba78f2d964fb82fce2e2d25a459d4ee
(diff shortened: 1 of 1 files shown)


tianon:
> We strongly recommend that most images create a single `Dockerfile` per entry in the library file that can be used for multiple architectures.

(from https://github.com/docker-library/official-images#multiple-architectures)

Can you please elaborate on why you believe this to be an exception?

cliveverghese:
We had to release a new version for Corretto-8 for `arm64v8`. This was an off-cycle release because the customer had issues that were blocking them in the previous release. 

We do not expect such off cycle releases for a specific architecture alone to common.

yosifkit:
I think we'd rather have any arch-specific logic, like downloading a different release, be addressed within a single Dockerfile; similar to how [`golang`](https://github.com/docker-library/golang/blob/b879b60a7d94128c8fb5aea763cf31772495511d/1.16/buster/Dockerfile#L24-L68) downloads pre-built binaries on some architectures versus the source and compilation on others.

cliveverghese:
Thank you for the feedback @yosifkit , I will update the docker images to reflect this.

cliveverghese:
I have updated the image to identify the `arch` and determine the version to install.

PR #5880, opened 2019-05-08
Update "empty" password handling in "no-hard-coded-passwords" to balk at an empty root password explicitly (CVE-2019-5021)

tianon:
See https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-5021, https://manpages.debian.org/stretch/passwd/passwd.5.en.html, https://manpages.debian.org/stretch/passwd/shadow.5.en.html.

PR #10286, opened 2021-06-01
Eggdrop v1.9.1 *STABLE*

vanosg:


tianon:
```diff
+RUN sed -i -e 's/^root::/root:!:/' /etc/shadow
```

This is a little eyebrow-raising.  I'm guessing this is related to your query on IRC around that bogus CVE?  See https://github.com/docker-library/official-images/pull/5880 where we made this part of our tests, so if any of the images actually had this today, our tests (and thus the CI on PRs) would be failing.

vanosg:
Precisely. Since there appears to be no way to contest it as a CVE (so frustrating, like you said its bogus), I figured I'd  add that to remove any future temptation to claim it against new images. I'm 110% open to other suggestions or changes, to include removal... just don't want some other "enterprising" individual to try it again

tianon:
I used https://cveform.mitre.org/ to dispute some of the previous ones
-- I never got a response, but they did remove/dispute/reject some of
the CVEs, so it might be worth a try.

PR #10300, opened 2021-06-03
Update rakudo-star official image to 2021.04

jmaslak:


tianon:
Welcome! :partying_face: 

cc @moritz (mostly for visibility; I don't think you'll mind having another maintainer listed here :smile:)

moritz:
No objections at all, glad to have @jmaslak on board!

PR #10301, opened 2021-06-03
Update Amazon Linux images

iliana:
Hi all, this will be my last update to this image on behalf of the Amazon Linux team. You'll hear from one of the other maintainers next time!

The AL1 image in this pull request does not actually have any package changes from the current AL1 image, however I went ahead and pushed it all the way through the process as part of handing this off to another team. If you'd like to avoid building something for lack of any changes that's fine, you can just close this pull request -- my maintainer line will get removed in the next PR we make.

Thanks!

### DIFF
--- a/library/amazonlinux
+++ b/library/amazonlinux
@@ -1,11 +1,10 @@
 Maintainers: Amazon Linux <amazon-linux@amazon.com> (@amazonlinux),
-             iliana destroyer of worlds (@iliana),
              Frédérick Lefebvre (@fred-lefebvre),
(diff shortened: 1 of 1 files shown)


tianon:
To be clear, I am going to be merging this with great speed because you're great and both notice and appreciate when we merge your PRs quickly, not because we're anxious to see you go. 🙈

Quite the opposite -- you will be greatly missed. ❤😭

(I've expressed this sentiment elsewhere, but definitely excited for where you're headed and hope it goes very well for you! 👀)

PR #10304, opened 2021-06-04
Update CentOS Linux images for 8.4.2105

bstinsonmhk:
We have a new CentOS Linux Release announced today:
https://lists.centos.org/pipermail/centos-announce/2021-June/048316.html

Here are the latest images.

Signed-off-by: Brian Stinson <bstinson@centosproject.org>

### DIFF
--- a/library/centos
+++ b/library/centos
@@ -2,13 +2,13 @@ Maintainers: The CentOS Project <cloud-ops@centos.org> (@CentOS)
 GitRepo: https://github.com/CentOS/sig-cloud-instance-images.git
 Directory: docker
 
-Tags: latest, centos8, 8, centos8.3.2011, 8.3.2011
+Tags: latest, centos8, 8, centos8.4.2105, 8.4.2105
 GitFetch: refs/heads/CentOS-8-x86_64
-GitCommit: ccd17799397027acf9ee6d660e75b8fce4c852e8
+GitCommit: 607af70702bacc6f46fab2ded055ab23d9113831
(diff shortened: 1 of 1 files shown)


tianon:
It looks like this removed `dhclient` but pulled in `NetworkManager` - is that intentional/desired?

yosifkit:
Friendly ping. 

> this removed `dhclient` but pulled in `NetworkManager` - is that intentional/desired?

thmo:
> It looks like this removed `dhclient` but pulled in `NetworkManager` - is that intentional/desired?

Don't know whether it is generally desired, but it is caused by this change: https://git.centos.org/rpms/dracut/blob/1bf31ccc/f/SOURCES/0133.patch , applied for https://bugzilla.redhat.com/show_bug.cgi?id=1916174.

tianon:
Great catch!  So maybe the kickstart scripts should be adjusted to include `dhclient` directly instead to satisfy that dependency?  Or do you think including NetworkManager here is generally acceptable and expected?

The image size difference seems pretty hefty -- ~209MB before and ~236MB after. :grimacing:

twistdroach:
Hi everyone - thanks for your work on this - any idea when it will be resolved?  Would love to start using 8.4 but need the docker image to switch my product over...

tianon:
Friendly ping @bstinsonmhk :pray:

bstinsonmhk:
This push doesn't affect the sizing much, we have a bit of minimization work to do I think. We at least got rid of NetworkManager

PR #10342, opened 2021-06-10
Backfill ELK 7.13.0

tianon:
Oops.


tianon:
https://github.com/elastic/dockerfiles/compare/v7.12.1...v7.13.0

tianon:
I guess in this case, https://github.com/elastic/dockerfiles/compare/v7.13.1..v7.13.0 is a better diff.

PR #10362, opened 2021-06-15
Gradle 7.1

keeganwitt:


tianon:
(Naughty failure is our good ol' AdoptOpenJDK flip-flopping architectures :see_no_evil:)

keeganwitt:
@tianon Yep. You heard any update on whether that's gonna improve? Wasn't there an issue for that?

TheSRK:
Hello, my project is on Gradle 7.1 but I don't see an official docker image for 7.1 in the docker hub. So, is it mandatory to update now? Or my 7.1 gradle will be compatible with gradle:7.5-jdk8?

yosifkit:
None of the old tags are deleted, so there are still `7.1` based images: [hub.docker.com](https://hub.docker.com/_/gradle/tags?page=1&name=7.1). But they are no longer updated. As far as compatibility, you could check the [release notes](https://gradle.org/releases/) or just trust that they are following their policy: https://docs.gradle.org/current/userguide/feature_lifecycle.html#backwards_compatibility

PR #10366, opened 2021-06-15
alpine: add 3.14.0 and remove EOL v3.10

ncopa:


yosifkit:
Since Alpine 3.10 is being removed, alert to affected image maintainers who still have an image based upon it: Haxe (@andyli) and Notary (@justincormack). Please update to a new Alpine release or remove the unsupported images.

PR #10376, opened 2021-06-16
Update postgres

yosifkit:
Changes:

- https://github.com/docker-library/postgres/commit/0ecab40: Merge pull request https://github.com/docker-library/postgres/pull/852 from infosiftr/jq-template
- https://github.com/docker-library/postgres/commit/517c64f: Add initial jq-based templating engine
- https://github.com/docker-library/postgres/commit/55001e0: Merge pull request https://github.com/docker-library/postgres/pull/861 from J0WI/alpine-3.14
- https://github.com/docker-library/postgres/commit/718c12a: Alpine 3.14

### DIFF
--- a/library/postgres
+++ b/library/postgres
@@ -1,55 +1,70 @@
-# this file is generated via https://github.com/docker-library/postgres/blob/dba8ec0bf97b220682d2cfe417e5d4f367df8a92/generate-stackbrew-library.sh
+# this file is generated via https://github.com/docker-library/postgres/blob/517c64f87e6661366b415df3f2273c76cea428b0/generate-stackbrew-library.sh
 
 Maintainers: Tianon Gravi <admwiggin@gmail.com> (@tianon),
              Joseph Ferguson <yosifkit@gmail.com> (@yosifkit)
 GitRepo: https://github.com/docker-library/postgres.git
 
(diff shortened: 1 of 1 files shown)


yosifkit:
For reference: 

> the bare `9`, `10`, and `11` aliases on version that are still on Stretch will stay on Stretch until it becomes EOL, at which point we no longer plan to update them (and they'll be dropped from the "supported" list completely, leaving only "10-buster" etc).
>
>\- https://github.com/docker-library/postgres/issues/582#issuecomment-850595817

So `postgres:9`, `postgres:10`. and `postgres:11` will always be Debian Stretch (and will stop being updated when that version is end of life or Stretch is end of life).

----


Summary of https://github.com/docker-library/postgres/issues/582: When upgrading from `9` (or `10` or `11`)  which is based on Debian Stretch to the same version but based on Debian Buster (`9-buster`), It should be enough to ensure you have a full backup (for safety) and then do a `postgres reindexdb --all` in the new container.

Very rough untested steps
1. make a postgres backup
2. stop `postgres:9` container
3. start new `postgres:9-buster` container using same data directory
4. run `postgres reindexdb --all` in the new container (perhaps via `docker exec`)
5. :tada:

PR #10381, opened 2021-06-16
Add Consul v1.10.0-rc

freddygv:


library/consul
```diff
@@ -1,9 +1,9 @@
 Maintainers: Consul Team <consul@hashicorp.com> (@hashicorp/consul)
 
-Tags: 1.10.0-beta4, 1.10.0-beta
+Tags: 1.10.0-rc, 1.10.0-beta
```
yosifkit, line 3: Just to double check that keeping `1.10.0-rc` tagged as `1.10.0-beta` is intentional

mikemorris: Yes, we're intentionally keeping the `beta` channel tag for this last prerelease - we'll be removing the beta tag when we cut a GA release. Thanks for checking!

PR #10387, opened 2021-06-17
caddy: v2.4.3 release

hairyhenderson:
https://github.com/caddyserver/caddy/releases/tag/v2.4.3

### DIFF
--- a/library/caddy
+++ b/library/caddy
@@ -1,97 +1,51 @@
 # this file is generated with gomplate:
 # template: https://github.com/caddyserver/caddy-docker/blob/c74bdca53a1efd5195a9c35005e5515afb5feeff/stackbrew.tmpl
-# config context: https://github.com/caddyserver/caddy-docker/blob/977c552bbe252398811886356ea00d546ecd4f1e/stackbrew-config.yaml
+# config context: https://github.com/caddyserver/caddy-docker/blob/e01e0a10625f95d031411215378b5f790047d990/stackbrew-config.yaml
 Maintainers: Dave Henderson (@hairyhenderson)
 
-Tags: 2.3.0-alpine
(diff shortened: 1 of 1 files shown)


hairyhenderson:
Hmm... CI's failing with:

```

 - caddy:2.3.0-builder-alpine (FROM golang:1.15-alpine3.12) -- completely unsupported base!
```

My instinct is that the best way to work around this is simply to stop building the 2.3.0 image. @mholt what do you think?

mholt:
Sure. I don't know what that means, but we have no need for 2.3.0 anymore.

hairyhenderson:
> I don't know what that means

I _think_ it's because of the `alpine-3.12` flavour - with 3.14 out now it's probably explicitly deprecated. But that's just a gut feeling...

yosifkit:
> > I don't know what that means
> 
> I _think_ it's because of the `alpine-3.12` flavour - with 3.14 out now it's probably explicitly deprecated. But that's just a gut feeling...

Correct, for our language images, we try to limit the number of OS versions they are based upon (usually max of two per distribution). So when a new OS release comes out, the older one is usually dropped.

> we have no need for 2.3.0 anymore

Feel free to drop versions that you no longer support. :+1:  That way they'll also drop out of the "Supported Tags" list that appears of the Docker Hub (https://github.com/docker-library/docs/tree/7d084e9d284243ad099a90ef2fbda09c9ad1e505/caddy#supported-tags-and-respective-dockerfile-links).

hairyhenderson:
_Finally_ found some time to get back to this 😓

Ok - I've removed the `:2.3.0` tag so hopefully it should go through now 😉

mholt:
You're awesome @hairyhenderson , thanks so much.

PR #10408, opened 2021-06-23
Update Monica to v3.1.1

asbiin:
Update monica to [v3.1.1](https://github.com/monicahq/monica/releases/tag/v3.1.1)
Also
- make entrypoint compatible with mysql8
- add tests for mysql8 and mariadb10

### DIFF
--- a/library/monica
+++ b/library/monica
@@ -2,17 +2,17 @@
 Maintainers: Alexis Saettler <alexis@saettler.org> (@asbiin)
 GitRepo: https://github.com/monicahq/docker.git
 
-Tags: 3.1.0-apache, 3.1-apache, 3-apache, apache, 3.1.0, 3.1, 3, latest
+Tags: 3.1.1-apache, 3.1-apache, 3-apache, apache, 3.1.1, 3.1, 3, latest
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
 Directory: apache
-GitCommit: 8118ac0f51564c16d68131dae301b7151d323a39
+GitCommit: 00a09024b1012db35f96af9c379829663b814853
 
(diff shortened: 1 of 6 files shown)


yosifkit:
From sks-keyservers.net:
> **Update 2021-06-21**: Due to even more GDPR takedown requests, the DNS records for the pool will no longer be provided at all.

So, my only note is that you'll likely want to drop all SKS keyservers at some point. They will not cause a problem for our builder servers and a few extra loops there wouldn't break a local build.

> [we] hijack DNS for our builds to point that at a local instance of [github.com/tianon/pgp-happy-eyeballs](https://github.com/tianon/pgp-happy-eyeballs), which is a project that takes incoming HTTP GET requests for keys and in turn forwards them out to several keyservers at once, returning back to the client the fastest successful response
>
>\- https://github.com/docker-library/faq/tree/aaadb744f06bd101c65a5d5988e21cf10c9a2f1f#openpgp--gnupg-keys-and-verification

yosifkit (approved):

asbiin:
@yosifkit OK so you recommend keeping only one sks server `hkps://keys.openpgp.org`? I made this change because the previous did not worked anymore ...

yosifkit:
`keys.openpgp.org` only works if the key owner has verified their identity with them and consents to distribution of the "identity information". `gpg` doesn't work when the response only contains "non-identity information". (https://keys.openpgp.org/about)

The loop is fine (and helps someone build locally if they desire) and these URLs all still work.
```
keyserver.ubuntu.com
hkp://keyserver.ubuntu.com:80
pgp.mit.edu
```

asbiin:
I can still reduce the list to 1 if this one is reliable ;)
Anyway, I'll also update this for Monica 3.1.1 which contains a fix.

PR #10450, opened 2021-06-29
[Swift] Add Swift 5.4.2 release images

shahmishal:


yosifkit:
(failure is just `rm` & `gpg` racing)

PR #10473, opened 2021-07-02
Update photon image

gerrit-photon:
Update photon image

### DIFF
--- a/library/photon
+++ b/library/photon
@@ -4,24 +4,24 @@ Maintainers: Fabio Rapposelli <fabio@vmware.com> (@frapposelli),
 GitRepo: https://github.com/vmware/photon-docker-image.git
 Directory: docker
 
-Tags: 4.0, 4.0-20210625, latest
+Tags: 4.0, 4.0-20210702, latest
 Architectures: amd64, arm64v8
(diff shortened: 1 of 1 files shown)


tianon:
Are `curl` and all those `ca-certificates` files being added to the base image intentionally?

yosifkit:
Friendly ping: @YustasSwamp or @frapposelli

> Are `curl` and all those `ca-certificates` files being added to the base image intentionally?

frapposelli:
I believe it was intentional but I'll let @YustasSwamp reply as I'm no longer in charge of the photon image.

YustasSwamp:
Hi @yosifkit and @tianon. Thanks for the attention to these extra files.

There was a regression, where curl and certs were removed from photon docker image accidentally:
https://github.com/docker-library/official-images/pull/10399
Tags photon:3.0-20210621 and photon:3.0-20210625 are affected by this regression.

Curl binaries and certificates are supposed to be in the image.
We intentionally restored it back by:
https://github.com/vmware/photon/commit/591c9847dffb25b62aae9f0521f9ca5436dda25d

Thanks!

PR #10477, opened 2021-07-03
Update develop to 1.9.2rc1

vanosg:
- Remove 1.8 version
- Update 1.9 to 1.9.2
- Remove "heavy" version

### DIFF
--- a/library/eggdrop
+++ b/library/eggdrop
@@ -3,7 +3,7 @@ GitRepo: https://github.com/eggheads/eggdrop-docker.git
 
 Tags: develop
 Architectures: arm32v6, arm64v8, amd64
-GitCommit: 8d3bb9a0a2f35b45d50b82e9c7dd7a5af66e4c9a
+GitCommit: f067a3393434028c1f1bf785a9c4ca062d477acc
 Directory: develop
 
 Tags: 1.8, 1.8.4
@@ -15,8 +15,3 @@ Tags: 1.9, 1.9.1, stable, latest
 Architectures: amd64, arm32v6, arm64v8
 GitCommit: 5444df31406ef15da9fa7e6b742d093179f66d60
 Directory: 1.9
-
-Tags: 1.9-libs, 1.9.1-libs
-Architectures: amd64, arm32v6, arm64v8
-GitCommit: 5444df31406ef15da9fa7e6b742d093179f66d60
-Directory: 1.9-libs
--- a/library/eggdrop
+++ b/library/eggdrop
@@ -6,11 +6,6 @@ Architectures: arm32v6, arm64v8, amd64
 GitCommit: f067a3393434028c1f1bf785a9c4ca062d477acc
 Directory: develop
 
-Tags: 1.8, 1.8.4
-Architectures: amd64
-GitCommit: 14411e45f599536a86d9f023c0fa09f3dd2f5454
-Directory: 1.8
-
 Tags: 1.9, 1.9.1, stable, latest
 Architectures: amd64, arm32v6, arm64v8
 GitCommit: 5444df31406ef15da9fa7e6b742d093179f66d60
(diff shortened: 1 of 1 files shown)


tianon:
Hmm, this new variant appears to be strictly additive?  Here's the approach I would suggest:

```dockerfile
FROM eggdrop:1.9.1

RUN apk add --no-cache sqlite-tcl mariadb-connector-c-dev tcl-tls \
# TODO tcl-lib -- mixing Alpine's "edge" with a released version does *not* go well in our experience 🙈
# do you need this full list just for building mysqltcl?  could probably go smaller 👀
  && apk add --no-cache --virtual egg-deps tcl-dev wget ca-certificates make tar gpgme build-base openssl-dev \
  && wget http://www.xdobry.de/mysqltcl/mysqltcl-3.052.tar.gz \
# really wish there could be some verification here, but upstream doesn't publish anything 😞 (given the release's age though, maybe it's reasonable to just calculate the sha256 of the tarball ourselves and embed it?)
  && tar -zxvf mysqltcl-3.052.tar.gz \
  && rm mysqltcl-3.052.tar.gz \
    && ( cd mysqltcl-3.052 \
    && ./configure \
    && make \
    && make install ) \
  && rm -rf mysqltcl-3.052 \
  && apk del --no-network egg-deps
```

vanosg:
Sorry, I don't know why I didn't explain this better in the PR. The intent of this image in indeed purely additive, to add commonly used Tcl libraries that Eggdrop scripts often require. They are mysqltcl for interacting with mysql databases, sqlite-tcl for interacting with sqlite databases,  tcl-tls for https/TLS connections, and tcl-lib for working with json and other miscellaneous Tcl libraries. (mariadb-connector-c-dev is the smallest package I could find that gives me mysql.h, required to be present at runtime by the mysqltcl library)

_"mixing Alpine's "edge" with a released version does *not* go well in our experience"_ << the package only exists in edge, not main (that I can see, at least). Do you prefer keeping it this way, or compiling?

_"do you need this full list just for building mysqltcl?  could probably go smaller"_ << assuming this references the line above it, these packages are all intended to remain as part of the image to extend the Tcl libraries available, required by advanced scripts

-"really wish there could be some verification here, but upstream doesn't publish anything"_ << I bow down to your wisdom, let me know how you would like to proceed. I agree, it hasn't changed in about a decade I think, so unless you tell me not to, I'll go ahead and add that check in tomorrow or so.

Thanks for the review, I look forward to your responses

tianon:
> _"mixing Alpine's "edge" with a released version does _not_ go well in our experience"_ << the package only exists in edge, not main (that I can see, at least). Do you prefer keeping it this way, or compiling?

Sadness :disappointed:

Yeah, I suppose compiling is our best bet unless you're comfortable rolling the dice on the ability for this build to continue to succeed in the future (usually it's "fine" initially but the dependencies, etc drift over time and become incompatible eventually).

> _"do you need this full list just for building mysqltcl? could probably go smaller"_ << assuming this references the line above it, these packages are all intended to remain as part of the image to extend the Tcl libraries available, required by advanced scripts

Ahhh, then we should ditch `--virtual` entirely and just have one `apk add` entirely :sweat_smile:

> _"really wish there could be some verification here, but upstream doesn't publish anything"_ << I bow down to your wisdom, let me know how you would like to proceed. I agree, it hasn't changed in about a decade I think, so unless you tell me not to, I'll go ahead and add that check in tomorrow or so.

Alright, here's a tweaked version that "just" needs `tcllib`:

```dockerfile
FROM eggdrop:1.9.1

RUN apk add --no-cache build-base ca-certificates gpgme make mariadb-connector-c-dev openssl-dev sqlite-tcl tar tcl-dev tcl-tls wget \
  && wget http://www.xdobry.de/mysqltcl/mysqltcl-3.052.tar.gz \
  && echo '5b6e04430b80fd4af54599551503bae681232be0bae3c55c1a93adeb66702007 *mysqltcl-3.052.tar.gz' | sha256sum -c - \
  && tar -zxvf mysqltcl-3.052.tar.gz \
  && rm mysqltcl-3.052.tar.gz \
    && ( cd mysqltcl-3.052 \
    && ./configure \
    && make \
    && make install ) \
  && rm -rf mysqltcl-3.052
```

PR #10502, opened 2021-07-08
Add Debian riscv64 builds

tianon:
The most relevant bit to note here is that `riscv64` is still a "Debian Ports" architecture, so unfortunately only Debian Unstable ("Sid") is supported.  However, it is in extremely good shape (especially right now while everything is deeply frozen for the imminent Bullseye release), and is impressively well-supported in spite of that.

However, this will unfortunately be a limiting factor on the downstream effect of adding this support (in other words, not many other official images are likely to gain the architecture), since Debian Unstable is *not* a great (stable) base for running other software on top of over time (especially after that imminent Bullseye release and Unstable gets to freely earn its name once again).

See https://www.debian.org/ports/ and https://ports.debian.org/ for more information about the Debian Ports project in general, and the following for information about ports becoming officially supported "release architectures" (lots of room for contribution here!):

- https://ftp-master.debian.org/archive-criteria.html
- https://dsa.debian.org/ports/hardware-requirements/
- https://release.debian.org/testing/arch_policy.html
- https://wiki.debian.org/PortsDocs/New

I personally also think https://buildd.debian.org/stats/ is extremely interesting -- all the release architectures are above 97% of the archive, and `riscv64` is not far behind (appears closer to 95% ATM).  Major kudos to everyone who's been involved in getting this port to where it is today!

### DIFF
--- a/library/debian
+++ b/library/debian
@@ -4,7 +4,7 @@
 Maintainers: Tianon Gravi <tianon@debian.org> (@tianon),
              Paul Tagliamonte <paultag@debian.org> (@paultag)
 GitRepo: https://github.com/debuerreotype/docker-debian-artifacts.git
-GitCommit: 90a83b6b730031399d45a49809364b07df0d0087
+GitCommit: 9c1a3db18852520a8b72a4795c7b197bcc93e8a0
 # https://github.com/debuerreotype/docker-debian-artifacts/tree/dist-amd64
 amd64-GitFetch: refs/heads/dist-amd64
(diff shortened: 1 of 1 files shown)


yosifkit (approved):

tianon:
Anyone looking for more places to contribute to `riscv64` might have fun going through https://udd.debian.org/cgi-bin/ftbfs.cgi?arch=riscv64 (there's even some container-related things there :eyes:).

PR #10543, opened 2021-07-15
Update Lightstreamer to version 7.2.0

gfinocchiaro:


yosifkit:
The build failures are due to our hijacking of dns for common gpg servers (like [`keyserver.ubuntu.com` and `pgp.mit.edu`](https://github.com/tianon/pgp-happy-eyeballs/blob/0f592b7ae20edead4a3c67a8780991d03e27a862/hack-my-builds.sh#L70)) to provided a more reliable source of keys by sending the request to multiple servers and responding with the fastest success.  As a consequence, we cannot handle builds with `hkps` to any of these servers (since they are TLS).


Ref: https://github.com/docker-library/faq/tree/6138d05aabf61563606d86f98d0ccbd99f162b33#openpgp--gnupg-keys-and-verification

gfinocchiaro:
I'm sorry, but not sure to understand. In the doc you linked I read:

"_The way we typically recommend image maintainers fetch those public keys to verify said artifacts is via gpg --batch --keyserver **hkps**://keys.openpgp.org --recv-keys XXXXX (where XXXXX gets replaced with the full key fingerprint, as in 97FC712E4C024BBEA48A61ED3A5CA953F73C700D)_".

So it looks like you do handle builds with `hkps`: what am I missing?

tianon:
Yes, `hkps://keys.openpgp.org` is fully supported and honestly recommended, but it requires the additional step of having they key/email owner verify their email on https://keys.openpgp.org/ -- we do not support `hkps` on `keyserver.ubuntu.com` because the added value is minimal (since the fingerprint gets cryptographically verified after being fetched, and omitting the `hkps` allows us to redirect the request to a collection of servers transparently instead of just one).

PR #10554, opened 2021-07-17
Friendia - Fix SMTP settings

nupplaphil:


tianon:
`sendmail_path` defaults to `/usr/sbin/sendmail -t -i` -- isn't moving this just masking a bug further down the stack?  For example, if something else were to get installed that provides `/usr/sbin/sendmail` (or if `msmtp` were to be updated to update alternatives for `sendmail`, which I'm honestly really surprised it doesn't), then whatever is happening in https://github.com/friendica/docker/issues/156 will come right back, right?

From my reading of https://github.com/friendica/docker/issues/156, it appears the underlying issue there was that `smtp` was set to `false`, which leads to https://github.com/friendica/friendica-addons/blob/4ff458ecf279d2002e51174679f60155131ad50c/phpmailer/phpmailer.php#L44-L66 not doing anything SMTP-related, so `PHPMailer` ends up using PHP's `mail()` instead which (correctly) shells out to `sendmail_path`?

I'm also very, very confused by the `kill -USR2 1` -- at the point this script is running, PHP isn't yet, right?

tianon:
Friendly ping?

nupplaphil:
The change chains the successful configuration of `msmtp ` to it's activation at `sendmail.ini`.

I tried to avoid the situation where `msmtp` isn't fully configured ant therefore this error occurs:
`msmtp: account default not found: no configuration file available,`

This could happen when phpmailer is used an therefore the successful configuration of `msmtp` doesn't matter.

I searched for the availability to reload the PHP configuration (`sendmail.ini`) after it's restart. I found this specific kill-command `kill -USR2 1` for it :-)

tianon:
I'm still confused -- this only shows up when the system is misconfigured and attempts to use `mail()`, so all this really does is trade `msmtp: account default not found: no configuration file available` for something like `sendmail: command not found` when the system is misconfigured in that way?

nupplaphil:
@tianon thx for the input, I have to take some time to think about what you said (and I'm currently a little bit too busy for it :X ).. But to not block the upcoming releases, I'm closing this PR. I reverted this commit at https://github.com/friendica/docker/pull/160 , so the upcoming PR for the new releases will be without it :-)

PR #10574, opened 2021-07-20
Update amazoncorretto with 8u302 and 11.0.12

benty-amzn:


library/amazoncorretto
```diff
@@ -6,23 +6,23 @@ GitRepo: https://github.com/corretto/corretto-docker.git
 GitFetch: refs/heads/master
 GitCommit: 88df29474df6fc3f3f19daa8c5515d934f706cd0
```
yosifkit, line 7: This line is missing the bump to the new commit. :wink:

It looks like it needs to be pointing to https://github.com/corretto/corretto-docker/commit/cb98621700a830682c8b3a50d74673a4838e37b2

benty-amzn: Good catch, PR updated.

PR #10584, opened 2021-07-22
Release of GeoNetwork v3.10.7 and v3.12.1

juanluisrp:


yosifkit:
I am a bit confused by the addition of a `SHELL` of `bash` with `pipefail`. I'm not certain what failure case this is trying to catch since the only pipe is an `echo | md5sum`. This will be inherited (likely unexpectedly) by any users building an image `FROM geonetwork`.

juanluisrp:
I was just trying to remove [this warning](
https://github.com/hadolint/hadolint/wiki/DL4006), but I can revert the
`SHELL` command if you think it isn't needed.

tianon:
Yeah, it doesn't appear to be necessary, and it does have potential for
breaking users since it persists and is inherited.

juanluisrp:
I've removed the `SHELL` with `pipefail` from the `Dockerfile` files

PR #10599, opened 2021-07-23
Update photon image

gerrit-photon:
Update photon image

### DIFF
--- a/library/photon
+++ b/library/photon
@@ -4,24 +4,24 @@ Maintainers: Fabio Rapposelli <fabio@vmware.com> (@frapposelli),
 GitRepo: https://github.com/vmware/photon-docker-image.git
 Directory: docker
 
-Tags: 4.0, 4.0-20210716, latest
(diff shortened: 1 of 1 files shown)


tianon:
Just to double check, Python 3.5 was added to 1.0 here on purpose, right?

yosifkit:
> Just to double check, Python 3.5 was added to 1.0 here on purpose, right?

Ping @frapposelli or @YustasSwamp

YustasSwamp:
Hi @tianon and @yosifkit thanks for the catch!
This is not on purpose. Please hold on with this change.

YustasSwamp:
We are going to stop automatic image push, until we fix extra stuff that got added. Will update you once it's ready. Please ignore above commits.

YustasSwamp:
Hi @tianon and @yosifkit , we fixed content of our images. Please consider latest commit (1 hour ago) for merging. Thanks.

PR #10601, opened 2021-07-23
chore: upgrade kapa to 1.6.1

docmerlin:


prashanthjbabu:
@yosifkit  we're waiting on kapacitor 1.6.1 image for our deployments . Can you please prioritize this?

tianon:
Now that the key loop only has one key and one server, it could definitely be simplified (in a future PR). :sweat_smile:

library/kapacitor
```diff
@@ -1,7 +1,7 @@
 Maintainers: Jonathan A. Sternberg <jonathan@influxdata.com> (@jsternberg),
              j. Emrys Landivar <landivar@gmail.com> (@docmerlin)
 GitRepo: git://github.com/influxdata/influxdata-docker
-GitCommit: 63fe0684150d96916514a84cf5978c3512dc15fa
+GitCommit: 0e03b615c48d7f089fff7bb2aa2e7d65e83bd2c2
 
 Tags: 1.4, 1.4.1
```
tianon, line 6: Just want to double check that 1.4.x is still actively supported (and not EOL now that 1.6.x is released).

library/kapacitor
```diff
@@ -10,9 +10,17 @@ Directory: kapacitor/1.4
...
+Tags: 1.5, 1.5.9
 Architectures: amd64, arm32v7, arm64v8
 Directory: kapacitor/1.5
 
-Tags: 1.5-alpine, 1.5.9-alpine, alpine
+Tags: 1.5-alpine, 1.5.9-alpine
 Directory: kapacitor/1.5/alpine
+
+Tags: 1.6, 1.6.1, latest
+Architectures: amd64, arm32v7, arm64v8
```
tianon, line 21: This should probably be minus `arm32v7`, right?  (looks like `armhf` was dropped from the relevant `case` statement in this version)

docmerlin: thanks.

tianon:
Friendly ping?

docmerlin:
> Friendly ping?

sorry, I was out on vacation last week.  I have made the changes suggested.  Thanks.

PR #10605, opened 2021-07-24
[Haskell] Bump Haskell to use stack 2.7.3

AlistairB:
Hi :wave: , just a minor dep bump for Haskell - https://docs.haskellstack.org/en/stable/ChangeLog/#v273 

Oh, this also includes some [dockerfile improvements](https://github.com/haskell/docker-haskell/pull/39) (feedback welcome), but that should not impact users much.


tianon:
LGTM, thanks! :+1:

PR #10612, opened 2021-07-26
Remove fsharp config file

stevedesmond-ca:
Per [the fsharp image repo](https://github.com/fsprojects/docker-fsharp), the custom Docker image is no longer maintained, replaced with the official Microsoft-maintained .NET images.

This PR removes the fsharp config file, which may or may not be the right move (not knowing how Docker Hub's infra works behind the scenes) but figured it could at least start the conversation!


tianon:
We're sorry to see you go :disappointed:

This needs to start with a PR adding a `deprecated.md` file in https://github.com/docker-library/docs (I think `opensuse` is probably a pretty decent example for the expected format) so that the image description gets updated to let users know where to go.

(cc @ninjarobot)

stevedesmond-ca:
https://github.com/docker-library/docs/pull/2000 has now been merged. Are there any additional/next steps?

tianon:
Nope, this should be good to go now.

PR #10618, opened 2021-07-27
Update CrateDB to 4.5.5 and 4.6.1

mkleen:


library/crate
```diff
@@ -4,10 +4,13 @@ Maintainers: Mathias Fußenegger <mathias@crate.io> (@mfussenegger),
...
+Architectures: amd64, arm64v8
+GitCommit: 64df3fb9fa57d41a3c165d555ab903dcc04a4cf2
 
-Tags: 4.5.4, 4.5, latest
+Tags: 4.5.5, 4.5
 Architectures: amd64, arm64v8
-GitCommit: bdb04c7c7ca2b391d2079738f4b06dd62edf42ee
+GitCommit: 53f8d6336816b88f09d831c073cab749762c7be6
 
 Tags: 4.4.3, 4.4
```
tianon, line 15: Any particular reason to keep so many older versions listed?  Are they all still actively supported? (as in, will they have updates if, for example, a security issue was discovered in them?)

amotl: Dear Tianon,

apologies for the late reply and thank you for notifying us about this. We will submit a patch to purge the older versions soon.

With kind regards,
Andreas.

amotl: Hi Tianon,

#11708 implements this update, thanks again!

With kind regards,
Andreas.

PR #10619, opened 2021-07-27
arangodb 3.8.0

KVS85:
Added ArangoDB 3.8 (3.8.0).


tianon:
Nice!  Looks like 3.6 EOL is right around the corner too? :eyes:

KVS85:
> Nice! Looks like 3.6 EOL is right around the corner too? 👀

Yes. In September.

PR #10625, opened 2021-07-28
Update Vault for 1.8.0

kalafut:


library/vault
```diff
@@ -3,9 +3,9 @@ Maintainers: Chris Hoffman <choffman@hashicorp.com> (@chrishoffman),
...
 GitRepo: https://github.com/hashicorp/docker-vault.git
 
-Tags: 1.8.0-rc2
+Tags: 1.8.0
 Architectures: amd64, arm64v8, arm32v6, i386
-GitCommit: bdd23b0c75add997d0247e3d7fb72338f240fa70
+GitCommit: aabefbe426379c0d45d6ad3d8565fd3a6b17ae62
 Directory: 0.X
 
 Tags: 1.7.3, latest
```
yosifkit, line 11: Do you want to move `latest` to `1.8.0`?

kalafut: Oops, good catch. Updated!

PR #10628, opened 2021-07-28
eclipse-mosquitto: Fix issue with libwebsockets http2 support.

ralight:
Disable http2 support in the compiled libwebsockets, it causes problems in some situations.

### DIFF
--- a/library/eclipse-mosquitto
+++ b/library/eclipse-mosquitto
@@ -1,6 +1,6 @@
 Maintainers: Roger Light <roger@atchoo.org> (@ralight)
 GitRepo: https://github.com/eclipse/mosquitto.git
-GitCommit: 5217863b8b210f22df81c6b95d1eb89ed4af9b50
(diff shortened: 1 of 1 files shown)


mike-sandstrom:
Thank you for this as I have been having trouble connecting WSS on Firefox due to firefox initializing the connections with http2.  I am still relatively inexperienced with github. So you changed the build of libwebsockets when installing mosquitto or is this just for the docker image of mosquitto?

ralight:
@mike-sandstrom this is just for the docker image. How do you install mosquitto? Depending on the package there might be not a lot I can do.

mike-sandstrom:
sudo apt-get. I can install it however makes it work though. I believe  you mentioned that building mosquitto with cmake and turning off libwebsockets http2 is how to accomplish this, but I also saw you were planning on moving mosquitto away from libwebsockets. Any thoughts?

PR #10641, opened 2021-07-31
[varnish] multiarch and alpine support

gquintard:
It changes a few things:

- packages are built in the images rather than merely downloaded
- alpine support has been added (only for the `fresh` image as the `stable` version needs a few extra things unavailable currently)
- since we now build the packages, we should be able to build for all architectures (or close enough)

### DIFF
--- a/library/varnish
+++ b/library/varnish
@@ -1,13 +1,18 @@
-# this file was generated using https://github.com/varnish/docker-varnish/blob/4cfd65e1d3bd4eb9518d8e7590d9fec1de6a50c4/populate.sh
+# this file was generated using https://github.com/varnish/docker-varnish/blob/32e4b6f9ba7a00d449205c5cff79325309f1d691/populate.sh
 Maintainers: Guillaume Quintard <guillaume@varni.sh> (@gquintard)
 GitRepo: https://github.com/varnish/docker-varnish.git
 
-Tags: 6.0, 6.0.8-1, 6.0.8, stable
-Architectures: amd64
-Directory: stable/debian
-GitCommit: 4cfd65e1d3bd4eb9518d8e7590d9fec1de6a50c4
-
-Tags: 6.6, 6.6.1-1, 6.6.1, 6, latest, fresh
-Architectures: amd64
+Tags: fresh, 6.6.1, 6.6, latest
+Architectures: amd64, arm32v7, arm64v8, i386, ppc64le, s390x
 Directory: fresh/debian
-GitCommit: 4cfd65e1d3bd4eb9518d8e7590d9fec1de6a50c4
+GitCommit: 32e4b6f9ba7a00d449205c5cff79325309f1d691
+
+Tags: fresh-alpine, 6.6.1-alpine, 6.6-alpine, alpine
+Architectures: amd64, arm32v7, arm64v8, i386, ppc64le, s390x
+Directory: fresh/alpine
+GitCommit: 32e4b6f9ba7a00d449205c5cff79325309f1d691
+
+Tags: stable, 6.0.8, 6.0
+Architectures: amd64, arm32v7, arm64v8, i386, ppc64le, s390x
+Directory: stable/debian
+GitCommit: 32e4b6f9ba7a00d449205c5cff79325309f1d691


gquintard:
the `Munge PR / Diff Comment ` test is failing, but I'm unsure what's happening?

yosifkit:
This multi-stage build doesn't offer improvements over a single `RUN` that purges build packages and source files.  It actually causes the resulting image to be unavoidably larger than necessary since the copied deb files are still part of the image in the `COPY` layer (an `rm` in a `RUN` after `COPY` would just create a whiteout in the image layering).

See also our [FAQ section on Multi-stage Builds](https://github.com/docker-library/faq/tree/6138d05aabf61563606d86f98d0ccbd99f162b33#multi-stage-builds).

Small points of feedback:
- ` -j16` should be `-j"$(nproc)"` (include `coreutils` in Alpine build dependencies to get `nproc` available)
- `FROM alpine` should probably include a specific version
- `echo | sha512sum` is easier than a `test` (aka `[`):
   ```bash
   echo "73ed2f465b... */tmp/orig.tgz" | sha512sum -c -; \
   #  yes, space asterisk between the sha value and the file path (" *")
   ```
- `--single-branch` might be useful on the `git clone`
- you could also possibly optimize a tiny bit by downloading a source archive from GitHub at that commit such as `https://github.com/varnishcache/pkg-varnish-cache/archive/1f139121b5bce0b5b8f5d104224e14880a921b6b.tar.gz` instead of using git clone, but that's not strictly necessary and you'd then need a checksum to verify it.

tianon:
(I'm happy to help out if you want someone else to take a first pass on combining these)

gquintard:
Thanks a bunch @yosifkit! @tianon, your help is greatly welcome! Some feedback inline.

> This multi-stage build doesn't offer improvements over a single RUN that purges build packages and source files.

Erf, I thought I was being clever with the multi-stage and forgot about the layering... It should be okay now. (please check though)

> -j16 should be -j"$(nproc)" (include coreutils in Alpine build dependencies to get nproc available)

I'm pretty sure it shouldn't. From what I understood, `abuild` does it on it's own (https://github.com/alpinelinux/abuild/blob/754270e4607d31f7ffed502411e616251cb49460/abuild.conf#L9).

> FROM alpine should probably include a specific version

Noted, I went with `3.14`, please let me know if that's not ok.

> echo | sha512sum is easier than a test (aka [):

I should have thought about it, thanks!

> --single-branch might be useful on the git clone

I decided against it as the commits for `fresh` and `stable` are on different branches, that forced to encode more information in `populate.sh`. The repo is small too, so the gain is limited.

>    you could also possibly optimize a tiny bit by downloading a source archive from GitHub at that commit such as https://github.com/varnishcache/pkg-varnish-cache/archive/1f139121b5bce0b5b8f5d104224e14880a921b6b.tar.gz instead of using git clone, but that's not strictly necessary and you'd then need a checksum to verify it.

I like that one, but the checksum seems like hassle

yosifkit:
You can get an improvement of ~100MB smaller with these changes (and some savings in `alpine` images with similar cleanup of source). Explanatory comments inline.

```diff
diff --git a/Dockerfile.debian b/Dockerfile.debian
index 0bec29a..47c97c0 100644
--- a/Dockerfile.debian
+++ b/Dockerfile.debian
@@ -8,22 +8,23 @@ RUN set -e; \
     export DEBCONF_NONINTERACTIVE_SEEN=true; \
     apt-get update; \
     apt-get install -y $BASE_PKGS; \
# move all the work into a known empty directory so that it is easy to clean up
+    tmpdir="$(mktemp -d)"; \
+    cd "$tmpdir"; \
     git clone https://github.com/varnishcache/pkg-varnish-cache.git; \
     cd pkg-varnish-cache; \
     git checkout @PKG_COMMIT@; \
     rm -rf .git; \
-    curl http://varnish-cache.org/_downloads/varnish-@VARNISH_VERSION@.tgz -o /tmp/orig.tgz; \
-    echo "@DIST_SHA512@  /tmp/orig.tgz" | sha512sum -c -; \
-    tar xavf /tmp/orig.tgz --strip 1; \
# "-f" ensures curl fails appropriately
+    curl -f http://varnish-cache.org/_downloads/varnish-@VARNISH_VERSION@.tgz -o "$tmpdir/orig.tgz"; \
+    echo "@DIST_SHA512@  $tmpdir/orig.tgz" | sha512sum -c -; \
+    tar xavf "$tmpdir/orig.tgz" --strip 1; \
     sed -i -e "s|@VERSION@|@VARNISH_VERSION@|"  "debian/changelog"; \
     yes | mk-build-deps --install debian/control || true; \
     tar cvzf debian.tar.gz debian --dereference; tar xavf debian.tar.gz; \
     sed -i '' debian/varnish*; \
-    dpkg-buildpackage -us -uc -j16; \
+    dpkg-buildpackage -us -uc -j"$(nproc)"; \
     apt-get -y install ../*.deb; \
-    apt-get -y remove varnish-build-deps $BASE_PKGS; \
-    apt-get -y autoremove; \
-    rm -rf /var/lib/apt/lists/*
# just shorter, either way works
+    apt-get -y purge --auto-remove varnish-build-deps $BASE_PKGS; \
# this is the extra cleanup to delete the temporary directory with pkg-varnish-cache and the deb output files
+    rm -rf /var/lib/apt/lists/* "$tmpdir"

```

gquintard:
thanks, I completely forgot about the source.

Also, apologies, reading back at your earlier comment, I realized I misunderstood the `-j16` item and didn't see it on the `dpkg-buildpackage` line.

yosifkit (approved):

tianon:
A couple minor things look a little odd to me:

```diff
+    yes | mk-build-deps --install debian/control || true; \
```

I get the `yes |` here (although `mk-build-deps --install --tool="apt-get -o Debug::pkgProblemResolver=yes --yes" debian/control` is probably a cleaner solution to that), but I'm very confused by the `|| true` -- in what situations does this fail?

```diff
+    tar cvzf debian.tar.gz debian --dereference; tar xavf debian.tar.gz; \
+    sed -i '' debian/varnish*; \
```

If I'm understanding correctly, these are both trying to make sure all symlinks are dereferenced?  (That's the best I can figure.)

```diff
+    BASE_PKGS="tar alpine-sdk sudo git"; \
+    apk add -q --no-progress --update $BASE_PKGS; \
...
+    apk del $BASE_PKGS; \
```

This would probably be slightly cleaner with APK's `--virtual` feature, ala:

```dockerfile
    apk add --no-cache --virtual varnish-build-deps tar alpine-sdk sudo git; \
...
    apk del --no-network varnish-build-deps; \
```

```diff
+    apk add --allow-untrusted ~builder/packages/pkg-varnish-cache/*/*.apk; \
```

If these don't install any _new_ dependencies, this could use `--no-network` as well (although otherwise, `--no-cache` is probably a good idea).

```diff
+    cp /pkg-varnish-cache/systemd/varnishreload /usr/bin/; \
```

This is interesting -- I'm assuming this isn't already part of the APK because it's not generally needed on an installed Alpine system?

gquintard:
> I get the yes | here (although mk-build-deps --install --tool="apt-get -o Debug::pkgProblemResolver=yes --yes" debian/control is probably a cleaner solution to that), but I'm very confused by the || true -- in what situations does this fail?

Adopted the `--tool=` switch, way better. I *think* at some point (and possibly for another package) I had the issue of `mk-build-deps` failing even though all the deps were installed. It could have been because of `yes |` actually...

> If I'm understanding correctly, these are both trying to make sure all symlinks are dereferenced? (That's the best I can figure.)

Right on, I think I tried both and forgot to take one out. I kept the `sed` line, is that ok?

> This is interesting -- I'm assuming this isn't already part of the APK because it's not generally needed on an installed Alpine system?

Yes, that possibly should be fixed in `pkg-varnish-cache` (it's not upstream, and `pkg-varnish-cache` is just a cleaned up version of it, removing the now useless patches). I guess the upstream logic is that since `alpine` is for `docker`, you'll trash the the container and spin up a new one with the new config, even though it doesn't necessarily make sense with a cache.

>     apk add --no-cache --virtual varnish-build-deps tar alpine-sdk sudo git; \

I have an issue with that line, not with `--virtual` which is quite brilliant, but with `--no-cache`, if I add it, it causes the `abuild` call below to fail with:

``` bash
>>> varnish: Building pkg-varnish-cache/varnish 6.6.1-r1 (using abuild 3.8.0_rc4-r0) started Thu, 05 Aug 2021 21:54:00 +0000
>>> varnish: Checking sanity of /pkg-varnish-cache/alpine/APKBUILD...
>>> WARNING: varnish: No maintainer
>>> varnish: Analyzing dependencies...
ERROR: unable to select packages:
  pcre-dev (no such package):
    required by: .makedepends-varnish-20210805.215400[pcre-dev]
  pcre2-dev (no such package):
    required by: .makedepends-varnish-20210805.215400[pcre2-dev]
  ncurses-dev (no such package):
    required by: .makedepends-varnish-20210805.215400[ncurses-dev]
  libedit-dev (no such package):
    required by: .makedepends-varnish-20210805.215400[libedit-dev]
  py-docutils (no such package):
    required by: .makedepends-varnish-20210805.215400[py-docutils]
  linux-headers (no such package):
    required by: .makedepends-varnish-20210805.215400[linux-headers]
  libunwind-dev (no such package):
    required by: .makedepends-varnish-20210805.215400[libunwind-dev]
  python3 (no such package):
    required by: .makedepends-varnish-20210805.215400[python3]
  py3-sphinx (no such package):
    required by: .makedepends-varnish-20210805.215400[py3-sphinx]
>>> ERROR: varnish: builddeps failed
>>> varnish: Uninstalling dependencies...
WARNING: Ignoring https://dl-cdn.alpinelinux.org/alpine/v3.14/main: No such file or directory
WARNING: Ignoring https://dl-cdn.alpinelinux.org/alpine/v3.14/community: No such file or directory
ERROR: No such package: .makedepends-varnish
```

gquintard:
thanks again for your patience and support you two!

PR #10642, opened 2021-08-01
Update Clear Linux OS to 34900

clrbuilder:
Update Clear Linux OS latest+base images to release 34900

@bryteise @mdhorn @phmccarty

### DIFF
--- a/library/clearlinux
+++ b/library/clearlinux
@@ -2,5 +2,5 @@ Maintainers: William Douglas <william.douglas@intel.com> (@bryteise)
 GitRepo: https://github.com/clearlinux/docker-brew-clearlinux.git
 
 Tags: latest, base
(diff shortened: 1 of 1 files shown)


yosifkit:
Removal of `etc/os-release` seems unexpected. Was this intentional?

mdhorn:
> Removal of `etc/os-release` seems unexpected. Was this intentional?

That should be a link to /usr/lib/os-release ... investigating.

phmccarty:
Looks like all of the files and directories created via `systemd-tmpfiles` (from `/etc` and `/var`) disappeared from the image. Certainly unexpected...

mdhorn:
@yosifkit We found to root cause; related to a new version of systemd.
Please close this PR and we will get the fixes in next week's automated PR.

PR #10643, opened 2021-08-01
Update ALT Linux - 20210802

Obirvalger:


library/alt
```diff
@@ -4,27 +4,39 @@ Maintainers: ALT Linux Team Cloud <alt-cloud@altlinux.org> (@alt-cloud),
              Mikhail Gordeev <obirvalger@altlunux.org> (@obirvalger)
 GitRepo: https://github.com/alt-cloud/docker-brew-alt.git
 
-Tags: p9, latest
```
yosifkit, line 7: Did you want to keep the `latest` tag to now point to `p10`?

Emiliano978 (approved):

library/alt
Obirvalger: Yes, latest should point to p10 from now.

Obirvalger:
How could I help with merging this PR?

yosifkit:
> How could I help with merging this PR?

Apologies, I was waiting for the update to the tags. There is no `latest` tag and you agreed that it should point to `p10`, so it will need to be added to [line 7](https://github.com/docker-library/official-images/pull/10643/files#diff-85fa9427fe663bd341cdb445669e92ca427eb4ef0d9bc705332b514ed427a08cR7).

Obirvalger:
> > How could I help with merging this PR?
> 
> Apologies, I was waiting for the update to the tags. There is no `latest` tag and you agreed that it should point to `p10`, so it will need to be added to [line 7](https://github.com/docker-library/official-images/pull/10643/files#diff-85fa9427fe663bd341cdb445669e92ca427eb4ef0d9bc705332b514ed427a08cR7).

Sorry, I thought that added latest to p10 and you just confirmed replacement.

Issue #2171, opened 2016-09-22
openssl: CVE-2016-2177, CVE-2016-2178, CVE-2016-2179, CVE-2016-2180, CVE-2016-2181, CVE-2016-2182, CVE-2016-2183, CVE-2016-6302, CVE-2016-6303, CVE-2016-6304, CVE-2016-6305, CVE-2016-6306, CVE-2016-6307, CVE-2016-6308


tianon:
# CVE-2016-2177, CVE-2016-2178, CVE-2016-2179, CVE-2016-2180, CVE-2016-2181, CVE-2016-2182, CVE-2016-2183, CVE-2016-6302, CVE-2016-6303, CVE-2016-6304, CVE-2016-6305, CVE-2016-6306, CVE-2016-6307, CVE-2016-6308
## https://www.openssl.org/news/secadv/20160922.txt
- [x] `alpine`: https://github.com/docker-library/official-images/pull/2174
  - http://git.alpinelinux.org/cgit/aports/log/main/openssl?h=3.4-stable
  - http://git.alpinelinux.org/cgit/aports/log/main/openssl?h=3.3-stable
  - http://git.alpinelinux.org/cgit/aports/log/main/openssl?h=3.2-stable
  - http://git.alpinelinux.org/cgit/aports/log/main/openssl?h=3.1-stable
  - http://git.alpinelinux.org/cgit/aports/log/main/openssl
  - http://bugs.alpinelinux.org/projects/alpine/search?q=CVE-2016-2177
  - http://bugs.alpinelinux.org/projects/alpine/search?q=CVE-2016-2178
  - http://bugs.alpinelinux.org/projects/alpine/search?q=CVE-2016-2179
  - http://bugs.alpinelinux.org/projects/alpine/search?q=CVE-2016-2180
  - http://bugs.alpinelinux.org/projects/alpine/search?q=CVE-2016-2181
  - http://bugs.alpinelinux.org/projects/alpine/search?q=CVE-2016-2182
  - http://bugs.alpinelinux.org/projects/alpine/search?q=CVE-2016-2183
  - http://bugs.alpinelinux.org/projects/alpine/search?q=CVE-2016-6302
  - http://bugs.alpinelinux.org/projects/alpine/search?q=CVE-2016-6303
  - http://bugs.alpinelinux.org/projects/alpine/search?q=CVE-2016-6304
  - http://bugs.alpinelinux.org/projects/alpine/search?q=CVE-2016-6305
[...]

PR #10666, opened 2021-08-04
Add OpenJDK + Eclipse OpenJ9 images from IBM Semeru Runtimes

narkedi:
This request is to add the OpenJDK+ Eclipse OpenJ9 images to the official Docker images.

# Checklist for Review

    - IBM
    - language stack
    - https://github.com/docker-library/docs/pull/2017

### DIFF
--- a/library/ibm-semeru-runtimes
+++ b/library/ibm-semeru-runtimes
@@ -1,0 +1,87 @@
+# IBM Semeru Runtimes official Docker images.
+
+Maintainers: Surya Narkedimilli <snarkedi@in.ibm.com> (@narkedi)
+GitRepo: https://github.com/ibmruntimes/semeru-containers.git
+GitFetch: refs/heads/ibm
+
+#-----------------------------openj9 v8 images---------------------------------
+Tags: open-8u302-b08-jdk-focal, open-8-jdk-focal
+SharedTags: open-8u302-b08-jdk, open-8-jdk
+Architectures: amd64, ppc64le, s390x
+GitCommit: a5e772ef095c8d185a4dbcb804e5d4a3c895f71e
+Directory: 8/jdk/ubuntu
+File: Dockerfile.open.releases.full
+
+Tags: open-8u302-b08-jdk-centos7, open-8-jdk-centos7
+Architectures: amd64, ppc64le
+GitCommit: a5e772ef095c8d185a4dbcb804e5d4a3c895f71e
+Directory: 8/jdk/centos
+File: Dockerfile.open.releases.full
+
+Tags: open-8u302-b08-jre-focal, open-8-jre-focal
+SharedTags: open-8u302-b08-jre, open-8-jre
+Architectures: amd64, ppc64le, s390x
+GitCommit: a5e772ef095c8d185a4dbcb804e5d4a3c895f71e
+Directory: 8/jre/ubuntu
+File: Dockerfile.open.releases.full
+
+Tags: open-8u302-b08-jre-centos7, open-8-jre-centos7
+Architectures: amd64, ppc64le
+GitCommit: a5e772ef095c8d185a4dbcb804e5d4a3c895f71e
+Directory: 8/jre/centos
+File: Dockerfile.open.releases.full
+
+#-----------------------------openj9 v11 images---------------------------------
+Tags: open-11.0.12_7-jdk-focal, open-11-jdk-focal
+SharedTags: open-11.0.12_7-jdk, open-11-jdk
+Architectures: amd64, ppc64le, s390x
+GitCommit: a5e772ef095c8d185a4dbcb804e5d4a3c895f71e
+Directory: 11/jdk/ubuntu
+File: Dockerfile.open.releases.full
+
+Tags: open-11.0.12_7-jdk-centos7, open-11-jdk-centos7
+Architectures: amd64, ppc64le
+GitCommit: a5e772ef095c8d185a4dbcb804e5d4a3c895f71e
+Directory: 11/jdk/centos
+File: Dockerfile.open.releases.full
+
+Tags: open-11.0.12_7-jre-focal, open-11-jre-focal
+SharedTags: open-11.0.12_7-jre, open-11-jre
+Architectures: amd64, ppc64le, s390x
+GitCommit: a5e772ef095c8d185a4dbcb804e5d4a3c895f71e
+Directory: 11/jre/ubuntu
+File: Dockerfile.open.releases.full
+
+Tags: open-11.0.12_7-jre-centos7, open-11-jre-centos7
+Architectures: amd64, ppc64le
+GitCommit: a5e772ef095c8d185a4dbcb804e5d4a3c895f71e
+Directory: 11/jre/centos
+File: Dockerfile.open.releases.full
+
+#-----------------------------openj9 v16 images---------------------------------
+Tags: open-16.0.2_7-jdk-focal, open-16-jdk-focal
+SharedTags: open-16.0.2_7-jdk, open-16-jdk
+Architectures: amd64, ppc64le, s390x
+GitCommit: a5e772ef095c8d185a4dbcb804e5d4a3c895f71e
+Directory: 16/jdk/ubuntu
+File: Dockerfile.open.releases.full
+
+Tags: open-16.0.2_7-jdk-centos7, open-16-jdk-centos7
+Architectures: amd64, ppc64le
+GitCommit: a5e772ef095c8d185a4dbcb804e5d4a3c895f71e
+Directory: 16/jdk/centos
+File: Dockerfile.open.releases.full
+
+Tags: open-16.0.2_7-jre-focal, open-16-jre-focal
+SharedTags: open-16.0.2_7-jre, open-16-jre
+Architectures: amd64, ppc64le, s390x
+GitCommit: a5e772ef095c8d185a4dbcb804e5d4a3c895f71e
+Directory: 16/jre/ubuntu
+File: Dockerfile.open.releases.full
+
+Tags: open-16.0.2_7-jre-centos7, open-16-jre-centos7
+Architectures: amd64, ppc64le
+GitCommit: a5e772ef095c8d185a4dbcb804e5d4a3c895f71e
+Directory: 16/jre/centos
+File: Dockerfile.open.releases.full
+
--- a/test/config.sh
+++ b/test/config.sh
@@ -19,6 +19,7 @@ testAlias+=(
 	[adoptopenjdk]='openjdk'
 	[eclipse-temurin]='openjdk'
 	[sapmachine]='openjdk'
+	[ibm-semeru-runtimes]='openjdk'
 
 	[jruby]='ruby'
 	[pypy]='python'


tianon:
Wasn't `ibmjava` deprecated in favor of `adoptopenjdk` (which is now being very slowly renamed to "Eclipse Temurin"...)?

I'm very confused about what this is and how it fits in with the existing OpenJ9 variants over there.

narkedi:
@tianon Eclipse Temurin delivers OpenJDK + Hotspot images, and IBM Semeru Runtimes going forward will deliver both Open and Certified (JCKed) versions of OpenJDK + Eclipse OpenJ9 images.

AdamBrousseau:
Hi @tianon what Surya said is correct. It is a confusing story and transition, I will try to add a bit of background.
AdoptOpenJDK up until now was producing OpenJDK binaries with both Hotspot and OpenJ9 VM's. With Adopt's move to Eclipse, legal restrictions prevent the new Eclipse Adoptium group from producing/releasing OpenJ9 based binaries. As a result, IBM will be producing OpenJ9 based binaries in 2 flavours, Open and Certified, both under the family name IBM Semeru Runtimes. Essentially the same binaries, released under different licenses.

In regards to IBM Java 8 (https://hub.docker.com/_/ibmjava). These are not going anywhere anytime soon. These are based on a different code base and will continue for the foreseeable future. The IBM Java11 builds however, will transition to IBM Semeru Runtime Certified edition for the next release.

tianon:
I guess a major point of confusion for me here is why this is being proposed as a new image instead of these being part of https://hub.docker.com/_/ibmjava then?  This seems to me (as an outsider) that OpenJ9 used to be part of IBM Java, was part of AdoptOpenJDK for a while, and is now going back to being part of IBM Java due to the collapse/move of AdoptOpenJDK?

What makes Semeru different from "IBM Java"?  Why are they separate?

narkedi:
@tianon There is a branding change as we are moving away from AdoptOpenJDK. IBM Semeru is purely OpenJDK+OpenJ9 which supports both Open Edition and Certified Edition of the packages.

As such, the current "ibmjava" supports only the Certified edition. Also, the code base has additional IBM specific changes which is not same as IBM Semeru Runtimes. 

In future, "ibmjava" repo will only support JDK8 and all the other future releases will be part of IBM Semeru Runtimes.

mstoodle:
Some background info that may be helpful: [https://developer.ibm.com/blogs/introducing-the-ibm-semeru-runtimes/](https://developer.ibm.com/blogs/introducing-the-ibm-semeru-runtimes/)

This article explains the new branding that motivates this request. It doesn't help with differentiating from the existing ibmjava images, but I think the 2 key aspects (as Adam and Surya mentioned) are:
1. ibmjava images are for something called "IBM SDK for Java" which contain proprietary components and will only go up to Java 8.
2. IBM is rebranding its JDK going forward as "IBM Semeru Runtime", so it makes sense that we use a name connected to that. This name starts from Java 8 and continues to Java 11 and on.

narkedi:
@tianon Can you please let us know if there are any other changes needed ?

narkedi:
@tianon Just a ping on whether this can be moved further and merged. Incase of any other changes needed, please do let us know. Thanks much !!

AdamBrousseau:
@tianon or @yosifkit is there anything else we need to action on this or can it be merged?
Thanks.

narkedi:
@tianon Request you to trigger the Test Run's against this PR as well.

narkedi:
@tianon Request you to trigger the Test Run's once again. Fixed the 2 issues which were seen during the last run. Thanks !!

narkedi:
@tianon Now that the checks are completed, can this be merged. Thanks !!

tianon:
Frankly, I'm still very confused by the image and tag naming -- what is `open` in there for?  Is that referring to "OpenJ9" or is it implying that there might be a "closed" version in the future?  Also, "IBM Semeru Runtimes" sounds like a family of different runtime implementations -- is that really the image name you want to go with?  Are you planning to add more implementations in the future?

narkedi:
> Frankly, I'm still very confused by the image and tag naming -- what is `open` in there for? Is that referring to "OpenJ9" or is it implying that there might be a "closed" version in the future? Also, "IBM Semeru Runtimes" sounds like a family of different runtime implementations -- is that really the image name you want to go with? Are you planning to add more implementations in the future?

@tianon You got it right, we will be delivering "certified" version in future. We are not including "OpenJ9" into the tagging as we only deliver "OpenJ9" implementation. 

Yes, this is the image name we want to go with as we have changed our Branding for Java deliverables, and all future releases will be with this name. 

Regarding the implementations, as mentioned we won't be adding any other implementations. We will stick only to OpenJ9 implementation.

mstoodle:
To clarify: "open" in the name refers to the IBM Semeru Runtime Open Edition. As @narkedi says, there will also soon be an IBM Semeru Runtime Certified Edition which will put "certified" in that same spot. For legal reasons, we need to have two distinct names for the two editions.

Together, the Open Edition and the Certified Edition are referred to as the IBM Semeru Runtimes. 

Both editions have the same OpenJDK versions and OpenJ9 versions in them, but have different licenses, and only the certified edition has been tested for Java certification.

tianon:
Why would users choose one over the other?

mstoodle:
Some people are familiar with and comfortable with open source license -> Open Edition

Some IBM customers are familiar with and comfortable with IBM license -> Certified Edition

Some people really care about Java certification -> Certified Edition

else Open Edition :) .

narkedi:
@tianon Hope your confusion is cleared out. Can we please get this PR merged ?

tianon:
That makes me even more confused -- if all that's different is the amount of testing and the license applied, why not provide the same thing under a dual license?  Wouldn't that provide the same benefits, but with less overhead (both technically and mentally for IBM *and* IBM's users)?

As is, the choice being presented to users sounds to me like "would you like the *open* thing or the *tested* thing?" and I don't think that dichotomy is the intention.

mstoodle:
I'll refer you to one line in my earlier post:

> For legal reasons, we need to have two distinct names for the two editions.

mstoodle:
I will grant that the names fall into the category of "worst option, except for all the alternatives". You aren't raising any issues we haven't gone around a hundred times already :( .

We're not willfully trying to be confusing here, but we're operating under some constraints that we're not at liberty to discuss publicly. Forgive me, but I'm also actually not quite sure where you're going with this line of questioning. You're challenging names that have been decided (months ago) by the people who have the right and responsibility to make those naming decisions. We want to use those names for our binaries consistently inside and outside Docker Hub. Surely your end game here isn't to make things even more confusing (I know, I'm also amazed that is actually possible) by forcing us to use different names at Docker Hub than the ones people identify with at our direct download site ([here's a link](https://ibm.com/semeru-runtimes/downloads)) ?

tianon:
Nope, not trying to be a pain; sorry if it feels that way.

I'm trying to make sure I understand fully, and that the end result is actually in line with the goals of the program this image is being proposed into: https://github.com/docker-library/official-images#what-are-official-images

From what I'm understanding so far, the "Open" edition is definitely right in line, but the "Certified" edition is a fair bit murkier.

I realize this PR currently only contains the "Open" edition, but it's phrased/proposed in such a way that implies the future existence of the "Certified" edition here as well, which is why I'm trying to make sure I understand correctly.

mstoodle:
I'm not sure what to say. I cannot argue it from a technical standpoint. But there are other considerations because "certified" is a loaded term in the Java ecosystem and there are legal implications associated with the license needed to be able to run the Java Certification Kit (JCK) that I can't get into. We tried to make it as simple as we could although, in retrospect, including both OpenJDK and OpenJ9 version numbers is an area we could probably improve going forward.

I don't suppose there's a way forward that lets us get the currently named images out and available for people to use (they've been waiting more than a month now) but we improve the versioning (and therefore tagging) in our subsequent releases? (the October update release, say) ?

yosifkit:
A quick review of the Dockerfiles:

We recommend against using blanket package upgrades (`apt upgrade`/`apk upgrade`/`yum upgrade`/`yum update`) for official-images. When package upgrades are applied in a dependent image, it duplicates content of the base image, making the image larger than necessary. It also only delays the inevitable "there are outdated packages". The Official Images build pipeline makes heavy use of `docker build` cache, so that just rebuilding the all of the Dockerfiles won't cause any unnecessary change. So we rely on periodic base image updates.

>We strive to publish updated images at least monthly for Debian and Ubuntu. We also rebuild earlier if there is a critical security need, e.g. [docker-library/official-images#2171](https://github.com/docker-library/official-images/issues/2171). Many Official Images are maintained by the community or their respective upstream projects, like Alpine and Oracle Linux, and are subject to their own maintenance schedule. These refreshed base images also means that any other image in the Official Images program that is `FROM` them will also be rebuilt (as described in [the project `README.md` file](https://github.com/docker-library/official-images#library-definition-files)).
>
> \- https://github.com/docker-library/faq/tree/73f10b0daf2fb8e7b38efaccc0e90b3510919d51#why-does-my-security-scanner-show-that-an-image-has-cves

---

> what does the CentOS variant add that is not served by the Ubuntu and Windows images?

See https://github.com/docker-library/official-images/pull/10662#issuecomment-897881758 for more details.

Later in the linked discussion from Tianon:

> Whether or not you keep/remove CentOS is up to you, but I would note that it does not get updated very frequently as potentially useful information in your decision.

Or maybe drop CentOS altogether and document how to get the jdk/jre in any base like this?

```Dockerfile
# Example
FROM <base image>
ENV JAVA_HOME=/opt/java/openjdk
COPY --from=ibm-semeru-runtimes:open-11-jdk $JAVA_HOME $JAVA_HOME
ENV PATH="${JAVA_HOME}/bin:${PATH}"
```

---

For multi-stage builds, see our [faq entry](https://github.com/docker-library/faq/tree/73f10b0daf2fb8e7b38efaccc0e90b3510919d51#multi-stage-builds). TL;DR: `nanoserver` based images should `COPY` from their counterpart Windows servercore based images. Something like this:
```Dockerfile
ENV JAVA_HOME=C:\\openjdk-11
...
ENV JAVA_VERSION jdk-11.0.12+7_openj9-0.27.0
COPY --from=ibm-semeru-runtimes:open-11.0.12_7-jdk-windowsservercore-1809 $JAVA_HOME $JAVA_HOME
...
```
It could be the less specific version tag in the `COPY`, if desired. Reminder that official images are only built when the Dockerfile (+build context) changes, or any parent image changes. The `openjdk` image has a complete [example](https://github.com/docker-library/openjdk/blob/48b669921a105090c2e227f308596752fb04f453/11/jdk/windows/nanoserver-1809/Dockerfile) of this `COPY` behavior.

---

`ENV PATH` in the `nanoserver` image is not needed since the `PATH` is already set in the registry via `setx`. `JAVA_HOME` is also being put in the registry and as a docker `ENV`, but is sufficient as only a Docker variable.

narkedi:
@yosifkit @tianon Updated the Dockerfiles based on the review comments. Can you please trigger another round of checks. Thanks !!

narkedi:
@yosifkit @tianon Can you please let me know if the above failure is because of the patch ?

narkedi:
@yosifkit @tianon Removed the windows entries, will add them separately after fixing the issues. Can you please trigger the checks once again. Thanks !!

narkedi:
@tianon @yosifkit Can you please merge the PR. Thanks !!

narkedi:
@yosifkit @tianon Can you please let us know if there are anymore changes required from our end. Thanks much !!

library/ibm-semeru-runtimes
```diff
@@ -0,0 +1,92 @@
...
+#-----------------------------openj9 v8 images---------------------------------
+Tags: open-8u302-b08-jdk-focal, open-8-jdk-focal
+SharedTags: open-8u302-b08-jdk, open-8-jdk
+Architectures: amd64, ppc64le, s390x
+GitCommit: 21d1d498b7dff9b58acd3c265f1bb3aca97ce23e
+Directory: 8/jdk/ubuntu
+File: Dockerfile.open.releases.full
+
+Tags: open-8u302-b08-jdk-centos7, open-8-jdk-centos7
+SharedTags: open-8u302-b08-jdk, open-8-jdk
```
yosifkit, line 15: Unfortunately `SharedTags` does not work for items of the same [`Platform`](https://specs.opencontainers.org/image-spec/image-index/?v=v1.0.1#:~:text=be%20safely%20ignored.-,platform%20object,-This%20OPTIONAL%20property) and so would create an invalid manifest when used for two Linux-based images or two Windows images with the same `os.version` (e.g. `nanoserver-1809` and `windowsservercore-1809`).

So, `SharedTags` is only used for differing platforms: up to one linux (e.g. `ubuntu`), and each supported Windows version (e.g. `windowsservercore-1809` and `windowsservercore-ltsc2022`).

See the shared tags used in the [`golang` library file](https://github.com/docker-library/official-images/blob/5d9206dd7cc736c41fd0f9eedb324e1fd00309e2/library/golang#L8-L67) as a fairly comprehensive example:
 - shared "generic" tags (like `1.17.1`) on `bullseye`, `windowsservercore-ltsc2022`, `windowsservercore-1809`, and `windowsservercore-ltsc2016`
     - but not any of the other Linux distributions (`buster`, `stretch`, `alpine*`) or `nanoserver`
 - shared `windowsservercore`-only tags for  `windowsservercore-{ltsc2022,1809,ltsc2016}`
 - shared `nanoserver`-only tags for `nanoserver-{ltsc2022,1809}`

narkedi: @yosifkit I have stopped using the SharedTags for CentOS. Thanks for referring to the example. !! Will handle it accordingly when we add support for Windows later. Thanks !!

tianon:
A few more Dockerization comments:

---

Are you _sure_ you want to support an explicit CentOS variant?  It's much easier to add later than it will be to deprecate and remove, and it's trivial for users to create for themselves:

> > We strive to publish updated images at least monthly for Debian and Ubuntu. We also rebuild earlier if there is a critical security need, e.g. [docker-library/official-images#2171](https://github.com/docker-library/official-images/issues/2171). Many Official Images are maintained by the community or their respective upstream projects, like Alpine and Oracle Linux, and are subject to their own maintenance schedule. These refreshed base images also means that any other image in the Official Images program that is `FROM` them will also be rebuilt (as described in [the project `README.md` file](https://github.com/docker-library/official-images#library-definition-files)).
> > - https://github.com/docker-library/faq/tree/73f10b0daf2fb8e7b38efaccc0e90b3510919d51#why-does-my-security-scanner-show-that-an-image-has-cves
> 
> > what does the CentOS variant add that is not served by the Ubuntu and Windows images?
> 
> See [#10662 (comment)](https://github.com/docker-library/official-images/pull/10662#issuecomment-897881758) for more details.
> 
> Later in the linked discussion from Tianon:
> 
> > Whether or not you keep/remove CentOS is up to you, but I would note that it does not get updated very frequently as potentially useful information in your decision.
> 
> Or maybe drop CentOS altogether and document how to get the jdk/jre in any base like this?
> 
> ```dockerfile
> # Example
> FROM <base image>
> ENV JAVA_HOME=/opt/java/openjdk
> COPY --from=ibm-semeru-runtimes:open-11-jdk $JAVA_HOME $JAVA_HOME
> ENV PATH="${JAVA_HOME}/bin:${PATH}"
> ```

```dockerfile
FROM centos:7
ENV JAVA_HOME=/opt/java/openjdk
COPY --from=ibm-semeru-runtimes:open-11-jdk $JAVA_HOME $JAVA_HOME
ENV PATH="${JAVA_HOME}/bin:${PATH}"
```

---

If you decide to keep CentOS, you'll want to review https://github.com/docker-library/official-images/pull/10907#issuecomment-920274282 (and likely stop using `uname -m` for architecture detection).

---

I'm still not a big fan of the Shared Class Cache generation code here -- I thought our understanding when we approved it for AdoptOpenJDK was that it was temporary while a cleaner/better solution was worked out.  Any progress on that?  It's _really_ hacky to download and run Tomcat as part of the build here just to prime that cache. :confused:

---

Setting all three of `LANG`, `LANGUAGE`, and `LC_ALL` feels like overkill -- I think `LANG` is probably enough?  It might also be a good idea to not set `LANG` until the locale is actually generated (to avoid potential "missing locale" warnings/errors that might crop up in the future).

yosifkit:
> we should be able to add more stages to conduct this population process rather than downloading it during the build process.

Unfortunately any [multi-stage build](https://docs.docker.com/develop/develop-images/multistage-build/) that leaves behind intermediate layers is usually a no-go in official images; the early stage just takes space on the build server and since the build is not bit-for-bit reproducible, anytime we have to prune images on our build servers, we lose `docker build` cache, and they can get unnecessarily rebuilt (and pushed). See also our [Multi-stage Builds](https://github.com/docker-library/faq/tree/73f10b0daf2fb8e7b38efaccc0e90b3510919d51#multi-stage-builds) entry in our FAQ.

Publishing a common/universal class cache tar/zip somewhere that the Dockerfile could use `curl/wget` to download (and then verify with `gpg` and/or `sha512sum`) is one possible alternative (https://github.com/docker-library/official-images/pull/8781#issuecomment-707936337).

mstoodle:
It will not be easy to provide a "standalone" shared classes cache that could just be copied into the container as part of the build process. The validation of the cache performed at run time depends on multiple timestamps for files in the JKD and user permissions and things like that. If we create the cache outside of the container where it will be used, it's very likely that cache won't actually be used by the JDK (because it will fail those validations I mentioned) and so we'll be adding payload to the image with no actual benefit.

If there's no practical way to use a tomcat image via a staged build rather than downloading it, then the hack you don't like may just be a bad option that's nevertheless better than all the alternatives.

bmarwell:
Hello @tianon, I think I should add some notes from a customer point of view.
Getting those images is really important to us, because without these, the progress of updating Liberty + Java stalls.
It is really nice that you want bit-perfect descriptions and Dockerfiles, but customers do have support, so we are 100% sure there will be updates fixing any nit at some point in the future. As for now, getting a working image would be "beneficial", so to say.

Customers of (Open)Liberty will not care too much whether this image name is  `ibm-semeru-runtimes:open-8u302-b08_openj9-0.27.0-jdk` ("quite a mouth full" as you put it) or `ibm-semeru-runtimes:8u302-jdk` `ibm-semeru:8u302-b08-jdk`. But, please note this: We know some fixes come from the OpenJDK, so we might request an u302, u304, etc. 
Some changes OTOH are related to the OpenJ9 engine (which replaces hotspot), so we might request OpenJ9 v0.27.1 (e.g. a bugfix release). 

We (customers) are already well aware of the changes in naming, and so are many non-subscribed consumers from the slack channes (adoptium, OpenJ9).
OTOH, I might also want to use this image for Apache Projects (as a PMC member) in the future.

Seeing this PR stall for months now does shed a light on dockerhub reliability, and it does not seem entirely positive.

Thank you very much for your understanding.
Ben

yosifkit:
Ah, new CI failure is due to #11071 -- to fix it, you'll want to add `GitFetch: refs/heads/ibm` right after `GitRepo:`. :+1:

narkedi:
> Ah, new CI failure is due to #11071 -- to fix it, you'll want to add `GitFetch: refs/heads/ibm` right after `GitRepo:`. 👍

Thanks @yosifkit. I will do that, i was thinking its because of an outdated SHA of our GIT Repo.

iNikem:
@narkedi @mstoodle any estimation when we can expect Windows images? :)

verymahler:
@AdamBrousseau The IBM Semeru Runtime downloads for Java 17 are available now https://developer.ibm.com/languages/java/semeru-runtimes/downloads/

@PascalSchumacher The docker images are available here https://hub.docker.com/_/ibm-semeru-runtimes

rdean:
Ubuntu images will work for initial testing, but I'll need UBI-based images for deployment.

Software-Noob:
> @PascalSchumacher we're still waiting on OpenJ9 to release jdk17. There are still 1 or 2 blockers. We expect it will be soon. Once OpenJ9 release is ready, Semeru will be released within about a week. And Docker containers shortly after the Semeru release.

@AdamBrousseau Any reason why it's not built for ARM64, unlike the other LTS releases?

narkedi:
@Software-Noob Will check and add the ARM64 support, it just got missed out. We will see that its included. Thanks !!

AdamBrousseau:
@rdean official UBI images can be found at Red Hat Registry
https://catalog.redhat.com/software/containers/search?q=semeru&p=1

rdean:
Those images carry this statement:

"Before downloading or using this Container, you must agree to the Red Hat subscription agreement located at redhat.com/licenses. If you do not agree with these terms, do not download or use the Container. If you have an existing Red Hat Enterprise Agreement (or other negotiated agreement with Red Hat) with terms that govern subscription services associated with Containers, then your existing agreement will control."

UBI and Semeru individually don't require a Red Hat subscription / Enterprise Agreement to use.  Why does the combination?

narkedi:
> @Software-Noob Will check and add the ARM64 support, it just got missed out. We will see that its included. Thanks !!

ARM64 images are now part of the official images for JDK17.

narkedi:
> UBI and Semeru individually don't require a Red Hat subscription / Enterprise Agreement to use. Why does the combination?

@rdean Not quite sure why its popping up, will check with the RH team if there are any changes to the terms and conditions.

iNikem:
> @iNikem, they will be there in a week's time.

@narkedi did your plans change? It has been several months, but there are still no Windows images :(

narkedi:
@iNikem Not exactly, you should hear some news by next week. We will support Windows, and will publish with the latest version of JDK releases. Thanks for the interest and we will make sure that they are there very shortly.

PR #10695, opened 2021-08-10
mariadb releases 20210806 with s390x

grooverdan:
s390x was missed in the release.

Sorry for missing this new architecture in the last PR.


tianon:
Nice!

PR #10721, opened 2021-08-13
RiP EOLed groovy for neurodebian

yarikoptic:


tianon:
Hey @yarikoptic!  It probably makes sense to make a new one of these for Hirsute, right? :eyes:

PR #10726, opened 2021-08-15
[varnish] switch to debian:bullseye

gquintard:


heroku-miraheze:
#10727 please do not merge until this PR mentioned is merged.

tianon:
@heroku-miraheze why?  When `debian` is updated, everything `FROM` it will rebuild again, so merging this now shouldn't have any ill effect?

(especially since the previous/currently published build of Bullseye is not really _that_ much different from the final GA build, by design)

PR #10727, opened 2021-08-16
Add Debian 11.0 and upgrade to the latest versions

heroku-miraheze:


heroku-miraheze:
@tianon

evgeni:
Isn't this missing the 11 and 11.0 tags? Or are these added somewhere else?

library/debian
```diff
@@ -33,108 +33,108 @@ riscv64-GitCommit: 87fa5e6c3caeac435f2a4af10aa3bd308da1d4ec
 s390x-GitFetch: refs/heads/dist-s390x
 s390x-GitCommit: 941fe0ebfee8123836d6e9ffe2740c76defcfd52
 
-# bullseye -- Debian x.y Testing distribution - Not Released
-Tags: bullseye, bullseye-20210721
+# bullseye -- Debian 11.0 Released 14 August 2021
+Tags: bullseye, bullseye-20210816, latest
```
ekohl, line 37: Like @evgeni said, I think this also needs the 11 and 11.0 tags.
```suggestion
Tags: bullseye, bullseye-20210816, 11.0, 11, latest
```

tianon:
See #10730 (which includes an actual rebuild).

PR #10732, opened 2021-08-16
Update clojure tools-deps to 1.10.3.943

cap10morgan:
New upstream release


cap10morgan:
Looks like some upstream openjdk release changes bit me here again. I'll sort it out and push an update.

PR #10730, opened 2021-08-16
Update Debian, esp. for Bullseye (+ Bookworm)

tianon:
Happy [Debian Day](https://wiki.debian.org/DebianDay)!

This includes the move to Debian Bullseye as stable and the new testing, Bookworm. :smile:

PR #10739, opened 2021-08-16
Update golang

tianon:
Changes:

- https://github.com/docker-library/golang/commit/0760c29: Merge pull request https://github.com/docker-library/golang/pull/381 from infosiftr/1.17
- https://github.com/docker-library/golang/commit/48a7371: Add bullseye variants
- https://github.com/docker-library/golang/commit/2cd4dab: Update to 1.17 (GA)

### DIFF
--- a/library/golang
+++ b/library/golang
@@ -1,132 +1,100 @@
-# this file is generated via https://github.com/docker-library/golang/blob/f300e60ca19c3b98cfcf01ca112af2ac10104320/generate-stackbrew-library.sh
+# this file is generated via https://github.com/docker-library/golang/blob/48a7371ed6055a97a10adb0b75756192ad5f1c97/generate-stackbrew-library.sh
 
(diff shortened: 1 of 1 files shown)


yosifkit:
**WARNING:** this will likely not be available for a bit since the build queue will be very full due to the `bullseye` release and thus a rebuild of all Debian based images (https://github.com/docker-library/official-images/pull/10730)

Watch the jobs in https://doi-janky.infosiftr.net/job/multiarch/view/images/view/golang/ for the architecture of your choice to see when it gets build an pushed to the architecture-specific namespace (e.g. https://hub.docker.com/r/amd64/golang). Once it is there, it will be a short while later that it'll then be at https://hub.docker.com/_/golang in the multi-architecture manifest.  See also our FAQ entry for [An image's source changed in Git, now what?](https://github.com/docker-library/faq/tree/73f10b0daf2fb8e7b38efaccc0e90b3510919d51#an-images-source-changed-in-git-now-what).

PR #10752, opened 2021-08-18
Gradle 7.2

keeganwitt:


tianon:
CI failure is the normal arch weirdness on `adoptopenjdk` (hopefully that's part of what will be fixed by [Temurin](https://hub.docker.com/_/eclipse-temurin)?? :grimacing:)

keeganwitt:
> CI failure is the normal arch weirdness on `adoptopenjdk` (hopefully that's part of what will be fixed by [Temurin](https://hub.docker.com/_/eclipse-temurin)?? 😬)

Maybe. I'm trying to figure out the right path forward. If it'd make sense to move to that or if there's any reason to consider going back to the openjdk image. I've also got openj9 images because a user specifically asked for them that I have to find a new home for (I guess ibmjava?).

tianon:
Yeah, I'm not sure I have a good answer for you. :disappointed:

See #10666 for IBM's plans for OpenJ9 going forward... :confused:

(maybe we can just put a "No More :cursing_face: Java" ban on all Official Images instead :joy:)

PR #10757, opened 2021-08-19
Update Clear Linux OS to 34940

clrbuilder:
Update Clear Linux OS latest+base images to release 34940

@bryteise @mdhorn @phmccarty


phmccarty:
This is a planned out-of-cycle PR to address a critical bug that is present in the current base image (version 34930).

In particular, we integrated a fix for the `glibc` package that was affecting `swupd` from functioning correctly in the docker environment for version 34930.

PR #10760, opened 2021-08-19
release 2.1.6

alcohol:


yosifkit:
>
> \- https://github.com/docker-library/official-images/tree/c6f79b67dfd5d56e12e9ffe134964bfd2c54e9d7#repeatability

`COPY --from=mlocati/php-extension-installer` is not allowed. You could `curl` and checksum the file directly (https://github.com/docker-library/official-images/tree/c6f79b67dfd5d56e12e9ffe134964bfd2c54e9d7#image-build), though I am wary of downloading a 2800 line shell script to replace 15 lines.

alcohol:
My bad, updated.

tianon:
FYI, we normally discourage `apk upgrade` (preferring instead to see timely base image updates which then trickle down, such as we see consistently for Debian and Ubuntu; https://github.com/docker-library/faq#why-does-my-security-scanner-show-that-an-image-has-cves).

alcohol:
If it is desirable, I can remove the call to apk upgrade. I sincerely doubt it makes any difference for us so that would be no problem at all.

PR #10804, opened 2021-08-28
consul: 1.10.2, 1.9.9 and 1.8.15

eculver:
New releases today.

### DIFF
--- a/library/consul
+++ b/library/consul
@@ -1,19 +1,19 @@
 Maintainers: Consul Team <consul@hashicorp.com> (@hashicorp/consul)
 
-Tags: 1.10.1, 1.10, latest
+Tags: 1.10.2, 1.10, latest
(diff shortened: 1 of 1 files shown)


tianon:

tianon:
Friendly ping?

Also going to cc some of the most recent previous image updaters:
@mikemorris @freddygv @rboyer @mkeeler
(since there aren't any explicit humans listed in the `Maintainers:` line and I can't ping the @hashicorp/consul group thanks to GitHub limitations :grimacing:)

mikemorris:
Thanks for the ping @tianon, I'll try to get this cleaned up and resolved!

mikemorris:
@tianon Migrated fully to the OCI spec labels in our source Dockerfile and pushed a commit with updated commit shas, hope that'll do it!

PR #10817, opened 2021-08-31
Update tomcat

tianon:
Changes:

- https://github.com/docker-library/tomcat/commit/d6c16db: Merge pull request https://github.com/docker-library/tomcat/pull/240 from infosiftr/jq-template
- https://github.com/docker-library/tomcat/commit/55d8abb: Add initial jq-based templating engine

PR #10824, opened 2021-09-01
XWiki 13.7 release

manuelleduc:


yosifkit:
Apologies, the `tomcat:adoptopenjdk` variants where dropped in https://github.com/docker-library/official-images/pull/10817 (https://github.com/docker-library/tomcat/pull/240) because of the rename/move of the adoptopenjdk project to "Eclipse Temurin" (https://github.com/docker-library/official-images/pull/10662).  So the xwiki images will need to swap to a different tomcat variant.

yosifkit:
:+1: Thanks for updating the `13.7` versions. I'll merge so we can get the new version building for users.

You'll likely want to update the rest soon since they'll no longer be rebuilt with updated base images (esp for things like the recent openssl fixes).

PR #10826, opened 2021-09-01
eclipse-mosquitto 2.0.12

ralight:
Also:
* Bump Alpine version to 3.14
* Remove self compiled cJSON - this is included in alpine:3.14
* Bump libwebsockets to 4.2.1

### DIFF
--- a/library/eclipse-mosquitto
+++ b/library/eclipse-mosquitto
@@ -1,12 +1,12 @@
 Maintainers: Roger Light <roger@atchoo.org> (@ralight)
 GitRepo: https://github.com/eclipse/mosquitto.git
-GitCommit: 17bbca22fb5dc57bfdc83ac60be67e34208825fd
+GitCommit: 9afeeb1a568470871511a78de6b5927010ede1f5
 Architectures: amd64, arm32v6, arm64v8, i386, ppc64le, s390x
 
-Tags: 2.0.11, 2.0, 2, latest
+Tags: 2.0.12, 2.0, 2, latest
 Directory: docker/2.0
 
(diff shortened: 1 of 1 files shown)


yosifkit:
We recommend against using `apt`/`apk`/`yum upgrade` for official-images. When `apt`/`apk`/`yum` upgrade is applied in a dependent image, it duplicates content of the base image, making the image larger than necessary. It also only delays the inevitable "there are outdated packages". The Official Images build pipeline makes heavy use of `docker build` cache, so that just rebuilding the all of the Dockerfiles won't cause any unnecessary change. So we rely on periodic base image updates.

>We strive to publish updated images at least monthly for Debian and Ubuntu. We also rebuild earlier if there is a critical security need, e.g. [docker-library/official-images#2171](https://github.com/docker-library/official-images/issues/2171). Many Official Images are maintained by the community or their respective upstream projects, like Alpine and Oracle Linux, and are subject to their own maintenance schedule. These refreshed base images also means that any other image in the Official Images program that is `FROM` them will also be rebuilt (as described in [the project `README.md` file](https://github.com/docker-library/official-images#library-definition-files)).
>
> \- https://github.com/docker-library/faq/tree/73f10b0daf2fb8e7b38efaccc0e90b3510919d51#why-does-my-security-scanner-show-that-an-image-has-cves

---

Alpine images were updated just yesterday in https://github.com/docker-library/official-images/pull/10821 (and so `apk upgrade` does nothing today), so the "libressl / libretls conflict" is no longer relevant?

ralight:
Thanks for the tip. Actually this is my fault, my 3.14 image wasn't quite up to date. I've removed the update commands now and they all build fine.

PR #10851, opened 2021-09-04
Friendica Bugfixings & Improvements

nupplaphil:
- Added sha256 check
- Fix permission problem for `cron.sh`

### DIFF
--- a/library/friendica
+++ b/library/friendica
@@ -5,32 +5,32 @@ GitRepo: https://github.com/friendica/docker.git
 
 Tags: 2021.04-apache, 2021.04
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
-GitCommit: f7ff6093182f3892d925414ae8d7a6e810545ce9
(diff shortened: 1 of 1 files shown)


yosifkit:
For Dockerfile best practices, we recommend:
> embedding checksums directly in the `Dockerfile`
>
>The purpose in recommending checksum verification is to verify that the artifact is as expected. This ensures that when remote content changes, the Dockerfile also will change and provide a natural `docker build` cache bust. As a bonus, this also prevents accidentally downloading newer-than-expected artifacts on poorly versioned files.
>
> [README.md#image-build](https://github.com/docker-library/official-images/tree/951503fc65d27acae64adba5b50c61bbafb944ec#image-build)

This helps prevent accidental (or malicious) changes on the download server from affecting the build, since a shasum change in the Dockerfile without a version bump would be reviewed before being built & pushed to users.

Likely the `update.sh` in the friendica/docker repo could be used to download the shasums files and embed the correct values in the Dockerfiles.

nupplaphil:
@yosifkit I included the sha256 directly as recommended :-)

PR #10875, opened 2021-09-09
Friendica: Add GPG check to official images

nupplaphil:
FollowUp #10851 

Inspired by https://github.com/docker-library/faq/#openpgp--gnupg-keys-and-verification :-)


tianon:
:metal:

PR #10876, opened 2021-09-09
Update clojure images for various new upstreams

cap10morgan:
- tools-deps 1.10.3.967 was released
- openjdk released Debian bullseye-based images
- openjdk released Alpine-based images for JDKs 16, 17, & 18

### DIFF
--- a/library/clojure
+++ b/library/clojure
@@ -200,28 +200,37 @@ Tags: openjdk-18-tools-deps-bullseye, openjdk-18-tools-deps-1.10.3.967-bullseye
 Directory: target/openjdk-18-bullseye/tools-deps
 
 Tags: openjdk-16-alpine, openjdk-16-lein-alpine, openjdk-16-lein-2.9.6-alpine
+Architectures: amd64
 Directory: target/openjdk-16-alpine/lein
 
 Tags: openjdk-16-boot-alpine, openjdk-16-boot-2.8.3-alpine
+Architectures: amd64
 Directory: target/openjdk-16-alpine/boot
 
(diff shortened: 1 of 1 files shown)


cap10morgan:
What's a reliable way to tell which images are unsupported? Those `openjdk:16-alpine` and `openjdk:17-alpine` tags are still up on Docker Hub and I can pull and build from them locally.

tianon:
The "supported" list is officially those listed in https://github.com/docker-library/official-images/blob/master/library/openjdk (which can be parsed programmatically via https://github.com/docker-library/bashbrew), although the full list is also embedded in the image description markdown at https://github.com/docker-library/docs/blob/master/openjdk/README.md#supported-tags-and-respective-dockerfile-links.

The underlying reason in the specific case of the `alpine` tags is that OpenJDK on Alpine is something that was only very recently officially supported (in 16, I think?) but the note on http://jdk.java.net/16/ explains a bit more:

> The Alpine Linux build previously available on this page was removed as of the first JDK 16 release candidate. It’s not production-ready because it hasn’t been tested thoroughly enough to be considered a GA build. Please use the early-access JDK 17 Alpine Linux build in its place.

(However, since that was written, http://jdk.java.net/17/ also lost the Alpine build it had, so I'm guessing there were more issues discovered.)

cap10morgan:
OK thanks @tianon that's very helpful. Hopefully this latest push will be 👍🏻.

tianon:
FWIW, `apk add --update --no-cache` is redundant -- `apk add --no-cache` is enough :+1:

PR #10879, opened 2021-09-09
Make "Munge PR" more intelligent (test PR)

tianon:
There are a few goals here:

- avoid failing "diff" when no images are changed (test only PR, for example)
- automatically explicitly CC image maintainers on PRs

(This PR exists only as a trigger for the changes I'm working on because the nature of `pull_request_target` is that the branch making the changes that I want to test has to live on the origin. :sweat_smile:)


tianon:
(Before we can merge this, we need to do some cleanup of a few `library/` files so we don't ping folks who are long gone unnecessarily :grimacing: :innocent:)

tianon:
I think #11079, https://github.com/docker-library/golang/pull/395, and https://github.com/docker-library/busybox/pull/119 are the last few things _I_ wanted to clean up before we officially considered this. :eyes:

PR #10882, opened 2021-09-10
Add a new emqx image

Rory-Z:
documentation PR: https://github.com/docker-library/docs/pull/2032

# Checklist for Review

    - https://github.com/emqx
    - Apache-2.0
    - service
    - https://github.com/docker-library/docs/pull/2032

### DIFF
--- a/library/emqx
+++ b/library/emqx
@@ -1,5 +1,6 @@
 Maintainers: Rory Z <rory-z@outlook.com> (@rory-z)
 GitRepo: https://github.com/emqx/emqx-docker.git
+GitFetch: refs/heads/main
 
 Tags: 4.3.5, 4.3, 4, latest
 GitCommit: 915d12f4a289def311fdc350f85539c74068bdb9
--- a/library/emqx
+++ b/library/emqx
@@ -1,0 +1,7 @@
+Maintainers: Rory Z <rory-z@outlook.com> (@rory-z)
+GitRepo: https://github.com/emqx/emqx-docker.git
+
+Tags: 4.3.5, 4.3, 4, latest
+GitCommit: 915d12f4a289def311fdc350f85539c74068bdb9
+Directory: 4.3
+File: Dockerfile


tianon:
Sorry for the delay!  New image reviews usually take a lot more effort than image update review, so they often take a back seat (we're trying to be better about that). :bow:

Here's a few Dockerization review comments -- let me know if there's anything I can do to help with these or explain them better!

---

```diff
+        apt-transport-https \
```

This shouldn't be necessary with newer versions of APT -- I _think_ even Buster was new enough, but if not, maybe considering Bullseye makes sense now? (Hopefully adding Bullseye to https://repos.emqx.io/emqx-ce/deb/debian/ is planned? :sweat_smile:)

---

```diff
+        software-properties-common \
...
+    && add-apt-repository "deb https://repos.emqx.io/emqx-ce/deb/debian/ ./$(lsb_release -cs) stable" \
```

With this usage, this `add-apt-repository` is essentially just going to take that string verbatim and drop it into a new file in `/etc/apt/sources.list.d`.  I would suggest dropping `software-properties-common`/`add-apt-repository` and just putting the appropriate contents directly into a file.  Additionally, you already know which release you're in (`FROM debian:buster-slim`), so you can avoid `lsb_release` too if you wanted to.

---

```diff
+RUN curl -fsSL https://repos.emqx.io/gpg.pub | apt-key add - \
```

Using `apt-key add` is deprecated by the APT maintainers, and they prefer to see either a file in `/etc/apt/trusted.gpg.d` or better, a file elsewhere and the use of `signed-by` in the `sources.list` line (`deb [ signed-by=/etc/apt/keyrings/something.gpg ] ...`).

Also, downloading this key without any verification (from the same place the packages come from, no less :grimacing:) is not going to be acceptable.  What we'd suggest instead is either pulling the key by full fingerprint from a keyserver or downloading this flie, importing it into `gpg`, then exporting the key you're looking for by the full fingerprint into a file that's then used by APT.

See https://github.com/docker-library/mysql/blob/e4b225b3eed5c774fa11799f04832e0ad351da62/8.0/Dockerfile.debian#L54-L67 for a pretty straightforward example of that.

---

```diff
+LOCAL_IP=$(hostname -i | grep -oE '((25[0-5]|(2[0-4]|1[0-9]|[1-9]|)[0-9])\.){3}(25[0-5]|(2[0-4]|1[0-9]|[1-9]|)[0-9])' | head -n 1)
```

That's quite a complicated regular expression -- my best read of this is that it's trying to filter `hostname -i` output to just IPv4 addresses?

---

```diff
+fill_tuples() {
```

This function looks interesting, but it doesn't appear to be used anywhere -- just a leftover?

Rory-Z:
Hi, @tianon Thank you for your reply
I've updated the Dockerfile and entryponit scripts as you suggested, could you please review them again?

tianon:
Looking at https://github.com/emqx/emqx-docker/compare/7a24e0e...75d44b6 (but in my terminal so it can give me a better diff from `4.3.3/Dockerfile` to `4.3/Dockerfile` that GitHub isn't willing to for some reason :sweat_smile:):

---

```diff
+RUN groupadd -r emqx && useradd -r -g emqx emqx
```

Given there's on-disk storage here, I'd suggest picking an explicit UID/GID here -- a common choice is to use the "default" or most common port number (since that's usually unused), but anything that isn't already taken by your current base image (and hopefully untaken by any base image you might use in the future, including just newer versions of Debian) should be fine (for example, `1000:1000` is a pretty safe choice).

---

```diff
+ENV EMQX_MAJOR 4.3
+ENV EMQX_VERSION 4.3.5
```

I would suggest moving these down to _right_ before the `RUN` line where they're actually used so we can maximize the build cache.

---

Other than that, I don't see much else!  This is looking good. :+1:
(Once again, I really appreciate your patience with us! :heart:)

Rory-Z:
Hi, @tianon 
I've assigned GID/UID for docker user, plz review again.
I'm looking forward to being part of official images, it will help us a lot
Thanks again

library/emqx
```diff
@@ -0,0 +1,7 @@
+Maintainers: Rory Z <rory-z@outlook.com> (@rory-z)
+GitRepo: https://github.com/emqx/emqx-docker.git
```
tianon, line 2:
```suggestion
GitRepo: https://github.com/emqx/emqx-docker.git
GitFetch: refs/heads/main
```

(This should fix the CI failure.)

Rory-Z:
Hi, @tianon 
Thanks for your comment

tianon (approved):

yosifkit (approved):

Rory-Z:
Hi，@tianon, @yosifkit Thanks for your work.
How long will it take to pull this image from the docker hub?

tianon:
Sorry for the delay - forgot about one additional bit we had to do manually to get this up.  It's now available to pull, but will take a bit of time before https://hub.docker.com/_/emqx exists - in the meantime, you can check https://hub.docker.com/r/amd64/emqx to get an idea of how the documentation will render once it does. :+1:

Rory-Z:
So cooool，now I get image for  https://hub.docker.com/_/emqx，this's great ！

PR #10884, opened 2021-09-10
[Swift] Update the Docker image to support Swift 5.4.3

shahmishal:


yosifkit:
We recommend fetching keys from a keyserver instead of downloading a file directly. While the file is hosted on trusted infrastructure, it is not verified in any way (except that it is a key when imported by `gpg`). When using a keyserver, `gpg` will verify that the key data received from the keyserver is the desired fingerprint.

```console
$ # when downloading a valid but unexpected key
$ gpg --batch --keyserver ... --recv-keys 'DEAD BEEF DEAD BEEF DEAD  BEEF DEAD BEEF DEAD BEEF'
gpg: key 036A9C25BF357DD4: rejected by import screener
(and a non-zero exit)
```

See also our [FAQ entry](https://github.com/docker-library/faq/#openpgp--gnupg-keys-and-verification) on PGP.

shahmishal:
@yosifkit Does this resolve the concern? Thanks for the feedback.

tianon:
That CI failure is due to `dirmngr` still running in the background -- our official builds will retry several times so it likely won't affect those too negatively, but if you want to improve that CI failure, I'd suggest adding something like `gpgconf --kill all` before the `rm` of `GNUPGHOME`.

PR #10890, opened 2021-09-13
Friendica dev/rc: Switch from GIT to file-archive

nupplaphil:
- removed all GIT dependencies
- Added additional GPG checks for dev/rc archives
- Added SHA256 check before downloading new archives

### DIFF
--- a/library/friendica
+++ b/library/friendica
@@ -5,60 +5,60 @@ GitRepo: https://github.com/friendica/docker.git
 
 Tags: 2021.04-apache, 2021.04
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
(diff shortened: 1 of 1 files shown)


yosifkit:
So the `sum256` files are used to prevent re-downloading the full `tar.gz` file, but aren't directly tied to the downloaded tar files. Is there a possible race where a new `sum256` is downloaded but one or both of the old `tar.gz` files are the old file, thus preventing an update when the newer tar is available?  Do you need a case where only one set of the `sum256`+`tar` files changes?

nupplaphil:
The `sum256` is calculated directly after generating each artifact and is part of the overall artifact upload. So the only race condition could be between the upload of sum256 and the upload of the archive (normally it takes about a second to upload both). So it's very unlikely

But they are dev/rc archives. So the worst thing would be that one release would be missed, which is acceptable I guess

PR #10009, opened 2021-04-20
Add Official AlmaLinux OS Image

ezamriy:
Hello, we would like to add an official image for the [AlmaLinux OS](https://almalinux.org/).

# Checklist for Review

**NOTE:** This checklist is intended for the use of the Official Images maintainers both to track the status of your PR and to help inform you and others of where we're at. As such, please leave the "checking" of items to the repository maintainers. If there is a point below for which you would like to provide additional information or note completion, please do so by commenting on the PR. Thanks! (and thanks for staying patient with us :heart:)

-	[x] associated with or contacted upstream?
    -  https://github.com/AlmaLinux/docker-images, I'm the AlmaLinux OS Foundation board member and one of the core developers
-	[x] does it fit into one of the common categories? ("service", "language stack", "base distribution")
    -  base distribution 
-	[x] is it reasonably popular, or does it solve a particular use case well?
    -  AlmaLinux OS was released on March 30th 2021 and, as the only currently available CentOS replacement, has almost 20K downloads. AlmaLinux OS solves multiple use cases where the use of RHEL would be not viable, impossible or not feasible. AlmaLinux OS provides a long-term supported enterprise-grade OS which allows for any type of stack or workload to run on.
-	[x] does a [documentation](https://github.com/docker-library/docs/blob/master/README.md) PR exist? (should be reviewed and merged at roughly the same time so that we don't have an empty image page on the Hub for very long)
    - https://github.com/docker-library/docs/pull/1920
-	[x] official-images maintainer dockerization review for best practices and cache gotchas/improvements (ala [the official review guidelines](https://github.com/docker-library/official-images/blob/master/README.md#review-guidelines))?
-	[x] 2+ official-images maintainer dockerization review?
-	[x] ~existing official images have been considered as a base? (ie, if `foobar` needs Node.js, has `FROM node:...` instead of grabbing `node` via other means been considered?)~
-	[x] if `FROM scratch`, tarballs only exist in a single commit within the associated history?
-	[x] passes current tests? any simple new tests that might be appropriate to add? (https://github.com/docker-library/official-images/tree/master/test)

PR #10891, opened 2021-09-13
AlmaLinux: apply security fixes and add base, init, micro images

ezamriy:
Fixes the following CVEs in default and minimal images:
    
* glib2: CVE-2021-27218
* libxml2: CVE-2021-3516, CVE-2021-3517, CVE-2021-3518, CVE-2021-3537, CVE-2021-3541
* lz4-libs: CVE-2021-3520
* rpm: CVE-2021-20271
    
Adds 3 new RHEL UBI-compatible image types:
    
* base: ubi (standard) UBI alternative
* init: ubi-init (multiservice) alternative
* micro: ubi-micro alternative

### DIFF
--- a/library/almalinux
+++ b/library/almalinux
@@ -2,15 +2,36 @@ Maintainers: The AlmaLinux OS Foundation <cloud-infra@almalinux.org> (@AlmaLinux
 GitRepo: https://github.com/AlmaLinux/docker-images.git
 
 Tags: latest, 8, 8.4
-GitFetch: refs/heads/almalinux-8-x86_64
-GitCommit: 875d1e6ce39cf5a5495148fce53471f9be49a4c3
-arm64v8-GitFetch: refs/heads/almalinux-8-aarch64
-arm64v8-GitCommit: 54188857c898d835b22e23d273f8022cb8d1f1ce
+GitFetch: refs/heads/almalinux-8-x86_64-default
+GitCommit: 226a6650f4713fa050b1eec92dd4b6e974493e5c
+arm64v8-GitFetch: refs/heads/almalinux-8-aarch64-default
+arm64v8-GitCommit: 63a41e97c4b96311e802cfff56cce6caacce399d
 Architectures: amd64, arm64v8
 
 Tags: minimal, 8-minimal, 8.4-minimal
 GitFetch: refs/heads/almalinux-8-x86_64-minimal
-GitCommit: bc74e29d58210f825911fe73cd7ec0f724d1f78f
+GitCommit: 62c988b53a408c9d1124cc4a17b0f471ce716b73
 arm64v8-GitFetch: refs/heads/almalinux-8-aarch64-minimal
-arm64v8-GitCommit: c0291ce4ff7d7bedb0fef86a863cb7405fb77520
+arm64v8-GitCommit: 269d87f8720165979b79ecfd9b70aace0297675c
+Architectures: amd64, arm64v8
+
+Tags: micro, 8-micro, 8.4-micro
+GitFetch: refs/heads/almalinux-8-x86_64-micro
+GitCommit: 8221e72561dc5d9ffd3e01664f47deeeca2c4678
+arm64v8-GitFetch: refs/heads/almalinux-8-aarch64-micro
+arm64v8-GitCommit: fb5f0d1869557cbfbe16748b9dacee3f73a74f5d
+Architectures: amd64, arm64v8
+
+Tags: init, 8-init, 8.4-init
+GitFetch: refs/heads/almalinux-8-x86_64-init
+GitCommit: cf3a890a339807ad121bc83ea613b597330a6e89
+arm64v8-GitFetch: refs/heads/almalinux-8-aarch64-init
+arm64v8-GitCommit: 78115cd85a0b46b4974eeb44849796b0eaa9d477
+Architectures: amd64, arm64v8
+
+Tags: base, 8-base, 8.4-base
+GitFetch: refs/heads/almalinux-8-x86_64-base
+GitCommit: ad28b9b6ad3672c86857d6320cc8dd41d5ca3c66
+arm64v8-GitFetch: refs/heads/almalinux-8-aarch64-base
+arm64v8-GitCommit: f0eb492b0f6ce61501251f18feb3ddde0bb64efe
 Architectures: amd64, arm64v8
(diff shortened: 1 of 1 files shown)


yosifkit:
A base image using a tarball is reserved for images that cannot be built from other images. It seems like some of these could be a derived image. Could some (or most/all?) of these different tags be `FROM` another and build upon the other image by just installing a few packages and maybe some config or `Dockerfile` directives?

Could you share some information on the goal of each of these tags for those unfamiliar?

ezamriy:
Hi @yosifkit.

> Could you share some information on the goal of each of these tags for those unfamiliar?

Yes, sure. We are going to create an additional PR for the [docs](https://github.com/docker-library/docs/) explaining the difference between images.

As you know, the goal of the AlmaLinux OS project is to provide a CentOS/RHEL alternative because CentOS 8 will be deprecated at the end of this year. So, first, we created a default `latest` image which is a drop-in replacement for the CentOS official one.

But Red Hat also provides a set of [Universal Base Images](https://www.redhat.com/en/blog/introducing-red-hat-universal-base-image) (UBI)  for which many of our container users also want to have alternatives:

- [ubi-micro](https://www.redhat.com/en/blog/introduction-ubi-micro) - a super minimal (only 18 packages) image which is used as a base layer for hardened environments or for containerizing of statically compiled binaries. There is no package manager/etc here.
- [ubi-minimal](https://catalog.redhat.com/software/containers/ubi8/ubi-minimal/5c359a62bed8bd75a2c3fba8) - a stripped-down (only 101 packages) base image which is using the `microdnf` package manager with a very limited feature set support. Only core tools and libraries are installed on that image.
- [ubi-base](https://catalog.redhat.com/software/containers/ubi8/ubi/5c359854d70cc534b3a3784e) (also called a platform) - is a generic container image with the full-featured DNF package manager, OpenSSL, and default commands installed.
- [ubi-init](https://catalog.redhat.com/software/containers/ubi8/ubi-init/5c359b97d70cc534b3a378c8) - a systemd-enabled container image. Actually, it's the `ubi-base` image with one additional package installed and a different CMD.

So, answering your question

> It seems like some of these could be a derived image. Could some (or most/all?) of these different tags be FROM another and build upon the other image by just installing a few packages and maybe some config or Dockerfile directives?

1. We must have separate tarballs for both `micro` and `minimal` images because those are unique and very different from others.

2. the `init` image can reuse the `base` image tarball but its Dockerfile will be a bit dirty/complicated (similar to this one https://github.com/AlmaLinux/docker-images/blob/1fa8053a32d5100524c5771e6f513b3c1f2d1210/al-mini/Dockerfile) because we need to clean up logs/machine-id/etc after. If it's okay for a Docker Library official image, let's do it.

3. the difference between our default `latest` and the `base` image is the following 9 packages added to the `base`:

  - libssh
  - libssh-config
  - libuser
  - libpsl
  - passwd
  - publicsuffix-list-dafsa
  - usermode
  - virt-what
  - which

I don't really think they are necessary for a container image, but you know, it's all about compatibility with the RHEL upstream.
And the following 4 packages are present  in our `latest` and removed from the `base`:

  - binutils
  - hostname
  - less
  - xz

That's why we made two different images there. But if you insist on reducing the number of images, we can add those 9 packages to the `latest` image (I remember we previously spent some time trying to optimize its size here https://github.com/docker-library/official-images/pull/10009) and then just tag it as `base`.

Then we will have 3 "from scratch" images with unique tarballs (`micro`, `mini` and `latest`) and the `init` image based on the `latest` one. What do you think @yosifkit?

yosifkit:
- `micro` We've resisted adding any "distroless" type images. These images do not fit our definition of "generally useful" since the way to use them is to externally install software into it. If a user must use some other environment (OS or container) to build/package their software, then that environment has knowledge of all the required files to run it as well. For users that want a tiny attack surface, it seems that `FROM scratch` would be a better target.

	For `dart`, a "distroless" image was just a matter of providing a runtime directory that could be copied to an empty image (see their example in [Docker Hub docs](https://github.com/docker-library/docs/tree/fea35f330e89ab552c41a1a40635b2d60af62301/dart#using-this-image)) instead of an extra image with only the runtime. This has the added benefit that all libraries will get updated every time the application binary does. Perhaps `dnf`/`yum` have something similar already for `--installroot` to include the "micro" set of packages in an easy to use way?

- `minimal` and `latest/8.4` are understandably separate tar files (just like Debian `slim` + `latest/bullseye`)
- Unfortunately any multi-stage build that leaves behind intermediate layers is usually a no-go in official images; the early stage just takes space on the build server and since the build is not bit-for-bit reproducible, anytime we have to prune images on our build servers, we lose `docker build` cache and they can get unnecessarily rebuilt (and pushed). See also our [Multi-stage Builds](https://github.com/docker-library/faq/tree/73f10b0daf2fb8e7b38efaccc0e90b3510919d51#multi-stage-builds) entry in our FAQ.

	It's unclear why `init` requires a multi-stage build if it is just "one additional package installed and a different CMD" with a little clean up. How is something like this not enough?
	```Dockerfile
	FROM almalinux:8.4-base
	RUN set -eux; \
		dnf install --nodocs -y [the packages]; \
		rm -rf /var/cache/*; \
		dnf clean all; \
	# not sure which of these are required, since there isn't a "COPY --from"
		# generate build time file for compatibility with CentOS
		date +%Y%m%d_%H%M > /etc/BUILDTIME; \
		# set DNF infra variable to container for compatibility with CentOS
		echo 'container' > /etc/dnf/vars/infra; \
		# install only en_US.UTF-8 locale files, see
		# https://fedoraproject.org/wiki/Changes/Glibc_locale_subpackaging for details
		echo '%_install_langs en_US.UTF-8' > /etc/rpm/macros.image-language-conf ;\
		> /etc/machine-id
	
	CMD ["/sbin/init"]
	STOPSIGNAL SIGRTMIN+3
	```
- Combining `base` and `latest` into a single image is fine. I think that makes sense from a user perspective, since they'll have smaller alternatives if that is their priority. The concern we had in the initial PR was making sure that things like vm or device related packages were not included unnecessarily and to help users moving from CentOS have minimal hiccups.

ezamriy:
Hi @yosifkit, thanks for your feedback. We are still discussing it internally, but probably we will keep only `minimal` and `default `images here (like we had before this PR) and move the UBI-like mages to dedicated `almalinux` user repositories to keep things simple and homogeneous for users.

ezamriy:
Yeah, decided. @srbala submitted an update

srbala:
@yosifkit Thank you for the PR Merge. I think it would be good to have date date on our builds which will help users have an option to fallback older builds. Please my changes in https://github.com/AlmaLinux/docker-library-official-images/commit/297b3b9f72c637e6bc50152835e8e3cdcf60b0af, are we allowed to add this one?

tianon:
Seems OK, but shouldn't it be version-specific, like "8-20210913" or
"8-minimal-20210913" (or "8-20210913-minimal" depending on your
cosmetic preference) ?

PR #10907, opened 2021-09-15
eclipse-temurin: add windows JRE 11 images and arm32v7

gdams:


tianon:
Unfortunately, these 32bit CentOS builds will not build successfully on our infrastructure due to using `uname -m` for architecture detection.  When `uname -m` is invoked, it returns the _kernel_ architecture, when what you're really looking for / needing is the _userspace_ architecture.  On our machines on those 32bit builds, it'll probably return `armv8l` (although even that's probably unreliable).  For Debian-based images, this is why we use/recommend `dpkg --print-architecture` (and `apk --print-arch` for Alpine-based images), because that returns a string which corresponds appropriately to the userspace architecture.

For resolving this, I'd suggest either simplifying the `centos` images to `COPY --from=` the Ubuntu images as in the example in the image description (at which point even maintaining them at all seems a little overkill IMO :smile:), or doing something like we ended up with in `openjdk` for our Oracle Linux variants:

https://github.com/docker-library/openjdk/blob/d83080b4f20d310638ea412049f8313e562eff0f/17/jdk/oraclelinux8/Dockerfile#L37

For `arch="$(objdump="$(command -v objdump)" && objdump --file-headers "$objdump" | awk -F '[:,]+[[:space:]]+' '$1 == "architecture" { print $2 }')"`, the values you're looking for in your `case` statement will be:

- `amd64`: `i386:x86-64`
- `arm64v8`: `aarch64`
- `arm32v7`: `arm` (counting our blessings that `v7` is the only `arm32` that CentOS even supports :grimacing:)
- `ppc64le`: `powerpc:common64`

gdams:
@tianon when I try using those Oracle Linux commands I get the following issue with bashbrew:

```
Tags "8u302-b08-jdk-centos7, 8-jdk-centos7, 8-centos7" has invalid Architectures: "'i386:x86-64', 'powerpc:common64'"
```

Any idea what I need to do to fix that?

tianon:
The values in the generated `library` file's `Architectures:` field shouldn't change -- those should still be `amd64`, `arm32v7`, etc.

gdams:
@tianon updated PTAL

Issue #2076, opened 2016-08-19
[Proposal] Deprecate `onbuild` tags


yosifkit:
From our docs description of onbuild:

> While the `onbuild` variant is really useful for "getting off the ground running" (zero to Dockerized in a short period of time), it's not recommended for long-term usage within a project due to the lack of control over _when_ the `ONBUILD` triggers fire (see also [`docker/docker#5714`](https://github.com/docker/docker/issues/5714), [`docker/docker#8240`](https://github.com/docker/docker/issues/8240), [`docker/docker#11917`](https://github.com/docker/docker/issues/11917)).

Users are often wanting customization on the order of the `onbuild` directives (some examples in this [search](https://github.com/issues?utf8=%E2%9C%93&q=onbuild+repo%3Adocker-library%2Fruby+repo%3Adocker-library%2Fdjango+repo%3Adocker-library%2Fpypy+repo%3Adocker-library%2Fgolang+repo%3Adocker-library%2Frails)).  Since we recommend them only to get started, I think it would be better to just provide an example Dockerfile in the docs for these images.

The following are the onbuild images:

``` console
$ bashbrew list --all --uniq | grep onbuild
clojure:onbuild
clojure:alpine-onbuild
django:python3-onbuild
django:python2-onbuild
erlang:19.0.3-onbuild
erlang:18.3.4.3-onbuild
golang:1.6.3-onbuild
golang:1.7.0-onbuild
iojs:1.8.4-onbuild
iojs:2.5.0-onbuild
iojs:3.3.0-onbuild
jruby:9-onbuild
jruby:1.7-onbuild
maven:3.3.9-jdk-7-onbuild
maven:3.3.9-jdk-7-onbuild-alpine
maven:3.3.9-jdk-8-onbuild
maven:3.3.9-jdk-8-onbuild-alpine
maven:3.3.9-jdk-9-onbuild
mono:3.10.0-onbuild
mono:3.12.1-onbuild
mono:3.8.0-onbuild
mono:4.0.5.1-onbuild
mono:4.2.1.102-onbuild
mono:4.2.2.30-onbuild
mono:4.2.3.4-onbuild
mono:4.2.4.4-onbuild
mono:4.4.0.182-onbuild
mono:4.4.1.0-onbuild
mono:4.4.2.11-onbuild
node:0.10.46-onbuild
node:0.12.15-onbuild
node:4.5.0-onbuild
node:5.12.0-onbuild
node:6.4.0-onbuild
[...]

PR #10943, opened 2021-09-22
Update to JRuby 9.3 but keep 9.2 support

headius:
JRuby 9.1 support is no longer provided in this config.


tianon:
I would suggest dropping the deprecated `onbuild` variants entirely in 9.3+ (pursuant to https://github.com/docker-library/official-images/issues/2076 :sweat_smile:).

headius:
@tianon Sure, I'll make that change.

headius:
@tianon Looks like we're ready to go again!

PR #10957, opened 2021-09-23
Update tomcat

tianon:
Changes:

- https://github.com/docker-library/tomcat/commit/d0fbc3b: Merge pull request https://github.com/docker-library/tomcat/pull/243 from infosiftr/jdk17
- https://github.com/docker-library/tomcat/commit/5635007: Add jdk17 variants


tianon:
Copying relevant note from https://github.com/docker-library/tomcat/pull/243:

> These should be the new default (since 17 is LTS), but the longevity of vanilla builds from http://jdk.java.net/17/ are not necessarily a given (Oracle usually only supplies binaries through that site for ~6 months IIRC).
>
> So, until that is confirmed to be a long-term maintained home for new 17 builds or 17 shows up somewhere like https://github.com/adoptopenjdk?q=upstream, we should stick to 11 as our default. :disappointed:

PR #10960, opened 2021-09-24
Update bonita

baptistemesta:
Update bonita version to 2021.2-u0

### DIFF
--- a/library/bonita
+++ b/library/bonita
@@ -7,18 +7,17 @@ Maintainers: Baptiste Mesta <baptiste.mesta@bonitasoft.org> (@baptistemesta),
 Architectures: amd64, arm64v8, ppc64le
 
 
-
-Tags: 7.10.6, 7.10
-GitRepo: https://github.com/Bonitasoft-Community/docker_bonita.git
-GitCommit: e6f9f1a5e57c35bbd833c9441a639f326dffb7d5
-Directory: 7.10
-
 Tags: 7.11.4, 7.11
 GitRepo: https://github.com/bonitasoft/bonita-distrib.git
 GitCommit: 231024c8290a9aa31a45b758a0765a684c21ed21
 Directory: docker
 
-Tags: 2021.1, 7.12.1, 7.12, latest
+Tags: 2021.1, 7.12.1, 7.12
 GitRepo: https://github.com/bonitasoft/bonita-distrib.git
 GitCommit: c9b816249504017bb3418252bf58ec9d4fc3e86e
+Directory: docker
+
+Tags: 2021.2-u0, 2021.2, 7.13.0, 7.13, latest
+GitRepo: https://github.com/bonitasoft/bonita-distrib.git
+GitCommit: 1cb3943ac5402d93e8f550f985ca1269b5b806e8
 Directory: docker


yosifkit:
Unfortunately any multi-stage build that leaves behind intermediate layers is usually a no-go in official images; the early stage just takes space on the build server and since the build is not bit-for-bit reproducible, anytime we have to prune images on our build servers, we lose `docker build` cache and they can get unnecessarily rebuilt (and pushed). See also our [Multi-stage Builds](https://github.com/docker-library/faq/tree/73f10b0daf2fb8e7b38efaccc0e90b3510919d51#multi-stage-builds) entry in our FAQ.

passga:
Hello, thanks for your answer. I have a question, actually, in our Dockerfile, we use a layer named "builder " and copy the content of "/opt" to the next definitive layer. What i understood that's what is causing the problem. So i would like to know if is it ok for you,  if we release this layer in our https://hub.docker.com/u/bonitasoft, and use it in the DockerFile to build the definitive ( official ) image?

tianon:
Unfortunately not; https://github.com/docker-library/official-images#repeatability:

baptistemesta:
Hello,

We changed the image to use no multi-stage build.
Please tell us if it solves that issue.

Thanks in advance! :)

tianon:
This is definitely closer, thanks!  I think we've still got a cache-ability issue with so much of (otherwise highly-cache-able) `RUN` lines getting moved down below the (much more volatile) `COPY` instructions -- can we move the package installations and other static bits back up above the `VERSION` and `COPY` lines so we can get better caching back?

baptistemesta:
Ok, I'll do that. Thank you for the explanation.

educhastenier:
Hi @tianon Is this better like this?
Thanks for the improvement recommentations.

educhastenier:
Hi @tianon @yosifkit Is this better like this?
Thanks for the improvement recommentations.

yosifkit:
There is still some improvement missing that was is the previous versions of the images.  For best cache-ability, volatile Dockerfile instructions like `ENV` or `ARG` should generally be as late in the Dockerfile as it can; i.e. defined just before the `RUN` line that uses it. This ensures that an earlier layer which is unrelated to a Bonita version bump, for example, would not be unnecessarily rebuilt.

baptistemesta:
We reverted most of the changes back to what they were in 7.12 (2021.1) 

We re-pushed most of the `ENV` and `ARG` as low as possible. We will be more careful regarding the order of those instructions in the futur.

I hope this time it's ok 🤞 

Thank you!

yosifkit (approved):

tianon:
Very close :sweat_smile: :see_no_evil:

I'm concerned about the `jattach` addition:

> ```diff
> +# Install Jattach
> +ARG JATTACH_VERSION=v2.0
> +RUN curl -fsSL https://github.com/apangin/jattach/releases/download/${JATTACH_VERSION}/jattach -o /usr/bin/jattach \
> +  && chmod +x /usr/bin/jattach
> ```

Besides downloading this without _any_ verification (:grimacing:), it also appears this is a precompiled C binary that probably only works on amd64? :sweat_smile:

educhastenier:
Thx for the feedbacks.

> 
> Besides downloading this without _any_ verification (:grimacing:), it also appears this is a precompiled C binary that probably only works on amd64?

Jattach only provides **one version** of the Linux binaries, so it may not work on all architectures, but that's ok to us, as the tool is only here in case there is an issue that need further inverstigations.
We will document this limitation.
Regarding the Hash verification, you are right, we will do that.

BUT ...
in order for this PR to be merged ASAP, we will remove Jattach from this version and will do another PR with it.

Thx for you reactivity.

PR #10967, opened 2021-09-24
Update ELK images

tianon:
elasticsearch:
- https://github.com/docker-library/elasticsearch/commit/9515924: Update to 7.15.0
- https://github.com/docker-library/elasticsearch/commit/12452b4: Update to 6.8.19
- https://github.com/docker-library/elasticsearch/commit/aca03f5: Update to 7.14.2

logstash:
- https://github.com/docker-library/logstash/commit/9242490: Update to 7.15.0
- https://github.com/docker-library/logstash/commit/6f3408c: Update to 6.8.19
- https://github.com/docker-library/logstash/commit/eef3332: Update to 7.14.2

kibana:
- https://github.com/docker-library/kibana/commit/8faf2e3: Update to 7.15.0
- https://github.com/docker-library/kibana/commit/5cb47fd: Update to 6.8.19
- https://github.com/docker-library/kibana/commit/d04fefc: Update to 7.14.2


tianon:
:weary: :disappointed:

https://github.com/docker-library/elasticsearch/runs/3677832209

Not sure how https://github.com/elastic/elasticsearch/issues/51670#issuecomment-648717390 ended up with https://github.com/elastic/elasticsearch/pull/52519 :confused:

tianon:
I guess it's only `elasticsearch`, so we could revert that one to track 7.14 instead of 7.x and get the rest updated to 7.15.

I really think we should bring back the deprecation warning on all these images directing users to switch their usage, even if we continue updating them, since they're not actually maintained as part of this program (and users really should be using the Elastic images these pull directly instead).

pugnascotia:
@tianon is there anything Elastic can help with here?

tianon:
I'm not sure -- I guess an issue like this was bound to happen eventually (and I think we've had at least one minor instance before), so I'm not sure how to proceed here. :confused:

Irrespective of this issue, how do you feel about adding something like this to the image description?

> These images are a mirror of images maintained, built, and published by [Elastic](https://elastic.co) in [their `docker.elastic.co` registry](https://www.docker.elastic.co). It is highly recommended that all users adjust their usage to pull directly from those images instead for continued and expanded support.

pugnascotia:
@tianon I'll ask the appropriate people about that.

It wasn't obvious to me what the problem was here with the latest image version, but if it's something I can help with then please just ping me.

tianon:
The overly shortened version is that we were asked about the `FROM scratch` approach in use here, we brought up a lot of concerns with it (and how it certainly wouldn't be accepted if it were proposed here as-is without the ELK special casing), which seemed to be enough to have shifted the conversation to other ideas, but now here we are. :confused:

pugnascotia:
@tianon we're looking at revising (simplifying) our build process for the next minor release, but we'd still like to unblock the 7.15.0 and 7.15.1 releases if there's any way to make that possible?

pugnascotia:
@tianon @yosifkit We've published 7.16.0, where we've reverted Elasticsearch to drop the `scratch` stuff. Can you do whatever is necessary to build 7.16.0 versions of the ELK images please? 🙏 🙇

tianon:
Nice, yep!  I opened https://github.com/docker-library/elasticsearch/pull/201, https://github.com/docker-library/logstash/pull/103, and https://github.com/docker-library/kibana/pull/95 to see where we're at, and that led me to https://github.com/elastic/dockerfiles/pull/89#issuecomment-990119630 and https://github.com/elastic/dockerfiles/pull/89#discussion_r766061555 :grimacing:

PR #10972, opened 2021-09-27
Update stable version to 13.8

surli:


surli:
@yosifkit Hi, I'm not sure to understand why the check is failing. Apparently it's related to `tomcat:8-jdk11-adoptopenjdk-hotspot` except that since #10824 we don't reference this image anymore. Sounds like the check is seeing it in other versions of XWiki images we already deployed?

yosifkit:
The test checks everything still listed in the library file, since that is the "supported" versions.

All official images are rebuilt when either their build context changes (e.g. library file updated for newer `Dockerfile`) or their base image is updated (e.g. `debian` -> `python:3`). We strive to publish updated images at least monthly for Debian and Ubuntu so that all dependent images get updates for OS packages (i.e. security and bugfix updates). See also our [FAQ entry on CVE's](https://github.com/docker-library/faq/tree/73f10b0daf2fb8e7b38efaccc0e90b3510919d51#why-does-my-security-scanner-show-that-an-image-has-cves).

To ensure that all supported images get updates when base images are updated, they need to be `FROM` a current image. Since `xwiki:12`, `xwiki:13.4`, and their `postgres` variants are `FROM` an image that is no longer being maintained, they will no longer be rebuilt for OS security updates. Since users expect the images to be updated in a timely fashion, if those older releases are no longer going to be updated (for either an xwiki bump or an OS change so they can keep receiving rebuild updates), then they should likely be removed from the `library/xwiki` file.

yosifkit:

surli:
> since #10824 we don't reference this image anymore

So apparently I was wrong on that: @manuelleduc fixed the xwiki docker files to use temurin in 13.7 and 12.10.x but apparently he forgot to put the new commit hash for 12.10.x and forgot to fix it for 13.4.x. In theory, all images should now rely on temurin.

PR #10981, opened 2021-09-29
maven: switch openjdk latest to 17

carlossg:


library/maven
```diff
@@ -41,7 +41,7 @@ Architectures: amd64, arm64v8
 GitCommit: 0a7b1ef030687a5d95ce3e7567a4c279ab87dc81
 Directory: openjdk-11-slim
 
-Tags: 3.8.2-openjdk-16, 3.8.2, 3.8.2-openjdk, 3.8-openjdk-16, 3.8, 3.8-openjdk, 3-openjdk-16, 3, latest, 3-openjdk, openjdk
+Tags: 3.8.2-openjdk-16, 3.8.2, 3.8-openjdk-16, 3.8, 3-openjdk-16, 3, latest
```
yosifkit, line 44: Did you also want `3.8.2`, `3.8`, `3`, and `latest` to move to JDK 17?

carlossg: yes

yosifkit: Is that a change that you are working on? It looks like I can add commits if you'd like me to just move them down.

carlossg: sorry I was reading it on the phone and didn't catch the issue, it should be fixed now

PR #10991, opened 2021-09-30
actions: auto cancel builds if user pushes another commit

gdams:
Currently, if a user pushes several commits to a PR the CI will try to build each commit. This wastes both time and resources.

This PR adds a task that will cancel any other builds for the same PR.

### DIFF
--- a/.github/workflows/test-pr.yml
+++ b/.github/workflows/test-pr.yml
@@ -10,6 +10,11 @@ defaults:
 env:
   # https://github.com/docker-library/bashbrew/issues/10
   GIT_LFS_SKIP_SMUDGE: 1
+  
+# Cancel existing runs if user makes another push
+concurrency:
+  group: "${{ github.ref }}"
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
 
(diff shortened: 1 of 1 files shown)


.github/workflows/test-pr.yml
```diff
@@ -10,6 +10,11 @@ defaults:
 env:
   # https://github.com/docker-library/bashbrew/issues/10
   GIT_LFS_SKIP_SMUDGE: 1
+  
+# Cancel existing runs if user makes another push
+concurrency:
+  group: "${{ github.ref }}"
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```
tianon, line 17: Just want to note since it's ambiguous from the docs, `github.ref` is going to be of the format `refs/pull/:prNumber/merge` (not `refs/heads/<branch_name>` as the https://docs.github.com/en/actions/learn-github-actions/contexts#github-context docs imply -> https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#pull_request).

Just a minor suggestion to adjust for consistency:

```suggestion
# cancel existing runs if user makes another push
concurrency:
  group: ${{ github.ref }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

tianon (changes requested):
Nice, this is great; thank you!  We should probably do the same in the other script too, but this is a great start (and that one's a huge pain to test because it has to live on the origin; see #10879 :weary:).

gdams:
Updated PTAL

tianon (approved):

PR #11004, opened 2021-10-01
Fedora 20211001

cverna:
Remove Fedora 32 - EOL
Fedora 35 is now in Beta
Fedora 36 is now rawhide


yosifkit:
Diff: https://gist.github.com/docker-library-bot/1d4180e91635608c3808b27f7610fcf0

PR #11008, opened 2021-10-02
Friendica Release 2021.09 | Wrong Commit

nupplaphil:
I accidentally used a wrong commit for the Friendica Release 2021.09, so at least the SHA256 wasn't right ;-)

Thanks to this check finding it: https://doi-janky.infosiftr.net/job/multiarch/job/amd64/job/friendica/126/console


nupplaphil:
@yosifkit / @tianon - Is there are problem with Jenkins or with my commits? The amd64 jobs constantly fail:
https://doi-janky.infosiftr.net/job/multiarch/job/amd64/job/friendica/127/console

yosifkit:
Sorry, the server got full but is cleaned up now.

PR #11019, opened 2021-10-05
Switch to Temurin base image and replace Java 16 with 17

keeganwitt:


keeganwitt:
> The Actions failure is that `eclipse-temurin:8-jdk` doesn't have `s390x` support:
> 
> https://github.com/docker-library/official-images/blob/bcdbf922f5803db113b936a98929c0a2a74ce910/library/eclipse-temurin#L7-L9

Sorry, I should have left a question about this: Is this one of those intermittently not available problems like we had with the adoptopenjdk image? Or is that really an invalid combo?

PR #11024, opened 2021-10-05
eclipse-temurin: add windows ltsc-2022 support

gdams:


gdams:
@tianon @yosifkit any idea why the action failed?

yosifkit:
I think it is just a transient GitHub actions issue; I'll restart it.

gdams:
@yosifkit failed again :/

tianon:
Yeah -- I've filed an issue with GitHub support; we've got another failing in the same obscure way :disappointed:

tianon:
One thing I have noticed here (that's totally unrelated) is that the Windows variants are in the wrong order -- for the manifest list to work correctly, you'll want them listed in descending order (ltsc2022, then 1809, then ltsc2016).

tianon:
(Otherwise when Windows pulls the manifest list, it'll see that 2016 is "supported" and just pull that one, giving a much worse experience overall.)

tianon:
Other than that, the changes look correct/sane to me.

gdams:
>One thing I have noticed here (that's totally unrelated) is that the Windows variants are in the wrong order -- for the manifest list to work correctly, you'll want them listed in descending order (ltsc2022, then 1809, then ltsc2016).

So do I need to change the order in which they appear in this `eclipse-temurin` file? Or is there something else I need to change?

gdams:
>Yeah, just adjusting the order in this file will fix that.

Thanks! I've pushed an update which changes the order

PR #11050, opened 2021-10-07
Some bumps for Haskell

AlistairB:
:wave: The Haskell images have been a [bit stuck on installing some dependencies](https://github.com/haskell/docker-haskell/issues/45) via debian packaging, which seems to be dead at this point.

This includes changes to switch away from the debian packaging and [directly install the haskell compiler and the cabal build tool](https://github.com/haskell/docker-haskell/pull/46). We are also now doing sha256 verification on top of the gpg verify step. With this done, it includes the missing version bumps as well.

I'm assuming each of these commits will be applied independently and the versions released? (or I might need multiple stacked PRs?)

cc @psftw


tianon:
Looks good!  As soon as the builds at https://doi-janky.infosiftr.net/job/multiarch/job/amd64/job/haskell/ finish, it should be safe to make a new PR dropping the backfilled versions. :+1:

PR #11063, opened 2021-10-07
Clarify gpg verification

hasufell:
Some docker image developers not understanding what a hash function is have previously told me my installer has to gpg-verify all the tarballs instead of their checksums.

Of course that's nonsense, because then the internet, all distros (including ubuntu) and even blockchains in this world would be broken.

Adding this as a note here should silence this.

Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,6 +236,7 @@ The purpose in recommending checksum verification is to verify that the artifact
 Below are some examples:
 
 -	**Preferred**: *download over https, PGP key full fingerprint import and `asc` verification, embedded checksum verified.*
+-	**Note:**: verifying the PGP signature of the hash file only is as secure as verifying the signature of the data itself and common practice (given that it's a strong hash)
 
 	```Dockerfile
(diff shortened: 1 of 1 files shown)


hasufell:
> However, to properly do it "justice" it ends up getting a bit wordy -- perhaps there's a decent guide or article someone has written elsewhere that we could link to instead of including all this detail here?
> 
> (Otherwise I'd probably prefer to punt this addition to be part of #6282 instead so we can have something like a dedicated file to describe low-level details like this better and keep the high-level overview short.)

Could you clarify whether you're interested in me improving this PR? And if so what exactly do you need?

I don't know of any guide explaining GPG+hash use. It's somewhat basic cryptography knowledge.

tianon:
Honestly, I don't have a really strong opinion on the best way to incorporate this yet. I was hoping we could discuss it a bit more and help tip the scales, but it sounds like you're not up for that (which is totally fair).

Given that, I'm going to add this to the backlog, and we'll come back to it later; thanks for the suggestion!

PR #11081, opened 2021-10-09
Navy Linux Docker image support

unixlabs:
Request for the Navy Linux Docker image.


unixlabs:
can you please help on this error 

tianon:
Sorry for the delay :bow:

I'll try to explain this more clearly -- what I expect to see for a base image like this is a repository with a branch on it (can be the main branch if you want, more on that in a second) that has a `Dockerfile` that isn't much more than `FROM scratch` + `ADD xyz-foo-bar.tar.gz /` + `CMD ["bash"]` side-by-side with the committed tarball.  The primary challenge is that the tarball needs to not be modified by any commits in the history, so what I usually do is have source code (sometimes the `Dockerfile` too) on a `main` or `master` branch, and then have a separate `dist` branch that is only ever one commit above my `main` branch which adds the necessary tarball (with a force push that includes the new tarball in the same single commit for every image update).  For supporting multiple architectures, I usually rename that `dist` branch after the architecture, such as `dist-amd64`.

If there's anything I can do to help set that up, please let me know!

unixlabs:
> Sorry for the delay bow
> 
> I'll try to explain this more clearly -- what I expect to see for a base image like this is a repository with a branch on it (can be the main branch if you want, more on that in a second) that has a `Dockerfile` that isn't much more than `FROM scratch` + `ADD xyz-foo-bar.tar.gz /` + `CMD ["bash"]` side-by-side with the committed tarball. The primary challenge is that the tarball needs to not be modified by any commits in the history, so what I usually do is have source code (sometimes the `Dockerfile` too) on a `main` or `master` branch, and then have a separate `dist` branch that is only ever one commit above my `main` branch which adds the necessary tarball (with a force push that includes the new tarball in the same single commit for every image update). For supporting multiple architectures, I usually rename that `dist` branch after the architecture, such as `dist-amd64`.
> 
> If there's anything I can do to help set that up, please let me know!

@tianon Please review the updated pull request. and Thanks for the help.

yosifkit:
Clearing out old pull requests without activity. We sincerely apologize for the long delays. :bowing_man: Not to excuse our slow or missing replies, we often have to prioritize maintaining the current images over adding new images. We thank you for your contribution. :heart: :heart:

> is it reasonably popular, or does it solve a particular use case well?

When this was opened, the Navy Linux project was very new with a small community (like all of the sudden CentOS fork/clones at the time), but in the interim, it doesn't look like Navy Linux was able to gain much popularity/usage and it seems like the distro isn't really maintained anymore since it hasn't had a release in years. So, unfortunately, I don't think Navy Linux is suitable for inclusion in the program at this time.

PR #11112, opened 2021-10-15
Update ubuntu images now that impish is released

alex:


alex:
cc: @tianon, @mwhudson

PR #11154, opened 2021-10-21
Update Amazon Linux images

kgharris:
Hi - we have a new update ready to merge - Thanks!


tianon:
Looks like `prcarpen` should actually be @Timidger? :sweat_smile:

PR #11203, opened 2021-10-29
Adding the support for aarch64 on JDK11 and JDK8

narkedi:


narkedi:
@yosifkit Not able to see the latest images for aarch64, which should have been there as this PR is merged. Can you please confirm if we are missing something. Thanks !!

yosifkit:
Apologies, the Jenkins job generator got stuck because we have to manually approve the job DSL script whenever there is a new image+arch combo. :open_mouth: :disappointed: They should be building shortly.

PR #11213, opened 2021-11-01
composer 2.1.10

alcohol:


alcohol:
We renamed `master` to `main` at https://github.com/composer/docker; this seems to be causing issues?

alcohol:
> We renamed `master` to `main` at https://github.com/composer/docker; this seems to be causing issues?

Resolved.

PR #11215, opened 2021-11-01
Update Rust to 1.56.1

sfackler:


teohhanhui:
This should be fast tracked as it's a security release...

https://blog.rust-lang.org/2021/11/01/cve-2021-42574.html

PR #11226, opened 2021-11-02
Telegraf v1.20.3 using su

reimda:


yosifkit:
As far as I can tell, `EUID` doesn't work in Alpine's `sh`. So you probably have to check `id -u` instead:

```sh
if [ "$(id -u)" -ne 0 ]; then
```

```console
$ docker run -it --rm alpine sh
Unable to find image 'alpine:latest' locally
latest: Pulling from library/alpine
Digest: sha256:e1c082e3d3c45cccac829840a25941e679c25d438cc8412c2fa221cf1a824e6a
Status: Downloaded newer image for alpine:latest
/ # env
HOSTNAME=3ea2ed9c337f
SHLVL=1
HOME=/root
TERM=xterm
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
PWD=/
/ # echo "$EUID"

/ # echo "$euid"

/ # id -u
0
```

reimda:
Hi @yosifkit, thanks for catching my error. I've updated to use `id -u`.

PR #11232, opened 2021-11-03
Update sl container for sl7 to current (build-date=20211103)

sl-team:


jcpunk:
Those changes look right to me.

PR #11236, opened 2021-11-04
[security] Update golang

tianon:
Changes:

- https://github.com/docker-library/golang/commit/d5ee058: Update 1.17 to 1.17.3
- https://github.com/docker-library/golang/commit/b93b23c: Update 1.16 to 1.16.10

### DIFF
--- a/library/golang
+++ b/library/golang
@@ -5,124 +5,124 @@ Maintainers: Tianon Gravi <admwiggin@gmail.com> (@tianon),
              Johan Euphrosine <proppy@google.com> (@proppy)
 GitRepo: https://github.com/docker-library/golang.git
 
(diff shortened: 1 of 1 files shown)


tianon:
The failures we're getting are because the builds are signed by `2F528D36D67B69EDF998D85778BD65473CB3BD13`, which is technically a _subkey_ of `EB4C1BFD4F042F6DDDCCEC917721F63BD38B4796`. :disappointed:

Maybe we should adjust our `--recv-keys` line to explicitly try fetching the subkey instead so any keyserver that doesn't have the subkey specifically fails the fetch and pgp-happy-eyeballs tries another?

However, this *did* successfully work/finish on our build test, so I think we're good to merge anyhow (I've restarted a few times and it fails on different entries, so it's just a keyserver availability issue and the auto-retries we do in the real builds should smooth it over).

PR #11245, opened 2021-11-05
Update photon image

gerrit-photon:
Update photon image

### DIFF
--- a/library/photon
+++ b/library/photon
@@ -4,24 +4,24 @@ Maintainers: Fabio Rapposelli <fabio@vmware.com> (@frapposelli),
 GitRepo: https://github.com/vmware/photon-docker-image.git
 Directory: docker
 
-Tags: 4.0, 4.0-20211029, latest
+Tags: 4.0, 4.0-20211105, latest
 Architectures: amd64, arm64v8
-GitFetch: refs/heads/x86_64/4.0-20211029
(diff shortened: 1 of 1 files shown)


library/photon
```diff
@@ -4,24 +4,24 @@ Maintainers: Fabio Rapposelli <fabio@vmware.com> (@frapposelli),
...
+arm64v8-GitCommit: 90d7767179863f1e72fd142afeb817d933f3185b
 
-Tags: 3.0, 3.0-20211029
+Tags: 3.0, 3.0-20211105
 Architectures: amd64, arm64v8
-GitFetch: refs/heads/x86_64/3.0-20211029
-GitCommit: 9ecec586340fa0ce18b656261c5499dfb36d0067
+GitFetch: refs/heads/x86_64/3.0-20211105
+GitCommit: 85635c4557b95a12ab72d947033399e4fb7a795f
 arm64v8-GitFetch: refs/heads/aarch64/3.0-20211029
```
yosifkit, line 18: Is `3.0` on `arm64v8` still forthcoming?

If it stays as-is, then the `20211029` build for `arm64v8` will now be tagged as `3.0-20211105`; is that ok?

yosifkit: Newest commit has the updated `3.0` on `arm64v8`.

PR #11251, opened 2021-11-08
Updated mainline nginx to 1.21.4

thresheek:


yosifkit:
> https://github.com/docker-library/official-images/tree/bc8fca373290a0796e63244eb2103b12ae95534d#image-build

Is there anywhere the release `tar.gz` for `pkg-oss` is published so that a signature could be checked or at least a checksum embedded in the Dockerfile?

We also recommend at least a `-f` on the `curl` to ensure that non-200 responses cause a failure.

thresheek:
Thanks for the review @yosifkit !

Indeed there are no signed releases for pkg-oss.  I've implemented the checksum checks for it, and made curl fail on errors.  PR updated.

PR #11288, opened 2021-11-12
Groovy 4.0.0-beta-2

keeganwitt:


tianon:
You probably want to update the tags to also be `-beta-2`, right? :innocent:

keeganwitt:
> 
> 
> You probably want to update the tags to also be `-beta-2`, right? 😇

Yep. Thank you!

I realized yesterday that the last couple commits I had "Gradle" in the message instead of "Groovy" also (doh!). Too late to fix now...

PR #11292, opened 2021-11-12
Update clojure lein variants to 2.9.8

cap10morgan:
New upstream release


cap10morgan:
Oops, some bad commits in here. I'll force-push the right one.

PR #11294, opened 2021-11-12
library/perl: Add Debian bullseye, drop Debian stretch

zakame:
This switches the base image for Perl to use Debian bullseye.  We also
drop updating the tags still using Debian stretch, and also add arm32v5
and mips64le for multiarch.

https://github.com/Perl/docker-perl/issues/88
https://github.com/Perl/docker-perl/pull/107


zakame:
Thanks @yosifkit ! :tada:

PR #11301, opened 2021-11-15
Add alpine images

keeganwitt:


library/groovy
```diff
@@ -6,36 +6,58 @@ GitRepo: https://github.com/groovy/docker-groovy.git
...
 Tags: 3.0.9-jdk11, 3.0-jdk11, jdk11
 Architectures: amd64, arm32v7, arm64v8, ppc64le, s390x
-GitCommit: e46b8fbd268e1b4559178b87358aac7385f694d2
+GitCommit: 573aabd7a3b1387db10c82806f72bed00bda2770
+Directory: jdk11
+
+Tags: 3.0.9-jdk11-alpine, 3.0-jdk11-alpine, jdk11-alpine
+Architectures: amd64, arm32v7, arm64v8, ppc64le, s390x
+GitCommit: 573aabd7a3b1387db10c82806f72bed00bda2770
 Directory: jdk11
```
yosifkit, line 20: All of the Alpine images should be pointing to a different directory. Otherwise, like the diff shows, they'll just be the Ubuntu based images tagged as "alpine".

keeganwitt: Good catch. That was a copy/paste error.

tianon:
Looks generally sane -- the only thing that's a little eyebrow-raising for me is the installation of the VCSes, especially so many of them.  I wonder both what the size impact of having all those pre-installed is on the overall image and what percentage of users actually wants/needs `mercurial` and/or `subversion` these days? :grimacing:

For Alpine variants, we generally recommend that they come with just the bare minimum, since users typically want to use them specifically because they're so small (see also our standard `<version>-alpine` docs blurb: https://github.com/docker-library/docs/blob/master/.template-helpers/variant-alpine.md), _and_ because packages on Alpine generally install very, very quickly (even if they're fairly large).

keeganwitt:
Shoot, that's another stupid copy/paste mistake.  There's no reason for this image to have VCSs.  I'll pull them.

tianon:
(And to be clear, that's not a blocker, just something I'd like to make sure you definitely want to do since they'll be harder to remove later if you so decide :+1:)

PR #11342, opened 2021-11-19
Update aerospike 5.7.0.8

mcoberly2:


tianon:
That (forked!) `tini` downloaded without any verification is definitely a bit alarming -- can you elaborate on what problem that's solving?  If `SIGUSR1` should restart/reload aerospike, shouldn't that be functionality that gets implemented in aerospike itself?

mcoberly2:
We have a new PR to address the concerns in https://github.com/docker-library/official-images/pull/11342#issuecomment-974540516

PR #2432, opened 2016-12-08
Add Vault 0.6.3

jefferai:

PR #11349, opened 2021-11-19
Telegraf: add capabilities to telegraf binary

powersj:
This adds the CAP_NET_RAW and CAP_NET_BIND_SERVICE capabilities to the telegraf binary during the image build.


yosifkit:
It's a bit confusing, but as far as I can tell, Docker still does not fully support extended attributes in image layers (https://github.com/moby/moby/issues/40375, https://github.com/moby/moby/issues/35699, https://github.com/moby/moby/issues/1070). See also the discussion in https://github.com/docker-library/official-images/pull/2432 where the suggested workaround is to conditionally apply the `setcap` in the entrypoint script.

powersj:
Thanks for those links and for catching this. I have put up a [follow-up PR](https://github.com/influxdata/influxdata-docker/pull/552) in our repo to make the change in the entrypoint script. I will go ahead and close this request for now.

Thanks again!

PR #11354, opened 2021-11-19
Remove mips64le from friendica

tianon:
It's been failing to build for a while:

    + apt-get install -y --no-install-recommends rsync bzip2 msmtp tini gosu
    Reading package lists...
    Building dependency tree...
    Reading state information...
    E: Unable to locate package gosu

### DIFF
--- a/library/friendica
+++ b/library/friendica
@@ -2,6 +2,7 @@
 
 Maintainers: Friendica <info@friendi.ca> (@friendica), Philipp Holzer <admin@philipp.info> (@nupplaphil)
 GitRepo: https://github.com/friendica/docker.git
+GitFetch: refs/heads/stable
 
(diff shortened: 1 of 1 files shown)


tianon:
To update your `generate-stackbrew-library.sh` for these changes:

For b112259, I'd suggest adding the same `GitFetch: refs/heads/stable` in https://github.com/friendica/docker/blob/4ff2cfa3d2713e10d458d548e6eabdadeab9d80d/generate-stackbrew-library.sh#L59

For excluding `mips64le`, I'd suggest adding something like this: https://github.com/docker-library/cassandra/blob/f58183960ff693e4ec0c28f9f85453d69f646243/generate-stackbrew-library.sh#L93-L96

PR #11372, opened 2021-11-23
Bump to latest Mono 6.12.0.122 release

akoeplinger:
Includes fix for https://github.com/mono/docker/issues/89

/cc @tianon


tianon:
Since these are all buster (or newer), you don't actually need `apt-transport-https` anymore: https://packages.debian.org/buster/apt-transport-https :eyes:

akoeplinger:
@tianon updated, thanks!

tianon:
FWIW, you can also move both `GitFetch` and `GitCommit` up next to `GitRepo` in the "global" entry, if you wanted to (if you're maintaining this file by hand, that might be helpful).

PR #11373, opened 2021-11-23
Add haskell 9.2.1

AlistairB:
Normally we only list the last 2 major versions, however 9.0.1 was a bit botched. The majority of users are still using 8.10 line, so I am leaving it in for now.


tianon:
Nice :+1:

PR #11375, opened 2021-11-23
Update clojure image

cap10morgan:
This changes a few things:

- Makes openjdk-17 the new default base image since it's the new LTS Java release
- Cleans up the consistency of the tags so that they all conform to what we've long been claiming about them (e.g. slim-bullseye as the default distro everywhere)
- Uses the various build tools as ENTRYPOINTs (but still allows for `docker run -ti image bash`) in openjdk-17+ images (still uses CMDs in older ones)
- Uses clj as the ENTRYPOINT in clojure:latest instead of lein
- Uses clj as the ENTRYPOINT instead of lein in all images that don't specify a built tool

### DIFF
--- a/library/clojure
+++ b/library/clojure
@@ -2,7 +2,7 @@ Maintainers: Paul Lam <paul@quantisan.com> (@Quantisan),
              Wes Morgan <wes@wesmorgan.me> (@cap10morgan)
 Architectures: amd64, arm64v8
 GitRepo: https://github.com/Quantisan/docker-clojure.git
-GitCommit: 562f43944e1f957f7456a6466ec6082aebb76922
+GitCommit: 813917fbfab83a6ae37f09c9d5a7bf4acb704896
 
 Tags: latest
 Directory: target/openjdk-17-slim-bullseye/latest
--- a/library/clojure
+++ b/library/clojure
@@ -7,10 +7,10 @@ GitCommit: 562f43944e1f957f7456a6466ec6082aebb76922
 Tags: latest
 Directory: target/openjdk-17-slim-bullseye/latest
 
-Tags: openjdk-8-buster, openjdk-8-lein-buster, openjdk-8-lein-2.9.8-buster
+Tags: openjdk-8-lein-buster, openjdk-8-lein-2.9.8-buster
 Directory: target/openjdk-8-buster/lein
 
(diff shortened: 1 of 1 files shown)


yosifkit:
```dockerfile
RUN chmod +x /usr/local/bin/entrypoint
```

Each `RUN` that modifies the files from a `COPY` is creating another copy of the file in the image layers; although it is small, it is best practice to not cause duplication. `chmod +x` can be tracked in git so that the file will have the correct execute permissions when checked out.

---

```dockerfile
COPY rlwrap.retry /usr/bin/rlwrap.retry
RUN mv /usr/bin/rlwrap /usr/bin/rlwrap.real && \
mv /usr/bin/rlwrap.retry /usr/bin/rlwrap && \
chmod +x /usr/bin/rlwrap
```

This is similar to the above, any `RUN` changing files already in the image is creating duplication in the image. It might be simplified by just putting `rlwrap.retry` as `/usr/local/bin/rlwrap` (since that is earlier in the `PATH`) and having it call the real one as `/usr/bin/rlwrap`.

Something like this?

```dockerfile
COPY rlwrap.retry /usr/local/bin/rlwrap

# or some clever symlinks too
COPY rlwrap.retry /usr/local/bin/rlwrap.retry
RUN ln -s /usr/bin/rlwrap /usr/local/bin/rlwrap.real && \
	ln -s /usr/local/bin/rlwrap.retry /usr/local/bin/rlwrap
```

---

> ```console
> clojure:openjdk-18-alpine
> Step 6/10 : RUN mv /usr/bin/rlwrap /usr/bin/rlwrap.real && mv /usr/bin/rlwrap.retry /usr/bin/rlwrap && chmod +x /usr/bin/rlwrap
> ---> Running in 37960e2ae7a6
> mv: can't rename '/usr/bin/rlwrap': No such file or directory
> ```

The `openjdk:18-alpine` image is based on Alpine `3.14` but that doesn't have an `rlwrap` executable available until the next alpine release ([currently in edge](https://pkgs.alpinelinux.org/contents?file=rlwrap&path=&name=&branch=edge)).

cap10morgan:
Hopefully a rebuild on top of the new alpine3.15-based openjdk images will rectify that rlwrap missing error. I'll address the other feedback shortly. Thanks!

cap10morgan:
OK hopefully everything is addressed now

PR #11377, opened 2021-11-24
[varnish] fix tags

gquintard:
the latest `7.0.1` release got wrongly tagged with `7.1`


gquintard:
let's forget about this, https://github.com/docker-library/official-images/pull/11380 is more urgent and important

PR #11378, opened 2021-11-24
added support for MW 1.37 

gesinn-it-gea:
... via https://github.com/wikimedia/mediawiki-docker/pull/103 and https://github.com/wikimedia/mediawiki-docker/pull/104

### DIFF
--- a/library/mediawiki
+++ b/library/mediawiki
@@ -2,16 +2,26 @@ Maintainers: David Barratt <dbarratt@wikimedia.org> (@davidbarratt),
              Kunal Mehta <legoktm@wikimedia.org> (@legoktm),
              addshore <addshorewiki@gmail.com> (@addshore)
 GitRepo: https://github.com/wikimedia/mediawiki-docker.git
-GitCommit: b19002eb9ee59f3e1856307b39b0d0303f16e962
+GitCommit: 8c813a7a252dce971d2e8a3c31bad67e6fd19850
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, ppc64le
 
-Tags: 1.36.2, 1.36, stable, latest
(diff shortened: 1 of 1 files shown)


library/mediawiki
```diff
@@ -2,16 +2,26 @@ Maintainers: David Barratt <dbarratt@wikimedia.org> (@davidbarratt),
...
+
+Tags: 1.36.2, 1.36
 Directory: 1.36/apache
 
-Tags: 1.36.2-fpm, 1.36-fpm, stable-fpm
+Tags: 1.36.2-fpm, 1.36-fpm
 Directory: 1.36/fpm
 
-Tags: 1.36.2-fpm-alpine, 1.36-fpm-alpine, stable-fpm-alpine
+Tags: 1.36.2-fpm-alpine, 1.36-fpm-alpine
```
tianon, line 24: Are the older releases (1.36, 1.35, and 1.31) still actively supported?  I guess 1.35 is "lts" and 1.31 is "legacylts" so they probably both are (worth double checking IMO :sweat_smile:), but 1.36 seems like maybe should be dropped?

Maybe just some tag updates?  I found on https://www.mediawiki.org/wiki/Download the following:

> To users of MediaWiki versions version 1.34 and earlier: These versions are no longer supported. Please update to a newer version of MediaWiki:
> MediaWiki 1.37.0 (download .zip, download .tar.gz) - stable
> MediaWiki 1.36.2 (download .zip, download .tar.gz) - legacy
> MediaWiki 1.35.4 (download .zip, download .tar.gz) - long-term support (LTS)

legoktm: It was on my list to drop 1.31, but didn't get around to it yet. Both 1.35 and 1.36 are still supported.

tianon: Alright, fair enough!

PR #11380, opened 2021-11-24
Revert "[varnish] release 7.0.1"

andrewwiik:
This reverts commit 8657dc71f43afb77e38c0a1268dc50642d03bd3b.


gquintard:
well, this is embarrassing. By introducing unprivileged containers, I broke older versions of varnish. We have a plan to de-correlate future versions, but it's going to take a bit of time, and Thanksgiving is upon us, so it'd be good to revert that commit to avoid breaking more user setups

gquintard:
thanks @yosifkit

soleson:
Yes, thanks for the quick resolve @yosifkit !

PR #11379, opened 2021-11-24
alpine: add 3.15.0 as latest, drop EOL 3.11

ncopa:

PR #11389, opened 2021-11-25
haxe: add alpine3.15, drop alpine3.11, add bullseye as default, drop haxe 3 for newer distros

andyli:
* add alpine3.15, drop alpine3.11 (https://github.com/docker-library/official-images/pull/11379)
 * add bullseye, replaced buster as the default
 * do not build haxe 3.x for the newer distros, since haxe 3.x is not compatible with the newer ocamls


andyli:
The failed build looks like a random network issue - failed to fetch a deb from http://deb.debian.org.

PR #11395, opened 2021-11-26
Rebuild XWiki 12.10.x using the latest source changes

vmassol:


vmassol:
See https://jira.xwiki.org/browse/XDOCKER-207 to understand why we need a rebuild of 12.10.x.

tianon:
I'm a little confused here -- there aren't any actual source changes, so this is going to be a fully-cached build?

vmassol:
@tianon good catch. I had changed our template files between the 2 sha1 but I had forgotten to rebuild to propagate the changes to the 12.x directories... I've now updated the sha1 to the latest one. See https://github.com/xwiki/xwiki-docker/compare/eb1c6b2527d4d24ae10f1dddf0066135bb9c6923..fab1516b838eeacc3c9b232b28936c690a06b1e9

You can apply this PR before or after https://github.com/docker-library/official-images/pull/11398 both order should work fine.

Thanks and sorry for the confuzion!

tianon:
Looks like my comments on https://github.com/docker-library/official-images/pull/11398#issuecomment-981931837 are relevant here now too: :sweat_smile:

> See [#3540 (comment)](https://github.com/docker-library/official-images/issues/3540#issuecomment-530925319), especially:
> 
> > We don't actively recommend using labels. If an image maintainer wants to have labels, that is fine, but label names should adhere to the image spec: https://github.com/opencontainers/image-spec/blob/v1.0.1/annotations.md
> 
> (I'd personally just drop the `maintainer` metadata entirely since it's inherited by any image that's `FROM` this one, but it's your call.)



> I guess I'd also recommend using a tag that's a _little_ more specific than `tomcat:9`, since that one is likely to change things like the JDK version pretty willy-nilly, but that's also up to you. :sweat_smile:

vmassol:
@tianon image updated with opencontainers LABELs. Thx!

PR #11397, opened 2021-11-28
Friendica Release 2021.12-rc & mips64le

nupplaphil:
- Release 2021.12-rc
- Includes `GitFetch`
- Includes way to build gosu even for mips64le (@tianon for debian, I'm now doing it this way https://github.com/tianon/gosu/blob/master/INSTALL.md#from-debian )


tianon:
Nice, thanks!

PR #11398, opened 2021-11-29
Update XWiki to 13.10

surli:


tianon:

(I'd personally just drop the `maintainer` metadata entirely since it's inherited by any image that's `FROM` this one, but it's your call.)

tianon:
I guess I'd also recommend using a tag that's a _little_ more specific than `tomcat:9`, since that one is likely to change things like the JDK version pretty willy-nilly, but that's also up to you. :sweat_smile:

vmassol:
@tianon Thanks for your pieces of advices, much appreciated.

About:

> (I'd personally just drop the maintainer metadata entirely since it's inherited by any image that's FROM this one, but it's your call.)

I'm not sure I understand. I wanted to let users of the xwiki docker image know who are the maintainers. If I remove this metadata, how can anyone find that info by looking at the docker image?

Thanks

vmassol:
@tianon I had replaced the `MAINTAINER` command by a `LABEL`  one because the former is deprecated https://docs.docker.com/engine/reference/builder/#maintainer-deprecated and the recommendation is to use a `LABEL`.

I couldn't find what `We don't actively recommend using labels` actually means. Could you elaborate and explain what are the problems/consequences of using `LABEL maintainer="XWiki Development Team <committers@xwiki.org>"`?

Thanks a lot

yosifkit:
> Could you elaborate and explain what are the problems/consequences of using `LABEL maintainer="XWiki Development Team <committers@xwiki.org>"`?

The largest drawback is that any user that creates an image `FROM` it inherits all labels so it can incorrectly attribute maintainership to dependent images. This is also true of the `MAINTAINER` instruction. For images we directly maintain, we rely on users knowing where they pulled an image from so that they can find help by looking at the readme content hosted on Docker Hub and GitHub rather that relying on metadata in the image.

The XWiki image is welcome to move from the deprecated `MAINTAINER` Dockerfile instruction to using `LABEL`, but image label names should adhere to the image spec: https://github.com/opencontainers/image-spec/blob/v1.0.1/annotations.md. I think `org.opencontainers.image.authors` is the equivalent to maintaner.

vmassol:
@yosifkit Thanks a lot for the explanation. I've now updated the LABELs to follow the opencontainers conventions. Waiting for @surli to update this PR with sha1 `1d2ea57cbe91d8eff11f19d1ca8ddd378a3b871b`  for XWiki 13.10.

PR #11417, opened 2021-11-30
Update Matomo

J0WI:


yosifkit:
`env` does not have safely parseable output since environment variables can have new lines in their values. I think `eval` is going to the equivalent for `sh` versus the original `bash` implementation of `file_env`. Maybe the size difference for adding `bash` to the alpine images would be ok?

AtworkDev:
Any info when version 4.6.1 will be available in a docker image? Im getting the following error **_Your codesbase is running the old version 4.5.0 and we have detected that your Database hasa already been upgraded to the newer version 4.6.1..._**. Many thanks in advance :)

yosifkit:
Ref: https://github.com/matomo-org/docker/pull/255#issuecomment-985136502

PR #11424, opened 2021-12-01
Update JRuby 9.2 to 9.2.20.1

headius:


headius:
I don't understand the error in the failed job. Does not seem to be related to my change.

tianon:
Yep, was just a GitHub flake. :+1:

PR #11432, opened 2021-12-02
Releasing Neo4j 4.4.0

jennyowen:
Now introducing ARM64 support!
Thank you.

### DIFF
--- a/library/neo4j
+++ b/library/neo4j
@@ -7,564 +7,465 @@
 Maintainers: Chris Gioran <chris@neo4j.com> (@digitalstain),
              Jenny Owen <jenny.owen@neo4j.com> (@jennyowen),
              Bledi Feshti <bledi.feshti@neo4j.com> (@bfeshti)
-
-
-Tags: 4.3.7, 4.3.7-community, 4.3, 4.3-community, community, latest
 GitRepo: https://github.com/neo4j/docker-neo4j-publish.git
+
+Tags: 4.4.0, 4.4.0-community, 4.4, 4.4-community, community, latest
+Architectures: amd64, arm64v8
+GitCommit: 91eaae2fb453548687bd41e483bfb49a04b08e4d
+amd64-Directory: 4.4.0/amd64/community
+arm64v8-Directory: 4.4.0/arm64v8/community
+
+Tags: 4.4.0-enterprise, 4.4-enterprise, enterprise
+Architectures: amd64, arm64v8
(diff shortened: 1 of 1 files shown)


yosifkit:
> Images must have a single `Dockerfile` per entry in the library file that can be used for multiple architectures.
>
> \- https://github.com/docker-library/official-images/tree/9979a09cb49a16a905c386c3675ff4adc67b6c9f#multiple-architectures

Separate Dockerfiles per architecture is reserved for base images (i.e. `FROM scratch`).  It looks like the only difference in the two sets of Dockerfiles is the `tini` url + `sha256`; I think that could be derived with like a `case` statement based on `dpkg --print-architecture` (like [pypy](https://github.com/docker-library/pypy/blob/698f5c6574d8ecd034dfc2acf84fb687dfe338e6/3.8/bullseye/Dockerfile#L27-L48)) or maybe just install [`tini`](https://packages.debian.org/bullseye/tini) from Debian packages?

yosifkit:
> apt-get -y purge --auto-remove curl

The removal of `curl` in a separate `RUN` layer from installation won't save any space.

jennyowen:
OK that's hopefully fixed now.

Another thing we'd like to do for the 4.4.0 release is provide another neo4j image, which is a slightly pared down version of neo4j, just used for operating the database (like backing it up and stuff). 
Would it be ok with you guys if we added that here as an official neo4j image (also under neo4j)? It would mean that each new release from 4.4.0 onwards would be 4 images instead of 2, and I'm aware we already have quite a lot of active neo4j images listed here.

I don't plan to include any of that stuff in *this* PR, I didn't want any issues with it to block this full 4.4.0 release.

tianon:
Yep, looks OK now, thanks!

For this pared down version, I wonder if there's a straightforward way to make the fat images be `FROM` the slim images somehow?  The number of (unique) images is definitely going to be a concern. :grimacing: (we're already a bit over 100 again :see_no_evil:)

jennyowen:
@tianon @yosifkit I thought adding the `Architectures` line would be enough to get a Neo4j image running on an ARM64 machine, but `docker run -it --rm neo4j:4.4.0` is telling me `docker: no matching manifest for linux/arm64/v8 in the manifest list entries.`.

Have I missed some important step?

yosifkit:
That's on us; our Jenkins pipeline generator requires script approval every time a new image+arch is added. I went and approved it and then started the build.

jennyowen:
thanks @yosifkit it's all there now and working wonderfully :bowing_woman: 
chef's kiss

PR #11441, opened 2021-12-02
Update Rust to 1.57.0

sfackler:


sfackler:
@tianon It looks like a bunch of the tags didn't get properly pushed up to docker hub (e.g. 1.57.0-buster for the linux/amd64 platform). Any idea what's going on there?

tianon:
There was a Debian update very recently, so the build queue is still very, very full and working its way through everything.

sfackler:
Got it, thanks!

PR #11449, opened 2021-12-04
Update aerospike 5.7.0.8

mcoberly2:

1. Based on your excellent feedback from the previous, now Closed, Pull Request (https://github.com/docker-library/official-images/pull/11342) we built a release of tini (tini-static), signed it (SHA256) and verified we are downloading what we built in the container.

2. We implemented a feature that allows tini to restart the child process without restarting the container. This allows our database to warm restart / config reload without losing the shared memory index. Implementing this in the database server would be messy at best and not something we want to implement. Although, I could be wrong.

3) We don't deal with all the edge cases of SIGTERMs in our app.  This is some of the idiosyncrasies of the container environment.  We expect tini to already do this. This is a reason why tini and dumb-init exist. We used dumb-init before this and there were no concerns previously.

### DIFF
--- a/library/aerospike
+++ b/library/aerospike
@@ -5,7 +5,7 @@ Maintainers: Lucien Volmar <lucien@aerospike.com> (@volmarl),
 Tags: ee-5.7.0.8
 Architectures: amd64
 GitRepo: https://github.com/aerospike/aerospike-server-enterprise.docker.git
-GitCommit: 2d11fe93cec1c1adc86968a26b1b8827f61dc81f
+GitCommit: 9c3d341e34497e1e1127c67c5c487eff4b76ed99
 
 Tags: ce-5.7.0.8
 Architectures: amd64


yosifkit:
From what I can see, signal handling is already in Aerospike's C code, so I am a bit confused what other edge cases or idiosyncrasies of containers you see for PID 1? As far as I understand it, any catchable, blockable, or ignorable signal that would normally kill the process (like ["Term" and "Core" dispositions](https://man7.org/linux/man-pages/man7/signal.7.html)) are instead ignored by default for PID 1, but Aerospike already appears to handle them?

Perhaps the Aerospike `SIGUSR1` handler or another signal handler could be modified to do something similar to `SIGTERM` and unlock the main deadlock thread but instead of the main thread exiting, it would do an `execve` of itself (`/proc/self/exe` can be useful here) with the original argv after all the cleanup?


We are mildly concerned about using a custom fork of `tini`. Because it is still called `tini`, it may confuse users when they use other images containing `tini` or via the `--init` flag in Docker. It looks like both `tini` and `dumb-init` upstream developers label this functionality as "out of scope" (https://github.com/krallin/tini/issues/71#issuecomment-272652178, https://github.com/krallin/tini/pull/146, https://github.com/krallin/tini/issues/186, https://github.com/Yelp/dumb-init/issues/156, https://github.com/Yelp/dumb-init/issues/254). So, if there are plans to keep the fork maintained instead of adding more handling into `aerospike-server`, then it should probably be under a different name than just `tini`.

---

> The `Dockerfile` should be written to help mitigate interception attacks during build. Our requirements focus on three main objectives: verifying the source, verifying author, and verifying the content; these are respectively accomplished by the following: using https where possible; importing PGP keys with the full fingerprint in the `Dockerfile` to check signatures; embedding checksums directly in the `Dockerfile`. All three should be used when possible. Just https and embedded checksum can be used when no signature is published. 
>[...]
>The purpose in recommending checksum verification is to verify that the artifact is as expected. This ensures that when remote content changes, the `Dockerfile` also will change and provide a natural `docker build` cache bust. As a bonus, this also prevents accidentally downloading newer-than-expected artifacts on poorly versioned files.
>
>\- https://github.com/docker-library/official-images/tree/2969d4f64a77bc8343486f44f149ce74b70bc5a6#security

Downloading the sha256sum from the same source as the artifact, doesn't provide the same level of content verification that embedding the checksum in the Dockerfile does.

mcoberly2:
Closing...

PR #11485, opened 2021-12-12
Solr fix for Log4j CVE-2021-44228

dsmiley:
This is a docker-only fix for Log4j CVE-2021-44228 by setting a system property.
https://solr.apache.org/security.html#apache-solr-affected-by-apache-log4j-cve-2021-44228

I verified the system property is set.  Technically the updates to 5.x & 6.x wasn't necessary as they are not vulnerable but the docker-solr project manages the scripts universally by a common template and I'd rather not address that at this time.

### DIFF
--- a/library/solr
+++ b/library/solr
@@ -8,12 +8,12 @@ GitRepo: https://github.com/docker-solr/docker-solr.git
 
 Tags: 8.11.0, 8.11, 8, latest
 Architectures: amd64, arm64v8
-GitCommit: 6596ada4279a9640e9e339d21469a3ff319fa0ce
+GitCommit: 2e10b447bd468aecc2992916a796c327f9c80209
 Directory: 8.11
 
 Tags: 8.11.0-slim, 8.11-slim, 8-slim, slim
 Architectures: amd64, arm64v8
-GitCommit: 6596ada4279a9640e9e339d21469a3ff319fa0ce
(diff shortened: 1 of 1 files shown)


tianon:
Can you rebase your branch? I *think* how far behind it is is throwing off the CI (although I'm not 100% positive).

Easiest way is something like `git pull --rebase https://github.com/docker-library/official-images.git master` and then a force push.

dsmiley:
Ok I think I did that correctly.  The commit hash is correct.

tianon:
Much better, thanks!

tianon:
Just a reminder on all those older versions (many of which might not be actively "supported" anymore?), removing tags here will remove them from the "Supported" section on the Hub readme (and will prevent us from spending cycles rebuilding them on the official build servers), but the tags will still be available to users who want them.  (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)

sylus:
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-45046

Note that previous mitigations involving configuration such as to set the system property log4j2.noFormatMsgLookup to true do NOT mitigate this specific vulnerability. Log4j 2.16.0 fixes this issue by removing support for message lookup patterns and disabling JNDI functionality by default

https://github.com/docker-solr/docker-solr/issues/397

PR #11487, opened 2021-12-13
[hitch] 1.7.2

gquintard:
new version, which notably fixes the build errors from the outdated test certificate


tianon:
Nice, thank you! :+1:

PR #11489, opened 2021-12-13
maven: eclipse-temurin alpine

carlossg:


tianon:
I'll also note that you could also move up the `RUN apk add --no-cache curl tar bash procps` line (which would then match your `ibmjava` variants :joy:) and get slightly better caching. :eyes:

PR #11491, opened 2021-12-13
Adding support for IBM Semeru Runtime Open Edition JDK17

narkedi:


five-iron:
@narkedi Is there a plan to include ubi-based Semuru images in the official repo?

narkedi:
@five-iron As there are no official UBI  base images at DH (will verify once again), we can't include the ubi based images.

five-iron:
@narkedi I see, so using an unofficial base is not allowed? Any chance of a java 17 version of this then https://hub.docker.com/u/ibmsemeruruntime?

narkedi:
@five-iron For the first point, yes using unofficial base image is not allowed. Yes, you can find them on the un-official repo mostly by next week.

PR #11526, opened 2021-12-15
Telegraf v1.21.0

MyaLongmire:


yosifkit:
This has the `setcap` changes from https://github.com/docker-library/official-images/pull/11472 but I think we were still waiting on the entrypoint improvement from https://github.com/influxdata/influxdata-docker/pull/558.

PR #11532, opened 2021-12-16
kong 2.7.0

gszr:
Bump Kong to 2.7.0.

### DIFF
--- a/library/kong
+++ b/library/kong
@@ -3,19 +3,19 @@ GitRepo: https://github.com/Kong/docker-kong.git
 # needs "Constraints" to match "centos" (which this image is "FROM")
 
 Tags: 2.7.0-alpine, 2.7.0, 2.7, alpine, latest
-GitCommit: 395889c3a30c54d7579d9dfba93e5ae8fe62547d
+GitCommit: 4896c55914e46d044183c6a6541cf50366f5cffa
 GitFetch: refs/tags/2.7.0
 Directory: alpine
 Architectures: amd64, arm64v8
 
 Tags: 2.7.0-ubuntu, 2.7-ubuntu, ubuntu
-GitCommit: 395889c3a30c54d7579d9dfba93e5ae8fe62547d
+GitCommit: 4896c55914e46d044183c6a6541cf50366f5cffa
 GitFetch: refs/tags/2.7.0
 Directory: ubuntu
 Architectures: amd64, arm64v8
 
 Tags: 2.7.0-centos, 2.7-centos, centos
-GitCommit: 395889c3a30c54d7579d9dfba93e5ae8fe62547d
+GitCommit: 4896c55914e46d044183c6a6541cf50366f5cffa
 GitFetch: refs/tags/2.7.0
 Directory: centos
 Architectures: amd64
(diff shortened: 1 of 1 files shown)


tianon:
`apt-get upgrade` and `yum update` are not used in official images since it causes images to be larger than they should with packages being duplicated across layers. We recommend that images instead rely on periodic rebuilds that come from base image updates to get updated packages from lower layers.

Background: 
>

Since our build system makes heavy use of Docker build cache, just rebuilding the the Dockerfile won't cause any change (to help minimize rebuilds that don't have newer packages installed). So we rely on periodic base image updates.

> We strive to publish updated images at least monthly for Debian and Ubuntu. We also rebuild earlier if there is a critical security need. Many Official Images are maintained by the community or their respective upstream projects, like Alpine and Oracle Linux, and are subject to their own maintenance schedule.
>
> \- from the same FAQ link

hutchic:
@tianon we're aware but had thought / hoped it would work since the `apt-get upgrade` / `yum upgrade` / `apk upgrade` are contained in `else` logic and therefore they're not in the official image 🙏 ?

tianon:
Yeah, my concern is that it might inadvertently move outside that `else` block and because it's part of a change we've already reviewed, we might not catch it.  Maybe you could add a comment next to it which makes it clear it needs to stay inside that `else` block?

```dockerfile
...
    else \
# this needs to stay inside this "else" block so that it does not become part of the "official images" builds (https://github.com/docker-library/official-images/pull/11532#issuecomment-996219700)
      apt-get upgrade -y ; \
    fi; \
...
```

kikito:
The aforementioned comment has been included.

yosifkit:
This change in the `2.7.0` Ubuntu Dockerfile (compared to `2.6.0`) makes it unable to build on `arm64v8`.

```diff
-      && curl -fL https://download.konghq.com/gateway-${KONG_VERSION%%.*}.x-ubuntu-$(cat /etc/os-release | grep UBUNTU_CODENAME | cut -d = -f 2)/pool/all/k/kong/kong_${KONG_VERSION}_$(dpkg --print-architecture).deb -o /tmp/kong.deb \
+      && curl -fL https://download.konghq.com/gateway-${KONG_VERSION%%.*}.x-ubuntu-$(cat /etc/os-release | grep UBUNTU_CODENAME | cut -d = -f 2)/pool/all/k/kong/kong_${KONG_VERSION}_amd64.deb -o /tmp/kong.deb \
```

Build error when building on `arm64v8`:

```console
Reading package lists...
Building dependency tree...
Reading state information...
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 kong:amd64 : Depends: libpcre3:amd64 but it is not installable
              Depends: perl:amd64 but it is not installable
              Depends: zlib1g-dev:amd64 but it is not installable
E: Unable to correct problems, you have held broken packages.
```

Are arm64v8 images still desired for `2.7.0`?
- If no, then it should be dropped from the `Architectures` list in the [library/kong file](https://github.com/docker-library/official-images/blob/c3f80209402174cc73015ebec5db6cf9d2525a5d/library/kong#L15)
- if yes, then the Dockerfile should be updated to be architecture independent and the `library/kong` should be updated.

PR #11537, opened 2021-12-17
Update Solr to 8.11.1

dsmiley:
This is a bug-fix release.  See the [official announcement](https://lists.apache.org/thread/nfbjb1oqwppl9zk2c5okygzjckd6p63l) that lists a number of bug fixes including a few dependency updates (including Log4j 2.16) for addressing transitive vulnerabilities.  I'd like to re-iterate that we don't think we're vulnerable to the reported vulnerabilities against Log4J < 2.16.0 any more but of course upgrading eases everyone's minds, and security scanners don't get these nuances.

I also updated the README.md to [include a header at the very top](https://github.com/docker-solr/docker-solr#note-not-vulnerable-to-log4j-2-log4shell) explaining the "Log4shell" situation.
Question: If we wanted to update the README at a higher frequency than these images, do we need to publish to official-images again?

### DIFF
--- a/library/solr
+++ b/library/solr
@@ -8,12 +8,12 @@ GitRepo: https://github.com/docker-solr/docker-solr.git
 
 Tags: 8.11.0, 8.11, 8, latest
 Architectures: amd64, arm64v8
-GitCommit: 6596ada4279a9640e9e339d21469a3ff319fa0ce
+GitCommit: 2e10b447bd468aecc2992916a796c327f9c80209
 Directory: 8.11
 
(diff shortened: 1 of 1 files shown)


yosifkit:
> Question: If we wanted to update the README at a higher frequency than these images, do we need to publish to official-images again?

Nope. Updating the `docker-solr/docker-solr` README can be updated as much as you like, it doesn't affect any of the builds or the docs on Docker Hub. Feel free to send PRs for the content in https://github.com/docker-library/docs/tree/master/solr to update the [Docker Hub content](https://hub.docker.com/_/solr) as needed.

PR #11071, opened 2021-10-08
Add check to naughty-commits to ensure that referenced commit exists in the upstream

yosifkit:
This uses the `GitFetch` branch to ensure that the commit does not come from outside the upstream repo.  This will cause a few images to need to set a default `GitFetch` since they use a branch other than `master`.

Also update the last single-line-format library files (`sourcemage` and `thrift`) to rfc2822 format to avoid `refs/heads/*`; should either of these just be emptied since they are deprecated? (https://github.com/docker-library/docs/pull/1433 and https://github.com/docker-library/docs/pull/1732)

```console
$ ./naughty-commits.sh
 - arangodb:
   - commit cd9cff3458077a32daaf1d6d31899056c08a07c6:
     - is not in the specified branch refs/heads/master
   - commit 42fc6401e569423cab28e71b0085dc78300caa84:
     - is not in the specified branch refs/heads/master

 - bonita:
   - commit a1d9ee5e31d38958aa553cc7f9d465f1151d902f:
     - is not in the specified branch refs/heads/master

fatal: couldn't find remote ref refs/tags/clear-linux-35090
 - clearlinux:
   - commit 89ec5b23516ddc81799aeb66e989fd21bcd679b6:
     - unable to to find specified branch refs/tags/clear-linux-35090

 - couchdb:
   - commit ee358e62d72bdd72fd69d67ba7fbc80580502270:
     - is not in the specified branch refs/heads/master

 - crate:
   - commit 896c3f63e8e3d4746019e379a7aefb5225b050e3:
     - is not in the specified branch refs/heads/master

fatal: couldn't find remote ref refs/heads/master
fatal: couldn't find remote ref refs/heads/master
fatal: couldn't find remote ref refs/heads/master
fatal: couldn't find remote ref refs/heads/master
fatal: couldn't find remote ref refs/heads/master
fatal: couldn't find remote ref refs/heads/master
fatal: couldn't find remote ref refs/heads/master
fatal: couldn't find remote ref refs/heads/master
 - eclipse-temurin:
   - commit b60534569e25b3e721441185c1c36ebac17cf2ce:
     - unable to to find specified branch refs/heads/master
   - commit b54d893a02d99a837374d6ca2195ba26fd9154ce:
     - unable to to find specified branch refs/heads/master
   - commit 4b9b98b68117a9476073d2d8461369e3624fac09:
     - unable to to find specified branch refs/heads/master
   - commit 37f56aac1070ee4c160d442e00d81bbd4c86bb8a:
[...]

PR #11538, opened 2021-12-17
Update Docker image for Silverpeas version 6.2.2

mmoqui:


mmoqui:
The naughty test fails because it doesn't find the tag 6.1.3 and 6.2.2 from refs/heads/master. It is normal because in our own process each stable version of Silverpeas comes from its own branch (and hence not from master). For example, the tag 6.1.3 comes from the branch 6.1.x and 6.2.2 comes from the branch 6.2.x. So that we can then release a patch version of Silverpeas for each minor version. 
Previously, this worked fine. But it seams you have changed some rules that break our process to maintain a different lifecycle for each minor versions of Silverpeas in their own branch.

Do you have a suggestion here how we can fix it without breaking our own process?

library/silverpeas
```diff
@@ -2,8 +2,8 @@
 Maintainers: Miguel Moquillon <miguel.moquillon@silverpeas.org> (@mmoqui)
 GitRepo: https://github.com/Silverpeas/docker-silverpeas-prod.git
 
-Tags: 6.2.1, latest
-GitCommit: e97e55f0a21aae880a0054ef764cea0cb37e1e05
+Tags: 6.2.2, latest
+GitCommit: c6b68c650da8a95d9b258b4fe1f930fa9be737a6
```
yosifkit, line 6: Yeah, we added a test in https://github.com/docker-library/official-images/pull/11071 that ensures commits exist in the upstream. The fix is just telling it where to fetch the commits from. Adding a `GitFetch: ref/heads/6.2.x` line for the respective branch on each block should work.

Like this:
```suggestion
GitCommit: c6b68c650da8a95d9b258b4fe1f930fa9be737a6
GitFetch: refs/heads/6.2.x
```

mmoqui: Ok, thanks

mmoqui:
> Are 6.1.x and 6.0.x still actively supported release lines?
> 
> Just as a reminder, removing tags here will remove them from the "Supported" section on the Hub readme (and will prevent us from spending cycles rebuilding them on the official build servers), but the tags will still be available to users who want them. (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)

Currently 6.1.x is actively supported. 6.0.x is indeed no more supported. We will remove it from the lib def at a next release of Silverpeas (normally in the month).

PR #11542, opened 2021-12-17
[RFC] Added Unit as an official Library image

thresheek:
Hello!

This PR is to start the process of adding [Unit](http://unit.nginx.org/) docker images under the umbrella of Docker Library.

Unit is a polyglot application server, and we're already using per-language Official Images to build our images upon. Basically for a given variant, we build Unit and the language module and ship it like that, enabling users to work in the already familiar environment for a language. That works really well for us as a vendor, so I would like to thank everyone involved in making Docker Library a success story as it is!

One of the issues we do have however shipping our own images is the overall finicky and cumbersome process of building multi-architecture images, which we see an increased demand for (especially for armv8a since Graviton2 and M1 introduction).  Enrolling Unit images under a Docker Library umbrella would be a great win for us in this case, similarly to what we (F5, Inc.) already enjoy with `nginx` images publishing.

At this point I would like to iron out technical issues we might encounter for this process, and when it's ready, work on Documentation part (which involves converting what we already have on http://unit.nginx.org/installation/#docker-images).

This PR currently points to my own fork of [unit repo](https://github.com/nginx/unit/).  The reason for that is I would like to have a formal review from Library maintainers for Dockerfiles/tags/library definition files/etc to happen before I actually push code to the main Unit repository (we're also using mercurial upstream, but that's another story - the github repository is automagically updated).  Once that is sorted out, I'll submit the patches for Unit internal review and will update this PR with a fresh set of library definitions. Hope that is OK, but let me know if you would like that plan to be amended or thrown out. :-)

To help fill out a review checklist:
 - I am an employee of F5, Inc. / NGINX Inc., and I have [commit rights](https://github.com/nginx/unit/commits?author=thresheek) to the repo. Our devteam leaders are OK with the enrollment to the Official Library.
 - Unit is published under [Apache 2.0](https://github.com/nginx/unit/blob/master/LICENSE) source code license.

cc @vbart @artemkonev

# Checklist for Review

    - > I am an employee of F5, Inc. / NGINX Inc., and I have [commit rights](https://github.com/nginx/unit/commits?author=thresheek) to the repo. Our devteam leaders are OK with the enrollment to the Official Library.
    - Apache 2.0
    - https://github.com/docker-library/docs/pull/2187

### DIFF
--- a/library/unit
+++ b/library/unit
@@ -1,0 +1,60 @@
+# this file is generated via https://github.com/nginx/unit/blob/b9bc222021e77bbdfb12576b3e315b962cf6b399/pkg/docker/Makefile
+
+Maintainers: Unit Docker Maintainers <docker-maint@nginx.com> (@nginx)
+GitRepo: https://github.com/nginx/unit.git
+
+Tags: 1.29.1-go1.20, go1.20, go1, go
+Architectures: amd64, arm64v8
+GitFetch: refs/heads/branches/packaging
+GitCommit: b9bc222021e77bbdfb12576b3e315b962cf6b399
+Directory: pkg/docker
+File: Dockerfile.go1.20
+
+Tags: 1.29.1-jsc11, jsc11, jsc
+Architectures: amd64, arm64v8
+GitFetch: refs/heads/branches/packaging
+GitCommit: b9bc222021e77bbdfb12576b3e315b962cf6b399
+Directory: pkg/docker
+File: Dockerfile.jsc11
+
+Tags: 1.29.1-node18, node18, node
+Architectures: amd64, arm64v8
+GitFetch: refs/heads/branches/packaging
+GitCommit: b9bc222021e77bbdfb12576b3e315b962cf6b399
+Directory: pkg/docker
+File: Dockerfile.node18
+
+Tags: 1.29.1-perl5.36, perl5.36, perl5, perl
+Architectures: amd64, arm64v8
+GitFetch: refs/heads/branches/packaging
+GitCommit: b9bc222021e77bbdfb12576b3e315b962cf6b399
+Directory: pkg/docker
+File: Dockerfile.perl5.36
+
+Tags: 1.29.1-php8.2, php8.2, php8, php
+Architectures: amd64, arm64v8
+GitFetch: refs/heads/branches/packaging
+GitCommit: b9bc222021e77bbdfb12576b3e315b962cf6b399
+Directory: pkg/docker
+File: Dockerfile.php8.2
+
+Tags: 1.29.1-python3.11, python3.11, python3, python
+Architectures: amd64, arm64v8
+GitFetch: refs/heads/branches/packaging
+GitCommit: b9bc222021e77bbdfb12576b3e315b962cf6b399
+Directory: pkg/docker
+File: Dockerfile.python3.11
+
+Tags: 1.29.1-ruby3.2, ruby3.2, ruby3, ruby
+Architectures: amd64, arm64v8
+GitFetch: refs/heads/branches/packaging
+GitCommit: b9bc222021e77bbdfb12576b3e315b962cf6b399
+Directory: pkg/docker
+File: Dockerfile.ruby3.2
+
+Tags: 1.29.1-minimal, minimal
+Architectures: amd64, arm64v8
+GitFetch: refs/heads/branches/packaging
+GitCommit: b9bc222021e77bbdfb12576b3e315b962cf6b399
+Directory: pkg/docker
+File: Dockerfile.minimal


thresheek:
Arf, I've clicked the Submit PR button before I've actually pushed to my fork. Mea culpa.

thresheek:
Now that CI checks are green, let's continue with the remaining checklist boxes.

> - does it fit into one of the common categories? ("service", "language stack", "base distribution")

Yes, it's a "service".  Basically a standard way to launch your application on top of your favroutie language stack, similar to uwsgi, tomcat, php-fpm, with some parts of nginx.

> - existing official images have been considered as a base? (ie, if foobar needs Node.js, has FROM node:... instead of grabbing node via other means been considered?)

 That's exactly what we're using for all our images:

| Image | FROM |
| --- | --- |
| unit:1.26.1-go1.17 | FROM golang:1.17 |
| unit:1.26.1-jsc11 | FROM openjdk:11-jdk |
| unit:1.26.1-minimal | FROM debian:bullseye-slim |
| unit:1.26.1-node16 | FROM node:16 |
| unit:1.26.1-perl5.34 | FROM perl:5.34 |
| unit:1.26.1-php8.0 | FROM php:8.0-cli |
| unit:1.26.1-python3.9 | FROM python:3.9 |
| unit:1.26.1-ruby3.0 | FROM ruby:3.0 |

> - passes current tests? any simple new tests that might be appropriate to add? (https://github.com/docker-library/official-images/tree/master/test)

At least all the images pass the "run.sh" test.  That being said, we do have our own tests for the images that basically put a well known working application and it's configuration to a docker image and verify the output.  Not sure it's worth moving to official tests, but I'm open for that if there is a need.

yosifkit:
It is not necessary to copy your tests over here. Our current tests are mostly a smoke test to ensure we don't merge something obviously broken or insecure. CI in your own repo will help you catch any problems earlier anyway.

---

The build looks pretty straightforward; couldn't these instead just delete the source and build dependencies in the same `RUN` instead of using multiple stages? See also our [Multi-stage Builds](https://github.com/docker-library/faq/tree/73f10b0daf2fb8e7b38efaccc0e90b3510919d51#multi-stage-builds) entry in our FAQ.

thresheek:
Thanks @yosifkit !

I'll work on adapting the Dockerfiles for what would be suitable for official images inclusion - I somehow missed the "single FROM" part - guess "multi-stages builds ... approved for use" had a blinding effect :laughing:

thresheek:
I think I'm finally ready with the final tech parts of this PR - the CI checks are now green, and I think I addressed all the points that were made.

May I ask for a final review of the dockerfiles?

We're preparing the docs PR and I hope it will land soon.  When it's all done, my plan is to get the patches in the upstream `unit` repo, and then force-push the proper library file pointing to it.

Thanks!

thresheek:
I've added the remaining part, the PR to provide some documentation on how to use the images: https://github.com/docker-library/docs/pull/2187

cc @tianon @yosifkit 

Thanks!

yosifkit:
Sorry for the delay!  Docs PR looks good to me.

Just a few items to consider :monocle_face: and then hopefully this can be updated and merged. :pray::tada:

- :thinking: unbounded loops give me pause; while it seems very unlikely to become an endless loop, I think some sort of limit might still be useful on them. :shrug::confused: What do you think?
	```shell
	while [ ! -S /var/run/control.unit.sock ]; do echo ..; sleep ..; done`
	```
- update base images (a little bit rot since being opened; go removed 1.18 just recently, others can be updated as desired)
- it isn't currently a problem, just something to consider when using Dockerfiles within the upstream development repo: https://github.com/docker-library/official-images/pull/11100#issuecomment-1023481834
    - ensure that the build context remains minimal (like the current `pkg/docker` directory) to keep our `diff-pr.sh` script from including Unit's source tree :scream:
    - has the ability to update Dockerfiles outside of upstream releases like most packaging (sometimes necessary, as you know)

thresheek:
Thank you @yosifkit !

Indeed, we can introduce a hard limit on the only remaining endless loop (entrypoint script did evolve a bit since my submission).  Good catch!

For the second point, we actually have a separate `packaging` branch in the upstream repo that we use exactly for the purposes of making changes atop releases when it comes to packaging problems, be it CVE fixes or similar things.  Also, we don't have plans on changing the layout of `pkg/docker` any time soon - so that's covered too.

I'll work on merging the patches upstream, and will submit updates on this MR.

thresheek:
Hello!

I've finally gotten the patches in the main tree and updated the commit in this MR to point to the upstream repo.

bashbrew build works fine, images look good to me, so I think it's ready to be pushed.

Going forward, the plan is to fix the documentation over at https://unit.nginx.org/ to point to official images instead of what we're using now and point https://github.com/docker-library/docs/pull/2187 to it.

yosifkit (approved):
LGTM

I think `adduser`/`addgroup` will soon(?) not be part of the default Debian install, so it should eventually use `useradd`/`groupadd`, but it's fine for now. @tianon, is that changing in Bookworm?

tianon:
I've got a few comments - most of them are non-blocking though (for your information or decision based on our experience); I'll try to make which is which clear in the sections below by noting which ones are blocking. 🙇

---

```diff
+FROM golang:1.20 as BUILDER
+FROM eclipse-temurin:11-jdk as BUILDER
...
```

This is what I like to call a "square wheel" - once there's a new stable version of the base each of these images is `FROM`, this generic tag will update and might break your build.  I'd suggest pinning to a particular release like `golang:1.20-bullseye` instead so that _you_ get to choose when that square wheel rotates (and do your own relevant testing before it potentially fails to build on our infra 😅).

(Additionally, the `as BUILDER` looks like an unnecessary leftover from a multi-stage build attempt, but totally harmless if you'd rather leave it.)

---

**BLOCKING**

```diff
+LABEL maintainer="NGINX Docker Maintainers <docker-maint@nginx.com>"
```

See https://github.com/docker-library/official-images/issues/3540, especially https://github.com/docker-library/official-images/issues/3540#issuecomment-530925319:

---

In `apt-get` for both Debian and Ubuntu, `--no-install-suggests` is already the default, but harmless if you prefer being explicit.

---

```diff
+    && hg clone https://hg.nginx.org/unit \
+    && cd unit \
+    && hg up 1.29.1 \
```

Does `hg clone` not have a way to specify which thing to clone? 😅

(Maybe downloading the release tarballs would be better long-term?  especially with some checksum verification?  👀)

---

```diff
+    && /bin/true \
```

I'm guessing this is a quirk of your templating? (harmless either way)

---

```diff
+    && addgroup --system unit \
+    && adduser \
```

Like @yosifkit mentioned, this is one of those places the "square wheel" is definitely going to bite (`adduser`/`addgroup` will no longer be in the `minbase` set in Debian Bookworm).

ALSO, I would highly highly highly recommend fixing these to a specific UID/GID instead of letting them be assigned automatically (which one you choose doesn't really matter much, but it's helpful to have a fixed value; common choices are `999`, `1000`, or the default port number of the relevant service, which I don't think applies here).

---

Running `apt-get clean` in Ubuntu and Debian based containers is harmless, but unnecessary (and should be a no-op because it happens automatically during `install`): https://github.com/debuerreotype/debuerreotype/blob/5cf7949ecf1cec1afece267688bda64cd34a6817/scripts/debuerreotype-minimizing-config#L85-L109

---

Would any of those repeated compilation steps benefit from being grouped more closely together?  For example, you do `./configure ... --debug` for `unitd-debug`, then `make clean`, then do it all over again for `unitd`, then _again_ for `./configure go` (after a `./configure ... --debug` - do those stack nicely like that or does the latter overwrite the former's result?)

---

```diff
+CMD ["unitd", "--no-daemon", "--control", "unix:/var/run/control.unit.sock"]
```

Are there any other command-line flags to `unitd` that might be interesting for users to specify?  (In other words, is there value in making `--no-daemon` and/or `--control` part of the entrypoint script's logic instead so that other flags can be more easily specified by users?  perhaps interest from Unit maintainers in making both of those more easy to configure new defaults for during compilation so that they can be pre-configured instead?)

---

```diff
+            # even when the control socket exists, it does not mean unit has finished initialisation
+            # this curl call will get a reply once unit is fully launched
+            /usr/bin/curl -s -X GET --unix-socket /var/run/control.unit.sock http://localhost/
```

Will it also exit appropriately if the server goes away? (I think it will, but without `-f` maybe it won't return a non-zero exit code in that case?)

---

**NOT BLOCKING, BUT POSSIBLY IMPORTANT**

```diff
+    && ./configure $CONFIGURE_ARGS --cc-opt="$CC_OPT" --modules=/usr/lib/unit/debug-modules --debug \
+    && ./configure  \
+    && make -j $NCPU version \
+    && make clean \
+    && ./configure $CONFIGURE_ARGS --cc-opt="$CC_OPT" --modules=/usr/lib/unit/modules \
+    && ./configure  \
+    && make -j $NCPU version \
```

This section of code in the `minimal` variant looks odd -- is that doing the right thing?  (did you intend to have `./configure` with no arguments a second time there?  what is `make version` doing?)

thresheek:
Thanks for such a thorough and detailed review!

I've fixed most of the problems outlined, except for those quoted below.

---
> 
> In `apt-get` for both Debian and Ubuntu, `--no-install-suggests` is already the default, but harmless if you prefer being explicit.

Indeed, I would rather leave it explicitely mentioned.

---
> 
> ```diff
> +    && hg clone https://hg.nginx.org/unit \
> +    && cd unit \
> +    && hg up 1.29.1 \
> ```
> 
> Does `hg clone` not have a way to specify which thing to clone? 😅

I remembered `hg` not having an option to do so back in the days, but that doesn't seem to be true anymore. Fixed.

> (Maybe downloading the release tarballs would be better long-term? especially with some checksum verification? 👀)

I was thinking about downloading both .tar.gz and tar.gz.sha512 from https://unit.nginx.org/download/ and then checking them;  however it seems official images prefer to embed the checksum directly into a Dockerfile... That's something to consider, though. (And mercurial is heavily cached server side anyway, so doing a repo clone is not very expensive on our side).

---

> ```diff
> +    && /bin/true \
> ```
> 
> I'm guessing this is a quirk of your templating? (harmless either way)

Yeah, I tried to write one-template-fits-all for our build variants.  There are small oddities like that in the build phases here and there.

---

> Would any of those repeated compilation steps benefit from being grouped more closely together? For example, you do `./configure ... --debug` for `unitd-debug`, then `make clean`, then do it all over again for `unitd`, then _again_ for `./configure go` (after a `./configure ... --debug` - do those stack nicely like that or does the latter overwrite the former's result?)

I wanted to make builds in Docker be identical to what we provide via binary packages.  This meant that I needed to mirror hardening CFLAGS/LDFLAGS here as well - and if we use a single pair of "./configure; ./configure $LANG" the language modules will fail to link due to `-pie` flag.  This is why there are separate configure/make blocks for non-debug/debug builds, and the same for modules (modules require a "main" configure to succeed).
I'm also preparing a semi-related change to enable [njs](http://unit.nginx.org/configuration/#scripting) scripting support and that will require a slightly different set flags for the "main" configure as compared to the modules one.

---

> ```diff
> +CMD ["unitd", "--no-daemon", "--control", "unix:/var/run/control.unit.sock"]
> ```
> 
> Are there any other command-line flags to `unitd` that might be interesting for users to specify? (In other words, is there value in making `--no-daemon` and/or `--control` part of the entrypoint script's logic instead so that other flags can be more easily specified by users? perhaps interest from Unit maintainers in making both of those more easy to configure new defaults for during compilation so that they can be pre-configured instead?)

Apart from cosmetics, I don't think there are benefits of moving the flags to entrypoint.

I explicitly chose to specify control socket location since it varies between different OS - not the case for Debian/Ubuntu as used in this PR - but for a user migrating their macOS/Homebrew install to Docker I think there is value to be verbose.

And definitely "--no-daemon" will never be a default since it will be pretty much a breaking change for your typical unix installs...

---

> ```diff
> +            # even when the control socket exists, it does not mean unit has finished initialisation
> +            # this curl call will get a reply once unit is fully launched
> +            /usr/bin/curl -s -X GET --unix-socket /var/run/control.unit.sock http://localhost/
> ```
> 
> Will it also exit appropriately if the server goes away? (I think it will, but without `-f` maybe it won't return a non-zero exit code in that case?)

Hmm, yeah, in our tests it will wait for the server to initialize and respond; however on e.g. 5xx errors it will still happily exit with  code `0`.

I'll prepare a patch for that some time later; pretty sure we can live with that behaviour for now.

---

> **NOT BLOCKING, BUT POSSIBLY IMPORTANT**
> 
> ```diff
> +    && ./configure $CONFIGURE_ARGS --cc-opt="$CC_OPT" --modules=/usr/lib/unit/debug-modules --debug \
> +    && ./configure  \
> +    && make -j $NCPU version \
> +    && make clean \
> +    && ./configure $CONFIGURE_ARGS --cc-opt="$CC_OPT" --modules=/usr/lib/unit/modules \
> +    && ./configure  \
> +    && make -j $NCPU version \
> ```
> 
> This section of code in the `minimal` variant looks odd -- is that doing the right thing? (did you intend to have `./configure` with no arguments a second time there? what is `make version` doing?)

For `minimal` variant that block is basically a waste of resources/cpu time since there is no language module to build.   With it coming from a templated Dockerfile, I don't see an obvious way to fix that except maybe adding an env var specifying what variant we're currently building and case/if switch to skip this part for minimal.

thresheek:
And now with actual changes propagated to Dockerfiles... 😅q

PR #11544, opened 2021-12-17
Add new VzLinux official image

ktrushkin:
Hello,

We would like to add new VzLinux image.

Thank you!
# Checklist for Review

    * https://github.com/vzlinux/vz-docker-images
    * GPL v2
    * base distribution
    * https://github.com/docker-library/docs/pull/2095
-	[x] if `FROM scratch`, tarballs only exist in a single commit within the associated history?

### DIFF
--- a/library/vzlinux
+++ b/library/vzlinux
@@ -1,0 +1,6 @@
+Maintainers: Denis Silakov <dsilakov@virtuozzo.com> (@dsilakov), Kirill Trushkin <kirill.trushkin@virtuozzo.com> (@ktrushkin)
+GitRepo: https://github.com/vzlinux/vzlinux-cloud-instance-images.git
+GitFetch: refs/heads/main
+GitCommit: aa4532d72e3f196696c01d9589ec096be7948fec
+
+Tags: vzlinux8.5, latest


ktrushkin:
Are there any remarks?

tianon:
Sorry for the delay!  New image reviews usually take a lot more effort than image update review, so they often take a back seat (we're trying to be better about that). :bow:

I'd like to start by asking about the "popularity" aspect -- some of the normal metrics we use aren't really available here, presumably because it's so newly open source? :sweat_smile:

tianon:
I've also got a few comments about the Dockerization.

---

When you update this, you're going to need to force push to the branch to ensure there's only ever at most one tarball in the history (that the tarball doesn't get "updated" in the history of the provided commit).  How I usually handle this in my own repositories is to have a "main" branch that includes the source code (`Dockerfile`, etc), and then a separate "dist" branch that only ever includes a single commit above "main" which includes the tarball (force pushed on each update).  If you don't mind force-pushing to "main" the current structure is fine, but something to think about.

---

Regarding the labels, see https://github.com/docker-library/official-images/issues/3540#issuecomment-530925319, especially:

(`label-schema` has been deprecated for a long time now)

---

I'm not sure what `RUN yum reinstall -y vzlinux-release` is trying to accomplish -- can you elaborate on that?  Why wouldn't the base tarball be updated instead?  (It's also currently failing for me with `Installed package vzlinux-release-3:8.5-36.vl8.x86_64 not available.`)

tianon:
Of the packages installed in the tarball, I'm curious about a few packages that maybe aren't necessary to be installed in the base container images?  For example, all the `boost-*` packages and `smartmontools`.

tianon:
I'm updating the status of this PR to "draft" for now.  When it's ready for re-review, please remove the draft status and leave a comment (GitHub unfortunately does not notify maintainers for draft state changes).

yosifkit:

PR #11581, opened 2021-12-25
Update to spiped 1.6.2

TimWolla:


TimWolla:
I believe your automation broke, because the default branch of the source repository is called `main` :weary:

TimWolla:
Is this `GitFetch` requirement new? The branch is called `main` [since at least February](https://github.com/TimWolla/docker-spiped/commit/2e8a234451ab23c76e55be96b08d4a55016ce2fa) and I last updated the image in August.

Anyway, I could figure it out and it works :smile:

tianon:
Nicely done.  We updated the CI in #11071 to validate the `GitFetch` values that were supposed to be right all along, but thanks to GitHub not being strict didn't have to be.  That led to some edge case situations where we managed to review and merge an image update that was still only a PR on the origin repo, so this adds a bit more safety to the process (making sure the image changes are actually committed where they're supposed to be / advertised to be).

PR #11610, opened 2022-01-05
Friendica - Upgrade PHP & introduce new settings

nupplaphil:


tianon:
> ```dockerfile
> ENV FRIENDICA_SYSLOG_FLAGS 39 // LOG_PID | LOG_ODELAY | LOG_CONS | LOG_PERROR
> ```

I think this is going to set `FRIENDICA_SYSLOG_FLAGS` to the literal value "`39 // LOG_PID | LOG_ODELAY | LOG_CONS | LOG_PERROR`", but it looks like that latter portion was probably intended to be a comment?  Perhaps something like this was the intention instead:

```dockerfile
# 39 = LOG_PID | LOG_ODELAY | LOG_CONS | LOG_PERROR
ENV FRIENDICA_SYSLOG_FLAGS 39
```

nupplaphil:
Ups.. Thank you

PR #11620, opened 2022-01-06
Add arm64v8 support to Haskell

AlistairB:
And some related docs changes - https://github.com/docker-library/docs/pull/2101

arm64v8 does not include haskell stack as [it does not support it](https://github.com/commercialhaskell/stack/issues/2103). Many Haskell users do not use stack, so I did not want to hold this up waiting on it.

I did a fair restructure of the dockerfiles to support multiple processor architectures, mostly modelled after the golang images. Feedback appreciated.

### DIFF
--- a/library/haskell
+++ b/library/haskell
@@ -3,13 +3,16 @@ Maintainers: Alistair Burrowes <afburrowes@gmail.com> (@AlistairB),
 GitRepo: https://github.com/haskell/docker-haskell
 
 Tags: 9.2.1-buster, 9.2-buster, 9-buster, buster, 9.2.1, 9.2, 9, latest
-GitCommit: 71b980bfedb718c656ab6152998da8f500202469
+Architectures: amd64, arm64v8
+GitCommit: 5f8acc1199035266b1cc74fcc47d8ea114268076
 Directory: 9.2/buster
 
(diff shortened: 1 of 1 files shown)


yosifkit:
>`curl -sSL https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -; \`

1. Downloading keys without any verification of the key is not great; we instead recommend grabbing keys via `gnupg` via a full fingerprint (https://github.com/docker-library/official-images/tree/c3cfa2bebfb259a833a1c719a4b53f7849332db7#image-build)
   - use `gpg --batch -recv-keys --keyserver ... && gpg --armor --export > /etc/apt/trusted.gpg.d/name-for-key.gpg.asc` instead of `apt-key add`
   - the key can be limited to a specific apt repo with `signed-by` in the `sources.list` file (similar to https://github.com/docker-library/postgres/pull/921)

>`gpg --batch --verify SHA256SUMS.sig SHA256SUMS; \`

It seems the `SHA256SUMS` file is unused; maybe grep it for the `GHC_SHA256` that was used?

AlistairB:
Thanks @yosifkit great feedback!

I believe I have made the requested changes. Let me know if it still isn't quite right. Thanks.

yosifkit (approved):

PR #11623, opened 2022-01-07
Update Amazon Linux images

sktomer:
Hi,

This change adds the first docker container images for Amazon Linux 2022 (Tech Preview).

Regards,
Sumit


sktomer:
will re-create a new one.

PR #4502, opened 2018-06-26
Update Amazon Linux images (2.0.20180622.1, 2018.03.0.20180622)

ilianaw:
Amazon Linux 2 is available and is out of the candidate stage. We dropped the YYYY.MM versioning scheme in this release. "latest" now points to AL2.

We've updated the container image for 2018.03.

I've dropped references to all other versions, as we no longer generally support those (but as I understand it, this does not keep users from pulling those images as tags never get deleted from Docker Hub).

PR #11627, opened 2022-01-07
Update Amazon Linux images: add 2022.0.20211222.0

sktomer:
Hello,

This commit adds a (tech-preview) container image for the next generation of Amazon Linux : AL2022.

Thanks,
Sumit

### DIFF
--- a/library/amazonlinux
+++ b/library/amazonlinux
@@ -11,7 +11,7 @@ Maintainers: Amazon Linux <amazon-linux@amazon.com> (@amazonlinux),
              Preston Carpenter (@timidger),
              Richard Kelly (@rpkelly)
 GitRepo: https://github.com/amazonlinux/container-images.git
-GitCommit: c9ec4113f694e139e5b2fc2eaf02da0aaa3c7f6c
+GitCommit: 74eec9ec400d191dbe3398ca0cbba0ac9f892929
 
 Tags: 2.0.20211201.0, 2, latest
 Architectures: amd64, arm64v8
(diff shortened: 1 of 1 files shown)


sktomer:
Hello,

We tried to add new Tags in this commit for a new Amazon Linux major release 2022.
We did not modify the `latest` or `with-sources` tags.

However in the Munge PR / Diff Commit check failure, we see it is comparing 2022 with 2018.03 and 1.
Is this expected ?

Overall is the diff commit size a blocker for this pull request or something expected when new tags are added ?

For reference, similar pull request for when we added major release 2 : https://github.com/docker-library/official-images/pull/4502

tianon:
Yes, sorry, that's both somewhat expected and a little bit unexpected, and hopefully I can explain well enough why it's a little bit of both.

For base images, we don't really have a very meaningful `Dockerfile` to review (`FROM scratch` + `COPY ... /`), so we review the file listing of the rootfs tar itself.  In the case of _new_ base images, that often leads to a diff that's too long for a GitHub comment, so we usually post them to a gist instead and just link them here (so passing the CI check is not actually strictly necessary for this to move forward).

Now let's touch on why this is also a little bit _un_-expected in this specific case.  These new images are **significantly** larger than the previous versions.  Part of the reason for this is that there appear to be many packages/binaries/files that are not often included in container images like systemd, sshd, dhclient, grub, etc.

Almost every time we see "grub" inside a container image, it's a bug in the container rootfs creation process ending up including more things than are actually necessary to be able to successfully install more things, and certainly are not things most users need out of the container image, so they represent both wasted disk and wasted bandwidth. :grimacing:

You can read the current diff at https://github.com/docker-library/official-images/runs/4742149169?check_suite_focus=true#step:5:9, but fair warning that it is _very_ long. :see_no_evil:

stewartsmith:
Hrrm, good catch. Looks like we should go and look at our container recipe a bit closer.

sktomer:
Thanks for your feedback Tianon. We will work on the next steps to fix the image and submit a new request soon.

PR #11628, opened 2022-01-07
Update haproxy

tianon:
Changes:

- https://github.com/docker-library/haproxy/commit/7a14969: Update 2.4 to 2.4.11


TimWolla:
As a heads up: It looks like HAProxy 2.4.11 includes a very bad regression: https://github.com/haproxy/haproxy/issues/1510

wdauchy:
> As a heads up: It looks like HAProxy 2.4.11 includes a very bad regression: [haproxy/haproxy#1510](https://github.com/haproxy/haproxy/issues/1510)

fixed in 2.4.12 now we can probably update it here as well

PR #11638, opened 2022-01-10
Update Rust and add Alpine tag for 3.15

parasew:


yosifkit:
It seems the maintainer (@sfackler) was waiting to publish this change until the next Rust release: https://github.com/rust-lang/docker-rust/pull/95#issuecomment-1007650353.

sfackler:
Yep, that's happening this Thursday so it'll probably be easier to just wait until then to avoid extra tag churn.

yosifkit:
Ok, I'll close this for now. Thanks @sfackler and @parasew.

PR #11648, opened 2022-01-11
Update Amazon Linux Images

kgharris:
Corrected pull request - please create new container images for our amzn2 containers - thanks!


yosifkit:
These additions seem a bit strange for a container image; are they on purpose?

```diff
+usr/libexec/vi
+usr/share/vim/
+usr/share/vim/vim82/
+usr/share/vim/vim82/defaults.vim
```

kgharris:
Those are coming from our vim-minimal package now, which is not new - only those two files are.  So yes it's on purpose.

tianon:
Fair enough, thanks for confirming!

kgharris:
You're welcome - thanks for making sure!

PR #11661, opened 2022-01-12
Mark ltsc2016 as an unacceptable base now that it is EOL

tianon:
See https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/base-image-lifecycle

This also reformats our list to match the order of the table we link to, and includes the non-EOL items commented out so it's easier to compare/update over time.

### DIFF
--- a/naughty-from.sh
+++ b/naughty-from.sh
@@ -21,28 +21,19 @@ _is_naughty() {
 		# https://techcommunity.microsoft.com/t5/Containers/Removing-the-latest-Tag-An-Update-on-MCR/ba-p/393045
 		*=mcr.microsoft.com/windows/*:latest) return 0 ;;
 
-		# 20H2 is not *technically* EOL until 05/10/2022, but its use is discouraged here given the existence of ltsc2022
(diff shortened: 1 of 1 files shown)


tianon:
Affected images/maintainers:

- `caddy`: @hairyhenderson
- `eclipse-temurin`: @gdams
- `golang`: @tianon @yosifkit @proppy
- `haxe`: @andyli
- `hylang`: @paultag @hylang/docker
- `julia`: @tianon @yosifkit
- `mongo`: @tianon @yosifkit
- `nats`: @derekcollison @kozlovic @wallyqs @variadico
- `nats-streaming`: @kozlovic
- `openjdk`: @tianon @yosifkit
- `python`: @tianon @yosifkit

To be explicitly clear, the expected fix is to remove the affected variants. :smile:
(I'll be working on PRs to remove this from those I'm involved in shortly.)

Additionally, I've gone and made sure that all affected images have been successfully (re-)built and pushed with the final updated image which was released by Microsoft yesterday. :+1:

PR #11670, opened 2022-01-13
Releasing Neo4j 4.4.3

jennyowen:
I also deleted the 4.0.X images because we don't support them any more.

Thanks!


jennyowen:
ok looks like I messed up the release process a bit on our internal systems. It should be fixed now :crossed_fingers:

tianon:
Nice, thank you! :+1:

PR #11678, opened 2022-01-14
haxe: add windowsservercore ltsc2019/ltsc2022

andyli:


tianon:
As seen on https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/base-image-lifecycle, 1809 and ltsc2019 are the same thing (so we prefer to use the former, since it's a more specific kernel revision, where the latter could potentially change, although at this point it's unlikely to do so).

You can see this clearly by comparing the two images:

```console
$ diff -u <(docker run --rm gcr.io/go-containerregistry/crane config --platform windows/amd64 mcr.microsoft.com/windows/servercore:1809 | jq .) <(docker run --rm gcr.io/go-containerregistry/crane config --platform windows/amd64 mcr.microsoft.com/windows/servercore:ltsc2019 | jq .)
--- /dev/fd/63	2022-01-14 09:39:38.869941110 -0800
+++ /dev/fd/62	2022-01-14 09:39:38.869941110 -0800
@@ -10,7 +10,7 @@
     },
     {
       "created": "2022-01-07T22:48:13.5841926+00:00",
-      "created_by": "Install update 1809-amd64"
+      "created_by": "Install update ltsc2019-amd64"
     }
   ],
   "rootfs": {
```

(The entire difference is a superficial metadata change -- that `"rootfs": {` key at the bottom of the diff then includes the full checksum of each layer and they're identical.)

andyli:
I've just removed ltsc2019.

It sounds weird to me that Windows' "ltsc" may have (breaking?) kernel changes...

PR #11680, opened 2022-01-14
Patch GeoNetwork with log4j 2.17.1

juanluisrp:
Patch maitained series of GeoNetwork with log4j 2.17.1 until new GN versions
are released (GN 3.10.11, 3.12.3 and 4.0.6).﻿

### DIFF
--- a/library/geonetwork
+++ b/library/geonetwork
@@ -6,7 +6,7 @@ GitRepo: https://github.com/geonetwork/docker-geonetwork
 
 Tags: 3.10.10, 3.10
 Architectures: amd64, arm64v8
-GitCommit: 31d40ad44face58778f31879bb0d0c5578f7cc6d
+GitCommit: 739141f53daa1d82442475e739cfe0776951d2c3
 Directory: 3.10.10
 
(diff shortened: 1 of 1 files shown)


tianon:
Couldn't these be downloaded/verified in the `Dockerfile` instead of committed directly to the Git repository? :grimacing:

(IMO the ideal would be to release an updated GeoNetwork upstream so that non-Docker users can get this bump too, but I realize that's probably more complicated :see_no_evil:)

juanluisrp:
@tianon Dockerfile now downloads log4j from Apache instead of using a local copy.

juanluisrp:
I've updated the default branch of https://github.com/geonetwork/core-geonetwork from `master` to `main` and now the [Test PR / Naughty (pull_request)](https://github.com/docker-library/official-images/runs/4838415008?check_suite_focus=true) is failing...

juanluisrp:
I've added a `GitFetch` command to fix it.

juanluisrp:
@tianon, can you check it please?

tianon:
Looks good, thanks!

PR #11689, opened 2022-01-18
Update python

tianon:
Changes:

- https://github.com/docker-library/python/commit/d7fa897: Update to 3.10.2, setuptools 57.5.0, pip 21.2.4
- https://github.com/docker-library/python/commit/6a2c0f4: Update to 3.9.10, setuptools 57.5.0, pip 21.2.4
- https://github.com/docker-library/python/commit/4755150: Update to 3.11.0a4, setuptools 57.5.0, pip 21.2.4

(This is a follow-up to https://github.com/docker-library/official-images/pull/11684, see it for more details)

### DIFF
--- a/library/python
+++ b/library/python
@@ -3,125 +3,124 @@
 Maintainers: Tianon Gravi <admwiggin@gmail.com> (@tianon),
              Joseph Ferguson <yosifkit@gmail.com> (@yosifkit)
 GitRepo: https://github.com/docker-library/python.git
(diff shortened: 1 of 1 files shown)


tianon:
Doh, we need to do https://github.com/docker-library/official-images/pull/11694 first so we fix the tags we clobbered with Windows-only builds. :innocent: :see_no_evil:

PR #1000, opened 2015-08-27
Upgrade to iojs-3.2.0

hmalphettes:
Upgrade: https://github.com/nodejs/docker-iojs/pull/86

GPG keys of 2 new releasers @jasnell and @sam-github added to the Dockerfiles:
https://github.com/nodejs/docker-iojs/pull/85

PR #11690, opened 2022-01-18
Update Joomla!

Llewellynvdm:
Changes:

- joomla-docker/docker-joomla@66e0400: Moves user check into Apache usecase.
- joomla-docker/docker-joomla@6b42540f5: Update images of Joomla! with user options added.
- joomla-docker/docker-joomla@e49b190a: Adds option to set the user_ID and user_group_ID for apache images.
- joomla-docker/docker-joomla@58b776e: Adds check to make sure the JOOMLA_DB_PASSWORD_FILE is set before trying to check if the path exist.
- joomla-docker/docker-joomla@05d90f4: Update images of Joomla! 4.0.5 to 4.0.6
- joomla-docker/docker-joomla@f629593: Update version of Joomla! 4.0.5 to 4.0.6
- joomla-docker/docker-joomla@3510137: Update images of Joomla! 3.10.4 to 3.10.5
- joomla-docker/docker-joomla@785ffd3: Update version of Joomla! 3.10.4 to 3.10.5

### DIFF
--- a/library/joomla
+++ b/library/joomla
@@ -4,62 +4,62 @@ Maintainers: Llewellyn van der Merwe <llewellyn.van-der-merwe@community.joomla.o
              Harald Leithner <harald.leithner@community.joomla.org> (@HLeithner)
 GitRepo: https://github.com/joomla-docker/docker-joomla.git
 
-Tags: 4.0.5-php7.4-apache, 4.0-php7.4-apache, 4-php7.4-apache, php7.4-apache
+Tags: 4.0.6-php7.4-apache, 4.0-php7.4-apache, 4-php7.4-apache, php7.4-apache
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
-GitCommit: 786a6f32362959d4de9096ab5e93ff122c54e35a
+GitCommit: 6b42540f50815c10ab991660c23ba764e91445a9
 Directory: 4.0/php7.4/apache
 
-Tags: 4.0.5-php7.4-fpm-alpine, 4.0-php7.4-fpm-alpine, 4-php7.4-fpm-alpine, php7.4-fpm-alpine
+Tags: 4.0.6-php7.4-fpm-alpine, 4.0-php7.4-fpm-alpine, 4-php7.4-fpm-alpine, php7.4-fpm-alpine
 Architectures: amd64, arm32v6, arm32v7, arm64v8, i386, ppc64le, s390x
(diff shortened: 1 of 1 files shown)


yosifkit:
Shouldn't the `if ! id "$user"` section be inside the `if [ "$uid" = '0' ]` and specifically inside the `apache2` part of the case since that is the place where the script can run `usermod`/`groupmod` (when running as `root`)?

The use case seems to be when setting an `APACHE_RUN_USER` to a user name or number that doesn't exist in the container's `/etc/passwd` and then maybe also having to set `USER_NAME`. Is that correct? Does the user need to exist in `/etc/passwd` for Joomla under Apache2 to work?

Llewellynvdm:
If the user does not exist, like passing an ID like `#1000` we get the following error:
```
[Thu Jan 20 01:48:19.257350 2022] [unixd:alert] [pid 30] AH02155: getpwuid: couldn't determine user name from uid 1000, you probably need to modify the User directive
```
So making sure the www-data user ID is just updated is what worked to avoid this, [I got the idea from here](https://github.com/linuxserver/docker-openssh-server/blob/master/root/etc/cont-init.d/10-adduser) where they update a user ABC created [here](https://github.com/linuxserver/docker-baseimage-alpine/blob/master/Dockerfile#L74).

About the `if ! id "$user"` you may be right, shorthand is not always clear on intent... but it still does the same. I have been running these images as [test here](https://hub.docker.com/r/llewellyn/joomla) for the last 20 odd days.

The greats reason for updating the user id is when mounting the volume to the parent system, to be able to edit the files without permission issues. In the past we had to make use of `setfacl` to dual set the permissions, and this is a huge pain. So this user ID setting for the Apache images is ideal to solve this issue.

Llewellynvdm:
We agree that `if id "$user" &>/dev/null; then` will only ever be true if the user does not exist, right?

So being inside the case block does not matter really, as it will only be true if the user does not exist. Do you really want me to move the code?

```shell
user=0
if id "$user" &>/dev/null; then
  echo "$user exist"
fi
user=1232
if ! id "$user" &>/dev/null; then
  echo "$user does not exist"
fi
```
Result:
```
0 exist
1232 does not exist
```

Llewellynvdm:
Okay I moved the user check into the Apache usecase, let me know if there is any more improvements needed? Thanks!

Llewellynvdm:
@yosifkit can you give me and indication why we are still being help back?

yosifkit:
Apologies, one of us just hadn't come back to review again.

PR #11692, opened 2022-01-18
[ros] rolling ros1-bridge demo package version bump

mikaelarguedas:
ros-rolling-demo-nodes-cpp: 0.17.0 -> 0.18.0
ros-rolling-demo-nodes-py: 0.17.0 -> 0.18.0

### DIFF
--- a/library/ros
+++ b/library/ros
@@ -140,6 +140,6 @@ Directory: ros/rolling/ubuntu/focal/ros-base
 
 Tags: rolling-ros1-bridge, rolling-ros1-bridge-focal
 Architectures: amd64, arm64v8
-GitCommit: 048b3f6392998df61c330b789602bf83aae9c1be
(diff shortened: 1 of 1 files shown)


yosifkit:
The change is unrelated to the build failures, so I will go ahead and merge.

I would recommend fixing the apt GPG signing issues on `noetic` and `galactic` soon (since they'll cause any rebuilds due to base OS updates to fail).

>W: GPG error: http://packages.ros.org/ros2/ubuntu focal InRelease: The following signatures were invalid: EXPKEYSIG F42ED6FBAB17C654 Open Robotics <info@osrfoundation.org>

>W: GPG error: http://packages.ros.org/ros/ubuntu focal InRelease: The following signatures were invalid: EXPKEYSIG F42ED6FBAB17C654 Open Robotics <info@osrfoundation.org>

PR #11703, opened 2022-01-19
Update python

tianon:
This should only be merged after https://github.com/docker-library/official-images/pull/11694 is successfully built and pushed across all architectures.


tianon:
Ok, tags are all fixed now, so this is good to go.

PR #11718, opened 2022-01-21
Registry v2.8.0-beta.1 release

milosgajdos:
Signed-off-by: Milos Gajdos <milosthegajdos@gmail.com>

### DIFF
--- a/library/registry
+++ b/library/registry
@@ -1,6 +1,11 @@
-Maintainers: David Wu (@davidswu),
-             Ryan Abrams (@caervs)
+Maintainers: Milos Gajdos (@milosgajdos),
+             Sebastiaan van Stijn (@thajeztah)
 GitRepo: https://github.com/docker/distribution-library-image.git
+
+Tags: 2.8.0-beta.1
+Architectures: amd64, arm64v8, arm32v6, arm32v7, ppc64le, s390x
+GitFetch: refs/heads/master
+GitCommit: c70c2acb365928f10bfc7706cc0f3f4d16f7427b
 
 Tags: 2.7.1, 2.7, 2, latest
 Architectures: amd64, arm64v8, arm32v6


library/registry
```diff
@@ -1,7 +1,29 @@
...
+             Sargun Dhillon (sargun)
+             Wayne Warren (waynr)
+             Wang Yan (wy65701436)
+             Steve Lasker (stevelasker)
+
 GitRepo: https://github.com/docker/distribution-library-image.git
 
+Tags: 2.8.0-beta.1
+Architectures: amd64, arm64v8, arm32v6, arm32v7, ppc64le, s390x
+GitCommit: TBD
```
crazy-max, line 19:
```suggestion
GitCommit: d73a547b6129fd90bbb91aa39a8c458960aa79a4
```

library/registry
```diff
@@ -1,7 +1,29 @@
-Maintainers: David Wu (@davidswu),
-             Ryan Abrams (@caervs)
+Maintainers: Chris Patterson (chrispat)
```
crazy-max, line 1: Missing commas between them I think

library/registry
```diff
@@ -1,7 +1,29 @@
...
+             Hayley Swimelar (deleteriousEffect)
+             He Weiwei (heww)
+             João Pereira (joaodrp)
+             Justin Cormack (justincormack)
+             Kyle Squizzato (squizzi)
+             Milos Gajdos (milosgajdos)
+             Sargun Dhillon (sargun)
+             Wayne Warren (waynr)
+             Wang Yan (wy65701436)
+             Steve Lasker (stevelasker)
```
crazy-max, line 13:
```suggestion
Maintainers: Chris Patterson (chrispat),
             Bryan Clark (clarkbw),
             Cory Snider (corhere),
             Hayley Swimelar (deleteriousEffect),
             He Weiwei (heww),
             João Pereira (joaodrp),
             Justin Cormack (justincormack),
             Kyle Squizzato (squizzi),
             Milos Gajdos (milosgajdos),
             Sargun Dhillon (sargun),
             Wayne Warren (waynr),
             Wang Yan (wy65701436),
             Steve Lasker (stevelasker)
```

thaJeztah:
Looks like there's also an issue with one of the entrypoint scripts (probably should be fixed before this is merged); see https://github.com/docker/distribution-library-image/pull/129#pullrequestreview-859840111

milosgajdos:
I cant mark this as Draft @thaJeztah ; let's fix the `chmod`s first and then update this PR

library/registry
```diff
@@ -1,7 +1,29 @@
...
+             Sargun Dhillon (@sargun),
+             Wayne Warren (@waynr),
+             Wang Yan (@wy65701436),
+             Steve Lasker (@stevelasker)
+
 GitRepo: https://github.com/docker/distribution-library-image.git
 
+Tags: 2.8.0-beta.1
+Architectures: amd64, arm64v8, arm32v6, arm32v7, ppc64le, s390x
+GitCommit: d73a547b6129fd90bbb91aa39a8c458960aa79a4
```
crazy-max, line 19:
```suggestion
GitCommit: c0394acc06c25675624d6b9bba698eeb4c26c925
```

milosgajdos: https://github.com/docker-library/official-images/pull/11718/commits/7a8f4024884c77574bdf6efb94315de24858dd25

yosifkit:
>Images must have a single `Dockerfile` per entry in the library file that can be used for multiple architectures. This means that each supported architecture will have the same `FROM` line (e.g. `FROM debian:buster`). See [`golang`](https://github.com/docker-library/official-images/blob/master/library/golang), [`docker`](https://github.com/docker-library/official-images/blob/master/library/docker), [`haproxy`](https://github.com/docker-library/official-images/blob/master/library/haproxy), and [`php`](https://github.com/docker-library/official-images/blob/master/library/php) for examples of library files using one `Dockerfile` per entry and see their respective git repos for example `Dockerfile`s.
>
>If different parts of the Dockerfile only happen in one architecture or another, use control flow (e.g.`if`/`case`) along with `dpkg --print-architecture` or `apk -print-arch` to detect the userspace architecture. Only use `uname` for architecture detection when more accurate tools cannot be installed. See [golang](https://github.com/docker-library/golang/blob/b879b60a7d94128c8fb5aea763cf31772495511d/1.16/buster/Dockerfile#L24-L68) for an example where some architectures require building binaries from the upstream source packages and some merely download the binary release.
>
> \- https://github.com/docker-library/official-images/blob/49f0abbd18f69493d6f24a8a2b4c75ad62699cd7/README.md#multiple-architectures

Separate architecture-specific Dockerfiles are only acceptable for base images (i.e. images `FROM scratch` that `ADD` a tar file like [Debian](https://github.com/debuerreotype/docker-debian-artifacts/blob/d5eb7c589d016973bce6f3e1827b5c315b7cefbc/bullseye/Dockerfile)).

---

> The `Dockerfile` should be written to help mitigate interception attacks during build. Our requirements focus on three main objectives: verifying the source, verifying author, and verifying the content; these are respectively accomplished by the following: using https where possible; importing PGP keys with the full fingerprint in the `Dockerfile` to check signatures; embedding checksums directly in the `Dockerfile`. All three should be used when possible. Just https and embedded checksum can be used when no signature is published. As a last resort, just an embedded checksum is acceptable if the site doesn't have https available and no signature.
>
> \- https://github.com/docker-library/official-images/blob/49f0abbd18f69493d6f24a8a2b4c75ad62699cd7/README.md#image-build

---

The build looks pretty straightforward; couldn't these instead just download and delete the tar and a single `RUN` instead of using multiple stages? See also our [Multi-stage Builds](https://github.com/docker-library/faq/tree/73f10b0daf2fb8e7b38efaccc0e90b3510919d51#multi-stage-builds) entry in our FAQ.

library/registry
```diff
@@ -1,7 +1,30 @@
...
+             He Weiwei (@heww),
+             João Pereira (@joaodrp),
+             Justin Cormack (@justincormack),
+             Kyle Squizzato (@squizzi),
+             Milos Gajdos (@milosgajdos),
+             Sargun Dhillon (@sargun),
+             Wayne Warren (@waynr),
+             Wang Yan (@wy65701436),
+             Steve Lasker (@stevelasker)
+
```
yosifkit, line 14: > [...] lines starting with `#` are ignored and "entries" are separated by a blank line.
>
>The first entry is the "global" metadata for the image. 

The failing "Diff Comment" CI is because of this blank line so the single `GitRepo` line below is being considered as a separate entry.

milosgajdos:
Whilst we look into addressing @yosifkit 's comments, can someone give us a hint about why the CI is failing? CC: @tianon

milosgajdos:
> Relevant Maintainers:

where can we update maintainers? the folks who are pinged by GHA are no longer distribution maintainers

yosifkit:
Yeah, the diff ping uses the current list of maintainers :disappointed:, so later PRs will use the new list.

library/registry
```diff
@@ -1,7 +1,23 @@
...
+             Hayley Swimelar (@deleteriousEffect),
+             He Weiwei (@heww),
+             João Pereira (@joaodrp),
+             Justin Cormack (@justincormack),
+             Kyle Squizzato (@squizzi),
+             Milos Gajdos (@milosgajdos),
+             Sargun Dhillon (@sargun),
+             Wayne Warren (@waynr),
+             Wang Yan (@wy65701436),
+             Steve Lasker (@stevelasker)
```
tianon, line 13: Did you mean to add this full list?  This should be just the maintainers of the image repository (ie, https://github.com/docker/distribution-library-image), not the full list of maintainers of the upstream software. :sweat_smile:

(We use these for things like automated notifications to existing maintainers when an image is updated or to ping folks on issues like the Alpine version you're based on going away, etc.)

milosgajdos: Ohhhh, I see! Well, it's.......complicated :)
But for this release, I guess it can be just me and @thaJeztah 

In the future the image repo will move under `distribution` org -- in which case it'll have to be either all maintainers or some select group hehe

milosgajdos: https://github.com/docker-library/official-images/pull/11718/commits/6f33409389c501db767c7da48ed1a3f5afcbea5a

tianon: Heh, fair enough -- we'd also be happy to maintain it under the @docker-library org if you'd like (just to make sure you know you've got options)

tianon:
Nice, LGTM :+1:

tianon:
(Hopefully once this one goes GA we can drop all the old versions as unsupported/EOL and clear that failing CI? :eyes:)

thaJeztah:
Yes we should remove the old versions then (remove as in: stop them being rebuilt of course)

tianon:
Pulling out my classic related copypasta for everyone else's benefit (since I know you've probably seen it once or twice :joy:):

PR #11727, opened 2022-01-23
[ros] noetic ros-comm 1.15.13 -> 1.15.14

mikaelarguedas:


mikaelarguedas:
I cannot reproduce the gpg key issue.
What is surprising to me is that the same key is used to sign the apt repo for all ros distributions but in your CI only some of them fail: In https://github.com/docker-library/official-images/pull/11692#issuecomment-1016916777 it was `noetic` and `galactic`. And on this PR it is `rolling` and `melodic` but `noetic` and `galactic` pass.

yosifkit:
:thinking: It must be related to our proxy to get more reliable keyserver responses (https://github.com/tianon/pgp-happy-eyeballs) and so some of the servers we attempt have the expired key still and are sometimes faster than the others.

PR #11733, opened 2022-01-24
Release of GeoNetwork 3.12.3

juanluisrp:


juanluisrp:
Sourceforge doesn't seem very stable today, but the 3.12.3 WAR URL is right and I've been able to build the image locally.

PR #11752, opened 2022-01-27
Adding Couchbase Server 6.6.5 EE

ceejatec:


tianon:
There are a few things about this new multi-stage addition that are kind of strange.  For one, using Amazon Linux to build a binary to use in Ubuntu seems a little weird.  Having an explicit variable for `ARCH` is also strange -- we recommend using tools which can provide the _userspace_ architecture of the image for architecture detection, such as `dpkg --print-architecture` or `apk --print-arch`.  It also seems unnecessary for this to be a multi-stage build at all?  (See https://github.com/docker-library/faq#multi-stage-builds for some info about that.)

A few more notes would be that `apt-get clean` doesn't do anything in these images (it's done automatically via configuration in the Debian and Ubuntu base images), and `apt-get clean all` is probably not what you intended (that seems like a `yum`-ism).

Additionally, deleting files / cleaning cache in a separate layer isn't going to save any space (it's only going to create a layer of whiteout files).

ceejatec:
The Dockerfile changes are largely because we generate them from a template, and we recently had to update the template to also support Amazon Linux 2 on aarch64. As the comments explain, the first part of the multi-stage build is only for amzn2. We couldn't find a straightforward way to avoid copying runit into the Ubuntu image as well, so if there are better-practices we could use there, please let us know. Agreed that the "rm -rf /tmp/runit*" isn't really adding much value, though.

That also explains the ARCH argument, as it is passed in by our build scripts (we have to build that image ourselves since Docker Hub autobuilds don't support aarch64 yet).

I'm not sure why you say "apt-get clean" doesn't do anything - it's there to clean up anything left over from the preceding commands. As for "clean all", that actually is a yum-ism - since Amazon Linux 2 is rpm-based, we couldn't use apt-get, but 99% of the arguments are the same between apt-get and yum for the commands we actually needed so the template could just have {{ .PKG_COMMAND }} .

yosifkit:
`apt-get clean` doesn't have an effect in the Debian and Ubuntu docker images because it is already automatic since both have a similar entry in their apt conf files. This is created in the [Dockerfile](https://github.com/tianon/docker-brew-ubuntu-core/blob/f2f3f01ed67bab2e24b8c4fda60ef035a871b4c7/trusty/Dockerfile#L21-L24) for older versions of Ubuntu and part of the rootfs tar for the rest (e.g. in the scripts used to create the [Debian](https://github.com/debuerreotype/debuerreotype/blob/32f390b0eff8eeb01fdec135ddc6eab06850c826/scripts/debuerreotype-minimizing-config#L89-L109) rootfs tar files).

It looks like your templates are using go text/template, so something like this can add some control flow and make it so only the `amazonlinux` parts are added to the generated Dockerfile in the `amazonlinux`-based image.

```diff
diff --git a/generate/templates/couchbase-server/Dockerfile.template b/generate/templates/couchbase-server/Dockerfile.template
index b0f659df..035dd96a 100644
--- a/generate/templates/couchbase-server/Dockerfile.template
+++ b/generate/templates/couchbase-server/Dockerfile.template
@@ -1,3 +1,4 @@
+{{ if eq .DOCKER_BASE_IMAGE "amazonlinux:2" -}}
 # Stage 0: compile runit.
 # This is ONLY used for amzonlinux which doesn't come with runit
 FROM amazonlinux:2 as runit_build
@@ -34,6 +35,11 @@ RUN set -x && \
     fi
 
 RUN rm -rf /tmp/runit*
+{{ else -}}
+FROM {{ .DOCKER_BASE_IMAGE }}
+
+LABEL maintainer="docker@couchbase.com"
+{{- end }}
 # Install dependencies:
 #  runit: for container process management
```

---

One other note is that the `apt-get autoremove -y` is unnecessary right after installing packages, since nothing has been removed to cause any other package to be "no longer necessary".

ceejatec:
@yosifkit Thank you, that was very helpful! I've revamped our Dockerfile template here and I believe was able to address all of the stated concerns. I'm getting the changes reviewed internally and I'll update this PR hopefully on Monday.

ceejatec:
So, no more multi-stage build for our x86_64 Dockerfile. No more copying runit stuff around or useless delete layers. I removed all invocations of `apt-get clean`. I removed all `apt-get autoremove` invocations except one that was actually after an `apt-get remove`. I removed the ARCH arg as it wasn't strictly necessary. I also cleaned up some of our multi-line `this && that` RUN directives to be formatted more consistently and neatly, and use `set -x` for useful echoing of the commands when building.

The one oddness that's still there due to the underlying multi-arch builds are new ARGs for UPDATE_COMMAND and CLEANUP_COMMAND, in particular that the latter is set to "true" in this x86_64 Dockerfile because no cleanup command is necessary. Hopefully that is not overly objectionable.

ceejatec:
@tianon I am done making updates based on the issues you raised - does the current Dockerfile look better? Anything else we should address?

yosifkit:
Yeah, the `ARG`'s are a little odd, but not a reason to stop the PR. One thing I noticed is that the two `rm -rf /var/lib/apt/lists/*` got lost. That is, unfortunately, not something done automatically by `apt`. Maybe that can be the `CLEANUP_COMMAND`?

Everything else looks reasonable to me.

ceejatec:
@yosifkit Good to know about /var/lib/apt/lists not getting cleared automatically, and good suggestion to use that as CLEANUP_COMMAND. I've implemented that and it did save about 32MB in the image size.

ceejatec:
@tianon @yosifkit Thanks for the assistance. Please excuse my using this PR to ask a follow-up question. Starting with Couchbase Server 7.1.0 we will support Linux aarch64 (arm64v8). However that image is based on amazonlinux:2, rather than ubuntu:20.04 like our amd64 image. I'm looking at the multi-arch instructions, and it's not clear if this is supported or permitted. We will be generating two Dockerfiles at different paths in our git repository. Is it possible for our Official Image to be assembled in this way?

yosifkit:
Since a multi-architecture image is considered a "single" item, it would be very confusing for users that each architecture is based on a completely different Linux distribution.

We have many image maintainers that choose to have two distributions with separate tags (e.g. `couchbase:7.1-amazonlinux2`) with one being chosen for the generic tags like `latest` and `7.1`. Why are each distribution only for a particular architecture? Both [`amazonlinux:2`](https://github.com/docker-library/official-images/blob/1caec5a42189ac1177b1fc886d72bc9127a244be/library/amazonlinux#L17) and [`ubuntu:20.04`](https://github.com/docker-library/official-images/blob/1caec5a42189ac1177b1fc886d72bc9127a244be/library/ubuntu#L36) are available for the two desired architectures. We'd instead recommend ensuring that the Dockerfiles are resilient enough that they can be built on multiple architectures.

We've had huge success creating single Dockerfiles that build across multiple architectures as well as having images for multiple Linux distributions. For example, [`golang:1.17-bullseye`](https://github.com/docker-library/golang/blob/20770e021d1931892a68710e290af8792d9754f9/1.17/bullseye/Dockerfile) and [`golang:1.16-alpine3.15`](https://github.com/docker-library/golang/blob/20770e021d1931892a68710e290af8792d9754f9/1.17/alpine3.15/Dockerfile)) build on eight and seven architectures, respectively.

> Images must have a single `Dockerfile` per entry in the library file that can be used for multiple architectures. This means that each supported architecture will have the same `FROM` line (e.g. `FROM debian:buster`). See [`golang`](https://github.com/docker-library/official-images/blob/master/library/golang), [`docker`](https://github.com/docker-library/official-images/blob/master/library/docker), [`haproxy`](https://github.com/docker-library/official-images/blob/master/library/haproxy), and [`php`](https://github.com/docker-library/official-images/blob/master/library/php) for examples of library files using one `Dockerfile` per entry and see their respective git repos for example `Dockerfile`s.
>
>If different parts of the Dockerfile only happen in one architecture or another, use control flow (e.g.`if`/`case`) along with `dpkg --print-architecture` or `apk -print-arch` to detect the userspace architecture. Only use `uname` for architecture detection when more accurate tools **cannot** be installed. See [golang](https://github.com/docker-library/golang/blob/b879b60a7d94128c8fb5aea763cf31772495511d/1.16/buster/Dockerfile#L24-L68) for an example where some architectures require building binaries from the upstream source packages and some merely download the binary release.
>
>\-https://github.com/docker-library/official-images/tree/09d624839d5f811973ff8d1234d8786cec2c4915#multiple-architectures

ceejatec:
Couchbase Server for aarch64 was targeted specifically for Amazon Linux 2, so that was the platform development and certification was done on; it is, at least for now, the only platform it is supported on. We don't even create a .deb packaging. So it made sense to create the Docker image accordingly. I'm sure it would be feasible to convert it, but it's at least a fair amount of testing and re-certification.

> Since a multi-architecture image is considered a "single" item, it would be very confusing for users that each architecture is based on a completely different Linux distribution.

That seems a bit odd too - aren't containers generally intended to be a black box? It's always been my interpretation that the base image for a Dockerfile should be "don't care", other than possible concerns of image size or updated security patches.

ceejatec:
@yosifkit @tianon Any further thoughts on the above? Doesn't it seem odd that the Official Image requirement implicitly requires that products with official images support the same set of distros across multiple architectures?

Should I file an issue here regarding this?

ceejatec:
@yosifkit @tianon If the current rule cannot be relaxed, would it be acceptable for us to use different tags on the Official Image for aarch64? eg., `couchbase:7.1.0-aarch64` (or perhaps `couchbase:7.1.0-arm64v8`) vs. `couchbase:7.1.0` (implicitly x86_64).

yosifkit:
I think the average user might not care about the inner details of an image, but official images are not meant to be a black box. Our [consistency](https://github.com/docker-library/official-images/tree/2c461918e7b262e75da8c0bea3a581deb3b3c406#consistency) requirement to ensure that `docker run [image] bash` works is there to make image exploration easier. They are open so that users can understand them and extend them.

Most open source projects providing distribution-specific releases that I've seen are usually similar to [PostgreSQL](https://www.postgresql.org/download/) in that they have a base set of architectures (like `amd64`) that are available on all supported distributions with extra architectures available on as many as they are willing to support.

Of the ~140 official images, this would be the first to not support `amd64` for all their chosen Linux distribution base images. Having `amd64` is not a hard requirement, nor is supporting the same set of architectures between disparate Linux distributions. We highly recommend supporting as many architectures that the base distributions supports as possible. This will ensure users can run the equivalent distribution-specific image when they debug locally whether or not their machine matches the production architecture.

Architecture names in the image tags would not be acceptable.

If `amd64` for Amazon Linux-based images is still a long way away, then something like this could work:

```yaml
# I'd suggest extending the tags to make it clear which Linux distro is used in the "default" tags
Tags: community, community-7.0.2, community-focal, community-7.0.2-focal
Architectures: amd64 # not necessary, since this is the default
GitCommit: ...
Directory: ...

Tags: community-amazonlinux2, community-7.0.2-amazonlinux2
Architectures: arm64v8
GitCommit: ...
Directory: ...
```

Putting both distributions under a single set of Tags would be unacceptable. This is the scenario my earlier comment was mostly trying to prevent.

ceejatec:
@yosifkit Thanks for the response. Again, for us, the point isn't supporting amazonlinux:2 as a base image per se; it's extending Server's architecture support. I wouldn't personally want to have an image named eg. `couchbase:7.1.0-amazonlinux2` that just so happened to only be available on aarch64, because it's burying the lede as it were. We also don't have Official (or unofficial) images based on Debian, or Centos 7, or OpenSUSE, or any of the other distros we support. That's why I suggested `couchbase:7.1.0-aarch64`. If that's not acceptable, then I think unfortunately our Official Image will go forward for some time without aarch64 support, even though the product supports it, and we even provide our "unofficial image" `couchbase/server` on aarch64.

PR #11759, opened 2022-01-28
Release SonarQube 9.3.0

wouter-admiraal-sonarsource:


tobias-trabelsi-sonarsource:
Hey there. I hope that this will now just update the 9.x image and not the LTS (8.x). If this is correct this should be okay. if this PR would update the 8.9 images as well please don't merge this PR. 
The goal is to just release 9.3 and not touch the 8.9 images :sweat_smile:

tianon:
Yep, that's correct!

PR #11760, opened 2022-01-28
eclipse-temurin: dockerfile updates

gdams:
`arm32v7` will be available shortly but we wanted to press on and get these published

### DIFF
--- a/library/eclipse-temurin
+++ b/library/eclipse-temurin
@@ -5,196 +5,196 @@ GitRepo: https://github.com/adoptium/containers.git
 GitFetch: refs/heads/main
 
 #------------------------------v8 images---------------------------------
-Tags: 8u312-b07-jdk-focal, 8-jdk-focal, 8-focal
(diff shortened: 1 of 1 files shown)


library/eclipse-temurin
```diff
@@ -257,34 +257,34 @@ GitCommit: f06ea446cee4810ab29df927d2d4c53f6270a5cd
 Directory: 17/jdk/centos
```
yosifkit, line 257: Were the 17 jdk/jre non-windows images also supposed to be updated?

Just to be clear about the consequences of still having these versions with `SharedTags: 17.0.1_12-jdk` is that the manifest for those tags will be updated to no longer include the Windows images of that version (similar to the issue we just had in https://github.com/docker-library/python/issues/681)..

gdams: Sorry, yes just spotted that! I'll get the JDK17 Linux dockerfiles added too

gdams:
@yosifkit this should be good to merge now

PR #11771, opened 2022-01-31
bump Kong to 2.7.1

kikito:


tianon:
Any ideas for why all three `centos` variants fail to build? :grimacing:

gszr:
> Any ideas for why all three `centos` variants fail to build? 😬

@tianon the reason is CentOS reached EOL. To get images building, we had to introduce the `sed` you can see in the diff, changing the repo references to vault.centos.org.

yosifkit:
I am fine with this swap to the centos vault mirror as a fix to help users stuck on that version to be able to use the image. Seeing that `centos:8` is end of life, we will likely be talking with the maintainer about removing it from official-images, so it will be required to not be `FROM` that image in the future. Once removed from the `library/` files here, both the `centos:8` image and any image `FROM` it would still be available on Docker Hub, just no longer updated.

PR #11773, opened 2022-01-31
Inital batch of Temurin updates for January 2022 releases

sxa:
Update all the platforms and versions that we have released so far for Temurin / OpenJDK 8u322, 11.0.14 and 17.0.2

32-bit platforms to follow later.

Signed-off-by: Stewart X Addison <sxa@redhat.com>

### DIFF
--- a/library/eclipse-temurin
+++ b/library/eclipse-temurin
@@ -5,196 +5,196 @@ GitRepo: https://github.com/adoptium/containers.git
 GitFetch: refs/heads/main
 
 #------------------------------v8 images---------------------------------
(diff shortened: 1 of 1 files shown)


tianon:
How does this relate to #11760?  Is it the replacement that addresses https://github.com/docker-library/official-images/pull/11760#discussion_r794920624? :eyes:

cc @gdams

gdams:
> How does this relate to #11760? Is it the replacement that addresses #11760 (comment)? 👀

It's essentially the same pull request (sorry bad communication on my end). I've pushed another update to my PR which adds ppc64le for JDK 17 so my change is likely the one you want to take. We'll follow up with arm32 shortly (and s390x on JDK17)

PR #6252, opened 2019-07-09
Be more patient with Nextcloud setups

J0WI:
Should be merged before #6236

PR #11787, opened 2022-02-02
Update Nextcloud

J0WI:


tianon:
It seems like these are _all_ consistently failing the `nextcloud` tests -- it seems to me that either those tests aren't actually meaningful/useful anymore (and we should likely finally remove them) or something is really wrong here. :see_no_evil:

J0WI:
I'm happy as long as not all database variants are failing at once... But any help to reduce the flakiness is appreciated. #6252 was not very successful.

PR #11792, opened 2022-02-03
Update redis

tianon:
Changes:

- https://github.com/docker-library/redis/commit/5800387: Ditch "7.0" alias for "7.0-rc"
- https://github.com/docker-library/redis/commit/fb8bdd7: Merge pull request https://github.com/docker-library/redis/pull/304 from infosiftr/7.0-rc
- https://github.com/docker-library/redis/commit/b90f94b: Add 7.0-rc1

### DIFF
--- a/library/redis
+++ b/library/redis
@@ -1,8 +1,18 @@
-# this file is generated via https://github.com/docker-library/redis/blob/231905d0841f52ee4f3a5b8b42d62cd6d14a1a93/generate-stackbrew-library.sh
+# this file is generated via https://github.com/docker-library/redis/blob/5800387c04b3a07957e578258e0438187dd9e5d2/generate-stackbrew-library.sh
 
 Maintainers: Tianon Gravi <admwiggin@gmail.com> (@tianon),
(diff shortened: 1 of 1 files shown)


library/redis
```diff
@@ -4,6 +4,16 @@ Maintainers: Tianon Gravi <admwiggin@gmail.com> (@tianon),
              Joseph Ferguson <yosifkit@gmail.com> (@yosifkit)
 GitRepo: https://github.com/docker-library/redis.git
 
+Tags: 7.0-rc1, 7.0, 7.0-rc, 7.0-rc1-bullseye, 7.0-bullseye, 7.0-rc-bullseye
```
yosifkit, line 7: Do we want to avoid plain tags like `7.0` on an `rc` release?

tianon: Ah yeah, I guess so -- good catch.

tianon: Updated via https://github.com/docker-library/redis/commit/5800387c04b3a07957e578258e0438187dd9e5d2 (updated commit here coming in just a second).

PR #11794, opened 2022-02-03
Add image for MariaDB

vmassol:


tianon:
Interesting (I guess MariaDB is finally diverging more strongly from "MySQL compatible" which I suppose was inevitable :see_no_evil:) -- this makes me wonder whether it's worth considering combining all three connectors into a single image since they're all almost identical minus the couple megabytes of each different connector?

I guess the hard part there would be detecting which type of database the (otherwise generic) `DB_...` variables are referencing, but maybe that could be solved via some clever use of `ENV`?

I'm thinking about something like this (where `xwiki:xxx` is a placeholder for a generic all-connectors image that respects a `DB_TYPE` parameter to choose an appropriate connector at runtime):

```dockerfile
FROM xwiki:xxx
ENV DB_TYPE mysql
```

```dockerfile
FROM xwiki:xxx
ENV DB_TYPE postgresql
```

etc?  Thoughts?

vmassol:
Hi @tianon. Yes it would be possible at the expense of making it slightly harder for the user since they'd need to pass an extra env property (there's also the issue of current users upgrading and finding that their usual command line doesn't work anymore). It's also a bit more heavy and we'd need to generate the proper config files (but that's doable).

Note that we might deprecate the XWiki/MySQL image soonish too.

Just to be sure: what's the reason for having a single image from your side? Less CPU and disk space? (that's a good reason for sure).

Is it possible to merge this PR for now and we can then decide what to do for the future?

Thx for checking and caring! :)

tianon:
I guess my thought was more around less conceptual complexity (and better caching).

I'm still suggesting you have tags for each type, but then the `Dockerfile` for each of those tags is a simplified `FROM`+`ENV DB_TYPE` instead of the full download everything.

I think the biggest argument against my idea is that the chances of a user _switching_ from one database type to another (or deploying multiple instances with _different_ database types) are very low. :sweat_smile:

PR #11801, opened 2022-02-03
Adjust nextcloud tests to avoid flakiness

tianon:
This adjusts `nextcloud-apache-run` and `nextcloud-fpm-run` to try up to three times for a successful result, and adds a new `nextcloud-cli` test that combines the other three (`nextcloud-cli-mysql`, `nextcloud-cli-postgres`, and `nextcloud-cli-sqlite`) into a single test that will pass if any one of them passes (but does attempt to run all three every time).

Also, account for change in the installation page (from "Finish setup" to "Install"): https://github.com/nextcloud/server/pull/30615

See also https://github.com/docker-library/official-images/pull/6252

(This includes #11787 to actually run the tests, especially since that update includes the above linked change to the installation page.)

cc @J0WI @tilosp


test/tests/nextcloud-apache-run/run.sh
```diff
@@ -1,37 +0,0 @@
...
-		"$clientImage" \
-		curl -fsL -X"$method" "$@" "http://apache/$url"
-}
-
-# Make sure that Apache is listening and ready
-. "$dir/../../retry.sh" --tries 30 '_request GET / --output /dev/null'
-
-# Check that we can request / and that it contains the pattern "Finish setup" somewhere
-# <input type="submit" class="primary" value="Finish setup" data-finishing="Finishing …">
-_request GET '/' | grep -i "Finish setup" > /dev/null
```
tianon, line 37: I'm not at a computer right this second, but did this message change, maybe? 😂😅🙈

tilosp: yes, see: https://github.com/nextcloud/server/pull/30615

tianon:
Whew, fully green now!  I've rebased at some point on https://github.com/docker-library/official-images/pull/11787 's commit exactly, so if we merge this it should mark that one as merged too. :eyes:

J0WI:
Thanks a lot!

PR #11817, opened 2022-02-07
Refresh Couchbase Server 7.0.3 for Log4Shell remediation

ceejatec:


tianon:
(CI failures are just the unchanged 6.6.x versions)

tianon:
Is 6.0.5 still actively supported?  (It looks like it hasn't been explicitly updated for a bit :see_no_evil:)

ceejatec:
@tianon Yes, we still offer support for 6.0.5 (although clearly we'd prefer our customers upgrade!). I'm not sure what updates you would expect to see? My understanding is that Docker Hub rebuilds the image whenever the Ubuntu base image is updated for security issues. We're not creating any newer builds of 6.0.5 itself; how could we? If necessary, we'd release a 6.0.6.

tianon:
Right, sorry, I should've included "log4shell remediation" in my comment. :sweat_smile:

(Docker's scanning is triggering on it, so I'm getting pinged :see_no_evil:)

Does it make sense to release a 6.0.6 with a similar log4j bump?

ceejatec:
Ah, got it. That decision is above my pay grade, but I'll ask.

PR #11819, opened 2022-02-07
Ubuntu 21.04 Hirsute EOL

toabctl:
As announced [1], Ubuntu Hirsute has reached the end of it's life.

[1]
https://lists.ubuntu.com/archives/ubuntu-security-announce/2022-January/006363.html


tianon:
This will be a slightly heavier child trigger, but it's a delete only so they should all be fully-cached no-ops anyhow. :+1:

tianon:
FYI for affected images:

- `neurodebian`: @yarikoptic

Issue #11679, opened 2022-01-14
Is the CentOS image still maintained?


raesene:
Looking at the CentOS images on Docker Hub at the moment, it looks like they've not had patched applied in 4 months, and this results in them having a number of [unfixed security issues](https://try.trivy.dev/results/zmX7RaQQ2Y5CgREKhLwkbxyE).

Is this image still maintained? I was just thinking that (especially as centos is a pretty heavily used base image) , if it's not, it might be worth indicating that it's unmaintained on Docker hub (like the other unmaintained official images).

tianon:
Yeah, it has been a bit since the last update :disappointed: :grimacing: 

@bstinsonmhk any plan to do an update soon? :pray:

jamshid:
This is weird I assumed centos:8 would just become a continuously updated (stream?) image. But nope, it's pretty useless now:
```
% docker run -ti centos:8 bash      
[root@5dcdae78e366 /]# yum update
Failed to set locale, defaulting to C.UTF-8
CentOS Linux 8 - AppStream                                                                                                        128  B/s |  38  B     00:00    
Error: Failed to download metadata for repo 'appstream': Cannot prepare internal mirrorlist: No URLs in mirrorlist
```
Is centos:8 really dead, RedHat is making everyone switch to stream8?
```
% docker run -ti quay.io/centos/centos:stream8 bash
[root@821b3f36fa02 /]# cat /etc/centos-release 
CentOS Stream release 8
[root@821b3f36fa02 /]# yum update
...hundreds of package need updating what kind of a stream is this?...
Complete!
```

tianon:
Ouch, I guess https://www.centos.org/centos-linux-eol/ means we should probably add a deprecation notice to Docker Hub?  cc @bstinsonmhk :pray: :bow:

yeroc:
The centos7 image has now been unpatched for 7 months.  I thought there was some expectation that images with the "Docker Official Image" label would be kept up to date?  Am I out to lunch here?

tianon:
Yeah, thanks for the reminder - I've opened https://github.com/docker-library/docs/pull/2205 now :disappointed:

yeroc:
Thanks.  Will there be a deprecation notice put on Docker Hub as well? 

The strange thing here is that CentOS 7 isn't actually EOL...

tianon:
[...]

PR #11831, opened 2022-02-08
Drop end of life CentOS versions

yosifkit:
CC `centos` image maintainer: @bstinsonmhk Are there going to be `stream8` or `stream9` images added?

Removing it here will remove it from the "Supported" section on the Hub readme, but the tag will still be available to users who want it.  (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)

This drops CentOS 6 and 8 images since they are both End of life. From the [CentOS](https://en.wikipedia.org/wiki/CentOS#End-of-support_schedule) page on Wikipedia:
- `centos:6` 2020-11-30
- `centos:8` 2021-12-31

This is a follow up to https://github.com/docker-library/official-images/pull/11771#issuecomment-1030293755 and https://github.com/docker-library/official-images/issues/11679.

### DIFF
--- a/library/centos
+++ b/library/centos
@@ -2,15 +2,6 @@ Maintainers: The CentOS Project <cloud-ops@centos.org> (@CentOS)
 GitRepo: https://github.com/CentOS/sig-cloud-instance-images.git
 Directory: docker
 
-Tags: latest, centos8, 8, centos8.4.2105, 8.4.2105
-GitFetch: refs/heads/CentOS-8-x86_64
-GitCommit: 607af70702bacc6f46fab2ded055ab23d9113831
-arm64v8-GitFetch: refs/heads/CentOS-8-aarch64
-arm64v8-GitCommit: e79ccf67325a31bf0bb79a8a0e82d3d8a4de2da8
-ppc64le-GitFetch: refs/heads/CentOS-8-ppc64le
-ppc64le-GitCommit: 76f876b31bb82108a1acf2cee1032c1d2ebc3bd9
-Architectures: amd64, arm64v8, ppc64le
-
 Tags: centos7, 7, centos7.9.2009, 7.9.2009
 GitFetch: refs/heads/CentOS-7-x86_64
 GitCommit: b2d195220e1c5b181427c3172829c23ab9cd27eb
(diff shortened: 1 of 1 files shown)


tianon:
I'm going to shamelessly cc @jperrin as well - I know this isn't your thing any more, but maybe you have thoughts/opinions/info? :pray: :bow:

jdbarnes-isi:
can't run dnf on official built latest image. what's up with EOL vs stream on docker hub?

library/centos
```diff
@@ -2,15 +2,6 @@ Maintainers: The CentOS Project <cloud-ops@centos.org> (@CentOS)
 GitRepo: https://github.com/CentOS/sig-cloud-instance-images.git
 Directory: docker
 
-Tags: latest, centos8, 8, centos8.4.2105, 8.4.2105
```
thaJeztah, line 5: Should the `latest` tag be moved to `centos:7` (as that will be the only supported version after this - until `stream8` and/or `stream9` are added that is)?

thaJeztah: Of course, that would be somewhat disruptive (but `:latest` by definition is "unpredictable", and not recommended, so perhaps that's "fine")

jperrin:
👋 @tianon! You're right, I'm not that involved anymore. Let me ask around though and I'll either cause someone to answer here or let you know what I find out.

bstinsonmhk:
👍 to removing these from the 'supported' list. I'll reach out separately about adding Stream tags.

library/centos
thaJeztah: I think removing it here means that it will still be there, just not updated (so continue to point to the `centos:8` image)

tianon:
@bstinsonmhk @jperrin thank you both!! :heart:

We're also probably due for a rebuild of the `centos:7` tag, right?
(It's been ... a minute. :grimacing:  It's got at least 42 outdated packages in it. :see_no_evil:)

q0rban:
This comment is for GitHub search, in case anyone else like me is running into `Error: Failed to download metadata for repo 'appstream': Cannot prepare internal mirrorlist: No URLs in mirrorlist` when running `yum install` on a centos:8 image. ☺️

PR #11832, opened 2022-02-08
Add devel tags to AL 2022 images

sktomer:
Adds devel tags to Amazon Linux 2022 container images.
We need this as currently AL 2022 is in Tech Preview phase.


rpkelly:
lgtm

PR #11835, opened 2022-02-09
Releasing Neo4j 4.2.15 and 4.4.4

jennyowen:
Thanks!

### DIFF
--- a/library/neo4j
+++ b/library/neo4j
@@ -11,12 +11,12 @@ GitRepo: https://github.com/neo4j/docker-neo4j-publish.git
 
 Tags: 4.4.4, 4.4.4-community, 4.4, 4.4-community, community, latest
 Architectures: amd64, arm64v8
-GitCommit: 1d29a489ad5bd4bb032d3cdf5769e4b5d318d3c0
+GitCommit: 0b2c95bcf4e2d9c26366890922dfd85d315cb96a
 Directory: 4.4.4/community
 
 Tags: 4.4.4-enterprise, 4.4-enterprise, enterprise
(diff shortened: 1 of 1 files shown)


yosifkit:
It seems there was some sort of templating issue that also copied the template Dockerfile into the `local-package/` directory. Small and harmless when copied into the image, but does conflate the diff a bit.

jennyowen:
Gah! I suppose this is what happens when you try to refactor old complicated Makefiles :cry: 

I've fixed it, and there was another Neo4j release overnight, so that's included in this PR too now.

yosifkit:
It looks like the files from the `COPY ./local-package/* /startup/` are no longer used by the `RUN` line, so it should be moved later in the Dockerfile to optimize `docker build` cache.

jennyowen:
Hi @yosifkit you're right that the COPY could be moved to after the RUN *in the Dockerfile we publish* but it would break our build process for when Neo4j is under development.

We only upload a Neo4j to `https://dist.neo4j.org/neo4j-<whatever>-unix.tar.gz` once a new version has been developed, tested and released. Until then, we make the docker image by putting neo4j in `./local-package/` and setting the `NEO4J_URI` build arg to point to the copied file when we do `docker build`.

If it's important to you that we keep the build cache as optimised as possible, I could still make the changes you request to just the published Dockerfiles. Although I'm hesitant to make them too different to the ones we test.

So TLDR: I would prefer not do that unless it's important for compliance with the docker official images library.

jennyowen:
Thank you both! :heart:

PR #11853, opened 2022-02-11
Update phpMyAdmin to version 5.1.3

ibennetch:
Signed-off-by: Isaac Bennetch <bennetch@gmail.com>


ibennetch:
Hm, am I reading this right that there seems to be some problem with the amd64 variant? I'm not sure what would cause this problem since our base builds seem to me to have problem amd64 support.

Can you offer any tips on this?

yosifkit:
It looks like it was unable to find the commit earlier. Restarted the jobs and it's working fine now.

ibennetch:
Thank you!

PR #11864, opened 2022-02-14
mariadb: feb2022 release

grooverdan:
10.2.43, 10.3.34, 10.4.24, 10.5.15, 10.6.7, 10.7.3, 10.8.2.

Yes, a version is missing, we discovered a fault part way though
releasing.

10.7 is now GA and 10.8 is new.

Notable changes include a mariadb-upgrade and a healthcheck script (but no HEALTHCHECK directive via policy). I made the rational that its better to provide something right that people can use than leave it to them invent their own.

A set of document updates coming soon

### DIFF
--- a/library/mariadb
+++ b/library/mariadb
@@ -1,35 +1,40 @@
-# this file is generated via https://github.com/MariaDB/mariadb-docker/blob/c75ff03efb6527d6c6ce06839867a06b6fa9387d/generate-stackbrew-library.sh
+# this file is generated via https://github.com/MariaDB/mariadb-docker/blob/2937a9bd36ac208e8322c8564f069b10261749f8/generate-stackbrew-library.sh
 
 Maintainers: Daniel Black <daniel@mariadb.org> (@grooverdan),
              Daniel Bartholomew <dbart@mariadb.com> (@dbart)
 GitRepo: https://github.com/MariaDB/mariadb-docker.git
 
-Tags: 10.7.1-focal, 10.7-focal, 10.7.1, 10.7
+Tags: 10.8.2-focal, 10.8-focal, 10.8.2, 10.8
 Architectures: amd64, arm64v8, ppc64le, s390x
-GitCommit: d4efdb8951f606fe2837e9cc4db37225a5b7a621
+GitCommit: cd10ba6c780e49e732ca0eacb76affaac6e87d74
+Directory: 10.8
+
+Tags: 10.7.3-focal, 10.7-focal, 10-focal, focal, 10.7.3, 10.7, 10, latest
+Architectures: amd64, arm64v8, ppc64le, s390x
+GitCommit: cd10ba6c780e49e732ca0eacb76affaac6e87d74
 Directory: 10.7
 
-Tags: 10.6.5-focal, 10.6-focal, 10-focal, focal, 10.6.5, 10.6, 10, latest
+Tags: 10.6.7-focal, 10.6-focal, 10.6.7, 10.6
 Architectures: amd64, arm64v8, ppc64le, s390x
-GitCommit: d711333929498046f354e14430cbe65e4767fc63
+GitCommit: cd10ba6c780e49e732ca0eacb76affaac6e87d74
 Directory: 10.6
 
-Tags: 10.5.13-focal, 10.5-focal, 10.5.13, 10.5
+Tags: 10.5.15-focal, 10.5-focal, 10.5.15, 10.5
 Architectures: amd64, arm64v8, ppc64le, s390x
-GitCommit: 8414e4ff9ebff49f28148a860d85131e11c049c6
+GitCommit: cd10ba6c780e49e732ca0eacb76affaac6e87d74
 Directory: 10.5
 
-Tags: 10.4.22-focal, 10.4-focal, 10.4.22, 10.4
+Tags: 10.4.24-focal, 10.4-focal, 10.4.24, 10.4
 Architectures: amd64, arm64v8, ppc64le
-GitCommit: 435a47c85a80be384524051f23cc3e8f132c31ff
+GitCommit: cd10ba6c780e49e732ca0eacb76affaac6e87d74
 Directory: 10.4
 
-Tags: 10.3.32-focal, 10.3-focal, 10.3.32, 10.3
+Tags: 10.3.34-focal, 10.3-focal, 10.3.34, 10.3
 Architectures: amd64, arm64v8, ppc64le
-GitCommit: a98e19f16db72ff1f6148cd797a0cb53dacacef3
+GitCommit: cd10ba6c780e49e732ca0eacb76affaac6e87d74
 Directory: 10.3
 
-Tags: 10.2.41-bionic, 10.2-bionic, 10.2.41, 10.2
+Tags: 10.2.43-bionic, 10.2-bionic, 10.2.43, 10.2
 Architectures: amd64, arm64v8, ppc64le
-GitCommit: 9c804d5875d9faa880133b7720fdf7ccf35d6393
+GitCommit: cd10ba6c780e49e732ca0eacb76affaac6e87d74
 Directory: 10.2


tianon:
Sorry, this is really big -- we're working our way through it, but it's a lot. :see_no_evil:  Hopefully we'll have a more useful comment soon. :+1:

yosifkit:
Here is my initial pass of feedback.

```bash
	docker_temp_server_start "$@" --skip-grant-tables \
		--loose-innodb_buffer_pool_dump_at_shutdown=0
	local pid=$!
```

This seems a bit error prone in ensuring nothing else in `docker_temp_server_start` starts something in the background. Maybe adding `--pid-file` would be better?

```bash
	local pidFile
	pidFile="$(mktemp)"
	docker_temp_server_start "$@" --skip-grant-tables \
		--loose-innodb_buffer_pool_dump_at_shutdown=0 \
		--pid-file="$pidFile"
```

---

```bash
	# docker_temp_server_stop needs authentication since
	# upgrade ended in FLUSH PRIVILEGES
	mysql_note "Stopping temporary server"
	kill "$pid"
	...
```

If `docker_temp_server_stop` needs authentication wouldn't it make sense to improve it to also handle the case where it may not have authentication? In the past, we've recommended to users that they can source the entrypoint to customize its execution and being able to reliably start and stop the temp server seems pretty important. Maybe it also does the `aria_log_control` wait too? (I'm uncertain if that is upgrade specific or not.)

---

There is quite a lot of custom shell code in the healthcheck script. Will the docker images be updated with fixes outside of MariaDB version bumps? For example, there is a typo in the `--su=` flag parsing: `u="${1#*-}"` -> `u="${1#*=}"`. Or that only some of the options work without `=`.

grooverdan:
> This seems a bit error prone in ensuring nothing else in docker_temp_server_start starts something in the background.

Fairly noted. I do have CI test around the upgrade that should catch this.

> Maybe adding --pid-file would be better

Testing with singularity used read only filesystems rather than overlay filesystems so if I go with a pid file, I'd probably use the default `/etc/mysql/mariadb.conf.d/50-server.cnf:pid-file                = /run/mysqld/mysqld.pid` (or however derived from `mysql_get_config pid-file $@`, as that has been shown not to have problems.

Is a global `PID=$!` that is set immediately after starting in `docker_temp_server_start` acceptable?

> `docker_temp_server_stop` needs authentication wouldn't it make sense to improve it to also handle the case where it may not have authentication

To handle this as a general case `docker_temp_server_start` should also depend on not having authentication (which is currently required). I can't use `--skip-grant-tables` as a default `docker_temp_server_start` as this prevents creating users until `FLUSH PRIVILEGES` is issued which would break compatibility.

I assume people currently using `docker_temp_server_{start,stop}` have the case covered with existing authentication. `MARIADB_MYSQL_LOCALHOST_USER` also had it mind giving a inside container privileged user available.

By implementing the `docker_mariadb_upgrade` I'm hoping I've covered one of the fairly substantial use cased of a temporary server start. I can't think of another use case for temporary start/stopping the server. Even features like data-at-rest encryption can be sufficiently controlled at configuration and runtime without a restart. Almost all system server variables are dynamic.

> Maybe it also does the aria_log_control wait too

`aria_log_control` locking was my attempt at working around that a `kill -0 $pid` was indicating no process however the file was still locked on when the main process did startup. I admit, I still don't understand why this deferred release, and even the deferred 2 second delay here. For the moment I'd like to leave this peculiarity to mariadb_upgrade only.

`mysqladmin shutdown` as just looked up, waits for the server pidfile to vanish. This happens after the pre-shutdown in the server when the server only has std(in,out,error) file descriptors open.

> There is quite a lot of custom shell code in the healthcheck script. Will the docker images be updated with fixes outside of MariaDB version bumps?

No. There needs to be a lot more CI around it, but for the most part I consider a convenience function. There are certainly higher priority fixes in the server itself that I'm not releasing pre-release updates for.

> For example, there is a typo in the --su= flag parsing: u="${1#*-}" -> u="${1#*=}". Or that only some of the options work without =.

Thank you corrected.

>  Or that only some of the options work without =

Right, `su` and `datadir` are missing a non-`=` implementation. I can correct for next time. I was getting to the point of wondering if it was going to be acceptable at all.

So my TLDR version of the next steps from the above initial review are:

1. `docker_temp_server_start` with `$!` as the pid, works for now, but better functionality/resilience for next release.
1. `docker_temp_server_{start,stop}` without authentication - seems like users could want it, but no-one has asked, and mariadb_upgrade covers the main case.
1. `aria_log_control` locking, seems a little fragile, moving back to a pidfile based like `docker_temp_server_stop`. (or just use `fg` to wait until the temp server competes shutdown?)
1. `healthcheck`, convenience script, aren't going to be doing upgrades on beyond server releases.
1. Typo, breaking `--su=`, consistent with above, not going to do a release based on it, `--su-mysql` is there (which I obviously tested more).
1. Consistent healthcheck `--X=Y` / `--X Y`, need to fix before next release.

So 3 is the only one needing immediate fixing? Based on discussion/future review do you have a preference?

grooverdan:
Note the test suite for the entrypoint `.test/run.sh {image} [testname]` is maintained.

tianon:
> I was getting to the point of wondering if it was going to be acceptable at all.

I mean...  this definitely isn't ideal.  To put how we're seeing this PR bluntly, we've got a pretty huge amount of added code here (outside the actual release of MariaDB Itself) that we don't expect to be touched again until/unless there is a new release of MariaDB upstream, which definitely won't be released just to fix issues with this script.  It kind of feels like this (the healthcheck script specifically) would be better serving users by having it available as an example and making it as easy as possible for them to get the latest version of it into their images than to try and pre-embed a potentially stale version?

For some added context, we _do_ rebuild and republish any and all "supported" tags listed in this file for things like base image updates, thus these tags are already not necessarily "immutable", so it feels strange to draw the line at updating a script (_especially_ one that's considered a convenience helper at best) when there are fixes that could be applied in between upstream updates.

The way container registries work, if a user wants/needs an "immutable" image, they can pull-by-digest and be assured that they will always get the exact same bits back every single time they pull (the digest being a cryptographic hash of the full image configuration/contents).

You can also think of it as similar to reasons behind the "Debian revision" (software version "x.y.z" will have a "-1", "-2", etc suffix for when there are relevant packaging updates but not necessarily a new actual upstream release) -- we just don't typically add a suffix on our tags because the digest provides a mechanism for users to pin more directly if they're sensitive to change, and calculating that suffix would be complicated given we rebuild for things like base image updates.

tianon:
With https://github.com/MariaDB/mariadb-docker/pull/414, the `docker-entrypoint.sh` changes look acceptable to me.

grooverdan:
I was under the impression from "Will the docker images be updated with fixes outside of MariaDB version bumps?" that you didn't particularly want frequent updates (despite as described, the ability to do so exists). As this is not the case I certainly will provide fixes in the healthcheck.sh scripts in the same way I would for docker-entrypoint.sh, and if significant, request non-version bump releases to correct these.

The examples that I've [come](https://stackoverflow.com/questions/70485527/mysql-docker-check-if-mysql-server-is-up-and-running/70486117?noredirect=1#comment124600307_70486117) [across](https://stackoverflow.com/questions/48007020/debugging-a-docker-compose-mariadb-healthcheck) on Stackoverflow have been to use an existing range of `mysqladmin ping` tests and [application code attempts](https://stackoverflow.com/questions/42014344/mariadb-docker-container-cant-connect-to-mysql-server-on-host-111-connection-r/42031017#42031017) to hack this.

I have seen the frequent comments around healthchecks on official images for user's to define and think about what healthiness/liveliness actually is. By providing this I'm hoping users will consider more than just a ping, if they need to do so. Addition items are provided, and with a extensible interface I'm hoping meaningful extensions to the currently implementation can be received as contributions without breaking compatibility for everyone.

I could consider moving this into part of the server packaging.

Thank you for the time reviewing changes here. I appreciate the feedback, however blunt it needs to be, to give our users the best image possible.

tianon:
Good enough for me, thanks! :heart:

Feel free to send PRs any time, even if the scripts are all they update (just, you know, if you're doing multiple per day for a few days, maybe reconsider what's happening in your process that's causing that and see if you can coalesce a little for all our sakes :joy:).

PR #11872, opened 2022-02-15
[Friendica] Fixing redis session handling and run_as

nupplaphil:
I accidentally introduced two major bugs :(

- Partly reverted the behavior for `run_as` and explicit `export` the two missing environment variables
- Using `expr "${REDIS_HOST}" : "/" 1>/dev/null;` to be `/bin/sh` compatible

btw I copied the Redis behavior from https://github.com/nextcloud/docker/blob/master/docker-entrypoint.sh#L60 , which is a `/bin/sh` script as well
--> shouldn't it get fixed there as well?

### DIFF
--- a/library/friendica
+++ b/library/friendica
@@ -6,45 +6,45 @@ GitFetch: refs/heads/stable
 
 Tags: 2021.09-apache, 2021.09
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
-GitCommit: 39ef8ac58708f49f47ca9c800f1262cfcdd2e754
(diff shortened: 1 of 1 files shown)


yosifkit:
There's still some rough edges with `$1` and `$@` in the `run_as`. I'd suggest simplifying it by installing a tiny tool meant only for stepping down from root in the same way that `docker run --user` does, like `gosu` (or `su-exec`) via `apt/apk`.

```sh
run_as() {
  set -- sh -c "cd /var/www/html; $*"
  if [ "$(id -u)" -eq 0 ]; then
    set -- gosu www-data "$@"
  fi
  "$@"
}
```

PR #11881, opened 2022-02-17
Update Nextcloud

J0WI:


yosifkit:
```bash
if [ "${REDIS_HOST:0:1}" = "/" ]; then
```

Friendica added this same check and found it didn't work in Debian `/bin/sh` (dash) but did work in Alpine `/bin/sh` (busybox).

PR #11896, opened 2022-02-18
Add new "python-no-pyc" test

tianon:
This test validates that the Python images do not contain any pre-compiled `.pyc` or `.pyo` files in `/usr/local` (or `/opt`, for good measure).

Ideally this would also include all of `/usr` (not just `/usr/local`) but then it would also have to filter out results that are from Debian packages (`python`'s default variants are `FROM buildpack-deps` which includes a fair amount of Python packages).

See https://github.com/docker-library/python/issues/207

(The goal is to simplify how we avoid `.pyc` and `.pyo` files in the images but with a safety net to validate that we don't regress.)

### DIFF
--- a/test/config.sh
+++ b/test/config.sh
@@ -196,6 +196,7 @@ imageTests+=(
 	[python]='
 		python-hy
 		python-imports
+		python-no-pyc
 		python-pip-requests-ssl
 		python-sqlite3
 		python-stack-size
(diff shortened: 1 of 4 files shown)


imomaliev:
Hi, could you please help me understand this [part](https://github.com/docker-library/official-images/pull/11896/files#diff-94b0ea7a5252b448573d2da70c7ef230b8d7604b3f98aa07b23375c4ec7c0910R1)
> rem N/A: Windows-based Python images may or may not contain ".pyc" and/or ".pyo" files because they are not a significant contributor to the overall image size

I never used Windows's docker images and see that they are quite large, I assume the main contributor to their size is the base Windows image. After a bit of research, I see that there is also `mcr.microsoft.com/windows/nanoserver` image and using it as a base in possible future images could require fighting for every byte as in Linux based images. Is similar test checking `py[co]` for Windows would be hard to implement/maintain or this would be just premature optimization and is not worth the effort?

PR #11899, opened 2022-02-20
Add haskell debian slim variants

AlistairB:
The non-slim variants now also include profiling support.

### DIFF
--- a/library/haskell
+++ b/library/haskell
@@ -4,15 +4,30 @@ GitRepo: https://github.com/haskell/docker-haskell
 
 Tags: 9.2.1-buster, 9.2-buster, 9-buster, buster, 9.2.1, 9.2, 9, latest
 Architectures: amd64, arm64v8
-GitCommit: 5f8acc1199035266b1cc74fcc47d8ea114268076
+GitCommit: 56d3f9996456bf633876fdb5d76bfb302b689f7c
 Directory: 9.2/buster
+
+Tags: 9.2.1-slim-buster, 9.2-slim-buster, 9-slim-buster, slim-buster, 9.2.1-slim, 9.2-slim, 9-slim, slim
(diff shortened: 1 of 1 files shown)


library/haskell
```diff
@@ -4,15 +4,30 @@ GitRepo: https://github.com/haskell/docker-haskell
 
 Tags: 9.2.1-buster, 9.2-buster, 9-buster, buster, 9.2.1, 9.2, 9, latest
 Architectures: amd64, arm64v8
-GitCommit: 5f8acc1199035266b1cc74fcc47d8ea114268076
+GitCommit: 2ef191675ddf332aa2a07ea80a4f0c10afdbbb0a
 Directory: 9.2/buster
 
+Tags: 9.2.1-buster-slim, 9.2-buster-slim, 9-buster-slim, buster-slim, 9.2.1-slim, 9.2-slim, 9-slim, slim
```
yosifkit, line 10: Just some thoughts about ordering within a tag. Tianon and I tag the images we maintain as `slim-buster` instead of `buster-slim`so as to specify that the slim applies to the general install (i.e. `haskell` in this case) and doesn't necessarily mean it is from `debian:buster-slim`.  We like to keep it open that even an image without `slim` could be from `debian:buster-slim` while a `slim` image focuses on being small whether or not it is from one of the `debian:*-slim` images. While it is not required, would it make sense for the haskell tags to follow this pattern?

AlistairB: Thanks, good question! I'm a bit busy right now but will do some investigation over the weekend and get back to you.

AlistairB: Good call out. I think the haskell images should remain consistent with established patterns used by the official images. I have changed this to `slim-buster` etc. I have also removed some non-essential packages from the slim images as this is also part of the common pattern it seems. Thanks!

PR #11900, opened 2022-02-20
Update yourls

LeoColomb:
* Fix an typo in a file name
* Fix getenv validation
* Renamed repository (previous name redirected)


LeoColomb:
I also have a question (@tianon?): is the build engine buildkit compatible?
Can I use `ARG`s to simplify the templating?
That would create file like this: https://github.com/YOURLS/docker/blob/274a62bff436d6cd3dd2d5544e8dceec84cd4477/fpm-alpine/Dockerfile#L1-L7

tianon:
You can use `ARG`, but not in `FROM` (because it makes it too complicated for us to calculate dependencies as we'd have to parse not just the `Dockerfile` but also evaluate the arguments).

We also don't support a mechanism for supplying `ARG` values in the `library/` files (as the `Dockerfile`s in the program are intended to be self-contained as much as possible).

PR #11917, opened 2022-02-23
ROS | Update rolling to jammy base image

ruffsl:
Also removes ros1-bridge tag, given no ROS1 distro targets Ubuntu jammy
https://github.com/osrf/docker_images/pull/601
https://discourse.ros.org/t/new-packages-for-ros-2-rolling-ridley/24431

PR #11924, opened 2022-02-24
Skip pgp-happy-eyeballs on build tests

yosifkit:
This is the workaround for stale servers that we are seeing in https://github.com/docker-library/official-images/pull/11917#issuecomment-1049236174 and https://github.com/docker-library/official-images/pull/11727#issuecomment-1020420044.

Using this line in [generate.sh](https://github.com/docker-library/bashbrew/blob/22e529f066b4bee5c6141f53c1059877b386bdbe/scripts/github-actions/generate.sh#L170) to skip pgp-happy-eyeballs setup.

I don't think we should remove `pgp-happy-eyeballs` from the build servers since we haven't hit the stale keyservers there. This can be re-enabled once https://github.com/tianon/pgp-happy-eyeballs/issues/4 has a proper solution.

### DIFF
--- a/.github/workflows/test-pr.yml
+++ b/.github/workflows/test-pr.yml
@@ -46,6 +46,8 @@ jobs:
       - id: generate-jobs
         name: Generate Jobs
         run: |
+          # https://github.com/tianon/pgp-happy-eyeballs/issues/4
+          export BASHBREW_GENERATE_SKIP_PGP_PROXY=1
           git clone --depth 1 https://github.com/docker-library/bashbrew.git -b master ~/bashbrew
           strategy="$(.github/workflows/generate.sh ~/bashbrew)"
           jq . <<<"$strategy" # sanity check / debugging aid


tianon (approved):

tianon:
> This can be re-enabled once https://github.com/tianon/pgp-happy-eyeballs/issues/4 has a proper solution.

IMO, we shouldn't re-enable this -- `pgp-happy-eyeballs` was designed for a time when SKS was relatively healthy, which really isn't the case anymore.  The "gossip network" might still be alive and active, but finding a good, up-to-date list of "which servers should we consider as OK" is a very challenging problem that I don't think is worth solving.

Both `hkps://keys.openpgp.org` and `hkps://keyserver.ubuntu.com` are great, stable replacements that work well and solve the problem to the extent any of our image builds actually require. :see_no_evil:

I think we should probably adjust the wording in https://github.com/docker-library/faq#openpgp--gnupg-keys-and-verification to make the current situation (and our suggested "solutions", ie those two keyservers, preferred in that order) more clear.

PR #11976, opened 2022-03-03
Update Amazon Linux images

kgharris:


kgharris:
Hi - we have some updates for our AL2 containers - please update them when you get a chance.  Thanks!

PR #11999, opened 2022-03-08
Update redmine

tianon:
Changes:

- https://github.com/docker-library/redmine/commit/13ee124: Update to 4.1.6
- https://github.com/docker-library/redmine/commit/44dd05c: Update to 4.2.4


tianon:
I've verified via `diffoscope` that the checksum changes here are all metadata/ordering changes (not anything material in the tarball contents).

Issue #10794, opened 2021-08-26
Explicitly disallow the use of "alpine-pkg-glibc"


tianon:
This is definitely a piece of #6282, but in light of https://ariadne.space/2021/08/26/there-is-no-such-thing-as-a-glibc-based-alpine-image/ (and https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/24647), we plan to enforce the no-glibc-in-Alpine-based-images (and even not mixing release pockets like 3.14 and edge) more strictly.

Fortunately, there are only two images which are currently using this:

- `sonarqube`: @christophelevis @pierre-guillot-sonarsource @tobias-trabelsi-sonarsource @lukasz-jarocki-sonarsource
- `ibmjava`: @jayasg12

Please make (short term) plans to either migrate off this solution or deprecate the affected image variants.

(See also https://github.com/docker-library/official-images/pull/10779#issuecomment-906240756 for an example of a concrete issue with the usage of this in the context of Java.)

J0WI:
Question (maybe to @kaniini?): does this affect any image using the `libc6-compat` package or is this still an acceptable workaround?

kaniini:
`libc6-compat` has largely been superseded by gcompat and needs to be retired, especially since musl no longer supports that hack

jayasg12:
Hi @tianon Is there a way to get stats of download for ibmjava Alpine images , so it helps us to take next step of action. Thanks !!

tianon:
Closing this now as anything left is really part of https://github.com/docker-library/official-images/issues/6282. :+1:

PR #12005, opened 2022-03-09
Update bonita, add 2022.1-b1

baptistemesta:
This adds the beta version of bonita: 2022.1-b1

I did not remove any previous version because this one is still a beta.

There was a lot of changes in that new version. Please tell us if anything is wrong in these changes.

Also previous images still uses sks (mentioned there: https://github.com/docker-library/faq/pull/26 ) 
Do we need to update the already existing previous versions that uses it?

Thank you!

### DIFF
--- a/library/bonita
+++ b/library/bonita
@@ -19,3 +19,7 @@ Directory: docker
 Tags: 2021.2-u0, 2021.2, 7.13.0, 7.13, latest
 GitCommit: a1d9ee5e31d38958aa553cc7f9d465f1151d902f
 Directory: docker
+
+Tags: 2022.1-b1
+GitCommit: d186e6d24a2dfd5b7020909b9e65aad14d1e59f9
+Directory: docker
--- a/library/bonita
+++ b/library/bonita
@@ -24,6 +24,6 @@ GitCommit: a1d9ee5e31d38958aa553cc7f9d465f1151d902f
 Directory: docker
 
 Tags: 2022.1-b1
-GitFetch: refs/heads/release-7.14.0
-GitCommit: b72469b76d55e0cbd3d8de293d940b79b5fd147d
+GitFetch: refs/heads/docker/2022.1
+GitCommit: a75229e9e6c73d8187f672406d3155abb5e0b192
 Directory: docker
--- a/library/bonita
+++ b/library/bonita
@@ -6,24 +6,20 @@ Maintainers: Baptiste Mesta <baptiste.mesta@bonitasoft.org> (@baptistemesta),
              Dumitru Corini <dumitru.corini@bonitasoft.org> (@DumitruCorini)
 Architectures: amd64, arm64v8, ppc64le
 GitRepo: https://github.com/bonitasoft/bonita-distrib.git
-
+Directory: docker
 
 Tags: 7.11.4, 7.11
 GitFetch: refs/heads/docker/7.11.4
 GitCommit: 7058084357dcd0fccf723ab2c7e21ec2b73f1f45
-Directory: docker
 
 Tags: 2021.1, 7.12.1, 7.12
 GitFetch: refs/heads/docker/2021.1
 GitCommit: bfdd527629063b73f053320e50e1f6bfcd135d0a
-Directory: docker
 
-Tags: 2021.2-u0, 2021.2, 7.13.0, 7.13, latest
+Tags: 2021.2-u0, 2021.2, 7.13.0, 7.13
 GitFetch: refs/heads/docker/2021.2
 GitCommit: a1d9ee5e31d38958aa553cc7f9d465f1151d902f
-Directory: docker
 
-Tags: 2022.1-b1
-GitFetch: refs/heads/docker/2022.1
-GitCommit: 6e39514bc4e558ac78930d87575c7502a89f1fec
-Directory: docker
+Tags: 2022.1-u0, 2022.1, 7.14.0, 7.14, latest
+GitFetch: refs/heads/master
+GitCommit: 694bf79347add872f8c6a4c0a7f5c3ef12c31dc8
--- a/library/bonita
+++ b/library/bonita
@@ -25,5 +25,5 @@ Directory: docker
 
 Tags: 2022.1-b1
 GitFetch: refs/heads/docker/2022.1
-GitCommit: a75229e9e6c73d8187f672406d3155abb5e0b192
+GitCommit: 6e39514bc4e558ac78930d87575c7502a89f1fec
 Directory: docker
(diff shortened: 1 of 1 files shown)


baptistemesta:
Ok the checks answers some of my questions! I'll fix that

baptistemesta:
We use a new base image eclipse-temurin alpine that is not multi arch. We are discussing if we need to change that

library/bonita
```diff
@@ -9,13 +9,18 @@ GitRepo: https://github.com/bonitasoft/bonita-distrib.git
 
 
 Tags: 7.11.4, 7.11
-GitCommit: 231024c8290a9aa31a45b758a0765a684c21ed21
+GitCommit: 7058084357dcd0fccf723ab2c7e21ec2b73f1f45
 Directory: docker
 
 Tags: 2021.1, 7.12.1, 7.12
-GitCommit: c9b816249504017bb3418252bf58ec9d4fc3e86e
+GitCommit: bfdd527629063b73f053320e50e1f6bfcd135d0a
```
tianon, line 16:
```suggestion
GitFetch: refs/heads/docker/2021.1
GitCommit: bfdd527629063b73f053320e50e1f6bfcd135d0a
```

library/bonita
```diff
@@ -9,13 +9,18 @@ GitRepo: https://github.com/bonitasoft/bonita-distrib.git
...
 Directory: docker
 
 Tags: 2021.2-u0, 2021.2, 7.13.0, 7.13, latest
 GitCommit: a1d9ee5e31d38958aa553cc7f9d465f1151d902f
```
tianon, line 20:
```suggestion
GitFetch: refs/heads/dev
GitCommit: a1d9ee5e31d38958aa553cc7f9d465f1151d902f
```

library/bonita
```diff
@@ -9,13 +9,18 @@ GitRepo: https://github.com/bonitasoft/bonita-distrib.git
...
 Directory: docker
+
+Tags: 2022.1-b1
+Architectures: amd64
+GitCommit: 89271fc68c286a4f845d5a4d0c28bfa079a5af67
```
tianon, line 25:
```suggestion
GitFetch: refs/heads/release-7.14.0
GitCommit: 89271fc68c286a4f845d5a4d0c28bfa079a5af67
```

tianon (commented):
See https://github.com/docker-library/official-images/pull/11071 for where the checks were updated to include verifying that the referenced `GitCommit:` value exists on the specified (or in your case, implied) `GitFetch:` branch -- I've included suggestions for each commit on which branch I _think_ they're supposed to apply to, but if I've got any wrong this should at least give you the appropriate format to get them pointing to the intended places.

I haven't done a _thorough_ review of the new changes (especially as you've indicated that they might change again once you've re-evaluated that base image choice), but here's a few things I noticed right off:

Regarding mixing packages from Alpine Edge and a release version, we unfortunately can't accept that (#10794).

Additionally, the tests on that new version are timing out repeatedly due to the change from `CMD` to `ENTRYPOINT` (see https://github.com/docker-library/official-images#consistency).

baptistemesta:
@tianon Hello, we integrated all feedbacks for github actions + we are now using alpine 3.15 as base on multiple architecture. Does it looks good to you?
There is no need to merge it, we only want to be sure it will be alright for the GA version that will come in few weeks!

tianon:
(marked as a draft to avoid accidental merging before you're ready -- still reviewing and hope to have more comments soon :heart:)

tianon:
> ```diff
> +RUN apk add --no-cache --upgrade curl unzip bash su-exec jattach openjdk11-jre-headless \
>    && rm -rf /var/lib/apt/lists/*
> ```

Removing `/var/lib/apt/lists` is unnecessary (leftover from being Ubuntu-based), and `--upgrade` shouldn't be here (if the base image needs updates, we encourage base image maintainers to update instead):

```dockerfile
RUN apk add --no-cache curl unzip bash su-exec jattach openjdk11-jre-headless
```

As a side note, you could also consider switching from `curl` to `wget` and avoid installing that (assuming Bonita itself doesn't require `curl` at runtime), since BusyBox includes a working `wget` (and the Alpine build of it even supports `https` URLs with validation out-of-the-box).

---

> ```diff
> +ENV HTTP_API_USERNAME ${HTTP_API_USERNAME:-http-api}
> +ENV HTTP_API_PASSWORD ${HTTP_API_PASSWORD:-""}
> ...
> ```

It doesn't really make much difference in the end (only in the interim maintainability) but this whole block's uses of `${...:-...}` are unnecessary/unused:

```dockerfile
ENV HTTP_API_USERNAME http-api
ENV HTTP_API_PASSWORD ""
...
```

(Should most/all of these be moved down next to `ENV HTTP_API false`?  They seem similar.)

---

> ```diff
> +## the double space is important inside the sha256sum on that version of busybox
> +  && echo "$BONITA_SHA256  /opt/files/$ZIP_FILE" | sha256sum -c - ; fi \
> ```

Instead of the error-prone double-space, I'd suggest using the explicit "binary mode" signifier of `*` (which then only requires a single space):

```dockerfile
  && echo "$BONITA_SHA256 */opt/files/$ZIP_FILE" | sha256sum -c - ; fi \
```

---

> ```diff
> +# Ensure there is no entrypoint, docker hub require to have CMD be overridable
> +ENTRYPOINT []
>  # command to run when the container starts
> -CMD ["/opt/files/startup.sh"]
> -
> +CMD ["/opt/files/startup.sh", "/opt/bonita/server/bin/catalina.sh", "run"]
> ```

Unless I'm misreading the logic in it, you appear to have updated `startup.sh` to meet the requirements for being `ENTRYPOINT`, so the following _should_ be OK (if not, we can help figure out where/why and how we suggest fixing it, if this is something you'd like to pursue -- I'm guessing it'll be something like making sure any `echo`/output goes to stderr instead of stdout, or that relevant things only run if `$1` is `catalina.sh`):

```dockerfile
ENTRYPOINT ["/opt/files/startup.sh"]
CMD ["/opt/bonita/server/bin/catalina.sh", "run"]
```

If it's going to be common for users to need to invoke `catalina.sh run` with different arguments _or_ if there are other tools/scripts in `/opt/bonita/server/bin` which are useful for users to invoke directly, it might be worth considering putting `/opt/bonita/server/bin` in `PATH` so that you could have `CMD ["catalina.sh", "run"]` instead (which is then much easier for users to accurately reproduce if they need to adjust it slightly).

baptistemesta:
Thank you for the great review!

I've integrated most of your comments. For now I've kept curl but it's definitively something that can be changed.
Regarding the entry point, I chose to ignore the initialisation when the CMD is not `catalina`, as you suggested.

I've put the PR back as "ready to review" because it looks like the Github Actions are not triggering otherwise

yosifkit:
Attempting to re-trigger github actions.

educhastenier:
@tianon we integrated all your remarks.
Can you give a final look? Thx.

PR #12022, opened 2022-03-11
MariaDB - fix aarch64 (10.2/bionic), healtcheck.sh

grooverdan:
We erroniously released a 10.2 aarch64 bionic package that
wasn't installable due to broken dependences. This is now fixed.

The healthcheck.sh script had spelling mistakes in its autogeneration
for the 10.6 tags that resulted in:
* The default datadir being /var/lib/mariadb instead of /var/lib/mysql
* the --su-mysql parameter got broken.


grooverdan:
thanks @tianon

PR #12030, opened 2022-03-14
Update httpd

yosifkit:
Changes:

- https://github.com/docker-library/httpd/commit/95c96d2: Update to 2.4.53


icemagno:
Where is a2enmod ?

yosifkit:
`a2enmod` is specific to Debian packaging ( `apt-get install apache2`) and not from the upstream `httpd` project. These images are a simple reflection of upstream `httpd` releases.

PR #12036, opened 2022-03-14
Update Rocky Linux Images - March 2022

NeilHanlon:
Hi all,

The plan with the images is to keep the minor version tag (e.g. 8.5) pinned to the initial release, and use latest as
well as the major version (8) tag to follow the most recent image.

I believe the change I've done should accomplish that, but would appreciate a second set of eyes, too.

Thank you!

Best,
Neil


tianon:
> The plan with the images is to keep the minor version tag (e.g. 8.5) pinned to the initial release, and use latest as well as the major version (8) tag to follow the most recent image.

This is generally OK (we're not going to block it if it's what you really want to do), but I would like to understand what the use case is?  Why would someone want to pin to 8.5 explicitly "as released" and not have any relevant security updates included in the base image?  (Wouldn't most additional package installs then likely also want to update packages from the base right away?)

PR #12037, opened 2022-03-14
Updates for neurodebian

yarikoptic:
per https://github.com/docker-library/official-images/pull/10721#issuecomment-1067113919
and another reminder to myself to fully automate this whole setup ;)

### DIFF
--- a/library/neurodebian
+++ b/library/neurodebian
@@ -1,6 +1,6 @@
 Maintainers: Yaroslav Halchenko <debian@onerussian.com> (@yarikoptic)
 GitRepo: https://github.com/neurodebian/dockerfiles.git
-GitCommit: a95bab388a1791db8ca2830d78417637bd0ca167
+GitCommit: 6cd018bf3cf473a152920194e9611ebd2af09f15
 
 Tags: xenial, nd16.04
 Directory: dockerfiles/xenial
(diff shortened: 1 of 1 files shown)


tianon:
Looks like the CI failures are related to https://github.com/docker-library/faq/pull/26 -- mind updating to either `hkps://keys.openpgp.org` or `keyserver.ubuntu.com` (depending on whether the key's email address is/can be verified at https://keys.openpgp.org/ ?) :pray: :heart:

yarikoptic:
ever saw

```
+ gpg --batch --keyserver hkps://keyserver.ubuntu.com --recv-keys DD95CC430502E37EF840ACEEA5D32F012649A5A9
gpg: keybox '/tmp/tmp.wUx5lSS2EW/pubring.kbx' created
gpg: keyserver receive failed: General error
```
@tianon ?

tianon:
Hmm, I think that generally just means "something happened in `dirmngr`" but in this case I think it's because you're trying to use `hkps` but don't have `ca-certificates`?  When you're using the full fingerprint, `gpg` will validate that the received key matches the fingerprint, so we don't typically worry about using `hkps` too much. :sweat_smile:

yarikoptic:
ok, seems to be happy now ;)

PR #12038, opened 2022-03-14
[Swift] Add Swift 5.6 Release to offical image

shahmishal:


shahmishal:
@yosifkit This is my first time including multiple arch in the official Docker image, please let me know if you see any issues. Thanks in advance.

tianon:
Regarding the CI failures, see #11831 (`centos:8` is EOL and thus unsupported), so those `centos:8` variants are failing to build and should be removed.

Regarding the multi-arch support, I'm all in favor and love to see it!

Unfortunately, `arch` is a shorthand for `uname -m`, which detects the kernel architecture instead of the userspace architecture (which is more what you're trying to detect; see also https://github.com/docker-library/official-images#multiple-architectures), so I'd prefer to see something like `dpkg --print-architecture`, `apk --print-arch`, etc used in images where that's possible, and I'd also recommend explicitly capturing the `amd64` branch in your `case` so your `*` fallback can be an error (see https://github.com/docker-library/julia/blob/8eaad1380b78ddd0b9148de2cdceb527624e8b3a/1.7/bullseye/Dockerfile#L42-L64 for a good example of what I mean).

For RPM-based distributions, they definitely have a "userspace architecture" but it's not normally recorded anywhere (that I'm aware of - I'd love to be wrong :see_no_evil:), so the best we've come up with is something clever around `objdump` (https://github.com/docker-library/openjdk/blob/76633a8f235476207b7ce7534826236e1ced626a/19/jdk/oraclelinux8/Dockerfile#L37-L48).  If that's not available in the image already, then `uname -m` is an acceptable fallback there.

I'm happy to help implement any of these changes if that would be helpful.

shahmishal:
> Regarding the CI failures, see #11831 (`centos:8` is EOL and thus unsupported), so those `centos:8` variants are failing to build and should be removed.

I will remove CentOS 8

> Regarding the multi-arch support, I'm all in favor and love to see it!
> 
> Unfortunately, `arch` is a shorthand for `uname -m`, which detects the kernel architecture instead of the userspace architecture (which is more what you're trying to detect; see also https://github.com/docker-library/official-images#multiple-architectures), so I'd prefer to see something like `dpkg --print-architecture`, `apk --print-arch`, etc used in images where that's possible, and I'd also recommend explicitly capturing the `amd64` branch in your `case` so your `*` fallback can be an error (see https://github.com/docker-library/julia/blob/8eaad1380b78ddd0b9148de2cdceb527624e8b3a/1.7/bullseye/Dockerfile#L42-L64 for a good example of what I mean).
> 
> For RPM-based distributions, they definitely have a "userspace architecture" but it's not normally recorded anywhere (that I'm aware of - I'd love to be wrong 🙈), so the best we've come up with is something clever around `objdump` (https://github.com/docker-library/openjdk/blob/76633a8f235476207b7ce7534826236e1ced626a/19/jdk/oraclelinux8/Dockerfile#L37-L48). If that's not available in the image already, then `uname -m` is an acceptable fallback there.
> 
> I'm happy to help implement any of these changes if that would be helpful.

For Ubuntu 20.04, I will update the Dockerfile to use `dpkg --print-architecture`.
For Amazon Linux 2, we can use `_arch` micro from rpm `rpm --eval '%{_arch}'`.

shahmishal:
@tianon @yosifkit - Not sure if we missed something, we are not seeing `arm64` images on hub.docker.com/_/swift. 

```
 % docker manifest inspect swift:5.6.0-focal
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 1162,
         "digest": "sha256:a34fb87d14544c49d5aa30c90fcf18347d301e7db1b1e917b23796d687c3e178",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      }
   ]
}
% docker manifest inspect swift:5.6.0-amazonlinux2
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 1162,
         "digest": "sha256:7bd26bcdd1d5fbcade07174f2cdb7edcd247d10cad93b8962c8603168c23289f",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      }
   ]
}
```

shahmishal:
The issue most likely related to Jenkins job not generating the arm64 job for swift. 
https://doi-janky.infosiftr.net/job/multiarch/job/_generator/2238/console

yosifkit:
Yeah, :disappointed: jenkins forces us to approve the job DSL every time there is a new image/arch combo.  I've triggered the builds now.

shahmishal:
Thank you @yosifkit

PR #12041, opened 2022-03-15
Eggdrop v1.9.2 *STABLE*

vanosg:


tianon:
Just to have a written record of it, as we discussed on IRC, I think `echo "$EGGDROP_SHA256 develop.tar.gz" | sha256sum -c -` needs to switch to `echo "$EGGDROP_SHA256 *develop.tar.gz" | sha256sum -c -` (for some reason, the BusyBox `sha256sum` in Alpine appears to be detecting this tarball as text and processing it stragely as a result, needing the explicit `*` binary file specifier to counteract that).

PR #12047, opened 2022-03-16
[varnish] release 7.1.0

gquintard:
this one deserves a bit more attention than usual as it adds two new packages: `vmod_dynamic` and `varnish-modules`. They are fairly light and the packaging should be ok, but as usual, I'll welcome all feedback. Cheers!

### DIFF
--- a/library/varnish
+++ b/library/varnish
@@ -1,26 +1,26 @@
-# this file was generated using https://github.com/varnish/docker-varnish/blob/d1212e4b8fd35b58c19b01ed389f8841d0a4ea38/populate.sh
-Maintainers: Guillaume Quintard <guillaume@varni.sh> (@gquintard)
+# this file was generated using https://github.com/varnish/docker-varnish/blob/693cddb1140d5c0e042793cbb92b6906fb8425a5/populate.sh
+Maintainers: Guillaume Quintard <guillaume.quintard@gmail.com> (@gquintard)
 GitRepo: https://github.com/varnish/docker-varnish.git
 
-Tags: fresh, 7.0.2, 7.0, latest
+Tags: fresh, 7.1.0, 7.1, latest
 Architectures: amd64, arm32v7, arm64v8, i386, ppc64le, s390x
 Directory: fresh/debian
-GitCommit: 6423e32e0afd20fa876276e1525cb318920fbefa
+GitCommit: bb05db3ed5d75a1d8402309336823e2d58330b1b
 
-Tags: fresh-alpine, 7.0.2-alpine, 7.0-alpine, alpine
+Tags: fresh-alpine, 7.1.0-alpine, 7.1-alpine, alpine
 Architectures: amd64, arm32v7, arm64v8, i386, ppc64le, s390x
 Directory: fresh/alpine
-GitCommit: 6423e32e0afd20fa876276e1525cb318920fbefa
+GitCommit: 693cddb1140d5c0e042793cbb92b6906fb8425a5
 
-Tags: old, 6.6.2, 6.6
+Tags: old, 7.0.2, 7.0
 Architectures: amd64, arm32v7, arm64v8, i386, ppc64le, s390x
 Directory: old/debian
-GitCommit: 6423e32e0afd20fa876276e1525cb318920fbefa
+GitCommit: 2012f1cbc7c401d603584ac0b601555af180eb09
 
-Tags: old-alpine, 6.6.2-alpine, 6.6-alpine
+Tags: old-alpine, 7.0.2-alpine, 7.0-alpine
 Architectures: amd64, arm32v7, arm64v8, i386, ppc64le, s390x
 Directory: old/alpine
-GitCommit: 6423e32e0afd20fa876276e1525cb318920fbefa
+GitCommit: 2012f1cbc7c401d603584ac0b601555af180eb09
 
 Tags: stable, 6.0.10, 6.0
 Architectures: amd64, arm32v7, arm64v8, i386, ppc64le, s390x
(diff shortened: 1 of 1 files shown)


tianon:
It's a little bit strange to have the `APKBUILD` and `debian/` files committed directly and included via `COPY` (which doesn't have super great cache semantics) -- is that intentional?

gquintard:
it's intentional, but mainly because I don't know better :-)
It didn't seem smart to create a new repo to hold those, so I went for simple.

What is the usual way of dealing with such a situation?

tianon:
I guess the better way to phrase that question might've been "are these new packages Docker specific?" (or are they things that might/will be used outside Docker too?)

If they're Docker specific, why are they packages instead of just building/installing the relevant upstream sources directly in the `Dockerfile` instead of doing the work to create explicit `debian/` and `APKBUILD` packaging scripts to then embed those inside a packaging script of a different kind?  (It's a little bit like creating an RPM spec file that then invokes `dpkg-buildpackge` or something. :sweat_smile:)

gquintard:
Ah ok, I can definitely answer those (and that is probably going to be wrong :-) )

From a general standpoint, I tend to prefer having packages as it's usually cleaner to install/remove. The point is probably moot here since the image is the end game.
I already had the `APKBUILD` for `varnish-modules` since I maintain it, and that was just me staying in my comfort zone, and because I was down that rabbit hole already, I did the same thing for the `debian` side for the sake of consistency.

Sooooo, I should just install brutally and be done with it?

gquintard:
there, it should be better :-) (it definitely looks simpler, at the very least)

yosifkit:
Looks pretty clean. :+1: 

I think there is duplicated work between lines [17-27](https://github.com/varnish/docker-varnish/blob/bb05db3ed5d75a1d8402309336823e2d58330b1b/next/alpine/Dockerfile#L17-L27) and [30-40](https://github.com/varnish/docker-varnish/blob/bb05db3ed5d75a1d8402309336823e2d58330b1b/next/alpine/Dockerfile#L30-L40) on the Alpine version.

gquintard:
ah, indeed, just fixed it

yosifkit:
Would the new `USER varnish` possibly break users that might've set a storage backend like `file`? Is that acceptable?

gquintard:
that's a good question, and yes, it will break it. I think that's acceptable as the `file` storage is lost anyway when `varnishd` restarts

tianon:
Did you want to update to `alpine:3.15` while you're at it, or was `3.14` intentional? :eyes:

gquintard:
completely unintentional, and now corrected, thanks!

library/varnish
```diff
@@ -1,26 +1,26 @@
...
 Directory: fresh/debian
-GitCommit: 6423e32e0afd20fa876276e1525cb318920fbefa
+GitCommit: bb05db3ed5d75a1d8402309336823e2d58330b1b
 
-Tags: fresh-alpine, 7.0.2-alpine, 7.0-alpine, alpine
+Tags: fresh-alpine, 7.1.0-alpine, 7.1-alpine, alpine
 Architectures: amd64, arm32v7, arm64v8, i386, ppc64le, s390x
 Directory: fresh/alpine
-GitCommit: 6423e32e0afd20fa876276e1525cb318920fbefa
+GitCommit: 693cddb1140d5c0e042793cbb92b6906fb8425a5
```
tianon, line 13: I think you meant to go a little bit higher here? (to pick up that `alpine:3.15` change in https://github.com/varnish/docker-varnish/commit/2ac414bf99af92cabbf4a924108f0c3af190f47f)

```suggestion
GitCommit: 2ac414bf99af92cabbf4a924108f0c3af190f47f
```

spantaleev:
Looks like running the 7.1 container image (which contains `USER varnish`) with `--tmpfs /var/lib/varnish:exec` (as recommended in the Docker Hub description) no longer works.

- **[OK]** ` docker run --rm --entrypoint=/bin/sh docker.io/varnish:7.1.0-alpine -c 'stat /var/lib/varnish'` says `0750 / varnish / varnish` -- the `varnish` user will be able to write to this

- **[not OK]** `docker run --rm --entrypoint=/bin/sh --tmpfs /var/lib/varnish:exec docker.io/varnish:7.1.0-alpine -c 'stat /var/lib/varnish'` says `0750 / root / root` -- the `varnish` user will not be able to write to a `root`-owned directory with these permissions

- **[not OK]** `docker run --rm --entrypoint=/bin/sh --mount type=tmpfs,destination=/var/lib/varnish,tmpfs-mode=1777 docker.io/varnish:7.1.0-alpine -c 'stat /var/lib/varnish'` says `0750 / root / root` -- same as the above - the `varnish` user will not be able to write to a `root`-owned directory with these permissions

- **[OK]** `docker run --rm --entrypoint=/bin/sh --mount type=tmpfs,destination=/var/lib/another,tmpfs-mode=1777 docker.io/varnish:7.1.0-alpine -c 'stat /var/lib/another'` says `1777 / root / root` -- the `varnish` user will be able to write to this **new** `root`-owned `/var/lib/another` directory

The Debian-based image suffers from the same problem.

---------

As a workaround, one might:

- remove the `--tmpfs` argument and have Varnish write to the container filesystem at `/var/lib/varnish` - not great
- start varnish with another `tmpfs` directory (e.g. `--tmpfs /var/lib/varnish2:exec`) by specifying another working directory: `-n /var/lib/varnish2` - ugly, but tolerable
- try to work around the `USER varnish` thing by using `--user` (possibly `--user=0:0`) - not great

gquintard:
can you open a ticket in `varnish/docker-varnish` please?

PR #12051, opened 2022-03-16
Plone: Fix zeo shared blob env

avoinea:
Refs https://github.com/plone/plone.docker/pull/167

### DIFF
--- a/library/plone
+++ b/library/plone
@@ -6,15 +6,5 @@ GitRepo: https://github.com/plone/plone.docker.git
 
 Tags: 5.2.7-python38, 5.2-python38, 5-python38, python38, 5.2.7, 5.2, 5, latest
 Architectures: amd64
-GitCommit: 49cbc9cf28091756ee5d0d908c232a4bb1e93425
+GitCommit: d17783d696042a3e9c2214c289306493e75f5ddc
 Directory: 5.2/5.2.7/debian
-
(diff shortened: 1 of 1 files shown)


tianon:
Just to be clear, this change is needed for those two older tags too? (Looking at https://github.com/docker-library/official-images/pull/11795, it appears you're up to three concurrent versions only temporarily? :sweat_smile:)

Also, that CI failure is a little weird -- it appears there's some extra unexpected whitespace in the output?  Do you think that's just a fluke that we should restart for, or something the test needs to be updated for?

avoinea:
> Just to be clear, this change is needed for those two older tags too? (Looking at #11795, it appears you're up to three concurrent versions only temporarily? sweat_smile)

True, we can skip the older tags.

> Also, that CI failure is a little weird -- it appears there's some extra unexpected whitespace in the output? Do you think that's just a fluke that we should restart for, or something the test needs to be updated for?

It seems a thing. I'll look into it :wink:

tianon:
It's probably been a minute since we shared our standard "copypasta" about older tags with you, so just for reference/reminder:

:+1:

tianon:
Hopefully https://www.githubstatus.com/incidents/fpk08rxnqjz2 is resolved soon :see_no_evil:

tianon:
(Rebased to trigger GitHub Actions :+1:)

PR #12055, opened 2022-03-16
alpine: edge snapshot 20220316, bump v3.15.1 (CVE-2022-0778)

ncopa:
fixes https://github.com/alpinelinux/docker-alpine/issues/243


yosifkit:
Should 3.14 and older versions also have a release/rebuild for CVE-2022-0778?

ncopa:
> Should 3.14 and older versions also have a release/rebuild for [CVE-2022-0778](https://github.com/advisories/GHSA-x3mh-jvjw-3xwx)?

yes. I am working on that too.

PR #12056, opened 2022-03-16
Update Amazon Linux 2022 images

sktomer:
New Release Update: 2022.0.20220308.1

Thanks,
Sumit


tianon:
This appears to include an OpenSSL library that's still vulnerable to CVE-2022-0778 (`openssl-libs-1.1.1l-2` both before and after this update) -- did you intend to include that update in this bump?

sktomer:
Hello Tianon,

Thanks for bringing this to our notice. I have confirmed that we are actively working on a new version of AL 2022 containing the fix for the mentioned CVE.

That will be another update after this one though.

Thanks,
Sumit

PR #12060, opened 2022-03-17
Update Amazon Linux images

boostyc:
New AL2 images release 2.0.20220316.0

Thanks,
Sam


tianon:
This was probably OK but just colliding with https://github.com/docker-library/official-images/pull/12056 a little bit, right?  (That's how I'm interpreting that CI failure.)

PR #12073, opened 2022-03-17
mariadb fix /docker-entrypoint-initdb.d replication

grooverdan:
A regression that a container master doesn't contain
the binary logs of its initdb.d data and cannot replicate this.

Fix details: https://github.com/MariaDB/mariadb-docker/pull/421.


grooverdan:
Thank you @tianon

PR #12075, opened 2022-03-17
Update Ubuntu

mwhudson:
20220315 (bionic)
20220316 (focal)
20220316 (impish)
20220315 (jammy)


mwhudson:
This is a bit ahead of when we'd usually update, but I'm told an official image update will kick off downstream builds to incorporate the openssl CVE fix from a couple of days ago.

tianon:
I'm seeing `libssl3:amd64  3.0.1-0ubuntu1` in the context (and one other that actually contains `libssl1.1`, but with an update) -- do you want to double check that we're in the clear across all of these, or would you rather merge as-is?

tianon:
Followed up on IRC -- @mwhudson says go. :smile:

mwhudson:
Well in general there is no security support for devel.

Issue #9445, opened 2021-01-18
Adding the Red Hat Universal Base Image as an official image


bstinsonmhk:
First I'd like to make some introductions, @fatherlinux and @brianwcook are the points of contact for the [Red Hat Universal Base Image](https://developers.redhat.com/products/rhel/ubi). 

It would be great to use this issue to discuss any questions they might have, and to reference against the PR when ready to add this entry to the library. 

We should review the [README](https://github.com/docker-library/official-images#contributing-to-the-standard-library) and the [Review Criteria](https://github.com/docker-library/official-images/blob/master/NEW-IMAGE-CHECKLIST.md) as we generate our Pull request. 

Scott, Brian: What questions do we have here? We chatted a little bit about signing the other day I remember.

grooverdan:
I'd like to see to see UBI in the notable exceptions https://github.com/docker-library/official-images#repeatability so that official images can be based on UBI

tianon:
We're definitely open to UBI becoming part of the program, and would love to have more conversation around that, but it will not be added as an exception.

We frankly don't love the Windows base image exceptions (and have considered even consuming them into top-level images), but they're necessary for running containers at all on Windows due to the way Windows containers work (which is also why that list is very specific about which of the `mcr.microsoft.com` images are considered acceptable, and the implementation details of our checks for that are even more particular about which specific tags are considered acceptable).

grooverdan:
Related, just saw the Red Hat announcement - https://developers.redhat.com/articles/2021/05/25/red-hat-universal-base-image-and-docker-hub-why-should-developers-care

LenzGr:
As we were just trying to submit updated images by ourselves (we migrated from CentOS 8 to UBI), I was wondering if there's any update or outlook on this topic? What's missing and how could we help in getting this resolved?

fatherlinux:
@justincormack and I discussed this a few weeks back, and I think Docker is working on a long term solution. It's a much more difficult problem than I had originally thought. Originally, I asked if we could just push the UBI image into an Official repository, but Justin pointed out that this creates other problems.

Just might have more information, or he might be able to point us in the direction of the right person to work with.

curiositycasualty:
Sorry to add a notification to everyone's inbox.

Are there anymore updates on this? Anything I can help with to get this over the finish line?
Correct me if I'm wrong, but it seems like we're waiting on feedback from @justincormack?

gdams:
[...]

PR #12093, opened 2022-03-21
[RFC] trying to migrate to debian-slim and ubi-slim

hutchic:
we at Kong would like to migrate our base images from ubuntu, centos (not building anymore already), redhat (only publishing to redhat catalog) to alpine, debian-slim and ubi-slim.

Please don't merge this PR it's just to trigger CI as well as a discussion about any of the guidlines / requirements I may have missed fullfilling.


library/kong
```diff
@@ -1,12 +1,23 @@
 Maintainers: Kong Docker Maintainers <support@konghq.com> (@thekonginc)
 GitRepo: https://github.com/Kong/docker-kong.git
 
-Tags: 2.8.0-alpine, 2.8.0, 2.8, alpine, latest
-GitCommit: c32a80c3dd332d096d1a6461d2816ead344d43c5
-GitFetch: refs/tags/2.8.0
-Directory: alpine
+Tags: 2.8.0-20220321-alpine, 2.8.0-alpine, 2.8-alpine, alpine, 2.8.0, 2.8, latest
```
hutchic, line 4: The calver tag is so we can have a mix of mutable (rebuild when a new base image is available) vs immutable.

I think this is OK as I see prior art examples though I couldn't find any official docs for or against this practice

hutchic:
cc @tianon for feedback or triage to others for feedback?

Also is there somewhere where it's documented what's allowed as an official image base and I'm just failing to find it (ubi isn't but windows core is)?

yosifkit:
> Also is there somewhere where it's documented what's allowed as an official image base and I'm just failing to find it (ubi isn't but windows core is)?

> No official images can be derived from, or depend on, non-official images with the following notable exceptions
> ...
> \- https://github.com/docker-library/official-images/tree/68032672b0f17d3b85d145e874c01353f0aad0cb#repeatability

As far as getting UBI into official-images, there was some discussion in https://github.com/docker-library/official-images/issues/9445.

---

The build looks pretty straightforward; couldn't these instead just download, install, and delete the installer/tar in a single `RUN` instead of using multiple stages? See also our [Multi-stage Builds](https://github.com/docker-library/faq/tree/73f10b0daf2fb8e7b38efaccc0e90b3510919d51#multi-stage-builds) entry in our FAQ.

---

```dockerfile
COPY download-package.sh /tmp/download-package.sh
RUN ... \
	./tmp/download-package.sh $KONG_VERSION
```

> It may be tempting, for the sake of brevity, to put complicated initialization details into a standalone script and merely add a `RUN` command in the `Dockerfile`. However, this causes the resulting `Dockerfile` to be overly opaque, and such `Dockerfile`s are unlikely to pass review. Instead, it is recommended to put all the commands for initialization into the `Dockerfile` as appropriate `RUN` or `ENV` command combinations.
>
> \-https://github.com/docker-library/official-images/tree/68032672b0f17d3b85d145e874c01353f0aad0cb#clarity

---

The `FROM` targets for the Debian and Alpine images need to be more specific. Both `alpine:latest` and `debian:stable-slim` are very "square" wheels that come with many package upgrades when the new release comes out. This has a very high likelihood to break builds. `alpine:3.15` and `debian:bullseye-slim` would be better targets.

library/kong
tianon: We don't actually support any "immutable" tags, so if Alpine is updated, `2.8.0-20220321-alpine` will be rebuilt.  What we suggest for users who require actual immutable image references is to use pull-by-digest instead, which is designed for cryptographic bit-for-bit reproducibility (ie, `docker pull kong:2.8@sha256:ec7199734e3af2d21dfcf566352b4cdf5c939ffcdf9a1cd0ad4f3ec6aac914b2` instead of just `kong:2.8`).

hutchic:
got the feedback I required -- thank you.

PR #12101, opened 2022-03-23
Update sapmachine to contain 18

chalstrick:


yosifkit:
We recommend against `ADD` for any remote URL. We only recommend using `ADD` for a base image to extract the tar contents. We recommend fetching keys via `gpg` from a keyserver rather than just a remote file since `gpg` does some verification of the content.

Related: https://github.com/docker-library/faq/tree/2c48cc77c354ec18edc13ac017a8e2a382fae74c#openpgp--gnupg-keys-and-verification

PR #12112, opened 2022-03-24
erlang/otp 23.3.4.12, 25.0-rc2

getong:


library/erlang
```diff
@@ -3,19 +3,19 @@
 Maintainers: Mr C0B <denc716@gmail.com> (@c0b)
 GitRepo: https://github.com/erlang/docker-erlang-otp.git
 
-Tags: 25.0-rc1.0.0, 25.0-rc1.0, 25.0-rc1, 25
+Tags: 25.0-rc2.0.0, 25.0-rc2.0, 25.0-rc2, 25
```
yosifkit, line 6: I just noticed that the `rc` tags are a little strange. Perhaps they are meant to be `25.0.0.0-rc2, 25.0.0-rc2`?

The `rc1` tags are already this format, so I think we can merge as-is and then fix/change them in a separate PR as necessary.

PR #12113, opened 2022-03-24
Releasing Neo4j 4.3.11 and 4.4.5

jennyowen:
Thank you!

### DIFF
--- a/library/neo4j
+++ b/library/neo4j
@@ -9,12 +9,22 @@ Maintainers: Chris Gioran <chris@neo4j.com> (@digitalstain),
              Bledi Feshti <bledi.feshti@neo4j.com> (@bfeshti)
 GitRepo: https://github.com/neo4j/docker-neo4j-publish.git
 
-Tags: 4.4.4, 4.4.4-community, 4.4, 4.4-community, community, latest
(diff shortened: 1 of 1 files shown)


tianon:
Looks OK!  The only concern that I'd like to verify (for your sake!) is that `get_neo4j_run_cmd` returns a properly _quoted_ string such that it's suitable/safe for `eval`?  (I'm guessing it does or that it isn't likely to include spaces or anything like `$` that would be bad if it's not quoted correctly, but I want to make sure :heart:)

jennyowen:
Oh that's very strange. That function has been in the entrypoint for nearly a year now, I wonder why it's suddenly showing up in the diff? I will investigate.

jennyowen:
OK I understand, that function is in the 4.3.0 and upwards Neo4j images, and the diff was between 4.2 and 4.3 ones, so it was flagged.

Since that function's been in production for a while I'll not fix it right now, but I think you may have a point about the quoting. It's supposed to return some big long java command, but who's to say there won't be feature creep in the future that introduces `$` signs or something. I'll see what I can do before our next release.

I'm glad there's someone checking my work!

PR #12151, opened 2022-03-30
composer 2.2.10

alcohol:


tianon:
That 1.10.25 failure is probably worth looking more into. :sweat_smile:

(However, 2.2.10 looks fine, so I'm going to merge anyhow.)

PR #12159, opened 2022-03-31
Fix for MariaDB images

grooverdan:
It was intention that the timezone initialization
wasn't written to the binary log as it can slow
down initialization. This unfortunately regressed.

A test case is now written for this.


tianon:
```diff
 			mysql_tzinfo_to_sql --skip-write-binlog /usr/share/zoneinfo \
-			| docker_process_sql --dont-use-mysql-root-password --database=mysql
+
+			echo "SET SESSION SQL_LOG_BIN=@save_sql_log_bin;"
```

I think you probably meant to remove that trailing `\` on the `mysql_tzinfo_to_sql` line, right?

(I _think_ the shell will have it combine with the blank line, not the echo, but probably better safe than sorry?  Per the shell spec, I think it's _technically_ undefined behavior. :smile:)

grooverdan:
Thanks @tianon.

PR #12205, opened 2022-04-07
arangodb 3.9.1

KVS85:
Updated ArangoDB 3.9 to 3.9.1.

### DIFF
--- a/library/arangodb
+++ b/library/arangodb
@@ -10,6 +10,6 @@ Tags: 3.8, 3.8.6
 GitCommit: af354ba85e29e2bee525677b45a321530235a38d
 Directory: alpine/3.8.6
 
-Tags: 3.9, 3.9.0, latest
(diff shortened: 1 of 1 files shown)


library/arangodb
```diff
@@ -10,6 +10,6 @@ Tags: 3.8, 3.8.6
 GitCommit: af354ba85e29e2bee525677b45a321530235a38d
 Directory: alpine/3.8.6
 
-Tags: 3.9, 3.9.0, latest
-GitCommit: de7e9211691614f105c7809cb13405ef5afa0be8
-Directory: alpine/3.9.0
+Tags: 3.9, 3.9.1 latest
```
yosifkit, line 13: It looks like the comma got lost :see_no_evil:

KVS85: Thanks for the catch!

PR #12212, opened 2022-04-08
bump emqx version to 4.4.2

Rory-Z:


yosifkit:
There is quite a bit added here and this is what I reviewed so far. These first two are things that need to be fixed, but the rest are just recommendations and discussion points based on our experience.
- The `curl` invocation should have a `-f` to make sure it fails on non-200 responses.
- Instead of the just a single space, I'd suggest using the explicit "binary mode" signifier of `*`:
  ```bash
  echo "$sha256 *$pkg" | sha256sum -c
  ```
- Would it be easier to just source `/etc/os-release` and use `"${ID}${VERSION_ID}"` instead of keeping `ENV OS debian11` in sync with the `FROM`?
- We had a long history in the RabbitMQ image of maintaining a bash script that would `sed` an Erlang code/config file; it was all a bit fragile and prone to user misconfiguration (since types and such aren't enforcible in environment variables) and we have since removed all of the functionality. The generated/edited config file caused issues for upstream RabbitMQ when helping users since it could create invalid combinations of configuration. I guess what I am trying to point out is that this is a path with some pitfalls when users come back for debugging help.
- `VOLUME`:
  The general rule _we_ follow is that you could have a volume for where the application stores data, but configuration is almost never appropriate in a volume (unless it's specifically managed by the application itself at runtime and _never_ supplied/managed manually by users, or there is an easy way to override the default in the image).
  
  The reason for this is that, once defined, Docker has no means of "`UNVOLUME`" -- thus, once a volume, always a volume.
  
  Logs are a bit more iffy -- those are probably appropriate for having in a volume if it's really necessary for the application to log on-disk (rather than simply using Docker's built-in logging mechanism via stdio).

Rory-Z:
Hi，@yosifkit thanks for reply
I have modified the code as you suggested, could you please review again
Also I have a question, I don't see any separate enterprise image in the library, do we not accept PR on enterprise image?

yosifkit:
> I don't see any separate enterprise image in the library, do we not accept PR on enterprise image

There's some discussion happening inside Docker as to whether enterprise images are an appropriate image to include in the official images program, given the software is not under an Open Source license. One of the the official images [tenets](https://github.com/docker-library/official-images/tree/df204551b689e37488d072663a6e54680851b977#what-are-official-images) is "Focus on Free and Open-Source Software".  I think it would probably be worthwhile to send an email to "justin.cormack @ docker.com" and "jeff_hadfield @ docker.com" to discuss further with them.

Rory-Z:
Thx for answer

PR #12213, opened 2022-04-08
bump Kong to 2.8.1, 2.7.2, 2.6.1

kikito:


library/kong
```diff
@@ -1,37 +1,73 @@
...
+Directory: alpine
+Architectures: amd64, arm64v8
+
+Tags: 2.8.1-ubuntu, 2.8-ubuntu, ubuntu
+GitCommit: fffcd0a942daba110ccb89878570b50fcff64d97
+GitFetch: refs/tags/2.8.1
+Directory: ubuntu
+Architectures: amd64, arm64v8
+
+Tags: 2.8.0-alpine, 2.8.0
```
yosifkit, line 16: Any particular reason to keep these old versions listed here?  Are they still supported releases?  Removing it here will remove it from the "Supported" section on the Hub readme (and will prevent us from spending cycles rebuilding it on the official build server), but the tag will still be available to users who want it.  (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)

yosifkit: Going to merge to get the new version out, but I'd recommend thinking about the older versions.

kikito: Hello, it was a combination of a script not doing its job completely correctly and me misunderstanding the outcome when I attempted to fix manually.

I have removed the older versions of 2.8.x, 2.7.x and 2.6.x, keeping only the most up-to-date ones. Thanks for your comments.

PR #12260, opened 2022-04-19
feat: update to CouchDB 3.2.2

janl:
we published a new CouchDB release just now, thanks!


big-r81 (commented):
+1

yosifkit:
It seems that the `3.2.2` Dockerfile still [references `3.2.1-1` ](https://github.com/apache/couchdb-docker/blob/d861e435acb35e191aaacaf12d6c1f148557d8ec/3.2.2/Dockerfile#L57).

janl:
@yosifkit oh duh, thanks for the heads up. It is fixed now on `main` o/

yosifkit:

janl:
thanks!

PR #12268, opened 2022-04-20
[percona] Update psmdb-5.0 version

vorsel:
1. update psmdb-5.0 version
2. change base image to oraclelinux.
3. move ENV declarations to the place where they were before.


yosifkit:
Ignoring failing builds that are unchanged in this PR.

yosifkit:

PR #12270, opened 2022-04-20
Update for Consul 1.12.0, remove 1.9.17

markan:
Signed-off-by: Mark Anderson <manderson@hashicorp.com>


yosifkit:
Not a requirement to change now, but you may want to consider when to do an Alpine Linux version bump since the new `1.12.0` is still `FROM alpine:3.13` which will reach "end of support" on 2022-11-01 (https://alpinelinux.org/releases/).

---

PR #12275, opened 2022-04-21
Release Ubuntu Jammy Jellyfish

darkdragon-001:
See tianon/docker-brew-ubuntu-core#235


mwhudson:
Jammy isn't released yet so this is a bit early.

PR #12281, opened 2022-04-21
Aerospike 5.7.0.17

pvinh-spike:

### DIFF
--- a/library/aerospike
+++ b/library/aerospike
@@ -2,12 +2,12 @@ Maintainers: Lucien Volmar <lucien@aerospike.com> (@volmarl),
              Michael Coberly <mcoberly@aerospike.com> (@mcoberly2),
              Phuc Vinh <pvinh@aerospike.com> (@pvinh-spike)
 
-Tags: ee-5.7.0.16, latest
+Tags: ee-5.7.0.17, latest
 Architectures: amd64
 GitRepo: https://github.com/aerospike/aerospike-server-enterprise.docker.git
-GitCommit: daa166776b86bf2c2dd19d5333c14a0c6b37b8f3
+GitCommit: 45a16710725f6666cf4fa9949fccfaaf7ee099b9
 
-Tags: ce-5.7.0.16
+Tags: ce-5.7.0.17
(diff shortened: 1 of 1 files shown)


library/aerospike
```diff
@@ -2,12 +2,12 @@ Maintainers: Lucien Volmar <lucien@aerospike.com> (@volmarl),
              Michael Coberly <mcoberly@aerospike.com> (@mcoberly2),
              Phuc Vinh <pvinh@aerospike.com> (@pvinh-spike)
 
-Tags: ee-5.7.0.16, latest
+Tags: ee-5.7.0.17, latest
```
mcoberly2, line 5:
```suggestion
Tags: ee-5.7.0.17
```

mcoberly2 (commented):
We need to update latest on docker hub. Please remove this as part of the link until we change the latest link from 6.0.0.0-rc7 to 5.7.0.17 in docker hub

yosifkit:
I'm a bit confused. You said "We need to update latest" and then removed the `latest` tag. :confused: So did you want to keep `latest` unchanged? As the PR is now, `latest` will still be the same as `ee-5.7.0.16` and neither will be rebuilt for base image changes.

The `*-5.7.0.17` images comment out the `paxos-single-replica-limit` setting that is `obsolete as of 6.0`; is that correct to comment out for `5.7`? (I just want to make sure that 6.0 prep changes don't have an adverse effect if applied to the earlier versions).

mcoberly2:
@yosifkit Great question regarding paxos-single-replica-limit.  Let me consult with my peers. Please do not merge this.

mcoberly2:
@yosifkit We are good to go with this Pull Request.

Thanks again for your attention to detail. 
* We are not listing latest as a link.
* The removal of `paxos-single-replica-limit` from the `aerospike.template.conf` is not an issue for any of the released versions 5.5.0.x, 5.6.0.x, 5.7.0.x, 6.0.0.0-rc# that now include that change.  This was verified by our development team and also we tested our docker build minus this configuration parameter.

PR #12289, opened 2022-04-22
Update eclipse-temurin for new releases (part 1)

sxa:
Updates to new quarterly release for  JDK11.0.15 and 17.0.3

### DIFF
--- a/library/eclipse-temurin
+++ b/library/eclipse-temurin
@@ -115,16 +115,16 @@ GitCommit: 78a1d4d015ee0759f269888a9d085ddc03f20b95
 Directory: 11/jdk/alpine
 File: Dockerfile.releases.full
 
-Tags: 11.0.14.1_1-jdk-focal, 11-jdk-focal, 11-focal
-SharedTags: 11.0.14.1_1-jdk, 11-jdk, 11
-Architectures: amd64, arm32v7, arm64v8, ppc64le, s390x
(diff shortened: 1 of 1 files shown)


yosifkit:
This is going to hit the same issue as https://github.com/docker-library/python/issues/681 (and  https://github.com/docker-library/official-images/pull/11760#discussion_r794920624). So, any `Shared Tags` listed on the older images will lose all images and architectures not still in the `library/eclipse-temurin` file. So, `11.0.14.1_1-jdk`, `8u322-b06-jdk`, and others will have only windows images.

sxa:
Thanks for letting me know. I hadn't been aware of that happening previously - will discuss with my team.

sxa:
@yosifkit Done - we've got the full set of platforms ready for 11 and 17 now so I've modified the PR to reflect that. We'll need to see if we can do anything to be able to split them out at some point later :-)

yosifkit:
The `library/` file needs to be a complete description of any specific tag. It is acceptable if a new tag doesn't have all of it's OS/architecture combinations, but any already published tags likely need to maintain their OS/architecture set until that tag is removed.
I think the two future options to avoid the tagging problem linked above is to either remove the old version completely from the `library/` file, or to keep all of that old version until the new version is complete. Both of these will ensure that the manifest list doesn't get updated to exclude previously supported OS/architecture combinations. (I'd probably lean toward dropping the old version, since it'll still be available to pull and doesn't likely need rebuilds triggered from base image updates.)

PR #12298, opened 2022-04-24
Update YOURLS

LeoColomb:
* Bump YOURLS to v1.9
* Bump PHP to 8.1
* Minor fixes and improvements

### DIFF
--- a/library/yourls
+++ b/library/yourls
@@ -1,21 +1,21 @@
-# this file is generated via https://github.com/YOURLS/docker/blob/194f6035a6f05671e649be6eb96d189353abcc16/bin/generate-stackbrew-library.sh
+# this file is generated via https://github.com/YOURLS/docker/blob/b4a6fa40eb3292e03df95776a8acaf6436ed5a5d/bin/generate-stackbrew-library.sh
 
 Maintainers: YOURLS <yourls@yourls.org> (@YOURLS),
              Léo Colombaro <git@colombaro.fr> (@LeoColomb)
 GitRepo: https://github.com/YOURLS/docker.git
 GitFetch: refs/heads/dist
 
-Tags: 1.8.2-apache, 1.8-apache, 1-apache, apache, 1.8.2, 1.8, 1, latest
+Tags: 1.9-apache, 1-apache, apache, 1.9, 1, latest
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
-GitCommit: 194f6035a6f05671e649be6eb96d189353abcc16
+GitCommit: b4a6fa40eb3292e03df95776a8acaf6436ed5a5d
 Directory: apache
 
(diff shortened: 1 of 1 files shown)


LeoColomb:
Oh, no BuildKit support? 😭

tianon:
Indeed, unfortunately we use plain BuildKit-less `docker build`.

LeoColomb:
Alright, reverted back to plain `docker build` compat.

By the way, as visible with the force-push, we now use a bot to generate the branch for this pull-request.
Do you think it would be possible to allow [this same app](https://github.com/apps/yourls-bot) to create the pull request as well? (it requires to install it on this repo)

yosifkit:
I'm not sure the reason for adding the `ARG` declarations early in the Dockerfile, but they should be declared as late in the Dockerfile as possible in order to minimize `docker build` cache bust when they change.

---

We have other maintainers using automation to submit pull requests and haven't needed a GitHub app so far. Here is our [current bot guidelines](https://github.com/docker-library/faq/tree/2c48cc77c354ec18edc13ac017a8e2a382fae74c#can-i-use-a-bot-to-make-my-image-update-prs).

LeoColomb:
> I'm not sure the reason for adding the `ARG` declarations early in the Dockerfile

This is to avoid multiple static references and reduce mistake surface, without breaking the cache as `ENV` does and allowing to reference it in the `LABEL`s.
But I get the point, and I guess I could move the labeling at a later line.

> other maintainers using automation to submit pull requests

Using "fake" personal accounts, like @docker-library-bot, right?
Yes, this an option, but actually more a workaround now, since it is much better if used with a GitHub App (secure with temp tokens, strict permissions, and not faking a user account). 
But it also come with a downside, which is the app must be installed on the repo.
Let me know what you think about this.

tianon:
> > I'm not sure the reason for adding the `ARG` declarations early in the Dockerfile
> 
> This is to avoid multiple static references and reduce mistake surface, without breaking the cache as `ENV` does and allowing to reference it in the `LABEL`s. But I get the point, and I guess I could move the labeling at a later line.

Unfortunately, `ARG` breaks the cache in the same way `ENV` does. :see_no_evil: :disappointed:

> > other maintainers using automation to submit pull requests
> 
> Using "fake" personal accounts, like @docker-library-bot, right? Yes, this an option, but actually more a workaround now, since it is much better if used with a GitHub App (secure with temp tokens, strict permissions, and not faking a user account). But it also come with a downside, which is the app must be installed on the repo. Let me know what you think about this.

Yes, unfortunately providing an application access to our repository is not something we can accept (it's a pretty wide security issue for us to do so).

LeoColomb:
> Unfortunately, `ARG` breaks the cache in the same way `ENV` does. 🙈 😞

Ok, fixed ✨

> Yes, unfortunately providing an application access to our repository is not something we can accept

Alright

PR #12307, opened 2022-04-26
Update Nextcloud

J0WI:


J0WI:
Where does the rename come from?

yosifkit:
> Where does the rename come from?

That's just how the `diff-pr.sh` script shows results sometimes.

---

This unbounded `until` loop is not ideal. If the one container that was doing the sync happens to exit before removing the `.lock` file, then there would be little indication that something was wrong since they would all be "up". I'd recommend putting a limit on the count so that there is at least a flapping container to help indicate it is not ready or there is some other problem.

PR #12308, opened 2022-04-26
Update XWiki to 14.3

manuelleduc:


tianon:
I think you probably intended to update some of the `GitCommit` values here, right?
(there don't appear to be any `Dockerfile` changes :smile:)

manuelleduc:
Thanks a lot for catching this. Fixed :slightly_smiling_face:

PR #12335, opened 2022-04-29
eclipse-temurin: remove JDK16

gdams:
JDK16 is no longer supported so there will be no further updates to the dockerfiles.

@tianon I assume that making this change won't remove the existing tags?


yosifkit:
> making this change won't remove the existing tags?

Correct, they will still be available to `docker pull`.

PR #12366, opened 2022-05-04
eclipse-temurin: add 8u332-b09

gdams:


yosifkit:
This is going to hit the same issue as https://github.com/docker-library/python/issues/681 (and  https://github.com/docker-library/official-images/pull/11760#discussion_r794920624). So, any `Shared Tags` listed on the older images will lose all images and architectures not still in the `library/eclipse-temurin` file. So, `8u322-b06-jdk` will have only windows images.

I think the two options to avoid the linked tagging problem is to either remove the old version completely from the `library/` file, or to keep all of that old version until the new version is complete. Both of these will ensure that the manifest list doesn't get updated to exclude previously supported OS/architecture combinations. (I'd probably lean toward dropping the old version, since it'll still be available to pull and doesn't likely need rebuilds triggered from base image updates.)

gdams:
@yosifkit windows has been added now, this should be good to merge

PR #12385, opened 2022-05-07
Complete Revamp of the TomEE Images

scriptmonkey:
In this PR, I have:

- Removed OpenJDK17
- Added Temurin 8, 11, and 17
- Added Semeru 11 and 17
- Added complete tags so users can be more selective

Given the significance of these changes, I would like to have comments on these changes from the official-images maintainers.  Also, I am going to try to get some comments from the TomEE maintainers.

Please review carefully

Thanks,
SM


tianon:
Whew, you are not kidding - that is a very large matrix.  It reminds me a lot of the `tomcat` list (https://github.com/docker-library/docs/blob/master/tomcat/README.md#supported-tags-and-respective-dockerfile-links) which is similarly totally out of control and we've been seriously considering slimming down to a very small set (like at most one variant per Tomcat version) and at most providing example `Dockerfile`s for users who need something outside the smaller matrix, because not only is it kind of obscene to maintain such a large matrix of possibilities, it also actively makes the life of users more difficult ("which tag should I use?", etc). :grimacing:

scriptmonkey:
> Whew, you are not kidding - that is a very large matrix. It reminds me a lot of the `tomcat` list (https://github.com/docker-library/docs/blob/master/tomcat/README.md#supported-tags-and-respective-dockerfile-links) which is similarly totally out of control and we've been seriously considering slimming down to a very small set (like at most one variant per Tomcat version) and at most providing example `Dockerfile`s for users who need something outside the smaller matrix, because not only is it kind of obscene to maintain such a large matrix of possibilities, it also actively makes the life of users more difficult ("which tag should I use?", etc). 😬

Tomcat TomEE are in the same boat.  Given that we could run on different OSes and Java providers and java major version, it gets complex very quickly.  However, TomEE has the added wrinkle of 4 separate flavors.  

I foresee the OpenJDK images going away, but probably not the near future.  The next lowest demand was probably Semeru.  If you said we had to drop something, it would be my guess we would drop those images.

yosifkit:
I noticed that the `8`, `9`, and `latest` tags (maybe others?) are moving from `webprofile` to `microprofile`; will that break users of those tags?

scriptmonkey:
> I noticed that the 8, 9, and latest tags (maybe others?) are moving from webprofile to microprofile; will that break users of those tags?

Yes, and that was NOT planned.  I will look in to it.

yosifkit (approved):
This looks good to me. @scriptmonkey, would you like us to wait for some other TomEE maintainers to comment?

jwhitesel:
Having created the recent issue requesting generic tags, the list is move extensive then I expected. The tags appear to be consistent among the many tomee images. [I will assume the "non-version specific" tags will float with each new release.]

scriptmonkey:
> This looks good to me. @scriptmonkey, would you like us to wait for some other TomEE maintainers to comment?

Please merge the changes.  It has been long enough for people to comment.

Thank you for your patience!

PR #12389, opened 2022-05-08
Update rethinkdb tags

gabor-boros:
## Description

This PR updates the rethinkdb tags. Also, it removes Centos8 as we didn't build packages for that.

### DIFF
--- a/library/rethinkdb
+++ b/library/rethinkdb
@@ -1,7 +1,7 @@
 Maintainers: RethinkDB <services@rethinkdb.com> (@gabor-boros)
 GitRepo: https://github.com/rethinkdb/rethinkdb-dockerfiles.git
 
-Tags: 2.4.2-buster-slim, 2.4-buster-slim, 2-buster-slim, buster-slim, 2.4.2, 2.4, 2, latest
-Directory: buster/2.4.2
+Tags: 2.4.2-bullseye-slim, 2.4-bullseye-slim, 2-bullseye-slim, bullseye-slim, 2.4.2, 2.4, 2, latest
+Directory: bullseye/2.4.2
 GitCommit: eda09076e55ff437fcb57ae7d52146f2bd5bc46b
 Architectures: amd64, arm64v8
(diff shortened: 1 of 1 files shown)


gabor-boros:
cc: @tianon @yosifkit as you helped me the last time with PRs. 😇

tianon:
No worries!  This looks fine to me (once you figure out what's up with your ARM builds :see_no_evil:).

For a future PR, I'd suggest considering an update to Debian Bullseye (not at all urgent - Buster still has a lot of life left) and switching off `apt-key adv` (it's officially deprecated/discouraged).  Specifically, I'd recommend something like this: https://github.com/docker-library/mysql/blob/68bc91b85ffde4ec9c0ca084b092acfb28463bfa/8.0/Dockerfile.debian#L54-L67 (using `gpg` to fetch/export the key via the full fingerprint, then adding `signed-by` to the `sources.list` line).

gabor-boros:
@tianon Updated to Bullseye which already had the following, so that resolves the deprecation warning too:

```Dockerfile
RUN curl -fsSL https://download.rethinkdb.com/repository/raw/pubkey.gpg | gpg --dearmor | tee /usr/share/keyrings/rethinkdb.gpg | gpg --armor \
    && echo "deb [signed-by=/usr/share/keyrings/rethinkdb.gpg] https://download.rethinkdb.com/repository/debian-bullseye bullseye main" > /etc/apt/sources.list.d/rethinkdb.list
```

gabor-boros:
@tianon Wow, thank you for the PR. I just merged it and updated the git commit hash on this PR to `826a4193366e7d0ff176d101679385125b8fa4f1`.

tianon:
My pleasure -- I figured a PR would help make it more directly actionable for you than just a comment here. :smile:

gabor-boros:
@tianon Although I annotated the tag in the PR that this version supports armv8, it seems that on hub.docker.com it only indicates x86_64. Am I missing something?

tianon:
I'm not sure what updates those category listings (I bet they'll be updated within an hour or so), but the tags all have `arm64` builds now. :+1:

gabor-boros:
@tianon Such automation tricks :D Thank you for checking it! It looks good 🤩

PR #12399, opened 2022-05-10
[percona] Update psmdb-5.0 version

vorsel:


tianon:
CI failures are on unchanged versions (although I'd suggest updating or removing them -- they're currently not getting the benefit of rebuilds but are still claimed as "supported" :see_no_evil:)

vorsel:
> CI failures are on unchanged versions (although I'd suggest updating or removing them -- they're currently not getting the benefit of rebuilds but are still claimed as "supported" 🙈)

Hi Tianon,
We'll create an internal task for checking the outdated images if we should update them or if it's better to remove them.

PR #12408, opened 2022-05-11
Update dart to 2.17.0

whesse:


kevmoo:
@tianon – I see 2.17 here - https://hub.docker.com/_/dart – but pulling still gets the old versions.

Am I missing something?

tianon:
> FYI, it's probably going to be a little bit before this builds thanks to #12393 (the queue is pretty full ATM). :sweat_smile:

:eyes:

(Updating the description isn't tied to the builds because the build matrix is frankly quite complex.)

kevmoo:
Unfortunately confusing, but thanks for letting me know. :-)

PR #12425, opened 2022-05-12
Update phpMyAdmin, fixing Docker errors

ibennetch:
We recently released phpMyAdmin 5.2.0, which changed the configuration system a bit. Those changes were not compatible with how we package Docker, so for the past day or so, users have not been able to import custom settings to their docker containers.

Correct some errors with phpMyAdmin that were preventing custom settings from being loaded, fixing https://github.com/phpmyadmin/docker/issues/361.

Signed-off-by: Isaac Bennetch <bennetch@gmail.com>


ibennetch:
Thanks @tianon for the quick release!

PR #12427, opened 2022-05-13

adivinho:


tianon:
https://github.com/docker-library/official-images/pull/12399#issuecomment-1123014876

> CI failures are on unchanged versions (although I'd suggest updating or removing them -- they're currently not getting the benefit of rebuilds but are still claimed as "supported" :see_no_evil:)

PR #12429, opened 2022-05-13
22.0.0.5 release

crpotter:
Updates for the release of websphere-liberty and open-liberty version 22.0.0.5.


tianon:
That `websphere-liberty` test build is taking _suspiciously_ long on _all_ versions/variants -- did something go wrong?  (Did you test a `docker build` of these before submitting this? :sweat_smile:)

crpotter:
Hm... yes, these build appropriately for us on our end. Is it possible it was something in the build environment here and they just need to be launched again? I don't want to try a relaunch and tie up resources unnecessarily.

PR #12435, opened 2022-05-16
Rocky Linux updates for 8.6

NeilHanlon:
* Introduce new minimal image


NeilHanlon:
do not merge, please. a couple fixes needed.

yosifkit:
1. It seems that the amd64 `8.6` Dockerfile had a generation bug: https://github.com/rocky-linux/sig-cloud-instance-images/blob/3728fffe3fe4b7e33b136d20fd084e137fbd06ab/Dockerfile#L2
2. > no other tag ever contains the string `latest`, since it's not something users are expected or encouraged to actually type out
   >
   > \- https://github.com/docker-library/official-images/tree/872198ac8cd8f70f7df0883d1cd6878e465603bf#tags-and-aliases

PR #12436, opened 2022-05-16
Add alibaba-dragonwell

joeyleeeeeee97:


joeyleeeeeee97:
@tianon , Hi, please take a look

joeyleeeeeee97:
ping.

joeyleeeeeee97:
@tianon ,please take a look

joeyleeeeeee97:
@tianon , any updates?

joeyleeeeeee97:
@tianon

joeyleeeeeee97:
@yosifkit @tianon 
`Dragonwell` is a customized JDK popular in Alibaba Cloud
https://dragonwell-jdk.io/
https://github.com/alibaba/dragonwell8

We are trying to add it to offical docker hub

tianon:
Thank you for your patience. :bow:

(As a note, please do not modify the checklist in the OP here -- we use that for our state tracking in a transparent way, but it only shows up in places like the PR list if it's in the OP, which is why it's there instead of in a comment :pray: :heart:)

Overall I think this isn't too bad -- I've got a few initial review comments and I hope that together we can make this even better. :smile:

---

```yaml
SharedTags: xxx
```

We use `SharedTags:` for overlapping Windows tags with Linux tags -- in this case, they're not necessary (and can be moved to just use `Tags:` in all instances I see).

---

```diff
+alibaba-dragonwell:dragonwell-8.11.12_jdk8u332-ga-jdk-centos
+alibaba-dragonwell:dragonwell-8.11.12_jdk8u332-ga-jdk-focal
+alibaba-dragonwell:dragonwell-11.0.15.11.9_jdk-11.0.15-ga-alpine
+alibaba-dragonwell:dragonwell-11.0.15.11.9_jdk-11.0.15-ga-jdk-alpine
+alibaba-dragonwell:dragonwell-11.0.15.11.9_jdk-11.0.15-ga-jdk-centos7
+alibaba-dragonwell:dragonwell-11.0.15.11.9_jdk-11.0.15-ga-jdk-focal
+alibaba-dragonwell:dragonwell-17.0.3.0.3-jdk-centos7
+alibaba-dragonwell:dragonwell-17.0.3.0.3-jdk-focal
```

These tags are all really repetitive -- is there a reason that you're duplicating `dragonwell` in the tag names on top of the repository?

---

```diff
+FROM alpine:3.15
```

These could probably be updated to Alpine 3.16 with some testing :eyes:

(Although staying on Alpine 3.15 for now is totally fine too -- both are actively supported so you could just as easily update later.  It might be a good idea to add something like `-alpine3.15` tag aliases for users to pin to a specific Alpine release so that they have a way to revert if/when you do upgrade that, but that's ultimately up to you.)

---

```diff
+RUN apk update && apk add ca-certificates && update-ca-certificates && apk --no-cache add openssl wget
+
+RUN apk add --no-cache tzdata musl-locales musl-locales-lang \
+    && rm -rf /var/cache/apk/*
```

I think this would have the same end-result with something simpler like this:

```dockerfile
RUN apk add --no-cache ca-certificates musl-locales musl-locales-lang openssl tzdata wget
```

(Using `--no-cache` should take care of `/var/cache/apk` and installing `ca-certificates` shouldn't then require running `update-ca-certificates` explicitly afterwards :+1:)

---

```diff
+ENV JAVA_VERSION jdk-11.0.15+10
```

This doesn't seem to be used (and appears to mis-match with the version number in `BINARY_URL`? :grimacing:) so could probably just be removed?

---

```diff
+FROM centos:7
```

Are you including CentOS variants specifically because they're well-supported and actively recommended by the Dragonwell project, or more "just because" ?  (The `centos` image is not actively maintained, so builds of this variant will become stale very quickly, and I would honestly suggest that you probably want to just remove them for now.)

---

```diff
+# ------------------------------------------------------------------------------
+#               NOTE: THIS DOCKERFILE IS GENERATED VIA "update_multiarch.sh"
+#
+#                       PLEASE DO NOT EDIT IT DIRECTLY.
+# ------------------------------------------------------------------------------
```

Looking at your repository, this doesn't seem to be accurate, and these `Dockerfile`s appear to all instead be hand-maintained?  I'd suggest removing this notice (or generating your `Dockerfile`s instead, which is going to help you with maintaining them long-term).

joeyleeeeeee97:
- [x] Removed SharedTags
- [ ] for `dragonwell` in tags: I thought using the exact value from github tag might be a good idea, let users easier to find and locate sources, so current tag is formatted as ${GITHUB_TAG}-jdk-{PLATFORM}
As tags in [dragonwell-8.11.12_jdk8u332-ga](https://github.com/alibaba/dragonwell8/releases/tag/dragonwell-8.11.12_jdk8u332-ga)
- [x] update tags with alpine-3.16
- [x]  update apk command
- [ ]  ENV JAVA VERSION
Frankly I am trying to be the same as openjdk https://github.com/docker-library/openjdk/blob/790a2a8e26ad738e1ce7b31220cb4c564d0bb2e1/11/jdk/oraclelinux7/Dockerfile#L35, as some users might expect the same behavior from dragonwell image
- [x] remove inaccurate note
- [ ]  centos images
After some discussion we thought might still support this, most of our users are using centos, and eclipse-termurin is  supporting centos also. So we might keep those.

yosifkit:
`alibaba-dragonwell:11-jdk-alpine3.16` still has the old `apk update/add` lines.

---

The `JAVA_VERSION` environment variable on the java 11 variants (`jdk-11.0.15+10`) doesn't seem to match the installed versions.

Output during build:
```console
javac --version
javac 11.0.15.11
java --version
openjdk 11.0.15.11 2022-04-19
OpenJDK Runtime Environment (Alibaba Dragonwell)-11.0.15.11+9-GA (build 11.0.15.11+9)
OpenJDK 64-Bit Server VM (Alibaba Dragonwell)-11.0.15.11+9-GA (build 11.0.15.11+9, mixed mode)
```
Downloaded jdk files:
```
dragonwell-11.0.15.11.9_jdk-11.0.15-ga/Alibaba_Dragonwell_11.0.15.11.9_x64_alpine-linux.tar.gz
dragonwell-11.0.15.11.9_jdk-11.0.15-ga/Alibaba_Dragonwell_11.0.15.11.9_x64_linux.tar.gz
...
```
Should it be something else like `ENV JAVA_VERSION jdk-11.0.15+11` or `jdk-11.0.15+9` or `jdk-11.0.15`?

joeyleeeeeee97:
@yosifkit 
- old apk update/add lines.
Yes. I forgot to change one file, thanks for noticing.

- corrected JAVA_VERSION to use the exact value with dragonwell
For 11, we are using `jdk-11.0.15.11+9`, this is from openjdk `jdk-11.0.15+9` but we added 11 as extended version for our development changes, for senarios like `still from latest ga upstream , but we did some major changes on our features`.

yosifkit:
It looks like the `dragonwell-releng/dragonwell-containers` repo uses a `main` branch, so the library files will need a `GitFetch: refs/heads/main` (failing github actions check). Otherwise, the rest looks good to me.

tianon:

yosifkit:

PR #12454, opened 2022-05-18
[oraclelinux] Changed tag from 8.5 to 8.6.

amsteinb-orcl:

### DIFF
--- a/library/oraclelinux
+++ b/library/oraclelinux
@@ -13,6 +13,13 @@ Tags: 8.6, 8
 Architectures: amd64, arm64v8
 Directory: 8
 
+# temporary 8.5 tag fix
+Tags: 8.5
+Architectures: amd64, arm64v8
+Directory: 8
+amd64-GitCommit: b2631b4c3944ae70c9463ae62aaa75a1a4877a16
+arm64v8-GitCommit: ae688d9eef59795cd485bd9d9f43ce9015c3c461
+
 Tags: 8-slim
 Architectures: amd64, arm64v8
 Directory: 8-slim
--- a/library/oraclelinux
+++ b/library/oraclelinux
@@ -9,7 +9,7 @@ amd64-GitCommit: 2190ef249f644a54672562adab58500744619a38
 arm64v8-GitFetch: refs/heads/dist-arm64v8
 arm64v8-GitCommit: 2b89de16f2d760fb883964fcec81b1281ae8e6c5
 
-Tags: 8.5, 8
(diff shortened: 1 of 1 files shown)


yosifkit:
Did you want to re-add the `8.5` section so that the tag can be rebuilt to point at the previous build? I think the built tests here might fail, but our builder will still have access to the old commit and do fine.

And then it can be removed once re-pushed to Docker Hub.

amsteinb-orcl:
I unfortunately do not know the correct action here.  The latest container images are for the Oracle Linux 8.6 release. I did not realize until after the upload that the library file needed to be updated.

yosifkit:
This PR will correctly update the `8.6` tag and we can just go ahead and merge, but the `8.5` tag will be stuck pointing at an image containing 8.6. Do you want the `8.5` tag corrected?

If we add an extra section like this temporarily, it should allow a rebuild and push to fix the `8.5` tag to be pointing at the correct older image. I took the commit ids from the values previous to https://github.com/docker-library/official-images/pull/12451.

```yaml

# temporary 8.5 tag fix
Tags: 8.5
Architectures: amd64, arm64v8
amd64-GitCommit: b2631b4c3944ae70c9463ae62aaa75a1a4877a16
arm64v8-GitCommit: ae688d9eef59795cd485bd9d9f43ce9015c3c461
```

amsteinb-orcl:
I made the change as you suggested. Thank you.

yosifkit:
It looks right to me and the failing builds are expected from my end. Those old commits are not referenced in the specified `GitFetch` branches, but they are still accessible via github directly (like https://github.com/oracle/container-images/commit/b2631b4c3944ae70c9463ae62aaa75a1a4877a16), so we should be safe to merge.

cc @tianon

library/oraclelinux
```diff
@@ -9,10 +9,16 @@ amd64-GitCommit: 2190ef249f644a54672562adab58500744619a38
...
 
-Tags: 8.5, 8
+Tags: 8.6, 8
 Architectures: amd64, arm64v8
 Directory: 8
 
+# temporary 8.5 tag fix
+Tags: 8.5
+Architectures: amd64, arm64v8
+amd64-GitCommit: b2631b4c3944ae70c9463ae62aaa75a1a4877a16
```
tianon, line 19: This one needs `Directory:` also :eyes:

```suggestion
Directory: 8
amd64-GitCommit: b2631b4c3944ae70c9463ae62aaa75a1a4877a16
```

PR #12470, opened 2022-05-21
MariaDB releases 20220520

dbart:
Hi,

We've done our 10.9.1, 10.8.3, 10.7.4, 10.6.8, 10.5.16, 10.4.25, 10.3.35, 10.2.44 maintenance releases.


grooverdan:
@dbart, [10.9.1](https://mariadb.com/kb/en/mariadb-1091-release-notes/) is listed as `rc` and [10.8.4](https://mariadb.com/kb/en/release-notes-mariadb-108-series/) as `ga` which is different from the tags here. If you need to manually edit, ensure `10, latest` end up on 10.8 for consistency.

The update.sh script is pulling the info from https://downloads.mariadb.org/rest-api/mariadb/ so updating there first if possible.

PR #12478, opened 2022-05-23
Add official image for SATOSA

niheconomoum:
SATOSA is a configurable proxy for translating between different authentication protocols such as SAML2, OpenID Connect, and OAuth2.  It implements the Access Protocol Translation layer of [the Authentication and Authorization for Research Collaborations (AARC) Blueprint Architecture](https://aarc-project.eu/architecture/), and it is the foundation of several spoke-hub-style identity federations.  Adopters include research institutions, higher education, and governments.

# Checklist for Review

    - https://github.com/IdentityPython
    - Apache 2.0, MIT, CC0
    - service
    - https://github.com/docker-library/docs/pull/2158

### DIFF
--- a/library/satosa
+++ b/library/satosa
@@ -1,4 +1,4 @@
-# This file is generated via https://github.com/IdentityPython/satosa-docker/blob/ce9ed1fe9762516c42b0315fd2c7ac7b01809f5a/generate-stackbrew-library.sh
+# This file is generated via https://github.com/IdentityPython/satosa-docker/blob/f39d2b97aee3a36753e40a486f531f421d328d43/generate-stackbrew-library.sh
 
 Maintainers: Matthew X. Economou <economoum@niaid.nih.gov> (@niheconomoum)
 GitRepo: https://github.com/IdentityPython/satosa-docker.git
@@ -7,15 +7,10 @@ GitFetch: refs/heads/main
 Tags: 8.1.1-bullseye, 8.1-bullseye, 8-bullseye, bullseye
 SharedTags: 8.1.1, 8.1, 8, latest
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
-GitCommit: f19ecfded89178f2a90626b87c16ebc8c3ffa94f
+GitCommit: ae2f317df322553b13fc07042348828969868bfa
 Directory: 8.1/bullseye
-
-Tags: 8.1.1-slim-bullseye, 8.1-slim-bullseye, 8-slim-bullseye, slim-bullseye, 8.1.1-slim, 8.1-slim, 8-slim, slim
-Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
-GitCommit: f19ecfded89178f2a90626b87c16ebc8c3ffa94f
-Directory: 8.1/slim-bullseye
 
 Tags: 8.1.1-alpine3.16, 8.1-alpine3.16, 8-alpine3.16, alpine3.16, 8.1.1-alpine, 8.1-alpine, 8-alpine, alpine
 Architectures: amd64, arm32v6, arm32v7, arm64v8, i386, ppc64le, s390x
-GitCommit: f19ecfded89178f2a90626b87c16ebc8c3ffa94f
+GitCommit: ae2f317df322553b13fc07042348828969868bfa
 Directory: 8.1/alpine3.16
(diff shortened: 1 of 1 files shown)


tianon:
> I don't know. I'm blindly supporting every variant the base Python container supports.

Just to reiterate more persistently here what we discussed on IRC: without an explicit use case for an end-user to need to pick a specific version/variant of Python, I'd suggest trimming this down to only one variant (maybe two, if you want to also support a "smaller" Alpine-based build) for now.  It's much less disruptive (for users) to add more variants later than it is to take them away, and supporting less variants will make both your and our lives easier. :smile:

niheconomoum:
@tianon are three variants ok?  This commit reduces the list of variants to bullseye, slim-bullseye, and alpine3.16, all based on the latest Python release.

library/satosa
```diff
@@ -0,0 +1,21 @@
...
+Tags: 8.1.1-bullseye, 8.1-bullseye, 8-bullseye, bullseye
+SharedTags: 8.1.1, 8.1, 8, latest
+Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
+GitCommit: f19ecfded89178f2a90626b87c16ebc8c3ffa94f
+Directory: 8.1/bullseye
+
+Tags: 8.1.1-slim-bullseye, 8.1-slim-bullseye, 8-slim-bullseye, slim-bullseye, 8.1.1-slim, 8.1-slim, 8-slim, slim
+Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
+GitCommit: f19ecfded89178f2a90626b87c16ebc8c3ffa94f
+Directory: 8.1/slim-bullseye
```
tianon, line 16:
```suggestion
Tags: 8.1.1-bullseye, 8.1-bullseye, 8-bullseye, bullseye, 8.1.1, 8.1, 8, latest
Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
GitCommit: f19ecfded89178f2a90626b87c16ebc8c3ffa94f
Directory: 8.1/slim-bullseye
```

niheconomoum:
Thanks again for the feedback!  Regarding the Dockerization comments:
1. I set the UID/GID to 1000/1000.
2. I'm keeping the yq install as is since it's just a jq wrapper, but I'm unsure how best to address the ldap3 library install.  Thinking out loud here: At the time of writing, SATOSA defines three extras—ldap, pyop_mongo, and pyop_redis, each of which pull in the appropriate dependencies.  I could rewrite the questionable pip command as `pip install --no-cache-dir satosa[ldap]==${SATOSA_VERSION}` and have pip install whatever version of ldap3 that version of SATOSA wants.  Or maybe it would be better to _not_ install ldap3 by default?  Let me ask the rest of the Identity Python project what they think.
3. I changed the tar command per your suggestion and will propose your suggested changes to the Identity Python project.
4. I removed the VOLUME directive.  I was using it incorrectly.  The files in question are solely configuration, not state.
5. I fixed the permissions of the entrypoint script.
6. I'm waffling on whether to change the entrypoint script as suggested.  SATOSA does have features in development that would require database initialization (e.g., a session store to support SAML 2.0 Single Logout), so the use case you described would be useful to people deploying those features.

Regarding the images, I will make the change you suggested.

niheconomoum:
I ended up removing the ldap3 Python package from the image.  It isn't installed by the project's old image, and like you said, deployers can always add any of SATOSA's predefined extras if they need them, e.g.,:

```dockerfile
FROM satosa
RUN pip install --no-cache-dir satosa[ldap,pyop_redis]==${SATOSA_VERSION}
```

niheconomoum:
@tianon, I've made all of the changes you suggested except for pinning a version of **yq**, which we use as a simple CLI wrapper for **jq**.

tianon (approved):
LGTM :+1:

---

As an aside, I'd suggest dropping `RUN chmod 0755 /usr/local/bin/docker-entrypoint.sh` because if the file is executable in Git (which Git can track properly), it's unnecessary, and this creates a copy of the file in the new layer (but it's not a blocker).

yosifkit (approved):

yosifkit:
Moving the `COPY` or `chmod` will still cause a duplicate of the file to exist in the end image layers (one in the `COPY` layer and one in a `RUN` layer). I think it is reasonable to assume that the build system will clone the commit with the permissive umask value (`0022`).

niheconomoum:
Thank you for all your help @tianon @yosifkit!

PR #12481, opened 2022-05-23
Add tests for ghost version 5

ajfriesen:
Adding tests for ghost version 5.


ajfriesen:
The pipeline for the ghost image test failed because of the missing test here if I understand this correctly:

https://github.com/docker-library/ghost/pull/306

tianon:
Nice, this looks fine for now -- we can slim things down as a follow-up. :+1:

Let's get this in so we can restart CI over there :eyes:

PR #12498, opened 2022-05-24
Alpine 3.16 support for golang images

techknowlogick:
Result of https://github.com/docker-library/golang/pull/425


tianon:
Oops, looks like we need to update https://github.com/docker-library/golang/blob/0a910ff948b308ac96b1c888814dad719244b3db/generate-stackbrew-library.sh#L12 too :eyes:

tianon:
Updated in https://github.com/docker-library/golang/commit/0f8a676 -- want to update? (or would you rather I take it from here?)

techknowlogick:
@tianon whoops, sorry for missing that. Feel free to take this one, if not I'll get to it in a few hours :)

PR #12516, opened 2022-05-26
eclipse-temurin: add Ubuntu 22.04 support

gdams:
Switches the latest tags to default to Ubuntu 22.04 (jammy)


yosifkit:
> @tianon is there a way to know the shas of the images built ? looking for the ones before this change was merged

The history of the files here should have it: https://github.com/docker-library/repo-info/tree/master/repos/eclipse-temurin/remote

PR #12542, opened 2022-05-30
Enable s390x for cassandra:4.0

yselkowitz:
cassandra:4.0 uses OpenJDK 11 which is supported on s390x.  Earlier versions use OpenJDK 8 which is not.


tianon:
(force pushed an update per https://github.com/docker-library/cassandra/issues/193#issuecomment-1142696307 -> https://github.com/docker-library/cassandra/commit/dee264d35d3efd9aa8368742605d2e48979ec1b1 :+1: :tada:)

PR #12543, opened 2022-05-30
Separate Jammy and Focal images

keeganwitt:
People have reported issues using Eclipse Temurin's upgrade to Jammy (22.04). So I'm now going to publish both Focal and Jammy versions, with the tags that don't specify using the latest LTS (Jammy).

### DIFF
--- a/library/gradle
+++ b/library/gradle
@@ -4,70 +4,103 @@ GitRepo: https://github.com/keeganwitt/docker-gradle.git
 
 # Gradle 7.x
 
-Tags: 7.4.2-jdk8, 7.4-jdk8, 7-jdk8, jdk8
+Tags: 7.4.2-jdk8, 7.4-jdk8, 7-jdk8, jdk8, 7.4.2-jdk8-jammy, 7.4-jdk8-jammy, 7-jdk8-jammy, jdk8-jammy
 Architectures: amd64, arm32v7, arm64v8, ppc64le
-GitCommit: e92e133806cc92601dd9254756a759c5f6b9e4ca
+GitCommit: c250821d163836c1abfe6c31f59c466cf76d2f69
 Directory: jdk8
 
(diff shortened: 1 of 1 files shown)


library/gradle
```diff
@@ -4,70 +4,103 @@ GitRepo: https://github.com/keeganwitt/docker-gradle.git
...
+Architectures: amd64, arm32v7, arm64v8, ppc64le
+GitCommit: 41ba6e026e80f20f515c41fe925a38912e0e6fcf
+Directory: jdk8
+
+Tags: 6.9.2-jdk8-focal, 6.9-jdk8-focal, 6-jdk8-focal
 GitFetch: refs/heads/6
 Architectures: amd64, arm32v7, arm64v8, ppc64le
-GitCommit: 52b6facc824989b809f42b71ea158b54e0402587
+GitCommit: 41ba6e026e80f20f515c41fe925a38912e0e6fcf
 Directory: jdk8
```
yosifkit, line 70: These two have the same `GitFetch` , `GitCommit` and `Directory` and so are essentially the same image, yet it will be tagged as both jammy and focal. I guess the one just needs an updated `Directory`?

From the diff output that normalizes the library file:
```diff
-Tags: 6.9.2-jdk8, 6.9-jdk8, 6-jdk8
+Tags: 6.9.2-jdk8, 6.9-jdk8, 6-jdk8, 6.9.2-jdk8-jammy, 6.9-jdk8-jammy, 6-jdk8-jammy, 6.9.2-jdk8-focal, 6.9-jdk8-focal, 6-jdk8-focal
 Architectures: amd64, arm32v7, arm64v8, ppc64le
 GitFetch: refs/heads/6
-GitCommit: 52b6facc824989b809f42b71ea158b54e0402587
+GitCommit: 41ba6e026e80f20f515c41fe925a38912e0e6fcf
 Directory: jdk8
```

keeganwitt: Sorry, yes, the directory needs fixed.

PR #12554, opened 2022-05-31
[oraclelinux] Updating 7 and 8 for ELSA-2022-4799 and ELSA-2022-4803

amsteinb-orcl:
This update incorporates fixes for CVE-2022-24903.

https://linux.oracle.com/errata/ELSA-2022-4803.html
https://linux.oracle.com/errata/ELSA-2022-4799.html


library/oraclelinux
```diff
@@ -17,8 +17,8 @@ Directory: 8
 Tags: 8.5
 Architectures: amd64, arm64v8
 Directory: 8
-amd64-GitCommit: b2631b4c3944ae70c9463ae62aaa75a1a4877a16
-arm64v8-GitCommit: ae688d9eef59795cd485bd9d9f43ce9015c3c461
+amd64-GitCommit: b784a7bd222393fbf1d8cf694a318029d41b5a83
+arm64v8-GitCommit: 1d328340b30b22137f59e50ae0c9d3c369d9de9f
```
tianon, line 21: I think you just wanted to remove this whole "temporary 8.5" block instead, right? :sweat_smile:

amsteinb-orcl: Probably, but need to get confirmation from someone.

amsteinb-orcl: I thought it could wait until next update, but I think you were waiting on this before merging.  I have removed it now.  Please proceed.

tianon: Yeah sorry, if we'd merged it as-is it would've clobberd the `8.5` tag on Docker Hub with new `8.6` content (which is why this block was added in the first place :sweat_smile:).

PR #12567, opened 2022-06-02
composer 2.3.6

alcohol:


alcohol:
That looks unexpected.. let me look into that. Could be that the "test" we wrote a long time ago is no longer applicable.

alcohol:
Ah yes, https://github.com/composer/composer/pull/10817. We'll skip this release.

PR #12568, opened 2022-06-02
Deprecate Corretto Alpine 3.12 images

caojoshua:
Alpine 3.12 is EOL. Deprecating Corretto Alpine 3.12 images.


yosifkit:
Removals only, ignoring random failure.

PR #12569, opened 2022-06-03
library/perl: Update for cpanm update and HTTPS install by default

zakame:
- https://github.com/Perl/docker-perl/pull/114
- https://github.com/Perl/docker-perl/pull/115


zakame:
Thanks again @tianon ! :100:

PR #12570, opened 2022-06-03
Ubuntu update June 2022

woky:


mwhudson:
> +etc/adduser.conf.dpkg-save

How did that happen?

woky:
> > +etc/adduser.conf.dpkg-save
> 
> How did that happen?

Nice find. :-) I think the cause is this change https://git.launchpad.net/ubuntu/+source/adduser/commit/?id=afb57f1d07
```
        - Enable private home directories by default (LP: #48734)
          + Set DIR_MODE=0750 in the default adduser.conf
          + Change the description and default value to select private home
            directories by default in debconf template
          + Change the DIR_MODE when private home directories is configured via
            debconf from 0751 to 0750 to ensure files are truly private
```

By following logic from scripts below it seems that's cause:
- https://git.launchpad.net/ubuntu/+source/adduser/tree/debian/config?id=afb57f1d07
- https://git.launchpad.net/ubuntu/+source/adduser/tree/debian/postinst?id=afb57f1d07

woky:
@tianon I've reported the bug here: https://bugs.launchpad.net/ubuntu/+source/adduser/+bug/1977710. I think it's not serious issue and we can proceed. @mwhudson do you agree?

mwhudson:
Oh yeah seems benign enough to me. Shouldn't release the distro like that, an update of a devel series docker image, fine :)

PR #12583, opened 2022-06-04
Add JDK 11 Focal tags

keeganwitt:
I missed adding this image variant in the last PR.


keeganwitt:
> ```diff
> diff --git a/gradle_6-jdk11-focal/Dockerfile b/gradle_jdk11-focal/Dockerfile
> similarity index 100%
> copy from gradle_6-jdk11-focal/Dockerfile
> copy to gradle_jdk11-focal/Dockerfile
> ```
> 
> This diff looked incorrect, and after a bit of digging, it seems that the `6` branch got some inadvertent bumps to `7.x` in the `focal` additions.

Good catch. I missed that the workflow on the other branch had failed. I fixed this now.

PR #12590, opened 2022-06-07
Update rust:alpine to 3.16

cd-work:
This re-generates the Rust docker tags based on the latest docker-rust
master, using the `x.py` script.


sfackler:
:+1:

PR #12477, opened 2022-05-23
alpine: add 3.16.0 as latest, drop EOL 3.12

ncopa:

PR #12600, opened 2022-06-08
erlang/otp 24.3.4.1

getong:


tianon:
CI failure is https://github.com/docker-library/official-images/pull/12477#issuecomment-1134957196 :see_no_evil:

(Alpine 3.12 is EOL)

yosifkit:
I'll merge to get the new release built since the images with outdated an alpine version are not being changed.

PR #12614, opened 2022-06-10
erlang/otp 25.0.1

getong:


tianon:
CI failure is still https://github.com/docker-library/official-images/pull/12477#issuecomment-1134957196 (should probably be dealt with at some point either by updating the Alpine version or removing the affected Erlang version :see_no_evil:).

PR #12622, opened 2022-06-12
update erlang/otp alpine version

getong:


tianon:
Thank you for updating!! :heart: :partying_face:

PR #12635, opened 2022-06-14
chore: Included githubactions in the dependabot config

naveensrinivasan:
This should help with keeping the GitHub actions updated on new releases. This will also help with keeping it secure.

Dependabot helps in keeping the supply chain secure https://docs.github.com/en/code-security/dependabot

GitHub actions up to date https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

https://github.com/ossf/scorecard/blob/main/docs/checks.md#dependency-update-tool
Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


tianon:
Is there a way to tell dependabot not to make branches on the origin repository?

tianon:
IMO, @dependabot would be much easier to trust if it were using forks in the same way as any other contributor :see_no_evil:

(both as a user of projects and as a maintainer of projects -- unexpected branches on the origin are always a red flag for me in both contexts)

tianon:
If there were a way to make this informational (similar to @dependabot's security recommendation functionality), I'd be more +1, but as-is, I'm unfortunately -1 on this.  If @dependabot modifies their methodology to use a proper fork like a normal contributor would or to allow this to be merely informational instead of directly actioning a PR from a branch on the origin, I'd love to reconsider. :+1:

That said, thanks for the contribution! :heart:

PR #12639, opened 2022-06-14
kong: add 2.5.2, add back arm64v8

gszr:
- Add Kong 2.5.2
- Add back arm64v8 to all of 2.5, 2.6, 2.7, and 2.8 series
- Change maintainers to Team Gateway aliases (@team-gateway-bot results
  in a notification being sent to the team email alias)

### DIFF
--- a/library/kong
+++ b/library/kong
@@ -1,86 +1,50 @@
-Maintainers: Kong Docker Maintainers <support@konghq.com> (@thekonginc)
+Maintainers: Kong Docker Maintainers <team-gateway@konghq.com> (@team-gateway-bot)
 GitRepo: https://github.com/Kong/docker-kong.git
 
 Tags: 2.8.1-alpine, 2.8.1, 2.8, alpine, latest
-GitCommit: fffcd0a942daba110ccb89878570b50fcff64d97
+GitCommit: 8a361ad78d4f84b67e893588fd3a6108bea67fe7
 GitFetch: refs/tags/2.8.1
 Directory: alpine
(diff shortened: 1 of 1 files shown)


gszr:
@yosifkit Apologies for pinging you over here, but I am curious what your take is, as a maintainer, on the usage of a `maintainer` label in Dockerfiles. I understand the `MAINTAINER` instruction is deprecated, so we should avoid that one. There's also the standardized `org.opencontainers.image.authors` label. There's also the `Maintainers` global key employed over here.

tianon:
For the "official" take, see https://github.com/docker-library/official-images/issues/3540#issuecomment-530925319, especially:

My personal take is that Docker image labels are a bit annoying as they're auto-inherited by any image that's `FROM` your image.  Imagine if I put `org.opencontainers.image.authors=tianon@debian.org` on the `debian` base image -- then every image that's `FROM debian:xxx` that doesn't explicitly set that label would inherit that and it's less useful to end-users because they can't trust that I actually maintain the image they're using. :see_no_evil:

Hopefully when the engine/tooling implements proper OCI annotations, they don't end up suffering from the same inheritance behavior. :sweat_smile:

yosifkit:
As for the `Maintainers` key in the `library/*` files, that is just data for @tianon and I to find who to ping on a GitHub issue (or via email if security embargoed) if there is an issue with the image or related Docker Hub docs; so it isn't part of any image metadata.

gszr:
Thank you both for the info; this is helpful! 💯

PR #12642, opened 2022-06-15
maven: 3.8.6

carlossg:


tianon:
Doh, looks like this is getting hit by the second commit in https://github.com/docker-library/official-images/pull/12555 :disappointed:
(`ibmjava:8-sdk-alpine` is no longer supported :see_no_evil:)

carlossg:
yep, forgot to pull *before* running naughty

PR #12649, opened 2022-06-16
Updating Dockerfile for older Couchbase Server images

ceejatec:
These have a workaround which should allow them to build on Docker's infrastructure.


ceejatec:
Thanks!

PR #12666, opened 2022-06-21
Update clojure tools-deps variants to version 1.11.1.1139

cap10morgan:


yosifkit:
This single PR for both changes is fine; it all shows up in the generated diff :+1:.

PR #12690, opened 2022-06-23
22.0.0.6 release

crpotter:


yosifkit:
It looks like your two repos migrated to `main` branches, so the library files will need a `GitFetch: refs/heads/main`

crpotter:
Thanks for the heads up. Line should be added to both files!

PR #12694, opened 2022-06-24
Update photon image

gerrit-photon:
Update photon image

### DIFF
--- a/library/photon
+++ b/library/photon
@@ -4,24 +4,24 @@ Maintainers: Fabio Rapposelli <fabio@vmware.com> (@frapposelli),
 GitRepo: https://github.com/vmware/photon-docker-image.git
 Directory: docker
 
-Tags: 4.0, 4.0-20220617, latest
+Tags: 4.0, 4.0-20220624, latest
 Architectures: amd64, arm64v8
-GitFetch: refs/heads/x86_64/4.0-20220617
-GitCommit: 26dc28cc9e3750ac7a782c47698ecef6e373969d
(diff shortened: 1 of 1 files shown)


library/photon
```diff
@@ -4,24 +4,24 @@ Maintainers: Fabio Rapposelli <fabio@vmware.com> (@frapposelli),
...
-GitFetch: refs/heads/x86_64/3.0-20220617
-GitCommit: dd8799ae99fc065c02ac8f2fb397f0d7d85b12c9
-arm64v8-GitFetch: refs/heads/aarch64/3.0-20220617
-arm64v8-GitCommit: bb3d18fc3de752b730cc0e8cf0f86ca329809d9e
+GitFetch: refs/heads/x86_64/3.0-20220624
+GitCommit: 41f22b9aadc0fc9be4ec32b60b961505b80e7005
+arm64v8-GitFetch: refs/heads/aarch64/3.0-20220624
+arm64v8-GitCommit: ea8eea96c94736f8c33c7783ebcc4f5b02cd866e
 
 Tags: 1.0, 1.0-20220429
```
tianon, line 21: This tag should probably be removed now, right? (https://blogs.vmware.com/vsphere/2022/01/photon-1-x-end-of-support-announcement.html)

YustasSwamp: Hi @tianon thanks for bringing our attention to this. We will remove 1.0 tag with the followup update.

tianon: Nice, looking forward to it! :smile:

(I'll merge this for now :+1:)

PR #12704, opened 2022-06-27
Update JRuby 9.2.21.0 and 9.3.6.0

headius:


headius:
I see that the jdk17 images are being rejected as base and will arrange to switch our base to one of the supported images.

headius:
Looks good!

PR #12730, opened 2022-07-01
Update photon image

gerrit-photon:
Update photon image


tianon:
1.0 is going away, right? :eyes:

PR #12612, opened 2022-06-10
library/perl: Add out-of-the-box SSL support and App::cpm

zakame:
- https://github.com/Perl/docker-perl/pull/118
- https://github.com/Perl/docker-perl/pull/119

PR #12731, opened 2022-07-01
Temporarily revert perl back to before cpm was added

tianon:
The "cpm" additions removed "cpanm" (accidentally) -- a fix is in progress but maintainers are already into the weekend. 🙈

This is a partial revert of https://github.com/docker-library/official-images/pull/12612 while https://github.com/Perl/docker-perl/pull/121 is pending :heart:

### DIFF
--- a/library/perl
+++ b/library/perl
@@ -1,7 +1,7 @@
 Maintainers: Peter Martini <PeterCMartini@GMail.com> (@PeterMartini),
              Zak B. Elep <zakame@cpan.org> (@zakame)
 GitRepo: https://github.com/perl/docker-perl.git
-GitCommit: 87a4b9df7a72c9874d000274ab806d00e9cfbf1a
+GitCommit: 37fe216e72bed2a80cf699478fa7b7a7fed656c2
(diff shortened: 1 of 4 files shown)


yosifkit (approved):

tianon:
(added a basic test so we can prevent something quite so obviously broken happening again :eyes:)

tianon:
For what it's worth: I chose `Mojolicious` as the test here because one of the explicit project goals is to keep it zero-to-minimal hard dependencies, so it should install quickly and reliably (and hopefully will continue to support _all_ officially supported Perl versions :eyes:).

zakame:
Thanks @tianon! https://github.com/Perl/docker-perl/commit/f70e8ace49994efef8e90cbf730554a3e3201da7 should now have your fix 🙏

tianon:
Aha, I did not expect to see you until Monday!!  Thank you :heart:

I'll replace this one with https://github.com/docker-library/official-images/pull/12732 which is a proper Update instead :+1:

PR #12733, opened 2022-07-02
[oraclelinux]: Release Oracle Linux 9

amsteinb-orcl:
https://docs.oracle.com/en/operating-systems/oracle-linux/9/index.html

### DIFF
--- a/library/oraclelinux
+++ b/library/oraclelinux
@@ -4,10 +4,18 @@ GitCommit: 5c8a1c296acd6e90487cd261d16cf85fd6bcb73f
 GitFetch: refs/heads/master
 # https://github.com/oracle/container-images/tree/dist-amd64
 amd64-GitFetch: refs/heads/dist-amd64
-amd64-GitCommit: a6257f962838edf2bf4814d2763e66e31d5191e2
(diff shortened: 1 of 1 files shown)


amsteinb-orcl:
Can someone explain what the Diff Comment failure means and how it was caused?

Djelibeybi:
There is no diff because there was no previous release of `9` or `9-slim`.

@tianon @yosifkit could we get a manual review and merge, please?

tianon:
Yeah, unfortunately there's a limit to comment length, and new major releases typically go over it :sweat_smile:

Diff: https://gist.github.com/docker-library-bot/4314a147736889616aa13a8b8665017b

tianon:
I'm really impressed that even though `systemd` appears to have forced itself into the minimum set (which is totally understandable), the overall image sizes are still pretty respectable!

- `oraclelinux:8`: ~226MB
- `oraclelinux:9`: ~217MB
- `oraclelinux:8-slim`: ~101MB
- `oraclelinux:9-slim`: ~126MB

(I believe it's only `slim` that _didn't_ have `systemd` previously but now does, which makes sense from the image size bump :sweat_smile:)

PR #12743, opened 2022-07-05
Update mysql

tianon:
Changes:

- https://github.com/docker-library/mysql/commit/f8d76ae: Merge pull request https://github.com/docker-library/mysql/pull/867 from infosiftr/oracle
- https://github.com/docker-library/mysql/commit/6cb7337: Merge pull request https://github.com/docker-library/mysql/pull/866 from infosiftr/bullseye

### DIFF
--- a/library/mysql
+++ b/library/mysql
@@ -1,28 +1,28 @@
-# this file is generated via https://github.com/docker-library/mysql/blob/c77f72025a5523a86102fe45524ca6f9b66306c4/generate-stackbrew-library.sh
+# this file is generated via https://github.com/docker-library/mysql/blob/9ed73fdd2f2fd77591d0020e13c37602a0600eb8/generate-stackbrew-library.sh
 
 Maintainers: Tianon Gravi <admwiggin@gmail.com> (@tianon),
              Joseph Ferguson <yosifkit@gmail.com> (@yosifkit)
 GitRepo: https://github.com/docker-library/mysql.git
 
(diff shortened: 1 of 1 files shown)


bryan-lee-git:
Hello there! Just wanted to let you know of an issue I experienced related to this PR. This PR changes the default image that gets used from debian to oracle. This will break any dockerfiles using the `mysql:5.7` base image followed by steps like `apt-get` and result in errors like this `/bin/sh: apt-get: command not found`. Had to specify `mysql:5.7-debian` instead of just `mysql:5.7` to fix the issue.

bryan-lee-git:
After looking at the other issues regarding this update, I have switched over to just using the 5.7 image with oracle and changing apt-get to yum. However, I will say that making a sweeping change like this to an image that has used debian as the default for a long period of time seems like a questionable decision given how wisely used this image is.

khassad:
I will just add to this that this unwise and unannounced/undocumented change had some major consequences for some of us around.

The switch to oracle from debian introduces some breaking change which in our case led to badly corrupted database and hours of downtime. What was due to be a minor mysql version upgrade in the 8.0 branch turned into a business-critical nightmare.

The lesson here is clear : when your upstream supplier is unreliable, switch away from it !
Time for us to move away from mysql and oracle business.

I don't think oracle cares for its user base nor monitor the feedback but I write this here for others, in case our experience can save them some sweat and blood.

PR #12750, opened 2022-07-06
Update to Flink 1.15.1

alpinegizmo:


tianon:
```diff
-    GPG_KEY=CBE82BEFD827B08AFA843977EDBF922A7BC84897 \
+    GPG_KEY=3EE012FEE982F098 \
```

This should still be a full fingerprint (so that `gpg` can verify it). :sweat_smile:

alpinegizmo:
@tianon Updated to use the full fingerprint as requested.

PR #12756, opened 2022-07-07
Rocky Linux updates for 8.6

NeilHanlon:
* Introduce new minimal image

### DIFF
--- a/library/rockylinux
+++ b/library/rockylinux
@@ -2,13 +2,26 @@ Maintainers: The Rocky Linux Foundation <infrastructure@rockylinux.org> (@rocky-
              Neil Hanlon <neil@resf.org> (@neilhanlon)
 GitRepo: https://github.com/rocky-linux/sig-cloud-instance-images.git
 
-Tags: latest, 8, 8.5.20220308
+Tags: latest, 8, 8.6, 8.6.20227707
+GitFetch: refs/heads/Rocky-8.6.20220707-Base-x86_64
+GitCommit: 09dbe262451ff0e93fe50df95f6811b234d3d509
+arm64v8-GitFetch: refs/heads/Rocky-8.6.20220707-Base-aarch64
+arm64v8-GitCommit: 411d541689cc277f778ce5a0aa260d5a82dd83f3
+Architectures: amd64, arm64v8
+
+Tags: 8-minimal, 8.6-minimal, 8.6.20220707-minimal
+GitFetch: refs/heads/Rocky-8.6.20220707-Minimal-x86_64
+GitCommit: ef6b5ec9cbad530e1ba673e44dcef9d4c5d5df6a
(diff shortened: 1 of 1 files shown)


NeilHanlon:
There was some conversation in my previous [pr](https://github.com/docker-library/official-images/pull/12435) about the usage of 'latest', however we've had the latest tag in there since the beginning and it hasn't been an issue. Please advise!

Best,
Neil

tianon:
Oh, maybe I can clarify what was meant there -- having a `latest` tag is totally fine, but the point is that users are not expected to `docker run foo:latest` (it only exists for enabling `docker run foo`), so `docker run foo:latest-something` is unexpected and likely misnamed; usually `foo:something` is more natural/expected (so in your case, `rockylinux:minimal` is totally sane, but up to you whether you want to include it). :+1:

The diff comment CI is failing here because it's trying to post a comment with a diff of the old tarball to the new tarball, and the diff is larger than it can successfully comment (:weary:).  Usually we only see that on major updates (like Rocky Linux 8 -> 9), and it's weird to see such a dramatic difference in the tarball file listing in a minor release.  However, it's not necessarily _entirely_ unexpected -- in cases like that, we usually just (manually) post the diff to a gist instead and link it here:

Full Diff: https://gist.github.com/docker-library-bot/6149d59f722c3f7ae57ae3ed6f2b7e47

(We use the diff to review for unexpected changes -- in the case of base images, that's typically looking for unexpected packages like `grub` accidentally sneaking in, which is something we catch frequently on other base images :sweat_smile: :heart:)

tianon:
Oh, the diff is primarily deletions because you're introducing a new minimal image -- that's entirely expected all around. :smile:

NeilHanlon:
> Oh, maybe I can clarify what was meant there -- having a `latest` tag is totally fine, but the point is that users are not expected to `docker run foo:latest` (it only exists for enabling `docker run foo`), so `docker run foo:latest-something` is unexpected and likely misnamed; usually `foo:something` is more natural/expected (so in your case, `rockylinux:minimal` is totally sane, but up to you whether you want to include it). +1


Ah, I see. I'll leave it out, as I doubt people will use that over the 8-minimal (and future 9-minimal) tags, anyways.

Thank you for the explanation.

Yes, I suspected the diff failures were due to the addition of the minimal image and it not finding an appropriate image to diff against -- along with a naming convention change to support multiple variants.

Best,
Neil

tianon:
Ah, and a case of the diff-generator getting confused and re-diffing 8.5 and 8.6 -- for avoiding that in the future, I'd suggest swapping the order of the tags listed in `Tags:` (such that least-specific tags like `latest` come last in the list and most-specific tags come first).

Also, I'd suggest removing the 8.5 tags unless you're going to do further updates to 8.5 -- here's our standard boilerplate about that:

I'll merge this for now (because the 8.5 conversation is really orthogonal), but if you determine that 8.5 is not going to receive further updates, I'd love to see a follow-up that removes it. :smile:

PR #12767, opened 2022-07-11
Force usage of openjdk since the automatic switch to eclipse-temurin makes XWiki image unusable for some users

vmassol:
See https://jira.xwiki.org/browse/XDOCKER-241


tianon:
Doh -- looks like https://github.com/docker-library/tomcat/pull/265 made `xwiki` not show up in the list for my ping on https://github.com/docker-library/docs/pull/2162.

The TLDR there is that this is OK for _now_, but it will not work for long (because those `tomcat:*-openjdk` variants are going to go away completely very soon). :see_no_evil:

yosifkit:
`tomcat:*-temurin-focal` variants might be an alternative once the debian-based images are gone and still fix the issue that users were hitting.

vmassol:
Thanks @yosifkit , very useful information, as always! I'll discuss this with the team and users.

PR #12768, opened 2022-07-11
22.0.0.7 release

crpotter:


yosifkit:
>  - websphere-liberty:22.0.0.7-full-java8-ibmjava (FROM websphere-liberty:22.0.0.7-kernel) -- completely unsupported base!

It looks like there is a missing tag in the new version? Or perhaps they need more specific `FROM` lines?

> sha1sum: WARNING: 1 computed checksum did NOT match

`open-liberty:beta`, `open-liberty:beta-java11`, and `open-liberty:beta-java17` are having a checksum mismatch. While they aren't being changed in this PR, they'll fail to build when their base image is updated.

PR #12779, opened 2022-07-12
Rocky Linux updates for 9.0

NeilHanlon:
* Also drop legacy 8.5.x image

### DIFF
--- a/library/rockylinux
+++ b/library/rockylinux
@@ -2,30 +2,30 @@ Maintainers: The Rocky Linux Foundation <infrastructure@rockylinux.org> (@rocky-
              Neil Hanlon <neil@resf.org> (@neilhanlon)
 GitRepo: https://github.com/rocky-linux/sig-cloud-instance-images.git
 
-Tags: latest, 8, 8.6, 8.6.20227707
+Tags: 9.0.20220712, 9, 9.0
+GitFetch: refs/heads/Rocky-9.0.20220712-Base-x86_64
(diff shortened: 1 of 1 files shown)


NeilHanlon:
> Diff: https://gist.github.com/docker-library-bot/5086f6a11b91b9e52899825ae0a6ae9b
> 
> (Was this a draft just to test the CI, or because it's still a WIP and you don't want us to review it yet? pray)

Was just about to message you on IRC :) 

Still a WIP. Rocky 9 has not been officially released yet but I'm preparing the container images... plus I messed up the last PRs a few times so I figured I'd cross my T's

tianon:
Ah, totally fair!  I'll keep going through the diff then and let you know if I come across anything unexpected. :+1:

tianon:
Looks pretty vanilla to me! :+1: (nothing like `grub` sneaking in or anything :joy:)

I did notice you removed `rockylinux:latest` -- if that was intentional, we might want to consider deleting that specific alias from Hub as well (since normally it would stick around and just no longer be updated), but if that was an oversight you know what to do. :smile:

NeilHanlon:
Thanks! Yeah it was intentional. What's the process for removing that tag from the repo? Especially with two majors around, I think it makes little sense to keep the latest tag explicitly added, as per the other PR.

tianon:
Just remind us once we merge the final version of this PR and we'll go get rid of it manually. :innocent:

PR #12796, opened 2022-07-15
Eggdrop v1.9.3rc1

vanosg:


tianon:
```diff
+RUN apk update
```

With `--no-cache`, `apk update` is automatically implied (including cleanup after the command completes). :sweat_smile:

(Thus you also don't need to combine `--update` and `--no-cache` -- the latter alone is sufficient)

PR #12820, opened 2022-07-19
NeuroDebian: no more EOLed stretch, add arm64v8 architecture

yarikoptic:
ping @tianon - have I done specification of architectures correctly?  I have removed i386 from ubuntus since they no longer have it but kept it for debians.

We do have https://neuro.debian.net/debian/dists/bookworm/Contents-armel.gz populated  so hopefully that arm64 gets built. let's see

### DIFF
--- a/library/neurodebian
+++ b/library/neurodebian
@@ -1,6 +1,6 @@
 Maintainers: Yaroslav Halchenko <debian@onerussian.com> (@yarikoptic)
 GitRepo: https://github.com/neurodebian/dockerfiles.git
-GitCommit: c673678c2010811f745b3bb2638dd060b0534939
+GitCommit: c9aa5f02de44d51bfcd81af986829550b227daeb
 
 Tags: xenial, nd16.04
 Architectures: amd64, arm64v8
--- a/library/neurodebian
+++ b/library/neurodebian
@@ -1,57 +1,75 @@
 Maintainers: Yaroslav Halchenko <debian@onerussian.com> (@yarikoptic)
 GitRepo: https://github.com/neurodebian/dockerfiles.git
-GitCommit: 97b3bc1860afafc294abbec1f95bc239ddeddd88
+GitCommit: c673678c2010811f745b3bb2638dd060b0534939
 
 Tags: xenial, nd16.04
+Architectures: amd64, arm64v8
 Directory: dockerfiles/xenial
 
 Tags: xenial-non-free, nd16.04-non-free
+Architectures: amd64, arm64v8
 Directory: dockerfiles/xenial-non-free
 
 Tags: bionic, nd18.04
+Architectures: amd64, arm64v8
 Directory: dockerfiles/bionic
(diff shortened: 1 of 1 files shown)


yosifkit:
The `sed` in `debian:sid` is failing due to this change: https://github.com/debuerreotype/debuerreotype/pull/128

yarikoptic:
Thank you @yosifkit !

tianon:
The `Architectures:` lines are definitely in the correct format, and we've got a CI test that verifies that they're a strict subset of the base image (which is passing), so I think you're OK -- the only question will be whether your repos have each of these in all these places, but we can find that out the fun way by getting this in and seeing what fails. :smile:

yarikoptic:
> , but we can find that out the fun way by getting this in and seeing what fails.

my bet would be -- they should have them since I just did that... could not figure out how to copy from another architecture so they would have only `nuitka` in them, but shouldn't fail ;) ..... but silly me:

image

so it fails for some older ubuntus but not for newer ones.  e.g. bionic is failing, I have https://neuro.debian.net/debian/dists/bionic/main/binary-arm64/ and arm64 present for bionic for a sample package on https://packages.ubuntu.com/search?keywords=dcm2niix ... looking in the logs shows the reason

```
Reading state information...
[91mE: Unable to locate package neurodebian-freeze
[0mRemoving intermediate container de05a3498fdf
The command '/bin/sh -c set -x 	&& apt-get update 	&& apt-get install -y --no-install-recommends neurodebian-freeze eatmydata 	&& ln -s /usr/bin/eatmydata /usr/local/bin/apt-get 	&& rm -rf /var/lib/apt/lists/*' returned a non-zero code: 100
```

so we didn't have it yet in those official ubuntus, but I have not copied from one arch to another our packages (couldn't figure out)... the easiest way would be to build new release of neurodebian package and reupload and just to ask to rebuild those images.  How should I do that (ask later to rebuild them)?

yosifkit:
Yeah, let us know and we can trigger them once you are ready.

yarikoptic:
I think I am done here:

```
neurodebian@falkor:~/www/debian/dists/bionic/main/binary-arm64$ grep neurodebian-freeze Packages
Recommends: ubuntu-keyring, python3, zerofree, moreutils, time, debian-archive-keyring, apt-utils, cowbuilder, neurodebian-freeze
Package: neurodebian-freeze
Filename: pool/main/n/neurodebian/neurodebian-freeze_0.41.1~nd18.04+1_all.deb
```

and ready to be "triggered" ;)

tianon:
Looks like your Give-Back is now Installed :joy:  success :tada:

tianon:
Yep, that's correct -- when `debian` and/or `ubuntu` update, they'll rebuild together.  If you want to be more specific, I'd suggest version pinning (`apt-get install foo=x.y.z`), but then you need to stay more on top of updates to the pins. :innocent:

PR #12849, opened 2022-07-25
Fix satosa build errors due to missing Rust toolchain

niheconomoum:
Builds on the arm32v5, arm32v6, arm32v7, i386, mips64le, ppc64le, and
s390x platforms are failing due to missing pyca/cryptography build
dependencies.  The referenced commit adds them.

This also includes a change that removes an extraneous layer from the
image.


tianon:
Nice, thanks for being proactive! :heart:

PR #12883, opened 2022-08-01
eclipse-temurin: bump to 8u342-b07 and 18.0.2+9

gdams:


yosifkit:
By keeping only some of the old tags (windows), this will run into the same problem as mentioned previously: https://github.com/docker-library/official-images/pull/12366#issuecomment-1118022840

gdams:
@yosifkit updated PTAL

PR #12926, opened 2022-08-08
Update ghost

tianon:
Changes:

- https://github.com/docker-library/ghost/commit/9afb165: Update to 5.8.2, ghost-cli 1.21.1
- https://github.com/docker-library/ghost/commit/a798013: Update to 5.8.0, ghost-cli 1.21.1


tianon:
Hmm, I wonder if we should use this for a breaking change to resolve https://github.com/docker-library/ghost/issues/310? :grimacing:

tianon:
Discussed in https://github.com/docker-library/ghost/issues/310#issuecomment-1208340703, we should go ahead with this as-is. :+1:

PR #12933, opened 2022-08-10
[varnish] Security release for VSV00009

gquintard:
https://varnish-cache.org/security/VSV00009.html#vsv00009


yosifkit:
:confused: This doesn't have any changes?

gquintard:
somebody™ forgot to run the script before committing...

PR #12949, opened 2022-08-12
Rollback Zookeeper 3.5, 3.6, 3.7 and 3.8 to openjdk base image

31z4:
Base image update caused several bug reports (see https://github.com/31z4/zookeeper-docker/issues/139 and https://github.com/31z4/zookeeper-docker/issues/140). I realised that it wasn't a great idea to override existing tags. Here I rollback version specific tags and add separate tags for `temurin` based images. I will also update image docs a bit later.

### DIFF
--- a/library/zookeeper
+++ b/library/zookeeper
@@ -3,20 +3,40 @@ GitRepo: https://github.com/31z4/zookeeper-docker.git
 
 Tags: 3.5.9, 3.5
 Architectures: amd64, arm64v8
+GitCommit: 060f64d28d1780373f90ac152bd2538abcf1924c
+Directory: 3.5.9
+
+Tags: 3.5.9-temurin, 3.5-temurin
+Architectures: amd64, arm64v8
 GitCommit: 5cf119d9c5d61024fdba66f7be707413513a8b0d
 Directory: 3.5.9
 
 Tags: 3.6.3, 3.6
+Architectures: amd64, arm64v8
+GitCommit: 060f64d28d1780373f90ac152bd2538abcf1924c
+Directory: 3.6.3
+
+Tags: 3.6.3-temurin, 3.6-temurin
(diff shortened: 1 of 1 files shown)


yosifkit:
Eventually you'll want to move the plain tags, right? Or will they be dropped when the respective `openjdk` tags are dropped? Perhaps adding a something similar to the `-temurin` for the openjdk images so that users have an escape once the plain tags move. Or maybe they also need some images to be `FROM eclipse-temurin:11-jre-focal` (instead of the `jammy` default)?

We've moved base images on quite a few images and have found that giving them an escape option has been useful for users (even if that image will no longer be updated).

31z4:
Thanks for your feedback @yosifkit and sorry for the delayed response 😊 

Yes, the plan is to move the plain tags during corresponding version upgrades. For example, currently `3.5.9` and `3.5` are based on `openjdk`. Once `3.5.10` is released `3.5.10` and `3.5` becomes `temurin` (see https://github.com/docker-library/official-images/pull/13894).

> Perhaps adding a something similar to the -temurin for the openjdk images so that users have an escape once the plain tags move.

I think it makes sense to have both `-temurin` and `-openjdk` tags to help user migrate. But it looks like new builds based on `openjdk` are not allowed (see https://github.com/docker-library/official-images/actions/runs/3928318640/jobs/6715803837). Is there any workaround for that?

> Or maybe they also need some images to be FROM eclipse-temurin:11-jre-focal (instead of the jammy default)?

I'm not sure about different versions of `temurin` based images. I may add it in the future if there is such demand.

tianon:
I think we could do a one-time build based on `openjdk` to backfill those tags, but we'd expect them to be removed from the supported list after the backfill is successfully built/pushed. :see_no_evil:

PR #12965, opened 2022-08-16
Releasing neo4j 4.3.17 and legacy versions 3.5.35, 4.1.12 and 4.2.19

jennyowen:
We had to release bug fixes to some old, unsupported neo4j versions. So there are some images here we need publishing, but don't really mind about getting security updates on them. I don't know if you have a procedure for that? Once  those images have been accepted and published in dockerhub, should I submit another PR to delete them from this library file?

### DIFF
--- a/library/neo4j
+++ b/library/neo4j
@@ -12,12 +12,12 @@ GitRepo: https://github.com/neo4j/docker-neo4j-publish.git
 
 Tags: 4.4.10, 4.4.10-community, 4.4, 4.4-community, community, latest
 Architectures: amd64, arm64v8
-GitCommit: 985060aa38c5b873b18cd1b2c522c69ddb25d91a
+GitCommit: 598a291bc442f218be772f377d029689dba39ab8
 Directory: 4.4.10/community
 
 Tags: 4.4.10-enterprise, 4.4-enterprise, enterprise
 Architectures: amd64, arm64v8
-GitCommit: 985060aa38c5b873b18cd1b2c522c69ddb25d91a
+GitCommit: 598a291bc442f218be772f377d029689dba39ab8
 Directory: 4.4.10/enterprise
 
 Tags: 4.4.9, 4.4.9-community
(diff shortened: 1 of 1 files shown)


jennyowen:
Oh no! I have a change fail because I needed to revert to using an older patch of Java 11, to fix an OOM bug.

I know that `eclipse-temurin:11.0.15_10-jdk-jammy` isn't a supported official image, but it's used in a COPY not FROM, so I thought it might be ok? Although, re-reading the documentation I see the line 
> No official images can be derived from, **or depend on**, non-official images 

Which fair enough, my mistake.

But this raises some questions though that I didn't see answered in the documentation.
* What would have happened if I'd originally used `eclipse-temurin:11.0.15_10-jdk-jammy` as the base image (or `COPY --from`), and later it goes out of support? Would that automatically push my image out of support?
* My only other option to install the previous java release patch is to install via apt, but that would depend on openjdk keeping an old version in their repository... which they could change at any time and stop my image from building. If my image stops being buildable after publication what happens?

yosifkit:
>Once those images have been accepted and published in dockerhub, should I submit another PR to delete them from this library file?

Yeah, if you just want to do a few one-off bugfix releases for older versions, then removing them once they are built and pushed to Docker Hub is the way to go.

> What would have happened if I'd originally used `eclipse-temurin:11.0.15_10-jdk-jammy` as the base image (or `COPY --from`), and later it goes out of support? Would that automatically push my image out of support?

If the image(s) that a particular `neo4j` image is "from" are removed from support in an upstream image (like `eclipse-temurin`), then that `neo4j` image won't be automatically removed from official-images. It won't get build triggers via a parent change (since the parent is no longer updated) and of course wouldn't get any security fixes from that parent. It will cause a failing test on every PR for the `neo4j` library file and we'd strongly recommend that the particular image tags be removed or updated.

> If my image stops being buildable after publication what happens?

It just stops being updated and republished. We try to let maintainers know when their images are failing to build and recommend that they be fixed as soon as possible. If it is failing on purpose, it can cause our build servers to churn over the failing build periodically. If the desire is to no longer update it, then it should just be dropped from the library file.

jennyowen:
thanks @yosifkit for the detailed reply, that is useful information!

PR #12980, opened 2022-08-18
Update rakudo-star to 2022.07

m-dango:


moritz:
I approve :+1:

yosifkit:
Side note: `gpg` invocations would be more correct to have `--batch`; e.g. `gpg --batch --import ${tmpdir}/key.asc` (https://manpages.debian.org/bullseye/gpg/gpg.1.en.html#batch)

PR #12988, opened 2022-08-19
Release OrientDB 3.2.9

tglman:


library/orientdb
```diff
@@ -1,15 +1,15 @@
 
 Maintainers: Luigi Dell'Aquila <l.dellaquila@orientdb.com> (@luigidellaquila)
 
-Tags: 3.2.8, 3.2, latest
+Tags: 3.2.8, 3.9, latest
```
tianon, line 4: I don't think you meant to change this line in this specific way :sweat_smile:

Maybe this?

```suggestion
Tags: 3.2.9, 3.2, latest
```

tglman:
Hi, 

yes this is the correct version change, thank you for the fix.

Regards

PR #12989, opened 2022-08-19
chore: influxdb 2.4

jeffreyssmith2nd:


library/influxdb
```diff
@@ -61,10 +54,18 @@ Tags: 2.2-alpine, 2.2.0-alpine
...
+Architectures: amd64, arm64v8
+Directory: influxdb/2.3/alpine
+
+Tags: 2.4, 2.4.0, latest
+Architectures: amd64, arm64v8
+Directory: influxdb/2.4
+
+Tags: 2.4-alpine, 2.4.0-alpine, alpine
 Architectures: amd64, arm64v8
 Directory: influxdb/2.3/alpine
```
yosifkit, line 71: It looks like this one is pointing to the wrong directory.
```suggestion
Directory: influxdb/2.4/alpine
```

yosifkit:
```Dockerfile
# Remove GNUPG Configuration
RUN gpgconf --kill all                                      && \
    rm -rf                                                     \
      "/root/.gnupg"                                           \
      influxdb2.key                                            \
      influxdb2-${INFLUXDB_VERSION}-linux-${ARCH}*             \
      influxdb2-client-${INFLUX_CLI_VERSION}-linux-${ARCH}*
```
This new `RUN` line will not clean up the image correctly. Since it is a separate layer, there is no gpg background processes to kill, and any files deleted will have no effect on the image size since they are already part of previous layers. To effect image size, the files need to be deleted in the layer in which they are downloaded. I think leaving in the `.gnupg` directory would be fairly harmless and small since you are trying to reuse it for downloads across multiple layers.

jeffreyssmith2nd:
The latest commit should update to one that deletes in the right layers.

Issue #12158, opened 2022-03-30
upcoming Debian Stretch end of life


yosifkit:
Debian Stretch was first released on 2017-06-17 (https://wiki.debian.org/DebianReleases). As announced by the Debian LTS team back in [2020](https://www.debian.org/News/2020/20200709), Debian 9, "stretch", LTS will end on **2022-06-30**.

For official-images, this means that `debian:stretch` and `debian:stretch-slim` and descendants will no longer be a valid `FROM` for any official-image after that date. We recommend that maintainers plan to remove any Stretch-based images on or before 2022-06-30. To aid user transition, we recommend that for any versions of the contained software that need to continue longer, that `buster` or `bullseye` based images be provided (see the [Debian Releases page](https://wiki.debian.org/DebianReleases) for their timelines).

Here are the current image repos and maintainers that still have `debian:stretch[-slim]` based images:
- `buildpack-deps`: @tianon @yosifkit
- `chronograf`: @jsternberg @hoorayimhelping @bnpfeife
- `debian`: @tianon @paultag
- `erlang`: @c0b
- `golang`: @tianon @yosifkit @proppy
- `haxe`: @andyli
- `influxdb`: @bnpfeife
- `kapacitor`: @jsternberg @docmerlin @bnpfeife
- `neurodebian`: @yarikoptic
- `node`: @nodejs/docker group: @hmalphettes, @tianon, @ttshivers, @yosifkit, @Starefossen, @LaurentGoderre, @nschonni, @SimenB, @PeterDaveHello
- `postgres`: @tianon @yosifkit

Just a reminder that removing a tag from the library file will remove it from the "Supported" section on the Hub readme, but the tag will still be available to users who want it (and no longer updated).

tianon:
Closing this as it's just an FYI (so it's already served the intended purpose). :+1:

tianon:
FYI, this is officially EOL now: https://lists.debian.org/debian-lts-announce/2022/07/msg00002.html

Please, either update or drop any `stretch`-based image variants!  (otherwise CI on your PR will fail and review may be delayed :eyes:)

PR #13009, opened 2022-08-23
chore(kapacitor): docker images for 1.6.5

docmerlin:
1.6.5 docker release for kapacitor


tianon:
CI failure is https://github.com/docker-library/official-images/issues/12158 :sweat_smile:

(Debian Stretch is EOL and now officially removed)

tianon:
Nice, thanks for updating!

(You should also consider either updating or dropping 1.5 from the supported list, but that's orthogonal to this PR :+1:)

PR #13059, opened 2022-08-27
eclipse-temurin: update to 18.0.2.1_1

zdtsw:
Signed-off-by: Wen Zhou <wenzhou@redhat.com>

### DIFF
--- a/library/eclipse-temurin
+++ b/library/eclipse-temurin
@@ -353,116 +353,116 @@ Constraints: nanoserver-1809, windowsservercore-1809
 
 
 #------------------------------v18 images---------------------------------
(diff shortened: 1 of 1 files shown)


gdams:
@zdtsw as things are standing you're missing the ppc64le binaries that we just published.

Please ensure that the dockerfiles in the containers repo have all architectures before raising the pull request upstream.

zdtsw:
> @zdtsw as things are standing you're missing the ppc64le binaries that we just published.
> 
> Please ensure that the dockerfiles in the containers repo have all architectures before raising the pull request upstream.

thanks for checking up! only saw status table was updated but should have checked binaries as well. I will wait for the final commit of "container"

PR #13065, opened 2022-08-29
Update to Flink 1.15.2

zentol:


zentol:
hmm, not really sure what to make of the naughty error message :/

tianon:
That's unfortunately the downstream effect of https://github.com/docker-library/docs/pull/2162#issuecomment-1172781329 -- `openjdk:8*` and `openjdk:11*` are no longer maintained/supported.

See the top notice on https://hub.docker.com/_/openjdk for a list of alternative images/builds of OpenJDK 8 and 11 that might meet your needs.

zentol:
Got it, unfortunate that none of us noticed your ping :/

We'll get right on that. 👍

zentol:
@tianon We've updated the images to be based on eclipse-temurin. Would it be possible to merge this on Wednesday so we can give our users some notice beforehand?

tianon:
Good to go now?  (or do you want to give it more time?)

zentol:
@tianon Yes, we're good to go 👍

PR #13084, opened 2022-09-01
Update php

tianon:
Changes:

- https://github.com/docker-library/php/commit/219454b: Update 8.2-rc to 8.2.0RC1
- https://github.com/docker-library/php/commit/94e5814: Update 8.0 to 8.0.23
- https://github.com/docker-library/php/commit/f854089: Update 8.1-rc
- https://github.com/docker-library/php/commit/27fd826: Update 8.1 to 8.1.10

### DIFF
--- a/library/php
+++ b/library/php
@@ -4,354 +4,214 @@ Maintainers: Tianon Gravi <admwiggin@gmail.com> (@tianon),
              Joseph Ferguson <yosifkit@gmail.com> (@yosifkit)
 GitRepo: https://github.com/docker-library/php.git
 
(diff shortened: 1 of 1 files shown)


tianon:
The release announcements for 8.0.23 claim it's a security release, but I don't see anything security-related in https://www.php.net/ChangeLog-8.php#PHP_8_0 :see_no_evil: (so my guess is that someone copied the wrong release notes template, but I'm not certain!  It could just be that the security fix isn't described yet to gives folks a chance to update first?)

Issue #13057, opened 2022-08-27
Question about official image `GitRepo`


Yikun:
I am doing some investigation on official image support for Apache Spark.

Just curious that `GitRepo` should be a separate repo (`apache/spark-docker`) or using existing repo (`apache/spark`)?

Such like:

https://github.com/docker-library/official-images/blob/69305f5afe8f7658aea67c6894bfd3678059f5ee/library/solr#L8

https://github.com/docker-library/official-images/blob/6537e5494de4dea6905569048842bab5931c9bad/library/couchdb#L5

https://github.com/docker-library/official-images/blob/d7fcb735af24ccdf8c8ed02f65d12de10b6a626c/library/flink#L4

I am also try some by reference existing and have a try: https://github.com/yikun/spark-docker

But I have a question that: Why not use `apache` main repo directly? I know there might be contains many history release dockerfile and specific content structure, But I believe if we contains history dockerfile (dockerfile/path) on master branch, then we can use main repo (`apache/spark`) directly and use it hash, right?

Or there are some other considerations have to make a separate repo like `apache/spark-docker`?

Yikun:
cc @yosifkit @tianon

tianon:
Unfortunately having the `Dockerfile` in the root of the upstream repository (while useful for local development) doesn't work well as the target for official-images. Since the `Dockerfile` in those cases typically uses `COPY . ...`, it essentially means the entire upstream source code is now part of the Docker build context, and thus becomes part of the reviewed artifacts (see `diff-pr.sh` for how we generate our diff; it is basically taking all docker build contexts from the starting `library/foo` file and diffing them to the new contexts from the PR).

What we suggest instead is creating a separate repository to hold the "packaging", especially since you might have, for example, changes to make to the `Dockerfile` in between upstream releases of the software.

Yikun:
> What we suggest instead is creating a separate repository to hold the "packaging", especially since you might have, for example, changes to make to the Dockerfile in between upstream releases of the software.

@tianon Sounds great! Thanks for your detail explaination!

PR #13089, opened 2022-09-02
Add Apache Spark Docker Official Image

Yikun:
This patch is adding Apache Spark Docker Official Image.

# Checklist for Review

-	[x] associated with or contacted upstream? [VOTE](https://lists.apache.org/thread/ro6olodm1jzdffwjx4oc7ol7oh6kshbl): [passed](https://lists.apache.org/thread/zgj1db8wht6o10prx49j5fbh092z934q) in Apache Spark Community
-	[x] available under [an OSI-approved license](https://opensource.org/licenses)? [Apache License  2.0](https://github.com/apache/spark-docker/blob/master/LICENSE)
-	[x] does it fit into one of the common categories? ("service", "language stack", "base distribution") service
-	[x] is it reasonably popular, or does it solve a particular use case well? https://spark.apache.org/ https://hub.docker.com/r/apache/spark
-	[x] does a [documentation](https://github.com/docker-library/docs/blob/master/README.md) PR exist? (should be reviewed and merged at roughly the same time so that we don't have an empty image page on the Hub for very long) https://github.com/docker-library/docs/pull/2200
-	[x] existing official images have been considered as a base? (ie, if `foobar` needs Node.js, has `FROM node:...` instead of grabbing `node` via other means been considered?) `eclipse-temurin`
-	~[ ] if `FROM scratch`, tarballs only exist in a single commit within the associated history?~
-	[x] passes current tests? any simple new tests that might be appropriate to add? (https://github.com/docker-library/official-images/tree/master/test) Extra [integration test](https://github.com/apache/spark-docker/blob/master/.github/workflows/main.yml#L171) in repo

### DIFF
--- a/library/spark
+++ b/library/spark
@@ -1,22 +1,22 @@
 Maintainers: Apache Spark Developers <dev@spark.apache.org> (@ApacheSpark)
 GitRepo: https://github.com/apache/spark-docker.git
 
-Tags: 3.4.0-scala2.12-java11-python3-ubuntu, 3.4.0-python3, python3, 3.4.0, latest
+Tags: 3.4.1-scala2.12-java11-python3-ubuntu, 3.4.1-python3, python3, 3.4.1, latest
 Architectures: amd64,arm64v8
-GitCommit: 7f9b414de48639d69c64acfd81e6792517b86f61
-Directory: ./3.4.0/scala2.12-java11-python3-ubuntu
+GitCommit: 58d288546e8419d229f14b62b6a653999e0390f1
+Directory: ./3.4.1/scala2.12-java11-python3-ubuntu
 
-Tags: 3.4.0-scala2.12-java11-r-ubuntu, 3.4.0-r, r
+Tags: 3.4.1-scala2.12-java11-r-ubuntu, 3.4.1-r, r
 Architectures: amd64,arm64v8
-GitCommit: 7f9b414de48639d69c64acfd81e6792517b86f61
-Directory: ./3.4.0/scala2.12-java11-r-ubuntu
+GitCommit: 58d288546e8419d229f14b62b6a653999e0390f1
+Directory: ./3.4.1/scala2.12-java11-r-ubuntu
 
-Tags: 3.4.0-scala2.12-java11-ubuntu, 3.4.0-scala, scala
+Tags: 3.4.1-scala2.12-java11-ubuntu, 3.4.1-scala, scala
 Architectures: amd64,arm64v8
-GitCommit: 7f9b414de48639d69c64acfd81e6792517b86f61
-Directory: ./3.4.0/scala2.12-java11-ubuntu
+GitCommit: 58d288546e8419d229f14b62b6a653999e0390f1
+Directory: ./3.4.1/scala2.12-java11-ubuntu
 
-Tags: 3.4.0-scala2.12-java11-python3-r-ubuntu
+Tags: 3.4.1-scala2.12-java11-python3-r-ubuntu
 Architectures: amd64,arm64v8
-GitCommit: 7f9b414de48639d69c64acfd81e6792517b86f61
-Directory: ./3.4.0/scala2.12-java11-python3-r-ubuntu
+GitCommit: 58d288546e8419d229f14b62b6a653999e0390f1
+Directory: ./3.4.1/scala2.12-java11-python3-r-ubuntu
--- a/library/spark
+++ b/library/spark
@@ -1,0 +1,22 @@
+Maintainers: Apache Spark Developers <dev@spark.apache.org> (@ApacheSpark)
+GitRepo: https://github.com/apache/spark-docker.git
+
+Tags: 3.4.0-scala2.12-java11-python3-ubuntu, 3.4.0-python3, python3, 3.4.0, latest
+Architectures: amd64,arm64v8
+GitCommit: 7f9b414de48639d69c64acfd81e6792517b86f61
+Directory: ./3.4.0/scala2.12-java11-python3-ubuntu
+
+Tags: 3.4.0-scala2.12-java11-r-ubuntu, 3.4.0-r, r
+Architectures: amd64,arm64v8
+GitCommit: 7f9b414de48639d69c64acfd81e6792517b86f61
+Directory: ./3.4.0/scala2.12-java11-r-ubuntu
+
+Tags: 3.4.0-scala2.12-java11-ubuntu, 3.4.0-scala, scala
+Architectures: amd64,arm64v8
+GitCommit: 7f9b414de48639d69c64acfd81e6792517b86f61
+Directory: ./3.4.0/scala2.12-java11-ubuntu
+
+Tags: 3.4.0-scala2.12-java11-python3-r-ubuntu
+Architectures: amd64,arm64v8
+GitCommit: 7f9b414de48639d69c64acfd81e6792517b86f61
+Directory: ./3.4.0/scala2.12-java11-python3-r-ubuntu


Yikun:
Finally, will move from `Yikun/spark-docker` to `apache/spark-docker` as suggested in https://github.com/docker-library/official-images/issues/13057 . 

Currently, the dockerfile is maintained by apache spark.

This dockerfile(https://github.com/Yikun/spark-docker/tree/master/3.3.0/scala2.12-java11-focal) is changed from [link](https://github.com/apache/spark/blob/master/resource-managers/kubernetes/docker/src/main/dockerfiles/spark/Dockerfile), We have made some improvements as requested by offciail image, but don't know how many gaps there are. Then next step, we will have further discussions in the spark community to make it happen.

So, could you help with a preliminary review? @tianon

Yikun:
Apache Spark community starts [SPIP](https://spark.apache.org/improvement-proposals.html) discussion: [[DISCUSS] SPIP: Support Docker Official Image for Spark](https://lists.apache.org/thread/l1793y5224n8bqkp3s6ltgkykso4htb3).

Yikun:
A quick update:
- Sep.2022: The [vote](https://lists.apache.org/thread/ro6olodm1jzdffwjx4oc7ol7oh6kshbl) for [Apache Spark DOI support](https://docs.google.com/document/d/1nN-pKuvt-amUcrkTvYAQ-bJBgtsWb9nAkNoVNRM2S2o) has been [passed](https://lists.apache.org/thread/zgj1db8wht6o10prx49j5fbh092z934q) in Apache Spark community.
- Oct.2022: The [apache/spark-docker](https://github.com/apache/spark-docker) repo had been created, this PR is also update to spark official repo

 @tianon @yosifkit So, it's ready for review now.

I know it would be slow for new images approved, would you mind giving a rough start time of reviewing? We (The Apache Spark Community) will spare no effort to help DOI support, many thanks!

also cc @HyukjinKwon @zhengruifeng

Yikun:
Quick update (TL;DR: **Adding more test and script to ensure official image quality**):

1. Add Spark on K8s integration test to ensure the docker image quality: https://github.com/apache/spark-docker/pull/9
2. Use `template` to generate dockerfiles to avoid low level typo: https://github.com/apache/spark-docker/pull/12

I believe we are ready. @tianon @yosifkit

yosifkit:
Hello! :sparkles:

gatorsmile:
[yosifkit](https://github.com/yosifkit) Today, we just published the latest release spark 3.4 https://spark.apache.org/releases/spark-release-3-4-0.html I am wondering when we can have our own official docker image?

yosifkit:
Ok, I finally have some feedback ready🎉😻. Sorry for the delay 🙇 Let me know if you have any questions. And thank you for your patience 🙇🥰💖

1. Would it be useful to save space by sharing layers by having one image from another? 🤔 Something like the `*java11-ubuntu` as the "base" with `r` and `python` variants `FROM` that and the `r-python` being `FROM`, probably, the larger one of those?

      - Rough example Dockerfiles
        ```Dockerfile
        FROM eclipse-temurin:11-jre-focal
        # user stuff, install common deps, etc
        ...
        # download/extract spark (maybe keeping python and R files too? they seem relatively small compared to the rest)
        ```
        ```Dockerfile
        # other images in separate Dockerfiles
        FROM spark:3.3.0-scala2.12-java11-ubuntu
        # get "/opt/spark/{python,R}/" contents if not kept in base
        # install python or R (and things like R_HOME)
        ```

---

```
rm /bin/sh && ln -sv /bin/bash /bin/sh
```

2. Does spark expect `/bin/sh` to be bash (in posix mode) and not just a generic posix compliant shell? Is there a bug with using `dash`? Minimally, this should do a `dpkg --divert` to ensure that users in dependent images don't accidentally revert it with package updates.

---

```
echo "auth required pam_wheel.so use_uid" >> /etc/pam.d/su &&
```

3. I am unsure what this is for? 😕 As far as I can tell, this means that only members of the administrative group `wheel` (or `0` if there is no wheel) can switch to another user using the su command. That might make sense on a regular multi-user system, but I am unsure why it would matter for a container.

---

```
chgrp root /etc/passwd && chmod ug+rw /etc/passwd
```

4. Wider permissions on `/etc/passwd` is concerning. What use case is broken if the running user id doesn't exist?

```
echo ... >> /etc/passwd
```

5. Having the entrypoint itself modify `/etc/passwd` is fragile. Are there features that are broken if the user doesn't exist in `/etc/passwd` (like PostgreSQL's `initdb` that refuses to run)?  Minimally, this should probably use `useradd` and `usermod` rather than hand editing.

---

```
env | grep SPARK_JAVA_OPT_ |...
```

6. This is susceptible to a few bugs particularly around newlines in values. I see two ways around it.
    - ensuring the matching name is at the start, `-E '^SPARK_JAVA_OPT_'`, and running all the commands with null-terminated input and output: `env -0` and `-z` on the other commands
    - or bash variable prefix expansion
        ```
        for v in "${!SPARK_JAVA_OPT_@}"; do
            SPARK_EXECUTOR_JAVA_OPTS+=( "${!v}" )
        done
        ```

---

7. `switch_spark_if_root` and `tini` should only happen when running a spark `driver` or `executor` and not for something like `docker run -it spark bash`.

8. To minimize duplication across layers, `chmod`'s should be done in the layer that creates the file/folder (or in the case of a file from the context via `COPY`, it should have the `+x` committed to git)

---

non-blocking:

9. using `set -ex` means you can use `;` instead of `&&` (really only matters for complex expressions, like the `||` in the later `RUN` that does use `;`)

Yikun:
@yosifkit Thanks for your reply! We will address soon as soon as possible!

> Would it be useful to save space by sharing layers by having one image from another? 🤔 Something like the *java11-ubuntu as the "base" with r and python variants FROM that and the r-python being FROM, probably, the larger one of those?

Good suggestion
- should we put the base image and r, python image into separate PR? (first base and left python and r, because after change that, r, python depends on base). 
- We are also consider support java17. Just wanna sure java11 / java17 should be two base image, right?

Yikun:
@erikerlandson Really much thanks for your reply, we added `useradd` so seems not a problem now?

https://github.com/apache/spark-docker/blob/fe05e38f0ffad271edccd6ae40a77d5f14f3eef7/3.4.0/scala2.12-java11-ubuntu/Dockerfile#L21

And any more background about question 3 (`/etc/pam.d/su`) [SPARK-25275](https://issues.apache.org/jira/browse/SPARK-25275), is it also for OpenShift?

Yikun:
@yosifkit  Question 1, 8, 9 resolved, question 3, 4, 5, 6, 7 wait for reply! Thanks!

Yikun:
> should we put the base image and r, python image into separate PR? (first base and left python and r, because after change that, r, python depends on base).

The CI failed due to python/r/all image depends on base image. so I guess the first PR should be the base image, right? @yosifkit

tianon:
Because our build system has to be able to query/calculate inter-image dependencies, we unfortunately do not support parameterized `FROM` values.

Yikun:
@tianon So the suggestion which @yosifkit mentioned before (question 1) is that first adding base image and then adding pyspark/sparkr.

When the version bump in future, should also be two separate PR (base first, and then pyspark/sparkr)?

yosifkit:
Related to 1:
> should we put the base image and r, python image into separate PR?

There is no need for them to be separate PRs. The `bashbrew` tool figures build order dependencies between images (both `eclipse-temurin:xyz` -> `spark` and `spark:foo` -> `spark:bar`). They just can't have variables in the `FROM` values and all images referenced need to be a current official image or part of the changes.

> We are also consider support java17. Just wanna sure java11 / java17 should be two base image, right?

Yes, thats fine. They can't really be the same base since they need different versions of java and so will need a different `FROM eclipse-temurin:[version]`.

---

Related to 2 and 6:
> It was introduced with question 6, https://github.com/apache-spark-on-k8s/spark/pull/444/files#r134075892

Oh, if that's the place using it, then just setting `SHELL ["bash", "-c"]` will fix that (and be used for any `RUN` lines after setting `SHELL`). Though I'd personally recommend just putting all that `CMD` into a proper shell script (and seems to already have been done?).

> For bash variable prefix expansion, it works but can it also support sh? (I try and it works, but I'm not sure it should the best way or not)?

It looks like your script is already running as bash. Are there instances where it doesn't run as bash? The image controls the environment, so you can ensure that `bash` is always available. It will not work with posix-compliant `sh`.

---

Related to 3:
> I found it was introduced in SPARK-25275 for security improvement. It should also for OpenShift, so if it's not a blocker, we prefer to keep.

Yeah, it is extra, but harmless. A stronger guarantee to prevent privilege escalation would be in recommending that users set `--security-opt=no-new-privileges` (or `allowPrivilegeEscalation: false` in Kubernetes).

---

Related to 4 and 5:
> It was introduced for OpenShift, apache-spark-on-k8s/spark#404.

Reading into that issue, it seems that `Spark will crash if it can't find a passwd entry` (maybe it doesn't now?), so that sounds a lot like what we have with PostgreSQL's `initdb` that requires the user to exist (https://github.com/docker-library/postgres/pull/448). Perhaps something similar with `libnss_wrapper` would work for spark?

> BTW, a case might related, would mind share some idea about how we would switch username in entrypoint.sh like apache/spark#40831 to meet the spark on k8s user switch requirement?

:thinking: Perhaps the `libnss_wrapper` trick would also work for this?

---

Related to 7:
> I tried others apache official image like solr / flink

The [`solr`](https://github.com/apache/solr-docker/blob/989825ee6dce2f6bf7b31051f1ba053b6c4426f2/9.2/Dockerfile#LL138C1-L138C6) image uses a `USER` directive in the Dockerfile, so it can be overridden with `--user` on `docker run`.  Unfortunately, it seems `flink` uses an ungated entrypoint user drop down that we have missed in review😔; that could make it frustrating to new users that see that they aren't root and try to switch but are frustrated when `docker run --user root [image] bash` doesn't work as expected (without also overriding the entrypoint too).

Yikun:
@yosifkit 

> Related to 1:

https://github.com/apache/spark-docker/pull/39 we removed paramaterized `FROM`

> Related to 2 and 6:

2 resolved in: https://github.com/apache/spark-docker/pull/41 by recovering `sh`
6 resolved in: https://github.com/apache/spark-docker/pull/42 by apply suggestion changes
>  Are there instances where it doesn't run as bash?

I think no, previous question was I thought DOI required also support `sh`, for spark I think bash support is enough.

> Related to 7:

Thanks for explaination, and one more question: what the **default user should be specified in Docerfile (root or spark?)** is recommanded by DOI?

For spark, we were using `spark` / `185` as default user (specified USER in dockerfile).
  - `docker run  --user spark  -ti spark bash`: the `spark` user will be used.
  - `docker run  --user root  -ti spark bash`: the `root` user will be used.
  - `docker run -ti spark bash`: the `spark` user will be used (rather than `root`). (is it regular behavior?)
  - If above behavior is right, it will be resolved by https://github.com/apache/spark-docker/pull/43

> Related to 4 and 5:

We will take a deep look, so the suggestion is that use the `libnss_wrapper` fake user when driver/executor is starting? But for other cmd, we will skip the `libnss_wrapper` step?

yosifkit:
> Thanks for explaination, and one more question: what the default user should be specified in Docerfile (root or spark?) is recommanded by DOI?

We don't yet have specific requirements on when to use `USER` vs a step down in an entrypoint script. I recommend that if the entrypoint needs to do setup work as `root` before stepping down then using the default `USER root` is appropriate (like `mysql`), otherwise `USER` configured to an app-specific user in the Dockerfile is better from a security perspective. If the app doesn't really have a user, we just leave it up to the user running it (like `python`).

Also making sure that it still works (as much as possible) when it is run as an unexpected/random user is helpful both for things like openshift or for local development where users might run as their user id in order to bind-mount files or directories into the container.

> We will take a deep look, so the suggestion is that use the `libnss_wrapper` fake user when driver/executor is starting? But for other cmd, we will skip the `libnss_wrapper` step?

Yeah, I'd recommend using `libnss_wrapper` only for Spark driver/executor. Hopefully it is a relatively simple drop-in.

Yikun:
@yosifkit Thanks for your patient review!

1. https://github.com/apache/spark-docker/pull/43 We use the `spark` USER by default finally and also addressed your inline comments.
2. https://github.com/apache/spark-docker/pull/45 Here is the libnss_wrapper PR to resolve `/etc/passwd` wider permissions issue. (the last issue!)

Now all issue had been addressed! I believe DOI for spark is almost ready!

yosifkit:
Just a few minor bits of feedback.
- Have you considered a `set -eo pipefail` on the entrypoint script to help prevent any errors from being silently ignored?
- Some of the used environment variables like `SPARK_EXECUTOR_MEMORY` `SPARK_DRIVER_BIND_ADDRESS` and `SPARK_DRIVER_URL` (maybe others) don't have a default in the entrypoint script and I can't find documentation on them. Will those be added to the Docker Hub readme or maybe your [Kubernetes page](https://spark.apache.org/docs/3.4.0/running-on-kubernetes.html) or [configuration page](https://spark.apache.org/docs/latest/configuration.html#environment-variables)?

tianon:
I've got a few question on the tagging.

Is it going to be common for users to actually want Python or R but not both?  Is the size difference significant?  (They're both just installed directly from distro packages, right?  Is their usage common enough to warrant an explicit variant, or should they rather be part of some generic documentation about how to expand the image with "additional tools" or something similar?  That'd be a ~2 line `Dockerfile` :sweat_smile:)

Similarly, the explicit Scala version in the tags seems strange to me - is that something that will change over time and users might want to pin to the older version?  If that's something that will remain fairly constant, then I'd recommend just dropping it from those already very verbose/long tag names. :see_no_evil:

You should also consider swapping `ubuntu` in your tags for the more explicit `focal` so that when you inevitably upgrade to `jammy`, users have a way to easily refer to the older version if things break for them. :eyes:

tianon:
```diff
+    ln -s /lib /lib64; \
```

:raised_eyebrow: this is interesting - can you explain a little bit what this is for and what it's supposed to be accomplishing? :thinking:

---

```diff
+    apt install ...
```

This should be `apt-get` (`apt` is not intended for unattended use, as the warning during build makes clear).

---

```diff
+    rm -rf /var/cache/apt/*; \
```

This is harmless, but should be unnecessary (the base image configuration already makes sure this directory stays empty).

---

```diff
+      --driver-url $SPARK_DRIVER_URL
```

You probably want to quote this (and many of the other variables in this execution); ala `--driver-url "$SPARK_DRIVER_URL"`

Yikun:
@tianon

> Is it going to be common for users to actually want Python or R but not both? 

Yes, there are three images type in spark:
https://hub.docker.com/r/apache/spark
https://hub.docker.com/r/apache/spark-py
https://hub.docker.com/r/apache/spark-r

The current tag design intend to replace these images. I think it's common enough for current spark user.

For python one, we might introduced more depends to enable more feature for PySpark. The PySpark one will be default one for spark DOI.

For scala one, it's mainly for who only want  to play spark or spark on k8s but not care pyspark/sparkr.

For sparkr, frankly, there are less users than pyspark, spark scala, so we just make it as separate one.

> Is their usage common enough to warrant an explicit variant, or should they rather be part of some generic documentation about how to expand the image with "additional tools" or something similar?

And also we want make docker image as a simple way to help user play spark with one command line (just like spark website today - Run now -
https://spark.apache.org/), so doc bring pontential complex.

> Is the size difference significant?

You can see the size info in https://hub.docker.com/r/apache/spark/tags. 

Scala (400) / PySpark(500) / SparkR (600MB) / all (650MB)

> the explicit Scala version in the tags seems strange to me -  that something that will change over time and users might want to pin to the older version?

Yes, each spark version has specific recommandation match scala version. such as in
https://spark.apache.org/downloads.html, we publish scala2.12/2.13 version for Spark.

So we prefer to keep the scala version in tag.


> You should also consider swapping ubuntu in your tags for the more explicit focal so that when you inevitably upgrade to jammy, users have a way to easily refer to the older version if things break for them.

The original logic is that we want to only want to users know the OS but ignore the OS version. There are also only one OS version to be published for each Spark version. (Thats also the version we validate in Apache Spark CI).

Consider above, Is a good choice to only use the -ubuntu rather than focal/jammy?

Yikun:
@yosifkit 

> Have you considered a set -eo pipefail on the entrypoint script to help prevent any errors from being silently ignored?

Yes, that's a good point. I will add soon.

> Will those be added to the Docker Hub readme or maybe your Kubernetes page or configuration page?

Good suggestion, I believe all var in entrypoint should be part of readme, or a link in readme to link to the spark doc, will update in doc PR.

tianon:
> https://spark.apache.org/downloads.html, we publish scala2.12/2.13 version for Spark.

Ah, so in the future you might have `spark:3.4.0-scala2.13` variants as well?

> The original logic is that we want to only want to users know the OS but ignore the OS version. There are also only one OS version to be published for each Spark version. (Thats also the version we validate in Apache Spark CI).

What's your plan for once focal reaches EOL?  (at which point we will ask/require that the images either be updated or removed from the "supported" set)

> It was introduced by PR: https://github.com/apache/spark/pull/23898 to address the issue about snappy

Given that's an Alpine-related fix, it seems extra weird to have it in Ubuntu-based images. :sweat_smile:

Yikun:
> Ah, so in the future you might have spark:3.4.0-scala2.13 variants as well?

Yes! I think it is. If we only support scala 2.12 (current), we just have `3.4.0-scala` in here, if we support multiple scala version in future, we will add something like `3.4.0-scala2.13` and `3.4.0-scala2.12`.

> What's your plan for once focal reaches EOL? What's your plan for once focal reaches EOL? (at which point we will ask/require that the images either be updated or removed from the "supported" set)

According to Apache Spark [Maintenance releases and EOL](https://spark.apache.org/versioning-policy.html) a version will be maintained with bug fix releases for a period of **18 months**. And OS release has more longer EOL time in general. So I believe removed from the "supported" set is fine for us. BTW, spark is less deps on specific OS, so upgrade to ubuntu latest is not a big problem(https://github.com/apache/spark/commit/170a8bca357e3057a1c37088960de31a261608b3), we can also update it too!

> Given that's an Alpine-related fix, it seems extra weird to have it in Ubuntu-based images.

Looks like reasonable, I will do a double confirm!

Yikun:
> Will those be added to the Docker Hub readme or maybe your [Kubernetes page](https://spark.apache.org/docs/3.4.0/running-on-kubernetes.html) or [configuration page](https://spark.apache.org/docs/latest/configuration.html#environment-variables)?

Added env descirption and configurtation link to Docker Hub readme [doc PR](https://github.com/docker-library/docs/pull/2200/files#diff-49dbd3b5c81f7c8c7e0e66f600d9714e045ece9f813f35f5def497a285c69a0bR51).

> Have you considered a set -eo pipefail on the entrypoint script to help prevent any errors from being silently ignored?

Addressed: https://github.com/apache/spark-docker/pull/49

> ```diff
> +    ln -s /lib /lib64; \
> ```

Addressed: https://github.com/apache/spark-docker/pull/48

> ```diff
> +    apt install ...
> ```

> ```diff
> +    rm -rf /var/cache/apt/*; \
> ```

Addressed: https://github.com/apache/spark-docker/pull/47

> 
> ```diff
> +      --driver-url $SPARK_DRIVER_URL
> ```

Addressed: https://github.com/apache/spark-docker/pull/49

tianon:
Sorry, did one more pass and noticed one last thing :bow: :bow: :heart:

```diff
+    GPG_KEY=34F0FC5C
```

Oops, this should be the full key fingerprint: `F28C9C925C188C35E345614DEDA00CE834F0FC5C` (generating a collision for such a short key ID is trivial :sweat_smile:); see https://github.com/docker-library/official-images#security

Yikun:
@tianon Oh, thanks for catch, I didn't realize before.

But seems `F28C9C925C188C35E345614DEDA00CE834F0FC5C` not working: https://github.com/apache/spark-docker/pull/50

I got the key from https://dist.apache.org/repos/dist/dev/spark/KEYS of [release manager](https://lists.apache.org/thread/fo5qlkqstjmszolzl9j6yk95jb61jr5t). I only get the short key for v3.4.1 (https://github.com/apache/spark-docker/blob/master/tools/template.py#L34).

Maybe a stupid question: how to map the short key to full key fingerprint?

Yikun:
@dongjoon-hyun Would you mind taking a look about why we can't validate the long key.

It seems short key (it can be validated) and long key (failed to be validated due to no public key found raise) is same one. 😂 Does it need  to import or some other operations?

tianon:
Oh fun -- the way I mapped it was by downloading and importing that full `KEYS` file (`gpg --import`), and then running `gpg --fingerprint 34F0FC5C`, which gives it to you with spaces (which is also technically valid in the input, so you could set your variable to the space-separated version as well if you're careful about quoting it properly).

I think what's happening in your PR is that previously, keys.openpgp.org was probably rejecting the short ID completely so your code was falling back to keyserver.ubuntu.com, but now it's accepting the full fingerprint, but because the email address isn't verified, it's not returning a UID (https://keys.openpgp.org/about), but GnuPG is quietly ignoring the key _and_ returning a successful exit code, which is less than ideal.

I've followed the steps in https://keys.openpgp.org/about/usage#gnupg-upload which should've emailed @dongjoon-hyun with a verification link that once verified, should allow you to restart that build and see success (as the path of least resistance here). :bow:

tianon:
(all your `gpg` invocations should also include `--batch`, which essentially puts GnuPG into "API mode" instead of "UI mode", but that's not related to or causing this issue :sweat_smile:)

Yikun:
@tianon Thanks, I have contact dongjoon to help address this issue, after above 2 PRs ^ passed and  merged, I will update.

Yikun:
All comments addressed. Ready for review again.

tianon (approved):

Yikun:
@yosifkit Would you mind taking a look recently? Thanks!

yosifkit (approved):

HyukjinKwon:
awesome!

zhengruifeng:
congrats! @Yikun

gatorsmile:
Thank you !

Yikun:
@yosifkit @tianon Much thanks for your help! I also have a [post test](https://github.com/apache/spark-docker/actions/runs/5605477658/jobs/10254648978), it passed all tests!

image

Thanks all cc @gatorsmile @erikerlandson @HyukjinKwon @zhengruifeng @pan3793 @emiliofernandes @julien-faye

PR #13091, opened 2022-09-02
elixir 1.14.0

getong:


yosifkit:
It looks like some of the older versions need to drop some architectures caused the transition of Debian Buster to LTS (https://github.com/docker-library/official-images/pull/13005, https://github.com/docker-library/official-images/pull/13077)

adamu:
@yosifkit It looks like the issues you mentioned have been resolved

PR #13099, opened 2022-09-05
add Apache APISIX image

soulbird:


library/apisix
```diff
@@ -0,0 +1,12 @@
+Maintainers: Apache APISIX Docker Maintainers <dev@apisx.apache.org> (@soulbird)
+GitRepo: https://github.com/apache/apisix-docker.git
+
+Tags: 2.15.0, 2.15.0-debian, latest
```
tao12345666333, line 4: 2.15.0 doesn't exist

soulbird: This is the tag of the final target image, not the repository's tag.

soulbird:
Has been successfully built using `bashbrew` tools.

image

tao12345666333:
@tianon @yosifkit PTAL, thanks!

yosifkit:
Hello! :sparkles:

tianon:
Hey, thanks for your patience!  We're trying to work through our queue, and before we get too deep into reviewing we'd like to make sure this is still something you're interested in pursuing (esp. since it's been so long since our last touch point)? :bow: :heart_hands:

PR #13118, opened 2022-09-09
[nats] Release v2.9.0

kozlovic:
Details can be found [here](https://github.com/nats-io/nats-server/releases/tag/v2.9.0)


kozlovic:
@tianon Was wondering if there is a problem with the PR or you are just too busy and did not get to it yet?

yosifkit:
Sorry, Friday was a day off, so we are a bit behind.

PR #13127, opened 2022-09-12
Update docker

tianon:
Changes:

- https://github.com/docker-library/docker/commit/c13cbee: Update 20.10 to 20.10.18


benz0li:
@tianon @yosifkit Thank you.

PR #13132, opened 2022-09-12
Update Debian to 20220912, esp. for 11.5 & 10.13

tianon:


tianon:
Marking as a draft for now - going to regenerate everything to make sure we include https://github.com/debuerreotype/docker-debian-artifacts/issues/174 :innocent:

p0358:
@tianon @yosifkit  Hello, there is a regression in this image that broke all images dependent on this for me:
```
root@test:~# docker run --rm debian:bullseye-20220822-slim bash -c 'echo test'
test
root@test:~# echo $?
0
root@test:~# docker run --rm debian:bullseye-20220912-slim bash -c 'echo test'
root@test:~# echo $?
132
```

Basically stuff from the new update is crashing with exit code 132 (SIGILL, illegal instruction). My CPU is 4th core Intel Core, so it supports AVX2 and SSE4.2, it does seem like the update though made these images dependent on some even newer instructions?

As a result right now I cannot run any Debian-based images anymore at all. I can still pull and run the Alpine ones just fine for example. I hope this matter can be looked into...

tianon:
(Discussing the SIGILL issue over in https://github.com/debuerreotype/docker-debian-artifacts/issues/175 instead :+1:)

PR #13136, opened 2022-09-13
OrientDB 3.1.20 Release

tglman:


tianon:
That CI failure is unfortunately the downstream effect of https://github.com/docker-library/docs/pull/2162#issuecomment-1172781329 -- `openjdk:8*` and `openjdk:11*` are no longer maintained/supported.

These older versions should either be updated to a supported distribution or removed (depending on whether they're still actively supported versions of OrientDB, would be my suggestion).  As a reminder:

PR #13137, opened 2022-09-13
bump Kong to 3.0.0

hanshuebner:


gszr (commented):
looks good on my end 👍

kikito:
lgtm too 👍

tianon:

(This is only informational - the label already exists so I'm not going to make it a blocker, but something to consider for the future.)

tianon:
A new major release is also a great time to revisit whether all the old versions listed here are still supported. :eyes:

As a reminder:

PR #13139, opened 2022-09-13
[Swift] Add support Swift 5.7 Release

shahmishal:


shahmishal:
@yosifkit @tianon Can you help merge this PR? Thanks!

library/swift
```diff
@@ -7,39 +7,83 @@ GitRepo: https://github.com/apple/swift-docker.git
 GitFetch: refs/heads/main
 
 
-Tags: 5.6.3, 5.6, 5.6.3-bionic, 5.6-bionic, bionic, latest
+Tags: 5.7.0, 5.7, 5.7.0-jammy, 5.7-jammy, jammy, latest
+GitCommit: 5d771d72823df0d2cd97692298efa20349ed2c9c
+Directory: 5.7/ubuntu/22.04
```
yosifkit, line 12: From the Dockerfile (and `jammy-slim` too), it seems like this version is missing `Architectures: amd64, arm64v8`. Is this version supposed to also have an `arm64v8` build?

shahmishal: Good catch: https://github.com/docker-library/official-images/pull/13145

PR #13150, opened 2022-09-14
hitch: fix the version overwrite

gquintard:
forgot to update the `populate.sh` script, so the image was updated, but not the tags. This is a hand-crafted file that should give the new image the correct tag, but also give back the `1.7.2` tags a proper image


tianon:
Oops :sweat_smile:

PR #13160, opened 2022-09-15
varnish: upgrade to 7.2.0

gquintard:
new version, the only two changes are that the `debian` images gets a new library (`getdns`) and we have a pair of env variables to set the default ports (which are privileged and it's apparently an issue on old `docker` versions)


tianon:
Is `install-vmod` both here in `scripts/` _and_ in toolbox on purpose? :eyes:

gquintard:
ah, no indeed

tianon:
FWIW, I don't think `sbuild` is actually necessary (I don't see where you're using it, but it's also not the most usable tool in this situation :sweat_smile:)

gquintard:
oh, right, probably a remnant of when we were building the packages. I'll remove it for next time. Cheers!

PR #13162, opened 2022-09-16
Update mongo-express for v1.0.0

BlackthornYugen:
Mostly security updates and new configuration to support breaking changes in mongo-db driver.

See related pull requests:
https://github.com/mongo-express/mongo-express/pull/920
https://github.com/mongo-express/mongo-express-docker/pull/82

Full release notes available here:
[v1.0.0 - Initial node18 support](https://github.com/mongo-express/mongo-express/releases/tag/v1.0.0)
~~[v1.0.1 - Convert to ESM](https://github.com/mongo-express/mongo-express/releases/tag/v1.0.1)~~

Edit 2023/10/02: Removed v1.0.1 from the scope of this MR. Once we get 1.0.0 working we will immediately open a new MR for 1.0.1.

### DIFF
--- a/library/mongo-express
+++ b/library/mongo-express
@@ -5,10 +5,13 @@ Maintainers: Nick Cox <nickcox1008@gmail.com> (@knickers),
 GitRepo: https://github.com/mongo-express/mongo-express-docker.git
 
 
-GitCommit: f5e8e026f1f6454e4b639f66f5b241936dfd49c8
-Tags: 1.0.0, latest
+GitCommit: fbaa75000664d7d986ffe8b184a4b24c621c5ee7
+Tags: 1.0.1, latest
 Architectures: amd64, arm64v8
 
+GitCommit: f5e8e026f1f6454e4b639f66f5b241936dfd49c8
+Tags: 1.0.0
+Architectures: amd64, arm64v8
 
 GitCommit: 4b43fe8a1206434cb32a006cd155dd71462f092f
 Tags: 0.54.0, 0.54
--- a/library/mongo-express
+++ b/library/mongo-express
@@ -3,13 +3,21 @@
 Maintainers: Nick Cox <nickcox1008@gmail.com> (@knickers),
              John Steel <john@jskw.dev> (@BlackthornYugen)
 GitRepo: https://github.com/mongo-express/mongo-express-docker.git
-
-
-GitCommit: 0782917e644220ba2ddd81590a9630074a362488
-Tags: 1.0.1, latest
+GitCommit: a375909f86e1cac3ecc7ce381dc2ef4086f111a9
+Tags: 1.0.0-20-alpine3.18, 1.0-20-alpine3.18, 1-20-alpine3.18
 Architectures: amd64, arm64v8
-
-GitCommit: f5e8e026f1f6454e4b639f66f5b241936dfd49c8
-Tags: 1.0.0
+GitCommit: 2a6d82f29978e82ab5685618dc34ba6043568917
+Directory: 1.0/20-alpine3.18
+1.0.0-20 1.0-20 1-20 1.0.0-20-alpine3.17 1.0-20-alpine3.17 1-20-alpine3.17
+Tags: 1.0.0-20, 1.0-20, 1-20, 1.0.0-20-alpine3.17, 1.0-20-alpine3.17, 1-20-alpine3.17
 Architectures: amd64, arm64v8
-
+GitCommit: 2a6d82f29978e82ab5685618dc34ba6043568917
+Directory: 1.0/20-alpine3.17
+Tags: 1.0.0-18-alpine3.18, 1.0-18-alpine3.18, 1-18-alpine3.18
+Architectures: amd64, arm64v8
+GitCommit: 2a6d82f29978e82ab5685618dc34ba6043568917
+Directory: 1.0/18-alpine3.18
+Tags: 1.0.0, 1.0, 1, 1.0.0-18, 1.0-18, 1-18, 1.0.0-18-alpine3.17, 1.0-18-alpine3.17, 1-18-alpine3.17, latest
+Architectures: amd64, arm64v8
+GitCommit: 2a6d82f29978e82ab5685618dc34ba6043568917
+Directory: 1.0/18-alpine3.17
--- a/library/mongo-express
+++ b/library/mongo-express
@@ -22,6 +22,3 @@ GitCommit: 2a6d82f29978e82ab5685618dc34ba6043568917
 Directory: 1.0/18-alpine3.18
 Tags: 1.0.0, 1.0, 1, 1.0.0-18, 1.0-18, 1-18, 1.0.0-18-alpine3.17, 1.0-18-alpine3.17, 1-18-alpine3.17, latest
 Architectures: amd64, arm64v8
-
-GitCommit: 2a6d82f29978e82ab5685618dc34ba6043568917
-Directory: 1.0/18-alpine3.17
--- a/library/mongo-express
+++ b/library/mongo-express
@@ -3,21 +3,25 @@
 Maintainers: Nick Cox <nickcox1008@gmail.com> (@knickers),
              John Steel <john@jskw.dev> (@BlackthornYugen)
 GitRepo: https://github.com/mongo-express/mongo-express-docker.git
+
 GitCommit: a375909f86e1cac3ecc7ce381dc2ef4086f111a9
 Tags: 1.0.0-20-alpine3.18, 1.0-20-alpine3.18, 1-20-alpine3.18
 Architectures: amd64, arm64v8
+
 GitCommit: 2a6d82f29978e82ab5685618dc34ba6043568917
 Directory: 1.0/20-alpine3.18
-1.0.0-20 1.0-20 1-20 1.0.0-20-alpine3.17 1.0-20-alpine3.17 1-20-alpine3.17
 Tags: 1.0.0-20, 1.0-20, 1-20, 1.0.0-20-alpine3.17, 1.0-20-alpine3.17, 1-20-alpine3.17
 Architectures: amd64, arm64v8
+
 GitCommit: 2a6d82f29978e82ab5685618dc34ba6043568917
 Directory: 1.0/20-alpine3.17
 Tags: 1.0.0-18-alpine3.18, 1.0-18-alpine3.18, 1-18-alpine3.18
 Architectures: amd64, arm64v8
+
 GitCommit: 2a6d82f29978e82ab5685618dc34ba6043568917
 Directory: 1.0/18-alpine3.18
 Tags: 1.0.0, 1.0, 1, 1.0.0-18, 1.0-18, 1-18, 1.0.0-18-alpine3.17, 1.0-18-alpine3.17, 1-18-alpine3.17, latest
 Architectures: amd64, arm64v8
+
 GitCommit: 2a6d82f29978e82ab5685618dc34ba6043568917
 Directory: 1.0/18-alpine3.17
--- a/library/mongo-express
+++ b/library/mongo-express
@@ -5,7 +5,7 @@ Maintainers: Nick Cox <nickcox1008@gmail.com> (@knickers),
 GitRepo: https://github.com/mongo-express/mongo-express-docker.git
 
 
-GitCommit: fbaa75000664d7d986ffe8b184a4b24c621c5ee7
+GitCommit: 0782917e644220ba2ddd81590a9630074a362488
 Tags: 1.0.1, latest
 Architectures: amd64, arm64v8
 
--- a/library/mongo-express
+++ b/library/mongo-express
@@ -1,11 +1,12 @@
 # this file is generated via https://github.com/mongo-express/mongo-express-docker/blob/4b43fe8a1206434cb32a006cd155dd71462f092f/generate-stackbrew-library.sh
 
-Maintainers: Nick Cox <nickcox1008@gmail.com> (@knickers)
+Maintainers: Nick Cox <nickcox1008@gmail.com> (@knickers),
+             John Steel <john@jskw.dev> (@BlackthornYugen)
 GitRepo: https://github.com/mongo-express/mongo-express-docker.git
 
 
-GitCommit: 26e7ea94e4d222de7d52531caee52149ac093c26
-Tags: 1.0.0-alpha.4, 1.0.0-alpha, latest
+GitCommit: 0fe77f884e873b344a61ca6184694042141c4fb6
+Tags: 1.0.0, latest
 Architectures: amd64, arm64v8
 
 
--- a/library/mongo-express
+++ b/library/mongo-express
@@ -5,7 +5,7 @@ Maintainers: Nick Cox <nickcox1008@gmail.com> (@knickers),
 GitRepo: https://github.com/mongo-express/mongo-express-docker.git
 
 
-GitCommit: 0fe77f884e873b344a61ca6184694042141c4fb6
+GitCommit: f5e8e026f1f6454e4b639f66f5b241936dfd49c8
 Tags: 1.0.0, latest
 Architectures: amd64, arm64v8
 
--- a/library/mongo-express
+++ b/library/mongo-express
@@ -13,6 +13,3 @@ GitCommit: f5e8e026f1f6454e4b639f66f5b241936dfd49c8
 Tags: 1.0.0
 Architectures: amd64, arm64v8
 
-GitCommit: 4b43fe8a1206434cb32a006cd155dd71462f092f
-Tags: 0.54.0, 0.54
-Architectures: amd64, arm64v8
--- a/library/mongo-express
+++ b/library/mongo-express
@@ -1,26 +1,26 @@
-# this file is generated via https://github.com/mongo-express/mongo-express-docker/blob/3971af1b3b2703d9b6f5f6ca12e33ea4ca99bc97/generate-stackbrew-library.sh
+# this file is generated via https://github.com/mongo-express/mongo-express-docker/blob/0b9b935b188caa945834a27f7475a059c4e60b13/generate-stackbrew-library.sh
 
 Maintainers: Nick Cox <nickcox1008@gmail.com> (@knickers),
              John Steel <john@jskw.dev> (@BlackthornYugen)
 GitRepo: https://github.com/mongo-express/mongo-express-docker.git
-GitCommit: 3971af1b3b2703d9b6f5f6ca12e33ea4ca99bc97
+GitCommit: 0b9b935b188caa945834a27f7475a059c4e60b13
 
-Tags: 1.0.0-20-alpine3.18, 1.0-20-alpine3.18, 1 latest-20-alpine3.18
+Tags: 1.0.0-20-alpine3.18, 1.0-20-alpine3.18, 1-20-alpine3.18
 Architectures: amd64, arm64v8
 GitCommit: 5fb537b2fd181e458f58137fdb55d061a8c23da5
 Directory: 1.0/20-alpine3.18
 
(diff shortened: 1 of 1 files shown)


BlackthornYugen:
Looks like we have a bug in our dockerfile, maybe an overloaded environment variable / build arg. It's trying to look for `github.com/mongo-express_latest/`. 

```
cp: cannot stat 'tar/mongo-express_latest/[https://github.com/${MONGO_REPOSITORY}/archive/refs/tags/${MONGO_EXPRESS_TAG}.tar.gz](https://github.com/$%7BMONGO_REPOSITORY%7D/archive/refs/tags/$%7BMONGO_EXPRESS_TAG%7D.tar.gz)': No such file or directory
```

BlackthornYugen:
Maybe it's just that docker official images don't support adding from a URL? Should we replace it with a RUN wget? Or maybe we should rollback to using npm install now that v1.0.0 is published to NPM... 

FYI @shakaran @rtritto

rtritto:
> Looks like we have a bug in our dockerfile, maybe an overloaded environment variable / build arg. It's trying to look for `github.com/mongo-express_latest/`. 
> 
> ```
> cp: cannot stat 'tar/mongo-express_latest/[https://github.com/${MONGO_REPOSITORY}/archive/refs/tags/${MONGO_EXPRESS_TAG}.tar.gz](https://github.com/$%7BMONGO_REPOSITORY%7D/archive/refs/tags/$%7BMONGO_EXPRESS_TAG%7D.tar.gz)': No such file or directory
> ```

Can you try to replace syntax `${var}` to `$var`?

MahdiAbbasi95:
> > Looks like we have a bug in our dockerfile, maybe an overloaded environment variable / build arg. It's trying to look for `github.com/mongo-express_latest/`.
> > ```
> > cp: cannot stat 'tar/mongo-express_latest/[https://github.com/${MONGO_REPOSITORY}/archive/refs/tags/${MONGO_EXPRESS_TAG}.tar.gz](https://github.com/$%7BMONGO_REPOSITORY%7D/archive/refs/tags/$%7BMONGO_EXPRESS_TAG%7D.tar.gz)': No such file or directory
> > ```
> 
> Can you try to replace syntax `${var}` to `$var`?
> 
> FYI @MahdiAbbasi95

I changed the variable syntax and I opened a new pull request in our repo.
Please check it and merge.

https://github.com/mongo-express/mongo-express-docker/pull/84

@BlackthornYugen @shakaran

BlackthornYugen:
> > Looks like we have a bug in our dockerfile, maybe an overloaded environment variable / build arg. It's trying to look for `github.com/mongo-express_latest/`. 
> > 
> > ```
> > cp: cannot stat 'tar/mongo-express_latest/[https://github.com/${MONGO_REPOSITORY}/archive/refs/tags/${MONGO_EXPRESS_TAG}.tar.gz](https://github.com/$%7BMONGO_REPOSITORY%7D/archive/refs/tags/$%7BMONGO_EXPRESS_TAG%7D.tar.gz)': No such file or directory
> > ```
> 
> Can you try to replace syntax `${var}` to `$var`?
> 
> FYI @MahdiAbbasi95

I've tested that the previous syntax works both on my machine and in a GitHub workflow, removing the {} doesn't seem safe.

Some interpreters might have difficulty deciding what chars are part of the variable name and what aren't. Maybe we could change add to run wget instead? I think that might be more well supported

MahdiAbbasi95:
> > > Looks like we have a bug in our dockerfile, maybe an overloaded environment variable / build arg. It's trying to look for `github.com/mongo-express_latest/`.
> > > ```
> > > cp: cannot stat 'tar/mongo-express_latest/[https://github.com/${MONGO_REPOSITORY}/archive/refs/tags/${MONGO_EXPRESS_TAG}.tar.gz](https://github.com/$%7BMONGO_REPOSITORY%7D/archive/refs/tags/$%7BMONGO_EXPRESS_TAG%7D.tar.gz)': No such file or directory
> > > ```
> > 
> > 
> > Can you try to replace syntax `${var}` to `$var`?
> > FYI @MahdiAbbasi95
> 
> I've tested that the previous syntax works both on my machine and in a GitHub workflow, removing the {} doesn't seem safe.
> 
> Some interpreters might have difficulty deciding what chars are part of the variable name and what aren't. Maybe we could change add to run wget instead? I think that might be more well supported

I'll change it

MahdiAbbasi95:
> > > > Looks like we have a bug in our dockerfile, maybe an overloaded environment variable / build arg. It's trying to look for `github.com/mongo-express_latest/`.
> > > > ```
> > > > cp: cannot stat 'tar/mongo-express_latest/[https://github.com/${MONGO_REPOSITORY}/archive/refs/tags/${MONGO_EXPRESS_TAG}.tar.gz](https://github.com/$%7BMONGO_REPOSITORY%7D/archive/refs/tags/$%7BMONGO_EXPRESS_TAG%7D.tar.gz)': No such file or directory
> > > > ```
> > > 
> > > 
> > > Can you try to replace syntax `${var}` to `$var`?
> > > FYI @MahdiAbbasi95
> > 
> > 
> > I've tested that the previous syntax works both on my machine and in a GitHub workflow, removing the {} doesn't seem safe.
> > Some interpreters might have difficulty deciding what chars are part of the variable name and what aren't. Maybe we could change add to run wget instead? I think that might be more well supported
> 
> I'll change it

@BlackthornYugen
I've changed and tested it, please check the PR.

BlackthornYugen:
> @BlackthornYugen I've changed and tested it, please check the PR.

Thanks! I've updated to use the latest version here.

rtritto:
@BlackthornYugen https://github.com/mongo-express/mongo-express-docker/pull/82#issuecomment-1249563687

BlackthornYugen:
> @BlackthornYugen [mongo-express/mongo-express-docker#82 (comment)](https://github.com/mongo-express/mongo-express-docker/pull/82#issuecomment-1249563687)

NPM install might be a nicer way to do it in the future but I don't think it needs to hold up this release. 

FYI: @knickers

BlackthornYugen:
I wasn't aware of the test script at https://github.com/mongo-express/docker-official-images/blame/master/test/tests/mongo-express-run/run.sh -- I guess we should make sure that passes. We must have broken it somehow? Or something that used to work no longer works...

BlackthornYugen:
It looks like it just tries to curl the root to get a 200OK, but it looks like basic auth is enabled so it is probably getting a 401?

BlackthornYugen:
We could fix it by updating the test to use `curl --user "admin:pass" --fail --silent --location --request "$method" "$@" "http://me:8081/$url"` ; however, having a default credential isn't really more secure than having no basic auth requirement. We are probably going to break deployments that are already secured through a reverse proxy (like I do). 

So we should probably should make sure that the default deployment of this container does not require basic auth for such users.

tldr: **basic auth will cause headaches for users and not increase security**

yosifkit:
The test is only a basic smoke test; I'll leave it up to the `mongo-express` image maintainers whether to change the test and/or the image. I am glad that it is helping find a possible UX issue with the new release.

The "Naughty" test is mentioning that the `mongo-express:0.54.0` image is `FROM node:12-alpine3.11`, which is an image that is no longer supported or updated. That means that it misses out on the automatic image rebuilds of the official images that are triggered by a parent image update.  So the `0.54.0` image should either be dropped or updated to a supported release of the `node` image.  Removing it here will remove it from the "Supported" section on the Hub readme, but the tag will still be available to users who want it.  (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)

BlackthornYugen:
> The test is only a basic smoke test; I'll leave it up to the `mongo-express` image maintainers whether to change the test and/or the image. I am glad that it is helping find a possible UX issue with the new release.
> 
> The "Naughty" test is mentioning that the `mongo-express:0.54.0` image is `FROM node:12-alpine3.11`, which is an image that is no longer supported or updated. That means that it misses out on the automatic image rebuilds of the official images that are triggered by a parent image update. So the `0.54.0` image should either be dropped or updated to a supported release of the `node` image. Removing it here will remove it from the "Supported" section on the Hub readme, but the tag will still be available to users who want it. (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)

Should be good now I think

jeremyjams:
Hello guys, thank you for your work :muscle:
Do you have any updates on this?

BlackthornYugen:
> Hello guys, thank you for your work 💪 Do you have any updates on this?

I don't know, I think only @knickers or a docker-library admin have permission to merge.

yosifkit:
Apologies for the delay. :bowing_man:

Blocking: 

1. Neither of the two below images have an up-to-date base image. The image it is `FROM` needs to be an active Docker Official Image. `node:12*` is not a maintained version of the node image. It should be `FROM` 18 or 20 since Node.js 14 and 16 are end of life at the end of April and September, respectively. (unless you want to update again when those are removed).  `node:12-alpine3.16` is not a valid image, since `node:12` never had an Alpine 3.16 image and Node.js 12 is end of life anyway.

	> \- mongo-express:0.54.0 (FROM node:12-alpine3.11) -- completely unsupported base!
	> \- mongo-express:1.0.1 (FROM node:12-alpine3.16) -- completely unsupported base!

1. For the `1.0.0` Dockerfile: anything downloaded (even over `https`) should be verified in some way, usually via a detached gpg signature or, if that's unavailable, via a checksum embedded in the Dockerfile (https://github.com/docker-library/official-images/tree/20ad9922430f293411aa6808c12a7e1ec73ff219#image-build).

Non-blocking:

3. We have a basic integration test for `mongo-express` in [test/tests/mongo-express-run/run.sh](https://github.com/docker-library/official-images/blob/20ad9922430f293411aa6808c12a7e1ec73ff219/test/tests/mongo-express-run/run.sh); it is failing on the `1.0.0` version. If it could be updated to support `1.x` that'd be great, but if not, we are happy to drop that test if it isn't useful to you for us to run.

4. If mongo-express `0.54` is no longer supported now that `1.x` is out, it should be removed from the `library/mongo-express` file. The current image will remain available to `pull` from Docker Hub. (`0.54` doesn't have to be removed, but it does need an updated base image if kept)

tianon:
Also blocking:

1. `chmod +x` on a file that's `COPY`'d previously will copy the file, so the executable bit should be applied in Git instead (and then `COPY` will apply it correctly)
2. the `COPY` of the entrypoint should happen as low in the `Dockerfile` as possible (at the very least, after the installation of `mongo-express`)

BlackthornYugen:
> 1. Neither of the two below images have an up-to-date base image. The image it is `FROM` needs to be an active Docker Official Image. `node:12*` is not a maintained version of the node image. It should be `FROM` 18 or 20 since Node.js 14 and 16 are end of life at the end of April and September, respectively. (unless you want to update again when those are removed).  `node:12-alpine3.16` is not a valid image, since `node:12` never had an Alpine 3.16 image and Node.js 12 is end of life anyway.

This PR should not include 0.54.0, moving forward we cannot support that version of alpine. I think we have also dropped support for node 12 but I'll need to double check. 



> 2\. For the `1.0.0` Dockerfile: anything downloaded (even over `https`) should be verified in some way, usually via a detached gpg signature or, if that's unavailable, via a checksum embedded in the Dockerfile (https://github.com/docker-library/official-images/tree/20ad9922430f293411aa6808c12a7e1ec73ff219#image-build).

That's a good point. We will update to pin a checksum.



> 3\. We have a basic integration test for `mongo-express` in [test/tests/mongo-express-run/run.sh](https://github.com/docker-library/official-images/blob/20ad9922430f293411aa6808c12a7e1ec73ff219/test/tests/mongo-express-run/run.sh); it is failing on the `1.0.0` version. If it could be updated to support `1.x` that'd be great, but if not, we are happy to drop that test if it isn't useful to you for us to run.

The 1.0.0 version is always goign to fail this test but I think future versions should pass. Maybe we don't need 1.0.0? But it might be nice to give people the option to use it in case something was broken in a future version...

> 4\. If mongo-express `0.54` is no longer supported now that `1.x` is out, it should be removed from the `library/mongo-express` file. The current image will remain available to `pull` from Docker Hub. (`0.54` doesn't have to be removed, but it does need an updated base image if kept)

I thought I already did this but maybe I mistakenly reverted or something. Will double check.

tianon:
The current PR's 1.0.1 can't build because `node:12-alpine3.16` doesn't exist (and never did :grimacing:).

If we can commit to updating these other things in the next PR, then I think it's reasonable for that to be the only real hard blocker here.

armond1978:
@BlackthornYugen Is there any further update on this PR?

BlackthornYugen:
> @BlackthornYugen Is there any further update on this PR?

I believe this will be unblocked after https://github.com/mongo-express/mongo-express-docker/pull/93 is through. Should be able to update publish 1.0.0 and 1.0.1 this week. 🤞

BlackthornYugen:
@LaurentGoderre any idea why it doens't like it? I'm trying to use the output from `./generate-stackbrew-library.sh`. It says it wants a ':', but the previous one didn't have colons in tags, it was comma delimited. 

```shell
+ docker run --rm --read-only --tmpfs /tmp oisupport/bashbrew:diff-pr ./diff-pr.sh 13162
+ tee /home/runner/work/official-images/official-images/oi-pr.diff
failed fetching repo "mongo-express"
Bad line: '1.0.0-20 1.0-20 1-20 1.0.0-20-alpine3.17 1.0-20-alpine3.17 1-20-alpine3.17
' has no ':'
failed fetching repo "mongo-express"
Bad line: '1.0.0-20 1.0-20 1-20 1.0.0-20-alpine3.17 1.0-20-alpine3.17 1-20-alpine3.17
' has no ':'
```

library/mongo-express
```diff
@@ -1,14 +1,23 @@
...
+Tags: 1.0.0-20-alpine3.18, 1.0-20-alpine3.18, 1-20-alpine3.18
 Architectures: amd64, arm64v8
-
-
-GitCommit: 4b43fe8a1206434cb32a006cd155dd71462f092f
-Tags: 0.54.0, 0.54
+GitCommit: 2a6d82f29978e82ab5685618dc34ba6043568917
+Directory: 1.0/20-alpine3.18
+1.0.0-20 1.0-20 1-20 1.0.0-20-alpine3.17 1.0-20-alpine3.17 1-20-alpine3.17
+Tags: 1.0.0-20, 1.0-20, 1-20, 1.0.0-20-alpine3.17, 1.0-20-alpine3.17, 1-20-alpine3.17
```
yosifkit, line 12: It looks like there is an extra line above the tags that it is getting hung up on.  Maybe a debug line accidentally left in?

BlackthornYugen: Oh right; didn't even notice that before. Let me take it out.

BlackthornYugen: Removed in dbc659b46f6d590e76e4c3e4863191a8635a49ce -- I also spaced out the variants so it's easier to read.

yosifkit:
I think dropping [line 110](https://github.com/mongo-express/mongo-express-docker/blob/a375909f86e1cac3ecc7ce381dc2ef4086f111a9/generate-stackbrew-library.sh#L110) and adding a plain `echo` before the cat on [line116](https://github.com/mongo-express/mongo-express-docker/blob/a375909f86e1cac3ecc7ce381dc2ef4086f111a9/generate-stackbrew-library.sh#L116) (or a newline to start the cat) should get `generate-stackbrew-library.sh` to the correct output.

[Line 51](https://github.com/mongo-express/mongo-express-docker/blob/a375909f86e1cac3ecc7ce381dc2ef4086f111a9/generate-stackbrew-library.sh#L51) could also be dropped but is harmless.

BlackthornYugen:
@LaurentGoderre it looks like it's not reading the path to the dockerfile correctly? Do we need to escape the forward slash? 

```diff
- mongo-express_1-20-alpine3.18/Dockerfile
+ mongo-express/1-20-alpine3.18/Dockerfile
```

```
Run docker run --rm --read-only --tmpfs /tmp oisupport/bashbrew:diff-pr ./diff-pr.sh "$GITHUB_PR_NUMBER" | tee "$GITHUB_WORKSPACE/oi-pr.diff"
+ docker run --rm --read-only --tmpfs /tmp oisupport/bashbrew:diff-pr ./diff-pr.sh 13162
+ tee /home/runner/work/official-images/official-images/oi-pr.diff
gawk: cmd. line:2: fatal: cannot open file `tar/mongo-express_1-20-alpine3.18/Dockerfile' for reading: No such file or directory
Error: Process completed with exit code 2.
```

yosifkit:
> it looks like it's not reading the path to the dockerfile correctly? Do we need to escape the forward slash?

The directories in the diff-pr script are not the same paths as the upstream repo but are just paths related to the image and tags it represents to help us compare across revisions.

BlackthornYugen:
> > it looks like it's not reading the path to the dockerfile correctly? Do we need to escape the forward slash?
> 
> The directories in the diff-pr script are not the same paths as the upstream repo but are just paths related to the image and tags it represents to help us compare across revisions.

But doesn't the error makes it look like it can't find the Dockerfile?

yosifkit:
> But doesn't the error makes it look like it can't find the Dockerfile?

Yes, because the first entry is missing its path (which is on the second entry, and the second is on the third, etc).

BlackthornYugen:
> > But doesn't the error makes it look like it can't find the Dockerfile?
> 
> Yes, because the first entry is missing its `Directory` and `GitCommit` (which is on the second entry, and the second is on the third, etc).

Ah! Thanks for the hints 😅

Going to try again with the changes suggested. PR here: 
https://github.com/mongo-express/mongo-express-docker/pull/100

library/mongo-express
```diff
@@ -1,14 +1,26 @@
...
+GitCommit: 3971af1b3b2703d9b6f5f6ca12e33ea4ca99bc97
 
+Tags: 1.0.0-20-alpine3.18, 1.0-20-alpine3.18, 1 latest-20-alpine3.18
+Architectures: amd64, arm64v8
+GitCommit: 5fb537b2fd181e458f58137fdb55d061a8c23da5
+Directory: 1.0/20-alpine3.18
 
-GitCommit: 26e7ea94e4d222de7d52531caee52149ac093c26
-Tags: 1.0.0-alpha.4, 1.0.0-alpha, latest
+Tags: 1.0.0-20, 1.0-20, 1 latest-20, 1.0.0-20-alpine3.17, 1.0-20-alpine3.17, 1 latest-20-alpine3.17
```
yosifkit, line 13: Ah, I missed all the `1 othertag,` here and on the other entries below.  Not sure where the fix is yet.

BlackthornYugen: Just noticed that. We shouldn't ever have a space in the tag right? 

image

yosifkit: Nope, no spaces in a tag.  It looks like it comes from the quotes added on https://github.com/mongo-express/mongo-express-docker/pull/100/files#diff-9dd720074107975da3d3140d5e27deb1e3870cb10fd74029ded8d6e1efe98346R85; which references a space separated string, so it was unqoted on purpose.

LaurentGoderre: those are two different tags from the script. Still a little bug it seems

LaurentGoderre: Good catch @yosifkit

BlackthornYugen: My bad! I should have followed my own advice to keep MRs small 😅

LaurentGoderre (approved):

yosifkit (approved):
LGTM

No worries on the `mongo-express-run` test. It needs changes to support mongo-express `1.x`, so we can just ignore them for now. I'm up for just dropping the test, but we can also see if there is a way to update it to work.

Edit: link to [`mongo-express-run` test](https://github.com/docker-library/official-images/blob/8a5c557a6a3fdbe8a83c09ec533bee7d3284f29b/test/tests/mongo-express-run/run.sh)

BlackthornYugen:
@armond1978 @rtritto @shakaran 1.0.0 is up but we have a bug to resolve in 1.0.1.  

https://github.com/mongo-express/mongo-express-docker/issues/97

PR #13194, opened 2022-09-20
Add official image for OpenSearch Dashboards

rishabh6788:
Signed-off-by: Rishabh Singh <sngri@amazon.com>

OpenSearch Dashboards is the default visualization tool for data in OpenSearch. It also serves as a user interface for many of the OpenSearch plugins, including security, alerting, Index State Management, SQL, and more.
To know more about OpenSearch Dashboards please read our [documentation](https://opensearch.org/docs/latest/dashboards/index/)

# Checklist for Review

-	[x] associated with or contacted upstream? We own the source https://github.com/opensearch-project/OpenSearch-Dashboards
-	[x] available under [an OSI-approved license](https://opensource.org/licenses)? [Apache License 2.0](https://opensource.org/licenses/Apache-2.0)
-	[x] does it fit into one of the common categories? ("service", "language stack", "base distribution") Service
-	[x] does a [documentation](https://github.com/docker-library/docs/blob/master/README.md) PR exist? https://github.com/docker-library/docs/pull/2197
-	[ ] ~~if `FROM scratch`, tarballs only exist in a single commit within the associated history?~~

### DIFF
--- a/library/opensearch-dashboards
+++ b/library/opensearch-dashboards
@@ -1,0 +1,11 @@
+Maintainers: OpenSearch Project Team <opensearch@amazon.com> (@opensearch-project),
+             Rishabh Singh <sngri@amazon.com> (@rishabh6788),
+             Peter Zhu <zhujiaxi@amazon.com> (@peterzhuamazon),
+             Prudhvi Godithi <pgodithi@amazon.com> (@prudhvigodithi)
+GitRepo: https://github.com/opensearch-project/dashboards-docker-images.git
+
+Tags: 2.3.0, 2.3, 2, latest
+Architectures: amd64, arm64v8
+GitCommit: c8f80771985daceb6dcbe9498790b2d848159786
+Directory: 2.x
+


bbarani:
Gentle ping @tianon @yosifkit

bbarani:
Friendly ping @tianon @yosifkit

bbarani:
Friendly ping @tianon @yosifkit

bbarani:
Can you please provide an update  @tianon @yosifkit

bbarani:
Friendly ping @tianon @yosifkit

yosifkit:
Hello! :sparkles:

tianon:
Thank you for your patience :bow:

I'd like to start with a bit of high-level feedback about Dashboards specifically, then I'll get into some more technical bits.

---

From what I've understood, Dashboards is _very_ closely tied to OpenSearch, right?  (even to the point of being pretty fairly version-locked?)  The reason I'm asking is that we do strive for each official image to be reasonably "generally useful", so I'm wondering if this should perhaps be tags of `opensearch` itself (`opensearch:dashboard`, for example), since opensearch _is_ pretty generally useful, but this is really somewhat more like optional functionality of that core product.

---

```diff
+FROM amazonlinux:2 AS linux_stage_0
```

The use of multi-stage here is a little bit dubious (https://github.com/docker-library/faq#multi-stage-builds), but I don't think that's necessarily a blocker.  If you want to keep it multi-stage, can you please include `Builder: buildkit` in either your "global" entry (where `Maintainers:` and `GitRepo:` are) or in your `Tags:` entry to help mitigate the associated build cache issues?  (We hope to turn that on for all builds eventually, but we're doing so cautiously to avoid potential breakage at a large scale.)

---

```diff
+ARG OSD_VERSION=2.3.0
```

Every time this changes (in _both_ instances of it :thinking:), it's going to cause a cache bust, so it should be moved down just before the layer/instruction which uses/requires/changes based on it (technically also true for the other `ARG`s, but I doubt they're going to change very often so I'm less concerned about them).

---

```diff
+    case "$(arch)" in \
+        aarch64) cur_arch='arm64' ;; \
+        x86_64)  cur_arch='x64' ;; \
+        *) echo >&2 ; echo >&2 "Unsupported architecture $(arch)" ; echo >&2 ; exit 1 ;; \
+    esac ; \
```

This is using kernel-based architecture detection, which won't work if you ever end up supporting 32bit architectures (so for those you'll want/need to do something more "clever" to do userspace detection -- for RPM-based distributions, https://github.com/docker-library/openjdk/blob/c54013c65130eb1c41268ab45a7a04e41b47c612/21/jdk/oraclelinux8/Dockerfile#L37 is the best I've currently come up with :sob:).  I doubt you're going to do that any time soon (if ever), but wanted to make sure you're aware.

---

```diff
+    curl --retry 10 -S -L --output $TEMP_DIR/opensearch-dashboards.tar.gz https://artifacts.opensearch.org/releases/bundle/opensearch-dashboards/$OSD_VERSION/opensearch-dashboards-$OSD_VERSION-linux-$cur_arch.tar.gz; \
+    curl --output $TEMP_DIR/opensearch-dashboards.pgp https://artifacts.opensearch.org/publickeys/opensearch.pgp; \
+    gpg --import $TEMP_DIR/opensearch-dashboards.pgp; \
```

As described a bit in https://github.com/docker-library/official-images#image-build, this PGP key's full fingerprint should be embedded/verified in the `Dockerfile` somehow.  That can be done via `gpg --import` + `gpg --export` with the full fingerprint, but IMO it's easier to pull from a keyserver, especially one like https://keys.openpgp.org/ which is both hyper reliable and provides some verification of the provenance of the key (in the form of email address verification).

Irrespective of which direction you choose to go, any and all invocations of `gpg` should include `--batch` (which puts `gpg` into their "API mode" vs "UX mode").

It would _also_ be ideal if you included a strong checksum (SHA256, etc) and verified that, since one verifies the artifact was published by the appropriate party and the other verifies that the artifact hasn't changed / was transferred successfully and correctly (but not strictly a blocker).

Additionally, every invocation of `curl` should include at _least_ the `-f`/`--fail` flag so that download failures propagate appropriately (without relying on side effects like PGP verification failing).

---

```diff
+# Update packages
...
+RUN yum update -y && ...
```

For official images, we generally recommend against updating libraries that are included in the respective base image. Since the item cannot be deleted from an earlier layer, the update can add extra size to the resulting image. In order to have the most up-to-date packages, official images instead rely on periodic base image updates. We then rebuild all dependent images.

`amazonlinux` images are fairly active ([PR history](https://github.com/docker-library/official-images/issues?q=label%3Alibrary%2Famazonlinux)) and so already include very up-to-date packages.

---

```diff
+# Label
+LABEL org.label-schema.schema-version="1.0" \
+  org.label-schema.name="opensearch-dashboards" \
+  org.label-schema.version="$OSD_VERSION" \
+  org.label-schema.url="https://opensearch.org" \
+  org.label-schema.vcs-url="https://github.com/opensearch-project/OpenSearch-Dashboards" \
+  org.label-schema.license="Apache-2.0" \
+  org.label-schema.vendor="Amazon"
```

> `label-schema` is deprecated; see https://github.com/label-schema/label-schema.org#readme -- some of those labels have analogs in the image-spec, some do not.

and:

---

```diff
+ENTRYPOINT ["./opensearch-dashboards-docker-entrypoint.sh"]
```

Using a relative path for this value is dangerous, as users changing `--workdir`/`-w` when they run the container will then cause it to fail (without a very compelling reason _or_ a very great error message).  I would suggest one of (in order of my own personal stylistic preference, but with no particular blocking reasons one way or another):

1. move it into `$OPENSEARCH_DASHBOARDS_HOME/bin` and drop the `./` (thus relying on `PATH`)
2. symlink into somewhere like `/usr/local/bin` (and drop the `./`, thus relying on `PATH`)
3. update `ENTRYPOINT` to use an absolute path

---

```diff
+#!/bin/bash
+
```

Not strictly _required_, but I would recommend including at least `set -e` here so that the script does not quietly continue running if any instruction in it fails.  This requires a bit more discipline while writing the script (things that might fail need to explicitly handle that failure via `if command-that-might-fail; then ...` or `if ! command-that-might-fail; then ...` or `command-that-might-fail-and-we-do-not-care || :`, etc), but it does increase robustness.  A related Bash-specific option that strengthens the protections offered by `set -e` is `set -o pipefail`.

I would also suggest `set -u` (and thus explicitly using things like `if [ -n "${CHECKING_VARIABLE_THAT_MIGHT_NOT_BE_SET:-}" ]; then variable-is-set-code; fi`), as it is extremely helpful for catching minor typos, but that one requires even more discipline, so is less strongly recommended.

---

```diff
+export OPENSEARCH_DASHBOARDS_HOME=/usr/share/opensearch-dashboards
```

This matches the value of the `ARG` set in the `Dockerfile` -- should that instead be `ENV` so that this script can just reference it directly?  (That's what I would recommend.)

---

```diff
+opensearch_dashboards_vars=(
```

This is a very long list to maintain by hand. :sweat_smile:  How often does it change?  Once you inevitably are maintaining more than one version at a time, will the lists diverge and increase your maintenance burden even further?  (Raising for visibility / as food for thought in case there are ways you can think of to help mitigate the maintenance cost on yourselves associated with this way of implementing this behavior.)

---

```diff
+function setupSecurityDashboardsPlugin {
+    SECURITY_DASHBOARDS_PLUGIN="securityDashboards"
+
+    if [ -d "$OPENSEARCH_DASHBOARDS_HOME/plugins/$SECURITY_DASHBOARDS_PLUGIN" ]; then
+        if [ "$DISABLE_SECURITY_DASHBOARDS_PLUGIN" = "true" ]; then
+            echo "Disabling OpenSearch Security Dashboards Plugin"
+            ./bin/opensearch-dashboards-plugin remove securityDashboards
+
+            # Remove all security related parameters as well as changing HTTPS to HTTP
+            # Temporary fix before security-dashboards plugin implement a parameter to disable the plugin entirely
+            # https://github.com/opensearch-project/security-dashboards-plugin/issues/896
+            UPDATED_CONFIG=`cat $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml | sed "/^opensearch_security/d" | sed "s/https/http/g"`
+            echo "$UPDATED_CONFIG" > $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml
+        fi
+    fi
+}
```

This is ... interesting (especially as the linked issue that describes why this was temporary is closed with a comment that sounds to me like it's a "wontfix").

I have a few thoughts on the specifics of the implementation here, but I'll keep them to myself for now because on a more general level I'm wondering whether you really want to have this logic at all? :grimacing:

---

```diff
+    for opensearch_dashboards_var in ${opensearch_dashboards_vars[*]}; do
+        # 'opensearch.hosts' -> 'OPENSEARCH_URL'
+        env_var=$(echo ${opensearch_dashboards_var^^} | tr . _)
```

Minor shell code improvements (if you have a proper Bash array, using `*` and re-parsing/argument separating it doesn't make a lot of sense and this `tr` can be pure-shell :sweat_smile:):

```bash
    for opensearch_dashboards_var in "${opensearch_dashboards_vars[@]}"; do
        # 'opensearch.hosts' -> 'OPENSEARCH_URL'
        env_var=${opensearch_dashboards_var^^}
        env_var=${env_var//./_}
```

---

```diff
+    # TO DO:
+    # Confirm with Mihir if this is necessary
```

:joy: maybe Mihir could chime in here? :innocent:

---

```diff
+++ b/opensearch-dashboards_latest/config/opensearch.example.org.cert
@@ -0,0 +1,20 @@
+-----BEGIN CERTIFICATE-----
...
+++ b/opensearch-dashboards_latest/config/opensearch.example.org.key
@@ -0,0 +1,28 @@
+-----BEGIN RSA PRIVATE KEY-----
```

It's a little bit weird to embed a certificate directly like this -- is this just an example?  Is it necessary for opensearch to run?  Used in the default (example?) configuration?  What's the rotation story for this credential?  How do we help make sure users don't end up running with this certificate as-is in production? :grimacing:

(Some example alternatives to this: explain to users how to create one easily in the error message when they're missing, generate fresh ones on startup with a warning, etc.)

---

```diff
+++ b/opensearch-dashboards_latest/config/opensearch_dashboards.yml
```

Does the official distribution of Dashboards (like if I downloaded a raw release straight off the website) not contain an example/default configuration?  If it does, how does that one differ from this one?  (Back to the maintainability discussion, do you really want to maintain/update over time a whole new copy of that upstream configuration/example?)

tianon:

yosifkit:

PR #13198, opened 2022-09-21
Update lightstreamer image

dario-weswit:
Add Lightstreamer version 7.3

### DIFF
--- a/library/lightstreamer
+++ b/library/lightstreamer
@@ -2,52 +2,52 @@ Maintainers: Lightstreamer Server Development Team <support@lightreamer.com> (@l
 GitRepo: https://github.com/Lightstreamer/Docker.git
 
 Tags: 6.0.3, 6.0
-Architectures: amd64
-GitCommit: eeab1ae534273b1b05c973e577a1f3eec8d58427
+Architectures: amd64, arm64v8
+GitCommit: 84e3f6588620183b48b7eb62a18070b793eff019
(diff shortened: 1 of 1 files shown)


dario-weswit:
The tests fail on the dockerfiles for the old versions, whereas the pull request is only meant to add the images for a new version (and remove a few tags from the old images).
Are we requested to update all the old images as well?

yosifkit:
```diff
-RUN apt -y update
-RUN apt -y install gnupg
+RUN apt-get -y update \
+     && apt-get -y install gnupg \
+     && rm -rf /var/lib/apt/lists/*
```
Use `apt-get` rather than `apt` because "Warning: apt does not have a stable CLI interface. Use with caution in scripts." Always combine `RUN apt-get update` with `apt-get install` in the same `RUN` statement. Also recommend to add the clean up after to help keep image size down.

>Are we requested to update all the old images as well?

If they are still in the library file, then they are called "Supported" in official images and should continue to be updated (like image fixes and version bumps). Since they are from official images that are no longer updated, they will also not get the benefit of rebuilds from base image updates. If the images are just old releases that are no longer updated, they could potentially be removed.  Removing it here will remove it from the "Supported" section on the Hub readme, but the tag will still be available to users who want it.  (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)

dario-weswit:
Thank you for the patience.

PR #13201, opened 2022-09-21
mariadb entrypoint fix

grooverdan:
I made a mistake and caused #MariaDB/mariadb-docker/issues/460

Correct fails to set password with MARIADB_ROOT_HOST=localhost,

Now includes test case.

### DIFF
--- a/library/mariadb
+++ b/library/mariadb
@@ -6,40 +6,40 @@ GitRepo: https://github.com/MariaDB/mariadb-docker.git
 
 Tags: 10.10.1-rc-jammy, 10.10-rc-jammy, 10.10.1-rc, 10.10-rc
 Architectures: amd64, arm64v8, ppc64le, s390x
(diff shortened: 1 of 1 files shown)


yosifkit:
Shouldn't https://github.com/MariaDB/mariadb-docker/blob/661429797fe3bbbb2aac6d0dc842a333c8782e5b/docker-entrypoint.sh#L353 also have the possibility of being set with `MARIADB_ROOT_PASSWORD_HASH`?

grooverdan:
quite right, fixing now. Thanks.

grooverdan:
I'm working on a solution for `/docker-entrypoint-initdb.d/` with `$MARIADB_ROOT_PASSWORD_HASH`, but I don't want to rush it too much. For now this fixes the regression and some other parts of PASSWORD_HASH.

grooverdan:
Thank you.

PR #13221, opened 2022-09-26
Update photon image

gerrit-photon:
Update photon image


yosifkit:
No worries. This seems fine as a one off. (plus it seems to only be changing a version that wasn't changed in the Friday PR)

PR #13233, opened 2022-09-27
Fix ghost test to not request "//" to prevent current unending curl connection

yosifkit:
Also prevent curl from having an unending connection by setting a `max-time`.

Fixes https://github.com/docker-library/ghost/issues/335

```console
$ ./test/run.sh ghost:5.16.2
testing ghost:5.16.2
	'utc' [1/4]...passed
	'no-hard-coded-passwords' [2/4]...passed
	'override-cmd' [3/4]...passed
	'ghost-basics' [4/4]...............passed
```

### DIFF
--- a/test/tests/ghost-basics/run.sh
+++ b/test/tests/ghost-basics/run.sh
@@ -57,7 +57,7 @@ _request() {
 	docker run --rm \
 		--link "$cid":ghost \
 		"$clientImage" \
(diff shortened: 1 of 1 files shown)


yosifkit:
`ghost:5.16.2` does reach the time-out if I revert the variable change that removes the extra `/`.

tianon (approved):

PR #13247, opened 2022-09-29
[security] Update php

tianon:
Changes:

- https://github.com/docker-library/php/commit/fa38b1c: Update 7.4 to 7.4.32


tianon:
(I think at least 8.0 and 8.1 will also be getting updated today, but I'm not certain.)

PR #13261, opened 2022-09-30
Add Debian-based variants of the clojure image

cap10morgan:
I ran into some issues using the upstream Ubuntu base of the eclipse-temurin images (namely, Ubuntu's tendency to replace dpkg packages w/ snaps that can't be easily installed in Docker images) but the temurin folks recommended this approach for those wanting a different base image. So everything in these new variants still comes from official images. Are these OK to publish?


cap10morgan:
Sorry, you can re-run CI. Our repo wasn't synced w/ our config but should be fixed now.

cap10morgan:
But of course it needs a new commit in here to actually see the update. Fix coming shortly!

yosifkit (approved):

PR #13298, opened 2022-10-06
Aerospike 6.1.0.3

mcoberly2:

### DIFF
--- a/library/aerospike
+++ b/library/aerospike
@@ -2,12 +2,12 @@ Maintainers: Lucien Volmar <lucien@aerospike.com> (@volmarl),
              Michael Coberly <mcoberly@aerospike.com> (@mcoberly2),
              Phuc Vinh <pvinh@aerospike.com> (@pvinh-spike)
 
-Tags: ee-6.1.0.2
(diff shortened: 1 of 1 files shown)


yosifkit:
>```dockerfile
>echo 'debconf debconf/frontend select Noninteractive' | debconf-set-selections \
>```

This is equivalent to `ENV DEBIAN_FRONTEND=noninteractive`, which we also don't recommend since it persists to the resulting image and so could interfere with a dependent image or an interactive user exploring the image. If it is necessary for the packages being installed, then it should be temporary, i.e. for just that `RUN`. So an `export DEBIAN_FRONTEND=noninteractive` at the beginning of the `RUN` line would work.

mcoberly2:
@yosifkit Great information regarding 
`echo 'debconf debconf/frontend select Noninteractive' | debconf-set-selections \`
and possibly using 
`export DEBIAN_FRONTEND=noninteractive` at the beginning of the RUN line.
Allow me me consult with my peers.

mcoberly2:
@yosifkit Closing.  We will submit again with the requested updates.

PR #13315, opened 2022-10-11
Update python

tianon:
Changes:

- https://github.com/docker-library/python/commit/8c92910: Update 3.7 to 3.7.15


uthark:
what about 3.8 and 3.9 updates?

PR #13345, opened 2022-10-16
Friendica 2022.10 Release

nupplaphil:
and remove some old archives as well


nupplaphil:
the error:
**2022-10-16 19:23:46 ERROR 503: Egress is over the account limit..**

:(

PR #13349, opened 2022-10-17
Gradle 6.9.3

keeganwitt:


yosifkit:
It looks like the failure is `eclipse-temurin:18` being removed (https://github.com/docker-library/official-images/pull/13222); do you want to update the PR to drop those based on `eclipse-temurin:18*` or do that in the next PR?

keeganwitt:
> It looks like the failure is `eclipse-temurin:18` being removed (#13222); do you want to update the PR to drop those based on `eclipse-temurin:18*` or do that in the next PR?

Ah. OK. I'm fine with doing that in this PR.

PR #13355, opened 2022-10-18
NeuroDebian: Ubuntu impish EOLed

yarikoptic:
But I was actually mostly interested to see if other images would regenerate as well since we updated/released neurodebian-freeze which fixes a bug and ideally want to see it shipped in all the images.


yarikoptic:
thank you @yosifkit ! With that in mind I for now just added an explicit 'apt install' to upgrade that nd_freeze where needed.

This PR I guess still pertinent since ubuntu 21.10 EOLD but feel welcome to close if in general 21.10 images are still out there and "maintained".

PR #13381, opened 2022-10-21
Releasing Neo4j 5.1

jennyowen:
It's a major neo4j release day!
Also, from now on we're only going to list the latest patch versions of neo4j as being official :+1:

### DIFF
--- a/library/neo4j
+++ b/library/neo4j
@@ -11,47 +11,26 @@ GitRepo: https://github.com/neo4j/docker-neo4j-publish.git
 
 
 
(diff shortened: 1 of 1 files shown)


tianon:
Nice, thank you! :tada:

tianon:
For curiosity, this new release wouldn't happen to be one that contains a newer apache commons-text (1.10+) to get CVE-2022-42889 off our back, would it? :eyes:

jennyowen:
@tianon sadly not.
The fix _should_ be in the upcoming 4.4 and 4.3 releases. The fixed 5.2 release won't be for maybe a month though :cry:

PR #13392, opened 2022-10-24
nginx: introduced another variant: alpine-slim.

thresheek:
This introduces a slim variant for an alpine image, lowering possible attack surfaces due to a smaller subset of packages installed to an image.

Refs:
 - https://github.com/nginxinc/docker-nginx/issues/681
 - https://github.com/nginxinc/docker-nginx-unprivileged/issues/63

### DIFF
--- a/library/nginx
+++ b/library/nginx
@@ -23,6 +23,11 @@ Architectures: arm64v8, arm32v6, arm32v7, ppc64le, s390x, i386, amd64
 GitCommit: fef51235521d1cdf8b05d8cb1378a526d2abf421
 Directory: mainline/alpine-perl
 
+Tags: 1.23.2-alpine-slim, mainline-alpine-slim, 1-alpine-slim, 1.23-alpine-slim, alpine-slim
+Architectures: arm64v8, amd64
+GitCommit: fef51235521d1cdf8b05d8cb1378a526d2abf421
+Directory: mainline/alpine-slim
+
 Tags: 1.22.1, stable, 1.22
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
 GitCommit: fef51235521d1cdf8b05d8cb1378a526d2abf421
@@ -42,3 +47,8 @@ Tags: 1.22.1-alpine-perl, stable-alpine-perl, 1.22-alpine-perl
 Architectures: arm64v8, arm32v6, arm32v7, ppc64le, s390x, i386, amd64
 GitCommit: fef51235521d1cdf8b05d8cb1378a526d2abf421
 Directory: stable/alpine-perl
+
+Tags: 1.22.1-alpine-slim, stable-alpine-slim, 1.22-alpine-slim
+Architectures: arm64v8, amd64
+GitCommit: fef51235521d1cdf8b05d8cb1378a526d2abf421
+Directory: stable/alpine-slim
--- a/library/nginx
+++ b/library/nginx
@@ -1,7 +1,7 @@
 # this file is generated via https://github.com/nginxinc/docker-nginx/blob/1dca42f99b3f032d862a1d35e8a5b951d629dc98/generate-stackbrew-library.sh
 
 Maintainers: NGINX Docker Maintainers <docker-maint@nginx.com> (@nginxinc)
-GitRepo: https://github.com/nginxinc/docker-nginx.git
+GitRepo: https://github.com/thresheek/docker-nginx.git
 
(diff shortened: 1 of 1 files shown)


tianon:
Thank you for splitting this from the version bump! :bow: :heart:

I was going to write a new comment and realized it's going to be essentially the same as what I wrote in https://github.com/nginxinc/docker-nginx/issues/681#issuecomment-1189451373 :sweat_smile:

In short, I'm not especially convinced that there are very many users that actually want Alpine for Alpine's sake, and that they're rather typically choosing Alpine for size, so I don't understand what about the use case warrants a _new_ image variant instead of making the existing `alpine` variants smaller?

Even the default "alpine variant" blurb we put in the documentation corroborates the idea that `alpine` variants are intended to be bare minimum:

https://github.com/docker-library/docs/blob/master/.template-helpers/variant-alpine.md

> To minimize image size, it's uncommon for additional related tools (such as `git` or `bash`) to be included in Alpine-based images.

Looking at the CI results, the image sizes are ~23MiB vs ~10MiB -- significant in comparison to one another, but comparatively very similar.

Do you have any idea how much of that ~13MiB difference is `curl` and `ca-certificates` vs the extra NGINX modules?

If we do go ahead with this, is there any way we could reasonably make the non-slim images `FROM` these new slim ones?

thresheek:
Thanks for taking your time to review and write out your concerns, @tianon ! Much appreciated!

My biggest gripe with making the current `alpine` images smaller by removing the modules and/or curl/ca-certificates would be breaking backwards compatibility.  But maybe I'm too cautious.

> Do you have any idea how much of that ~13MiB difference is curl and ca-certificates vs the extra NGINX modules?

removing curl and ca-certificates frees ~1.3 megabytes, 22.4 vs 23.7 in my case (aarch64).

> If we do go ahead with this, is there any way we could reasonably make the non-slim images FROM these new slim ones?

Do you have an example on how the similar is done on any other official image?

Again, thank you for your comments.

tianon:
> Thanks for taking your time to review and write out your concerns, @tianon ! Much appreciated!


:heart:

> Do you have an example on how the similar is done on any other official image?

Sure, https://github.com/docker-library/docker/blob/1231bb8e025bdc715b8a8ca122ad20b5cf7ee469/20.10/dind/Dockerfile is a pretty decent example -- I don't know whether something like this could be accomplished without too much hassle here, but it seems at least worth considering (to decrease the amount of load building these new variants).

thresheek:
Ah, so you mean I can just drop in another Dockerfile with a corresponding FROM, and the magic behind official image build process will build a graph of dependencies and provide proper builds?  Impressive, and it's definitely worth exploring.

tianon:
This looks really neat!  This will also be really interesting for users who
end up pulling multiple variants for whatever reason (alpine and
alpine-perl, for example) since their layers will explicitly overlap. 👀

I didn't see anything obviously wrong, but I think we'll get a slightly
better diff from current to new once you get the commits listed here (you
could temporarily swap "GitRepo" and "GitFetch" to point at your branch if
you wanted to just get a diff generated here before you merge it into your
mainline branch).

sambernet:
If I understand correctly, we are now just looking "for a cleaner/neater diff view" in this PR, but there is nothing actually blocking left?

I'd be glad if we can move this forward. Thanks to @thresheek the final solution even brings (initially unintended) efficiency improvements now :) 

Is there anything I could support with from our end (as a non-committer)?

thresheek:
Thanks for pinging @sambernet !

@tianon @yosifkit happy holidays! I've updated the PR to include a temporary links to a fork.

yosifkit:
I think these look great! :+1: And the sizes are the same as on the hub, so no new bugs related to cleanup or duplication.

tianon:
Agreed, looks great to me too :eyes:

sambernet:
thanks @tianon and @yosifkit for the review and inputs. I like the updated approach for the image builds and think that extra round was clearly worth the effort 👍 

@thresheek do we need any extra confirmation here?
Or is this ready to merge?

Sorry I'm not fully aware of the contributors guidelines and who can do what in this repo. But I'd be happy to see this proceed before we run into competing changes or merge conflicts that require more extra work 😇

thresheek:
Thanks for the nudge @sambernet !

We're going to release NGINX 1.23.2 next week, December 12; I think it's best to combine the parts of this PR with the version updates? So we can have 1.23.1/1.22.1 as they are now.

yosifkit:
Closing in favor of the new PR: https://github.com/docker-library/official-images/pull/13733

PR #13405, opened 2022-10-25
node: bump LTS to NodeJS 18 as scheduled

noamcohen97:
NodeJS 18 is scheduled to become LTS today
https://github.com/nodejs/release#release-schedule


tianon:
Whatever you're trying to accomplish here needs to happen in https://github.com/nodejs/docker-node (their PRs are all automated). :sweat_smile:

PR #13432, opened 2022-10-28
Adding AmazonCorretto AL2022 RC images

lutkerd:


lutkerd:
The workflow failures are due to javac not being installed on the runtime images, but passed in the full JDK images.

yosifkit:
> The workflow failures are due to javac not being installed on the runtime images, but passed in the full JDK images.

We can adjust the tests to better know that `headless` and `headful` tags don't have the compiler.

---

These package installs need the same sort of verification that the `al2` images have. For example, the `gpgcheck` provided by `yum` rather than just installing RPM's from a URL. (https://github.com/docker-library/official-images/tree/42b2aecff45bc0324066e33d1a061c187a6ca522#image-build)

Not required, but perhaps to improve shared layers across images, the sets of three could be `FROM` one another? Rough example:
- headless: `FROM amazonlinux:2022`
- headful: `FROM amazoncorretto:11-RC-al2022-headless`
- jdk: `FROM amazoncorretto:11-RC-al2022-headful`

Curious as to whether the `headful` image makes sense for general container use cases? Is it just the addition of display-related libraries (e.g. `UIManager`)?

lutkerd:
> These package installs need the same sort of verification that the `al2` images have. For example, the `gpgcheck` provided by `yum` rather than just installing RPM's from a URL. (https://github.com/docker-library/official-images/tree/42b2aecff45bc0324066e33d1a061c187a6ca522#image-build)

I will work on that and see what I can verify without creating a yum repo.

> Not required, but perhaps to improve shared layers across images, the sets of three could be `FROM` one another? Rough example:
> 
>     * headless: `FROM amazonlinux:2022`
> 
>     * headful: `FROM amazoncorretto:11-RC-al2022-headless`
> 
>     * jdk: `FROM amazoncorretto:11-RC-al2022-headful`
> 

How will that work when the images aren't published yet or built across different machines? The specific version of Corretto has to match across the images.

tianon:
Our infrastructure that builds these will build them in the correct order
-- see
https://github.com/docker-library/docker/blob/1231bb8e025bdc715b8a8ca122ad20b5cf7ee469/20.10/dind/Dockerfile
or
https://github.com/docker-library/openjdk/blob/f52dc45c090a13e29a95739522a47a001560b770/20/jdk/windows/nanoserver-1809/Dockerfile#L25
for a couple examples of that.

lutkerd:
> Our infrastructure that builds these will build them in the correct order

I'll have to see if our internal verification infrastructure can handle that as well before changing the `FROM`

lutkerd:
Any other questions comments on this or can it be merged?

yosifkit:
Apologies for the delay. I'm still looking into the best solution for the test changes, but that shouldn't be a blocker for this update.  I'd still love if they were from one another to give registries and users the cross-image space saving benefit, but it is still fine.  I think my remaining question is does the `localpkg_gpgcheck=1` still being in `/etc/dnf/dnf.conf` cause troubles for any consumers of the images?

lutkerd:
Thanks! Cleaning up the dnf.conf is a great suggestion, someone would run into an issue because of it. I'll get the changes reviews and merged and then update this PR. https://github.com/corretto/corretto-docker/pull/124

lutkerd:
Updated to latest commit ID with the dnf.conf cleanup .

tianon:
Looks like this question was missed, and it's what sticks out to me:

> Curious as to whether the `headful` image makes sense for general container use cases? Is it just the addition of display-related libraries (e.g. `UIManager`)?

(I'm not convinced that UI-related libraries are really all that common in container-based use cases?  Could this additional image instead be documentation for how to install the missing "headful" libraries?)

I'm also still very nervous about this block:

```diff
+    && VERIFIED_COUNT=$(rpm -K ${CORRETO_TEMP}/*.rpm | grep -c ": digests signatures OK$") \
+    && [ 3 -eq ${VERIFIED_COUNT} ] \
```

I think the `grep -c` and just looking for the right number of "OK signatures" lines is going to be fiddly and hard to make sure the number we're checking for is correct vs something like a small `for` loop over each `rpm` file that then verifies/installs each one in turn?

```dockerfile
    && for rpm in ${CORRETO_TEMP}/*.rpm; do \
# might need to adjust path in the grep here, depending on the output of "rpm -K"
      rpm -K "$rpm" | grep -F "$rpm: digests signatures OK" || exit 1; \
    done \
```

I'd also push again on making the images `FROM` one another, but I'd be less concerned about that if there were only the two JDK+JRE images not three separate layers of images (that are all each strictly additive).

I would also add that invocations of `curl` should definitely include at least `-f` such that they fail appropriately if a download fails (I'm not sure how that interacts with multiple URLs for a single invocation, so perhaps the `for` loop I'm proposing above should instead loop over URLs and download + verify each together? then we make absolutely sure we either download, verify, and install _all_ the artifacts we expect to, or the build fails, and decreases the chances of us accidentally pushing a partial image because some URL failed to download and/or `curl` somehow downloaded more than one thing and thus the `grep -c` still returned the "correct" number of results)

lutkerd:
Thanks for the feedback.

>(I'm not convinced that UI-related libraries are really all that common in container-based use cases? Could this additional image instead be documentation for how to install the missing "headful" libraries?)

There are some libraries like Apache XML Graphics Commons which are used by some of our customers that require the headful install for the SVG and PDF rendering capabilities, it is definitely not a majority of users who need this, but it could be a breaking change for some customers.

I will make the changes to use a for loop over the list of RPMs, I think that makes sense and would simplify the checks a bit between the different images. But, I am not going to make changes to the images inherit from one another now as other internal infrastructure that consumes this code doesn't handle it yet, I will open an issue to make sure we track it.

PR #13443, opened 2022-10-31
Update R to 4.2.2 released this morning

eddelbuettel:
Standard update of r-base to the new upstream release 4.2.2 made this morning using the updated Debian binaries

One change of `addgroup` to `adduser` for same effect, no other changes

### DIFF
--- a/library/r-base
+++ b/library/r-base
@@ -2,8 +2,8 @@ Maintainers: Carl Boettiger <rocker-maintainers@eddelbuettel.com> (@cboettig),
              Dirk Eddelbuettel <rocker-maintainers@eddelbuettel.com> (@eddelbuettel)
 GitRepo: https://github.com/rocker-org/rocker.git
 
-Tags: 4.2.1, latest
+Tags: 4.2.2, latest
 Architectures: amd64, arm64v8, ppc64le, s390x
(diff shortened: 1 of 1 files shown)


yosifkit:
>```diff
>+        && apt-get upgrade -y \
>```

We recommend against using blanket package upgrades (`apt-get upgrade`/`apk upgrade`/`yum upgrade`/`yum update`) for official-images. When package upgrades are applied in a dependent image, it duplicates content of the base image, making the image larger than necessary. It also only delays the inevitable "there are outdated packages". The Official Images build pipeline makes heavy use of `docker build` cache, so we make periodic base image updates to then fully rebuild all dependent images (e.g. the [Debian image updates](https://github.com/docker-library/official-images/pulls?q=is%3Apr+label%3Alibrary%2Fdebian)).

eddelbuettel:
Sorry!  I believe you pointed that out to me before. I had it in one of the two and not the other.  I'll make a fix, mark it and hopefully will not forget next time.  PR will likely be updated a in 20 or 30 mins.

eddelbuettel:
@yosifkit -- Two containers (`rocker/r-base:{4.2.2,latest}`) built and pushed, change commited.  PR here updated and force-pushed with the updated sha1.  This should take care of it.

PR #13473, opened 2022-11-03
[security] Update php

tianon:
Changes:

- https://github.com/docker-library/php/commit/b9f1715: Update 7.4 to 7.4.33


tianon:
(GHA hiccup)

PR #13481, opened 2022-11-05
Bump Haskell GHC to 9.4.3

AlistairB:
https://www.haskell.org/ghc/download_ghc_9_4_3.html

Thanks


AlistairB:
Hmm seems like a random timeout failure. If the check could be re-run?

AlistairB:
Nevermind I force pushed to get it to re-run.

PR #13492, opened 2022-11-08
mariadb release Q4 for < 10.10

grooverdan:
10.10 and 10.11 are still coming.

For the moment 10.10 is a partially updated image.


yosifkit:
If you move any temporary packages added on lines [25-31](https://github.com/MariaDB/mariadb-docker/blob/9bfaada593ad99c33a497a1bb84ccfd44bbd3977/Dockerfile.template#L25-L31) to the `apt-get install` on line 33 (i.e. after the `savedAptMark` variable from line 32), then you will not need an extra `apt-get remove`, since it would happen automatically by the apt-mark auto/manual & auto-remove parts.

grooverdan:
Thanks @fauust for the fix.

PR #13499, opened 2022-11-09
kong: 3.0.1

hutchic:


library/kong
```diff
@@ -1,13 +1,25 @@
...
+Architectures: amd64, arm64v8
+
+Tags: 3.0.0-alpine, 3.0.0
 GitCommit: 5a2610d900717a4a6c8ee98bc421d0d4599b126f
 GitFetch: refs/tags/3.0.0
 Directory: alpine
 Architectures: amd64, arm64v8
 
-Tags: 3.0.0-ubuntu, 3.0-ubuntu, ubuntu
+Tags: 3.0.0-ubuntu
```
yosifkit, line 22: Any particular reason to keep the `3.0.0` versions listed here?  Removing it here will remove it from the "Supported" section on the Hub readme (and will prevent us from spending cycles rebuilding it on the official build server), but the tag will still be available to users who want it.  (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)

hutchic: No probably not :thinking:

library/kong
```diff
@@ -1,13 +1,25 @@
...
+File: Dockerfile.apk
+Architectures: amd64, arm64v8
+
+Tags: 3.0.1-ubuntu, 3.0-ubuntu, ubuntu
+GitCommit: b2057cd3432726c2122873a3740e31a68a03cecd
+GitFetch: refs/tags/3.0.1
+Directory: ubuntu
+Architectures: amd64, arm64v8
+
+Tags: 3.0.0-alpine, 3.0.0
```
hutchic, line 16:
```suggestion
Tags: 3.0.0-alpine
```
Is this what you were thinking @yosifkit ? I think the reason I left it behind is we always have from the looks of it ...

yosifkit: `3.0.0-alpine` and `3.0.0-ubuntu` are the images and tags I was referring to (though I now noticed that there are `2.8.1*` versions to which this comment would also apply). We recommend that the tags kept in the library file are the set of tags that are supported by the upstream (i.e. a new version bump would be released to fix bugs, CVEs, etc). So, when I see a version with a new release that is just patch releases newer (e.g. `3.0.1`), then I usually assume that the older release (`3.0.0`) is no longer supported. From what I've seen, semver based projects don't release sub-patch version bumps and usually focus effort to a limited set of `{major}.{minor}` branches with the latest patch of each `{major}.{minor}` series being the supported version.

hutchic: OK. Reduced down to the most recent patch versions we support so

Instead of 3.0.0, and 3.0.1 just 3.0.1 as well as the 2.8

PR #13526, opened 2022-11-14
Update python

yosifkit:
Changes:

- https://github.com/docker-library/python/commit/7ee8404: Update 3.12-rc to 3.12.0a2, pip 22.3.1
- https://github.com/docker-library/python/commit/7a895a1: Update 3.9
- https://github.com/docker-library/python/commit/d24131c: Update 3.8
- https://github.com/docker-library/python/commit/bd5f913: Update 3.7
- https://github.com/docker-library/python/commit/429706d: Update 3.12-rc
- https://github.com/docker-library/python/commit/902e9de: Update 3.11
- https://github.com/docker-library/python/commit/770cecb: Update 3.10

### DIFF
--- a/library/python
+++ b/library/python
@@ -4,230 +4,230 @@ Maintainers: Tianon Gravi <admwiggin@gmail.com> (@tianon),
              Joseph Ferguson <yosifkit@gmail.com> (@yosifkit)
 GitRepo: https://github.com/docker-library/python.git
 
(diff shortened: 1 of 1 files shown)


tianon:
I usually skip `get-pip.py`-only updates because it's typically just a new version of PIP embedded in the script for us to use to install the old version of PIP, so actual change to the image is low. :innocent:

PR #13529, opened 2022-11-14
Update Debian (20221114)

tianon:


tianon:
(Minor mistake, corrected in https://github.com/docker-library/official-images/pull/13530)

PR #13533, opened 2022-11-15
eclipse-temurin: add UBI minimal support

gdams:
@Tianon you'll need to add UBI as an approved base OS to satisfy the naughty PR check


tianon:
Hey!  Sorry for the delay :bow:

We've been working on a couple tooling changes to allow us to deal with external dependencies more correctly, especially for rebuilds/triggering in our infrastructure (https://github.com/docker-library/bashbrew/pull/55, https://github.com/docker-library/bashbrew/pull/56).

For now, I think we'll just accept this as-is since it will build just fine, and then we'll need to make sure we deal with the other half of the problem (rebuilding appropriately) separately. :+1:

PR #13546, opened 2022-11-16
Aerospike 6.1.0.4

mcoberly2:

### DIFF
--- a/library/aerospike
+++ b/library/aerospike
@@ -6,9 +6,9 @@ Maintainers: Lucien Volmar <lucien@aerospike.com> (@volmarl),
 Tags: ee-6.1.0.4
 Architectures: amd64
 GitRepo: https://github.com/aerospike/aerospike-server-enterprise.docker.git
-GitCommit: e4713117b8dec3635be5cb4c27619e18433cd6e5
+GitCommit: 0f6f634a5d7b14b76fb4b4ddabf714c9d4bf346f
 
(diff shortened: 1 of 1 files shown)


yosifkit:
From the test failures:
- referenced commits are not in the respective `master` branches of `github.com/aerospike/aerospike-server.docker` and `github.com/aerospike/aerospike-server-enterprise.docker`
    ```console
     - aerospike:
       - commit 9b5c7deb452ce646f2b6f064fe00225f5ac245de:
         - is not in the specified ref GitFetch: refs/heads/master
       - commit d1e8610192813c410f605cb86f392f8155739daa:
         - is not in the specified ref GitFetch: refs/heads/master
    ```
- `entrypoint.sh` missing for `ce-6.1.0.4`
    ```console
    Step 8/11 : COPY entrypoint.sh /entrypoint.sh
    COPY failed: file not found in build context or excluded by .dockerignore: stat entrypoint.sh: file does not exist
    ```

---

As for any Dockerfile changes that I noticed:  In official images, we do not recommend using such a specific `debian` image tag (`debian:bullseye-20221114-slim`); the `debain:bullseye-slim` tag is the correct one. The more specific tag would be out of scope as soon as the next Debian update.

Background: 
>

We update the Debian base images often ([Debian update PRs](https://github.com/docker-library/official-images/pulls?q=is%3Apr+label%3Alibrary%2Fdebian)):

>

PR #13550, opened 2022-11-16
Update clojure tools-deps variants to version 1.11.1.1200

cap10morgan:
Upstream version bump

### DIFF
--- a/library/clojure
+++ b/library/clojure
@@ -2,15 +2,11 @@ Maintainers: Paul Lam <paul@quantisan.com> (@Quantisan),
              Wes Morgan <wes@wesmorgan.me> (@cap10morgan)
 Architectures: arm64v8, amd64
 GitRepo: https://github.com/Quantisan/docker-clojure.git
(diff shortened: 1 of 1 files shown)


cap10morgan:
I think the CI failure was an intermittent connection issue. Hopefully trying again would fix it.

yosifkit:
I can reproduce the failure locally. After some digging to make sure it wasn't a change in Alpine, it seems to be related to the jdk bump of Eclipse temurin `8u345-b01` to Eclipse temurin `8u352-b08`.

cap10morgan:
> I can reproduce the failure locally. After some digging to make sure it wasn't a change in Alpine, it seems to be related to the jdk bump of Eclipse temurin `8u345-b01` to Eclipse temurin `8u352-b08`.

Huh, OK. I'll dig into it.

cap10morgan:
@yosifkit I think (if it's alright with you) I'm just going to exclude that combo for now (i.e. boot, JDK 8, Alpine). I would bet that its usage is very low to nonexistent.

yosifkit:
Yeah, that's fine.

PR #13567, opened 2022-11-19
Add explicit "external-pins"

### LINKED ISSUE #9445: Adding the Red Hat Universal Base Image as an official image

tianon:
The goal of this directory is two-fold: to list the explicitly allowable external dependencies and to pin those external dependencies to a specific content-addressable digest such that we can update them in the same way we do everything else (and then trigger rebuilds properly based on them, too).

Closes #9445

(This is possible thanks to the tooling updates in https://github.com/docker-library/bashbrew/pull/55 and https://github.com/docker-library/bashbrew/pull/56 :metal:)


yosifkit (approved):

fatherlinux:
@yosifkit I think the docs still need updated? It only shows scratch, windows core, and windows server as the exceptions

tianon:
Yeah, the docs do still need to be updated, but we also need to revert this and take a different filenaming approach thanks to Windows (which I tested, but not via `git clone`, only via `touch` that comes with Git Bash and I wrongly assumed that Git itself would follow the same rules and make the same smart conversions that the MSYS `touch` did :facepalm:).

https://github.com/docker-library/official-images/issues/13608 :disappointed:

PR #13569, opened 2022-11-20
Update rocky linux images for 9.1 and 8.7

NeilHanlon:


library/rockylinux
```diff
@@ -2,18 +2,18 @@ Maintainers: The Rocky Linux Foundation <infrastructure@rockylinux.org> (@rocky-
...
-arm64v8-GitFetch: refs/heads/Rocky-9.0.20220720-Minimal-aarch64
-arm64v8-GitCommit: bbea3587002f91144f3195b8b0372f05257fd6f5
+Tags: 9.1.20221119-minimal, 9-minimal, 9.1-minimal
+GitFetch: refs/heads/Rocky-9.1.20221119-Minimal-x86_64
+GitCommit: ec9fb969a767fe74591f9ce39e09b44659a56cf4
+arm64v8-GitFetch: refs/heads/Rocky-9.1.20221119-Minimal-aarch64
+arm64v8-GitCommit: 624fe1e35788615384eaf94345cc8dcd3919e594
 Architectures: amd64, arm64v8
 
 Tags: 8.6.20227707, 8.6, 8
```
tianon, line 19: I think you probably meant to drop/update this entry instead of just adding a new one?  (See the PR CI failures :eyes:)

NeilHanlon: I think my latest commit should address this.

My intention is to: build 8.7 and 9.1 images, and move the respective 8 & 9 tags to point to the newer ones.

NeilHanlon:
Hrm. good catch let me evaluate that difference.

NeilHanlon:
@tianon - the new 8.7 images I pushed there should be fixed in terms of their size for Minimal.. was a stupid mistake I made.

tianon:
Nice, looks good!  It looks like you managed to make it even slimmer :smile:

| Image       | Before | After |
|-------------|--------|-------|
| `8`         | ~196M  | ~197M |
| `8-minimal` | ~95M   | ~93M |
| `9`         | ~199M  | ~161M |
| `9-minimal` | ~129M  | ~95M  |

PR #13592, opened 2022-11-25
Add official image for KrakenD Api Gateway

taik0:
Signed-off-by: Daniel Ortiz <dortiz@krakend.io>

KrakenD is a stateless Open Source API Gateway built on top of the Lura Framework (formerly KrakenD Framework).
To know more about KrakenD, please read our [documentation](https://www.krakend.io/docs/)

https://github.com/krakendio/krakend-ce

# Checklist for Review

     - Apache 2.0: https://github.com/krakendio/krakend-ce/blob/master/LICENSE
    - https://github.com/docker-library/docs/pull/2082


taik0:
Documentation PR: https://github.com/docker-library/docs/pull/2082

taik0:
I didn't realize that that repository used the `main` branch instead of `master`. I pushed a `master` branch, so we don't have to change anything on this PR.
A re-run should work now.

taik0:
Hi, there's something in my hand that can be done to unblock this PR?

Thank you!

yosifkit:
Hello! :sparkles:

JAORMX:
Hey folks! As a krakend user we'd really benefit from getting this PR accepted. The project is growing fast and this would be very beneficial not only for the project itself but for the community that's growing around it.

taik0:
Hi team,

Are new images still blocked?

taik0:
Hi @yosifkit 
Is something wrong on our side? There hasn't been any response in almost 1.5 years, are new docker images still being accepted?

LaurentGoderre:
I opened a PR to optimize the image: https://github.com/krakend/docker-library/pull/4

Could you also open a docs PR as explained here: https://github.com/docker-library/docs/blob/master/README.md

LaurentGoderre (approved):

LaurentGoderre:
Small comments on the docs

alombarte:
Comments on the docs are addressed. Thank you very much for your review!

alombarte:
Hello @yosifkit. I see that @LaurentGoderre approved the PR almost one year ago. What are the chances of the image being added as an official image? Shall we continue updating this PR after two years?

yosifkit:
Apologies for the long delay. 🙇

I had a few notes typed up but forgot to flesh them out and post them. 🤦 So here they are!

Blockers:
- `gpgconf --kill all` and `rm -rf "$GNUPGHOME"` needs to be added somewhere after the `gpg --verify` (but within the `RUN` line)
- these two images are `FROM alpine:3.18` which is end of life in just two months (https://alpinelinux.org/releases/); but it looks like the Dockerfiles for newer `krackend` versions in https://github.com/krakend/docker-library are already updated. So the ask, do you want to stick with these older `krackend` releases and replace them in a follow up PR or just update this PR with the latest supported release(s)?
- the user conditional should be swapped so that any non-root user can work (e.g., `docker run --user 1234:5678 ...`) and so that the `su-exec` is only run as `root` since that is the only time it can work; it could also fall through to the `exec "$@"` at the end:
  ```diff 
  -    if [ "$(id -u)" = 1000 ]; then
  -        exec "$@"    
  -    else
  +    if [ "$(id -u)" = 0 ]; then
           # use su-exec to drop to a non-root user
           exec su-exec krakend "$@"
  +    else
  +        exec "$@"
       fi

  # alternative

  -    if [ "$(id -u)" = 1000 ]; then
  -        exec "$@"    
  -    else
  -        # use su-exec to drop to a non-root user
  -        exec su-exec krakend "$@"
  +    if [ "$(id -u)" = 0 ]; then
  +        set -- su-exec krakend "$@"
       fi
  ```
- the `sha256sum` should be done before the `tar` extraction

Suggestions:
- the `tar xvf` should avoid "traditional" flags and just use Unix and GNU style flags: https://manpages.debian.org/bookworm/tar/tar.1.en.html
  ```diff
  -    tar xzf krakend.tar.gz -C / --strip-components 1; \
  +    tar -xzf krakend.tar.gz -C / --strip-components 1; \
  ```
- perhaps the entrypoint could be added to the `PATH` (like `/usr/local/bin/`)?
  ```dockerfile
  ENTRYPOINT [ "docker-entrypoint.sh" ]
  ```
- :thinking: is the `-c /etc/krakend/krakend.json` in the default `CMD` important for users adding their own `run` flags or perhaps already the default config file location such that the flag is unnecessary?

taik0:
Hi @yosifkit 

Thank you for the review! I've addressed all the required changes and updated the KrakenD versions to the latest (which include Alpine 3.21). 
Regarding the `CMD`, this has been the default in all KrakenD history, and users expect that running the container without arguments will start KrakenD with those settings (the software does not enforce a default config file, but that parameters are widely used as a default).

Thank you!

yosifkit (approved):

tianon:
To be explicit/clear, I don't think anything below is an outright blocker; mostly food-for-thought and suggestion-for-improvement. :+1:

---

```diff
+Tags: 2.9.2
+Architectures: amd64, arm64v8
+GitCommit: a84f9f39fb6dc30354dee6a3ecd35666791aeb9e
+Directory: 2.9.2
```

This is only here to backfill the version, and it'll be removed after it's built/pushed, right?
(It's not "supported" in the sense that there might be a "2.9.2.1" release if there were a bug found in it, for example.)

---

It doesn't affect your usage of `su-exec`, but you probably want to be aware of https://github.com/ncopa/su-exec/pull/26 (and if you end up needing more `su-exec` in the future, I would strongly suggest finding alternatives).

---

```diff
+commandRun="run"
+commandVersion="version"
+commandCheck="check"
+commandPlugin="check-plugin"
+commandHelp="help"
+commandValidate="validate"
+commandAudit="audit"
+
+# this if will check if the first argument is a flag
+# but only works if all arguments require a hyphenated flag
+# -v; -SL; -f arg; Also check if the first argument is any
+# of the KrakenD commands
+if [ "$#" -eq 0 ] || [ "${1#-}" != "$1" ] ||
+   [ "$1" = "${commandRun}" ] || 
+   [ "$1" = "${commandVersion}" ] || 
+   [ "$1" = "${commandCheck}" ] || 
+   [ "$1" = "${commandPlugin}" ] ||
+   [ "$1" = "${commandValidate}" ] ||
+   [ "$1" = "${commandAudit}" ] ||
+   [ "$1" = "${commandHelp}" ]; then
+    set -- krakend "$@"
+fi
```

Just as a suggestion, this might be easier to maintain/understand as a `case` statement, something like this:

```sh
#!/bin/sh
set -eu

# this if will check if the first argument is nonexistent or a flag
# but only works if all arguments require a hyphenated flag
# -v; -SL; -f arg; Also check if the first argument is any
# of the KrakenD commands
case "${1:-}" in
	''       | \
	-*       | \
	run      | \
	version  | \
	check    | \
	plugin   | \
	validate | \
	audit    | \
	help     )
		set -- krakend "$@"
		;;
esac

# ...
```

---

```diff
+CMD [ "krakend", "run", "-c", "/etc/krakend/krakend.json" ]
```

When I read this line in the proposed docs:

> The configuration files are taken from the current directory (`$PWD`). Therefore, all examples expect to find at least the file `krakend.json`.

I get the sense that if I supply `--workdir`/`-w` to my container, that it will read `krakend.json` from the new directory, not from `/etc/krakend` anymore, but because this default command uses a full path, that's not actually true (and if I keep reading, it's clear that `$PWD` there was referring to the host, not the container).

I'd suggest either updating the documentation to clarify this, or perhaps better (and the solution I'd suggest) is removing the full path from `CMD` (and perhaps from all those examples, if you are willing to commit to `/etc/krakend` being the `WORKDIR` long-term, which doesn't seem like a stretch given your other replies here :smile:):

```dockerfile
CMD [ "krakend", "run", "-c", "krakend.json" ]
```

(this then also makes it easier to reproduce for users who want to run the original command but with some extra flags/arguments)

taik0:
Let me address those recommendations before merging since I find them pretty good improvements. Thank you!

taik0:
Hi,

I've changed the entrypoint to use the `case` you recommended (really helps adding new commands).
I also changed the `CMD` to remove the `/etc/krakend` from the config file.

Regarding having `2.9.2`, I removed it. We always recommend updating to the latest version (we never remove functionality), and since the latest version (`2.9.3`) has security fixes, it wouldn't be right to publish an insecure version.

In future updates, we will try to address the change of `su-exec` for an alternative.

Thank you!

tianon (approved):

PR #13593, opened 2022-11-25
Releasing neo4j 4.4.15 and 4.3.22

jennyowen:
Now that we're only supporting 3 images, would it be possible to get the supported tags listed on https://hub.docker.com/_/neo4j?tab=description again?

Thanks.

### DIFF
--- a/library/neo4j
+++ b/library/neo4j
@@ -22,22 +22,22 @@ GitCommit: 51c1670004a4c2d511a342a01a7963b78cc2c5c4
 Directory: 5.2.0/enterprise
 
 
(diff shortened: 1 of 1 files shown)


tianon:
> Now that we're only supporting 3 images, would it be possible to get the supported tags listed on https://hub.docker.com/_/neo4j?tab=description again?

Sorry, missed this when I did this merge (in my defense, it was a "quick merge spree while on PTO" :smile:).

I'll get a PR up shortly to remove that special case. :+1:

jennyowen:
Thanks @tianon you're a star!

PR #13609, opened 2022-11-29
update emqx to 4.3.22, 4.4.11 and add 5.0.11

id:


yosifkit (approved):

tianon:
I would recommend pushing that `COPY` back down as far as it reasonably can be (ideally below that last `RUN` at least) since `COPY` has weird cache behavior sometimes, but I think that's probably a pretty harmless `RUN` if it ends up getting repeated unnecessarily. :+1:

tianon:
Ah crap, just noticed that `RUN` is _not_ harmless, and hasn't been -- these lines at least need to move up into the line that creates most of `/opt/emqx` or it's unnecessarily duplicating the installation:

```dockerfile
     chgrp -Rf emqx /opt/emqx; \
     chmod -Rf g+w /opt/emqx; \
     chown -Rf emqx /opt/emqx
```

Rory-Z:
Hi, @tianon Thanks a lot for your advice, could you please review it for me: https://github.com/emqx/emqx-docker/pull/16/files ?

PR #13611, opened 2022-11-29
Telegraf v1.24.4

powersj:
Update for big fix release and removal of maintainers due to layoff :(


popey:
Getting tagged in this is amusingly ironic ;)

tianon:
Yeah, this particular case is very unfortunate and I'm sorry for the noise :disappointed: :heart: :bow:

In general I think it does make sense to ping existing maintainers on maintainer changes for visibility/awareness, though; for example, imagine a hostile change to a similar-looking repo that still looks pretty legitimate -- pinging existing maintainers makes sure that they see it and can comment that it's not legit if we somehow miss it :see_no_evil: :heart:

Issue #13608, opened 2022-11-29
Git checkout fails on Windows


keeganwitt:
Cloning (or pulling) the repository now fails.

```
$ git clone https://github.com/docker-library/official-images.git
Cloning into 'official-images'...
remote: Enumerating objects: 93744, done.
remote: Counting objects: 100% (228/228), done.
remote: Compressing objects: 100% (128/128), done.
remote: Total 93744 (delta 132), reused 183 (delta 100), pack-reused 93516
Receiving objects: 100% (93744/93744), 21.82 MiB | 15.65 MiB/s, done.
Resolving deltas: 100% (64747/64747), done.
error: invalid path '.external-pins/mcr.microsoft.com/windows/nanoserver:1809'
fatal: unable to checkout working tree
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry with 'git restore --source=HEAD :/'
```

The Windows file naming [convention](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions) states that the following are reserved characters
* `<` (less than)
* `>` (greater than)
* `:` (colon)
* `"` (double quote)
* `/` (forward slash)
* `\` (backslash)
* `|` (vertical bar or pipe)
* `?` (question mark)
* `*` (asterisk)

So in this case, the issue is with the colon character. Is it possible to rename those files (nanoserver:1809, nanoserver:ltsc2022, servercore:1809, servercore:ltsc2022)?

tianon:
Doh, yep -- as noted in https://github.com/docker-library/official-images/pull/13567#issuecomment-1331205266, I tested this, but only with MSYS `touch` that comes with Git for Windows, so I (wrongly) assumed Git would make the same intelligent conversions automatically that MSYS does.

PR #13613, opened 2022-11-29
Revert "Add explicit "external-pins""

tianon:
Reverts docker-library/official-images#13567
Closes https://github.com/docker-library/official-images/issues/13608
Opens https://github.com/docker-library/official-images/issues/9445


tianon:
(Merged to fix immediate breakage)

PR #13614, opened 2022-11-29
Add explicit "external-pins", take two

tianon:
This is a second attempt at https://github.com/docker-library/official-images/pull/13567:

> The goal of this directory is two-fold: to list the explicitly allowable external dependencies and to pin those external dependencies to a specific content-addressable digest such that we can update them in the same way we do everything else (and then trigger rebuilds properly based on them, too).

See `.external-pins/list.sh` for a useful summary of the assumptions that went into the naming scheme (see also https://github.com/docker-library/official-images/issues/13608).

Closes https://github.com/docker-library/official-images/issues/9445 (mostly; there are still some tooling/pipeline updates necessary, but this gets us back to a passing CI check on images that are `FROM redhat/ubi9-minimal`)


tianon:
Windows tests are passing!! :eyes:

.external-pins/list.sh
```diff
@@ -0,0 +1,18 @@
...
+dir="$(dirname "$BASH_SOURCE")"
+
+find "$dir" -mindepth 2 -type f -printf '%P\n' | sed -e 's/___/:/' | sort
+
+# assumptions which make the "___" -> ":" conversion ~safe (examples referencing "example.com/foo/bar:baz"):
+#
+#   1. we *always* specify a tag ("baz")
+#   2. the domain ("example.com") cannot contain underscores
+#   3. the repository ("foo/bar") can only contain singular or double underscores (never triple underscore), and only between alphanumerics (thus never right up next to ":")
+#   4. we do *not* use the "g" regex modifier in our sed, which means only the first instance of triple underscore is replaced (in pure Bash, that's "${img/:/___}" or "${img/___/:}" depending on the conversion direction)
```
yosifkit, line 13: And that we will never pin to a registry that runs on an alternate port: `example.com:5000`

tianon: Oops, great catch -- added an additional assumption to the file:

> `we do not pin to any registry with a non-443 port ("example.com:8443")`

yosifkit (approved):

keeganwitt:
Confirming that, as you'd expect, this doesn't break on Windows.

PR #13622, opened 2022-11-30
Update php

tianon:
Changes:

- https://github.com/docker-library/php/commit/d461611: Add Alpine 3.17 variant (https://github.com/docker-library/php/pull/1348)
- https://github.com/docker-library/php/commit/2a7223d: Merge pull request https://github.com/docker-library/php/pull/1349 from jnoordsij/remove-7.4

### DIFF
--- a/library/php
+++ b/library/php
@@ -1,4 +1,4 @@
-# this file is generated via https://github.com/docker-library/php/blob/f13631afb18ffbb187740cb0e237314f0733e4a1/generate-stackbrew-library.sh
+# this file is generated via https://github.com/docker-library/php/blob/f2999de80bd83a2ab1e0f4b14208deff6e08936f/generate-stackbrew-library.sh
 
 Maintainers: Tianon Gravi <admwiggin@gmail.com> (@tianon),
              Joseph Ferguson <yosifkit@gmail.com> (@yosifkit)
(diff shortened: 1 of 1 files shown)


tianon:
FYI to affected maintainers still on PHP 7.4 -- it is now EOL and affected versions/variants should either be updated or removed:

- `adminer`: @TimWolla
- `backdrop`: @pirog @serundeputy @jenlampton @klonos
- `drupal`: @tianon @yosifkit
- `friendica`: @friendica @nupplaphil
- `joomla`: @Llewellynvdm @HLeithner
- `mediawiki`: @davidbarratt @legoktm @addshore
- `postfixadmin`: @DavidGoodwin
- `wordpress`: @tianon @yosifkit

TimWolla:
> `adminer`: @TimWolla

Yeah, it's literally all of them. I've previously attempted to upgrade Adminer to 8.0, but needed to revert in https://github.com/TimWolla/docker-adminer/commit/0f6d3781a93e274faaee2020b110b7bf701dc5da, as it isn't cleanly compatible and still isn't to the best of my knowledge. Last upstream commit and last upstream release both were in May 2021.

Not sure what the best course of action is, here. Either I revert the revert of the upgrade to PHP 8.0 (introducing bugs into some of the features) or it would need to be removed entirely (which would be sad, because I personally still use it almost daily).

tianon:
That's unfortunate :disappointed:

From https://www.adminer.org/#requirements, they do _claim_ to support PHP 8, but yeah, it doesn't seem super actively maintained. :sob:

There are several related bugs in their tracker (https://sourceforge.net/p/adminer/bugs-and-features/ :see_no_evil:) that don't have any official response. :confounded:

I don't think there are any Linux distributions committing to longer-term support for PHP 7.4, right?  For the purposes of official images, I think our options are probably either PHP 8+ or image deprecation. 😞

TimWolla:
> they do _claim_ to support PHP 8

It's … complicated. The issue with PHP 8 is newly added Warnings that make it into SQL dumps (which completely break the dump of course). I could technically just slap a `php.ini` configuration onto there to suppress those Warnings, but that might paper over some *real* issues and a database management tool is not really something where personally I'd like to see undefined behavior. In fact my smoke tests explicitly contain a check for Warnings to catch the most obvious issues: https://github.com/TimWolla/docker-adminer/blob/4f3e355425bffd35db7bc0487eb3166d94316578/.github/workflows/image.yml#L53

It's also not likely to get better with future PHP versions, every PHP version adds new warnings and often also introduces small incompatibilities (it's unavoidable with the huge API surface PHP exposes in it's standard library).

> There are several related bugs in their tracker (https://sourceforge.net/p/adminer/bugs-and-features/ see_no_evil) that don't have any official response. confounded

There's even pull requests, such as: https://github.com/vrana/adminer/pull/450 and https://github.com/vrana/adminer/pull/429

> I don't think there are any Linux distributions committing to longer-term support for PHP 7.4, right?

Debian Bullseye (the current stable) includes PHP 7.4 and I believe that PHP will continue to receive security patches as any other package in Debian. In fact there's also an Adminer package in Debian (https://tracker.debian.org/pkg/adminer), even in the current testing which even is PHP 8.1.

tianon:
> Debian Bullseye (the current stable) includes PHP 7.4 and I believe that PHP will continue to receive security patches as any other package in Debian. In fact there's also an Adminer package in Debian (https://tracker.debian.org/pkg/adminer), even in the current testing which even is PHP 8.1.

Yeah, that's interesting -- I'm surprised there aren't patches applied (https://salsa.debian.org/debian/adminer/-/tree/debian/sid/debian currently has no `patches/` subfolder).

If you wanted to switch to Bullseye's PHP 7.4 packages, that seems reasonable (although slightly more work for you in the initial conversion).  You'll want to keep an eye on https://packages.debian.org/bullseye/debian-security-support / https://salsa.debian.org/debian/debian-security-support/-/blob/bullseye/security-support-ended.deb11 to make sure it stays supported, but it's probably somewhat safe (certainly safer than the images we won't be maintaining anymore :no_mouth:).

TimWolla:
> If you wanted to switch to Bullseye's PHP 7.4 packages, that seems reasonable (although slightly more work for you in the initial conversion).

Okay, switching to the native Bullseye packages sounds reasonable to me as well. I'll put this somewhere near the top of my ToDo list.



> but it's probably somewhat safe (certainly safer than the images we won't be maintaining anymore no_mouth).

Generally I would consider Adminer to be less affected than other PHP images, as it's a development tool. If there’s an admin interface to your production database exposed on the Internet then that’s already a security issue by default.

PR #13625, opened 2022-11-30
MariaDB update - work around for old-mode=""

grooverdan:
A upstream bug report https://jira.mariadb.org/browse/MDEV-30126 is causing troubles with utf8mb4 testing.


grooverdan:
Thank you.

PR #13665, opened 2022-12-06
Aerospike 6.2.0.1

mcoberly2:

### DIFF
--- a/library/aerospike
+++ b/library/aerospike
@@ -1,13 +1,16 @@
 Maintainers: Lucien Volmar <lucien@aerospike.com> (@volmarl),
              Michael Coberly <mcoberly@aerospike.com> (@mcoberly2),
-             Phuc Vinh <pvinh@aerospike.com> (@pvinh-spike)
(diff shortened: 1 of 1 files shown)


mcoberly2:
Would you please also delete the "latest" build from the tag listing: https://hub.docker.com/_/aerospike/tags

yosifkit:
Yeah, we have deleted the `latest` tag from Docker Hub.

---

Unfortunately the new `bootstrap.sh` falls under this section of our `README`:

> It may be tempting, for the sake of brevity, to put complicated initialization details into a standalone script and merely add a `RUN` command in the `Dockerfile`. However, this causes the resulting `Dockerfile` to be overly opaque, and such `Dockerfile`s are unlikely to pass review. Instead, it is recommended to put all the commands for initialization into the `Dockerfile` as appropriate `RUN` or `ENV` command combinations.
>
> \-https://github.com/docker-library/official-images/tree/6f8c8066b90e4f81d999d088f4081cc1723b2387#clarity

mcoberly2:
Closing, will replace with Version 6.2.0.2 shortly

PR #13667, opened 2022-12-06
Pulsar 2.10.2

peiit:


yosifkit:
Hello! :sparkles:

tianon:
My first question would be whether you've been in contact with Apache Pulsar upstream to see if they're either interested in being involved, maintaining the image, providing their blessing for your ownership/maintainership of it, or would even rather there weren't an official image for Apache Pulsar at all (which has happened before)? :pray: :sweat_smile:

PR #13678, opened 2022-12-07
Updated swipl devel to 9.1.1

JanWielemaker:
Note that I updated the stable and devel versions in two PRs.   Is it ok to do so in a single PR when both are updated next time?


tianon:
A single PR is actually preferred because the file is so short that your other PR will often have conflicts if we merge this one. :smile:

(It's then less overhead for us to review both changes together, especially since they're both pretty basic version bumps.)

PR #13679, opened 2022-12-07
Update to most recent upstream commit

dkadioglu:
The last couple of recently built images are behind upstream. It would be great, if the missing commits could be included. This would i.a. fix https://github.com/nextcloud/docker/pull/1789.

### DIFF
--- a/library/nextcloud
+++ b/library/nextcloud
@@ -5,45 +5,45 @@ GitRepo: https://github.com/nextcloud/docker.git
 
 Tags: 23.0.11-apache, 23.0-apache, 23-apache, 23.0.11, 23.0, 23
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
-GitCommit: 739d6996409825bc61000f0c901ea05fbe3debf7
+GitCommit: 6103e074d81396c235fb453a88f518104cf35382
 Directory: 23/apache
 
 Tags: 23.0.11-fpm, 23.0-fpm, 23-fpm
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
-GitCommit: 739d6996409825bc61000f0c901ea05fbe3debf7
+GitCommit: 6103e074d81396c235fb453a88f518104cf35382
 Directory: 23/fpm
 
 Tags: 23.0.11-fpm-alpine, 23.0-fpm-alpine, 23-fpm-alpine
 Architectures: amd64, arm32v6, arm32v7, arm64v8, i386, ppc64le, s390x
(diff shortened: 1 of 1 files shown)


library/nextcloud
```diff
@@ -5,45 +5,45 @@ GitRepo: https://github.com/nextcloud/docker.git
 
 Tags: 23.0.11-apache, 23.0-apache, 23-apache, 23.0.11, 23.0, 23
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
-GitCommit: 739d6996409825bc61000f0c901ea05fbe3debf7
+GitCommit: 6103e074d81396c235fb453a88f518104cf35382
```
yosifkit, line 8: These commit ids don't quite match the output of [`./generate-stackbrew-library.sh`](https://github.com/nextcloud/docker/blob/6103e074d81396c235fb453a88f518104cf35382/generate-stackbrew-library.sh).  It chooses https://github.com/nextcloud/docker/commit/48f223fc7f3131f2f1526ddb12e779396acadb20 since that is the last commit to modify each build context.

yosifkit: The output of their `generate-stackbrew-library.sh` is designed to be the `library/nextcloud` file to ensure consistency and help prevent small errors that can creep in when doing things manually (like missing one of the tags on a version bump). So the script has a deterministic way to choose a commit id so that it doesn't change when nothing changes in that build context (folder).

dkadioglu: Ok, thanks for that explanation. However, I still don"t understand the whole approach, I think. What exactly do you mean with 'build context'? In the script, one of the comments says '
# Get the most recent commit which modified "$1/Dockerfile" or any file that
# the Dockerfile copies into the rootfs (with COPY). .' which I would resolve to the commit hash in this PR, as this is the last one changing the Dockerfiles. What has to happen, that this script also resolves to this commit?

tianon: The build context is the argument to `docker build`; https://docs.docker.com/engine/reference/commandline/build/

> The `docker build` command builds Docker images from a Dockerfile and a “context”. A build’s context is the set of files located in the specified `PATH` or `URL`. The build process can refer to any of the files in the context. For example, your build can use a [COPY](https://docs.docker.com/engine/reference/builder/#copy) instruction to reference a file in the context.

The result of running that script locally should be something like https://github.com/docker-library/official-images/blob/7d7804fd42126894c01048d95e7bf1e3ebf27b4b/library/nextcloud (although that's got some ordering weirdness that I think we've seen before; cc @J0WI).

tianon: Oh, that `Maintainers:` field is also a bit general, let's cc @tilosp too :innocent:

PR #13710, opened 2022-12-10
Aerospike 6.2.0.2

mcoberly2:

### DIFF
--- a/library/aerospike
+++ b/library/aerospike
@@ -1,13 +1,16 @@
 Maintainers: Lucien Volmar <lucien@aerospike.com> (@volmarl),
              Michael Coberly <mcoberly@aerospike.com> (@mcoberly2),
-             Phuc Vinh <pvinh@aerospike.com> (@pvinh-spike)
+             Phuc Vinh <pvinh@aerospike.com> (@pvinh-spike),
+             Kevin Porter <kporter@aerospike.com> (@kportertx)
 
-Tags: ee-6.1.0.3
-Architectures: amd64
-GitRepo: https://github.com/aerospike/aerospike-server-enterprise.docker.git
-GitCommit: 01871453aad7390a02a51dee11efe8e6a55f95cf
-
-Tags: ce-6.1.0.3
-Architectures: amd64
+Tags: ee-6.2.0.2
+Architectures: amd64, arm64v8
 GitRepo: https://github.com/aerospike/aerospike-server.docker.git
-GitCommit: 4557452ca88bcad1929e91bd303b522a1c242b12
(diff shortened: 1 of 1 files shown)


yosifkit:
Closing/re-open to trigger GitHub actions.

kportertx:
Looks like their build doesn't use `buildx`. @yosifkit, is there a way to signal that these actions should enable `buildx`?

tianon:
When we requested that the "how to install" commands be embedded in the `Dockerfile`, taking the long installation script and embedding it verbatim via heredoc is not exactly what we had in mind. :grimacing:

Is all this complexity really necessary to properly download and install Aerospike securely?

Additionally, you'll probably be interested in https://github.com/docker-library/official-images/issues/3540#issuecomment-530925319, especially:

kportertx:
> Is all this complexity really necessary to properly download and install Aerospike securely?

It was much easier debugging with the `bootstrap.sh` since it provided line numbers for what command failed. The typical list of command separated by `&&` just says there is a failure on line one which isn't useful. The `heredoc` results in the same behavior as the `&&` version but at least I can switch back to executing `bootstrap.sh` for local debugging.

Some of the complexity in the script is to keep the package dependencies close to the items that depend on them.

I personally dislike that any tools are installed on the server package - I'd like it to be just the server, but overwhelming demands insisted that we install our `asinfo` and `asadm` tools on the server package. Installing just this subset adds quite a bit of complexity but it was insisted that we only install this subset.

> Additionally, you'll probably be interested in https://github.com/docker-library/official-images/issues/3540#issuecomment-530925319, especially:

We intend to manage UBI images from the same repository in the future (which will likely further complicate the bootstrap code). The goal is to simplify our release process. Seems dockerhub's recommendation for labels may differ from UBI - I'll need to discuss with our the OpenShift team.

kportertx:
There are also different server editions, some editions have differences in their package dependencies. In an effort to keep the containers minimal, we only install the dependencies required for a particular edition. I could simplify the scripts a bit by instead installing all dependencies required by any edition - if I'm required not to use `heredoc` that is probably what I will do to keep the obscene list of command separated by `&&` readable/manageable :). Though not being able to use `heredoc` will also make it more difficult to also manage UBI images in the same repository.

kportertx:
> Additionally, you'll probably be interested in https://github.com/docker-library/official-images/issues/3540#issuecomment-530925319, especially:

I guess the simple response to this is to not include any labels for the dockerhub releases. I was attempting to bring over labels from our UBI images that seemed applicable to the dockerhub images.

mcoberly2:
Closing, will submit an new PR.

PR #13718, opened 2022-12-12
Update CrateDB to 5.1.2

matriv:


library/crate
```diff
@@ -4,9 +4,9 @@ Maintainers: Mathias Fußenegger <mathias@crate.io> (@mfussenegger),
              Andreas Motl <andreas.motl@crate.io> (@amotl)
 GitRepo: https://github.com/crate/docker-crate.git
 
-Tags: 5.1.1, 5.1, latest
+Tags: 5.1.2, 5.1.1, 5.1, latest
```
yosifkit, line 7:
```diff
-Tags: 5.1.2, 5.1.1, 5.1, latest
+Tags: 5.1.2, 5.1, latest
```

There is an extra tag here that doesn't match.

PR #13733, opened 2022-12-13
nginx: introduced another variant: alpine-slim and updated to 1.23.3.

thresheek:
alpine-slim is currently only enabled for mainline images.  Will propagate to stable when there is a new release.

Refs:
 - https://github.com/nginxinc/docker-nginx/issues/681
 - https://github.com/nginxinc/docker-nginx-unprivileged/issues/63

PR #13738, opened 2022-12-13
Slightly improved "diff-pr" output, especially around build-order

tianon:
This should help avoid "empty cat" output we saw recently when something failed to fetch but was otherwise valid.

The current implementation is trying to do two things with one output:

1. normalized `bashbrew cat` diff
2. visualize `--build-order` changes

This change splits those into two separate things because combining them ends up making the diff harder to read than it should be ("changed tags + changed commit" leading to a full reordering of the diff because build-order also changed).

### DIFF
--- a/_bashbrew-cat-sorted.sh
+++ b/_bashbrew-cat-sorted.sh
@@ -5,8 +5,7 @@ set -Eeuo pipefail
 
 images="$(
 	bashbrew list --repos --uniq "$@" \
-		| sort -uV \
-		| xargs -r bashbrew list --repos --uniq --build-order
+		| sort -uV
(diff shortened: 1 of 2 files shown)


tianon:
(I'm going to borrow the commit from #13733 to test this in just a second.)

tianon:
(apologies to @thresheek in advance for the ping he's about to get in the name of science)

tianon:
Oh boo, I forgot that it runs the scripts from the _target_ branch, not from the source (for security reasons).

tianon:
:heart_eyes:  that `_bashbrew-cat` output is much better

PR #13750, opened 2022-12-15
Migrate Adminer to Debian Bulleye's native PHP 7.4 packages

TimWolla:
see TimWolla/docker-adminer#129


TimWolla:
I'm intentionally leaving it in the branch for now, because I hope to switch back to the `php` base image with the next Adminer update (if any).

PR #13577, opened 2022-11-22
alpine: bump 3.17.0 and update latest

ncopa:

PR #13772, opened 2022-12-20
erlang/otp 25.2

getong:


tianon:
Are you sure you don't want to add yourself to the `Maintainers:` list so you get included in pings like https://github.com/docker-library/official-images/pull/13577#issuecomment-1324292585? :sweat_smile:

tianon:
(Something like `19年梦醒 (@getong)` should work and be enough, FWIW)

getong:
> (Something like `19年梦醒 (@getong)` should work and be enough, FWIW)
No, but thanks.

I delete alpine support for erlang 21, as it was released 4 years ago.

tianon:
It probably makes sense to consider going further and removing more of those old tags from the supported/rebuilt list -- is Erlang 21 still actively supported by the project? :sweat_smile:

PR #13773, opened 2022-12-20
composer 2.5.0

alcohol:


yosifkit:
Docker's classic builder (`docker build` as in `20.10.x`) will still create the extra untagged images, even though they are unused by the build. The newer builder, [`buildkit`](https://docs.docker.com/build/buildkit/), does skip it appropriately. We've recently added support for specifying the newer builder (https://github.com/docker-library/bashbrew/pull/43). So this will need `Builder: buildkit` added to the Library file like [rabbitmq](https://github.com/docker-library/official-images/pull/13728).

alcohol:
To be honest, I am not sure what extra untagged images you are referring to. But looking at the change made for rabbitmq; if that is all that is desired, I'll serve that up in a minute.

tianon:
He means that the classic builder will build the `standalone-binary` stage whether it needs to or not, so without switching to BuildKit that'll result in fresh "dangling" images being built every time we build this repo (we do plan to switch everyone to BuildKit by default eventually, but it's a pretty dramatic change so we're trying to take it slow given the level of impact if anything goes wrong).

PR #13775, opened 2022-12-20
Aerospike 6.2.0.2

pvinh-spike:

### DIFF
--- a/library/aerospike
+++ b/library/aerospike
@@ -1,13 +1,16 @@
 Maintainers: Lucien Volmar <lucien@aerospike.com> (@volmarl),
              Michael Coberly <mcoberly@aerospike.com> (@mcoberly2),
-             Phuc Vinh <pvinh@aerospike.com> (@pvinh-spike)
+             Phuc Vinh <pvinh@aerospike.com> (@pvinh-spike),
+             Kevin Porter <kporter@aerospike.com> (@kportertx)
 
-Tags: ee-6.1.0.3
-Architectures: amd64
-GitRepo: https://github.com/aerospike/aerospike-server-enterprise.docker.git
-GitCommit: 01871453aad7390a02a51dee11efe8e6a55f95cf
-
-Tags: ce-6.1.0.3
-Architectures: amd64
+Tags: ee-6.2.0.2
+Architectures: amd64, arm64v8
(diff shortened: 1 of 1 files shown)


yosifkit:
By using semicolon as your command separator, you need to use `-e` so that failing commands exit instead of continuing to the next command. This should let you drop any of the `|| exit 1;` and similar behaviour since that is what `set -e` will do. You can add it to your `SHELL` declaration:

```dockerfile
SHELL ["/bin/bash", "-eo", "pipefail", "-c"]
```
---

Using `uname -m` or `arch` for platform detection is very strongly discouraged, as it tests the host kernel's architecture, when what you really wanted to test was the image's userspace architecture, so `apk --print-arch` or `dpkg --print-architecture` are going to be safer/more reliable options.

---

I'm unsure what the `grep` is here for. It is just to hide some informational `stderr` text?

> apt-get install -y --no-install-recommends apt-utils 2>&1 | grep -v "delaying package configuration"; \

kportertx:
> Using uname -m or arch for platform detection is very strongly discouraged, as it tests the host kernel's architecture, when what you really wanted to test was the image's userspace architecture, so apk --print-arch or dpkg --print-architecture are going to be safer/more reliable options.

I see another project that uses `ARG TARGETARCH` - can/should I use that instead of the `apk` or `dpkg` commands?

pvinh-spike:
Closing, will replace with Version 6.2.0.2 shortly

tianon:
In the future, can you please instead force push to your existing branch (which will then update the existing PR)?  When you close the PR and open a new one, it splits the conversation, which makes it much harder to follow, and generates more separate email threads making it harder to sort through.

kportertx:
@tianon, @yosifkit,

Any recommendations for how to query the `arch` for images that do not have `dpkg` or `apk`?

I currently have the following:

```
file -Lb /usr/bin/awk | awk -F'[, \t]*' '{ print $6 }'
```

tianon:
Here's one I've used in the past:

`arch="$(objdump="$(command -v objdump)" && objdump --file-headers "$objdump" | awk -F '[:,]+[[:space:]]+' '$1 == "architecture" { print $2 }')"`

(effectively asking `objdump` to give us the ELF architecture of itself and thus userspace)

PR #13776, opened 2022-12-21
add new image for nanomq

Rory-Z:


yosifkit:
Hello! :sparkles:

Rory-Z:
Hi, @yosifkit Thanks for your reply.
I will wait patiently for your review and thank you for your work!

tianon:
Hey, thanks for your patience! We're trying to work through our queue, and before we get too deep into reviewing we'd like to make sure this is still something you're interested in pursuing (esp. since it's been so long since our last touch point)? 🙇 🫶

JaylinYu:
That's for sure. 
I am with @Rory-Z, also the maintainer of NanoMQ

> Hey, thanks for your patience! We're trying to work through our queue, and before we get too deep into reviewing we'd like to make sure this is still something you're interested in pursuing (esp. since it's been so long since our last touch point)? 🙇 🫶

tianon:
Ok, I'll start with just a couple high-level first-pass thoughts, because the resolution of these might end up making some of my other comments moot, so I'll save them for now. :smile:

---

```diff
+GitRepo: https://github.com/emqx/nanomq-docker.git
```

This is probably the most obvious thing to update :eyes:

This repository redirects to https://github.com/emqxarchive/nanomq-docker now, and https://github.com/emqx/nanomq now redirects to https://github.com/nanomq/nanomq -- hopefully that's all actually good news for the project? :sweat_smile:

---

```diff
+Tags: 0.14.8, 0.14, latest
+Tags: 0.14.8-full, 0.14-full, full
+Tags: 0.14.8-slim, 0.14-slim, slim
```

I'm trying to sort out what each of these are and what end-user they each service and I have to admit that I'm really struggling to "get it" -- `latest` appears to be `FROM alpine`, `full` is `FROM ubuntu`, and `slim` is `FROM debian`, for example, and none of that is obvious from the tags, nor what variant users should pick based on their use case.

Is there a reason to have all three of these?  Is there one in particular that's more "supported" upstream than the others?

PR #13777, opened 2022-12-21
Update rocky linux 9.1 images to fix regressions from 9.0

NeilHanlon:


NeilHanlon:
(Small chicken/egg problem.. i'm pushing the container branches now but the checks might need a re-run)

PR #13778, opened 2022-12-21
Release Neo4j 4.3.23 and 4.4.16

bfeshti:


tianon:
For the next PR, I'd suggest starting your branch from the latest `master` on https://github.com/docker-library/official-images, and then pushing it to the `neo-technologies` fork when you're finished making changes.  (Otherwise, we're going to keep doing squash-and-merge because we don't want those unnecessary merge commits in here.)

Something like this is a pretty typical example of a standard PR workflow:

```console
$ git clone https://github.com/docker-library/official-images.git oi
$ cd oi
$ git remote add fork https://github.com/neo-technology/docker-official-images.git
$ git checkout -b some-change-branch-name
$ # edit "library/neo4j" here
$ git commit ...
$ git push --set-upstream fork some-change-branch-name
```

(This is Git, so there are other valid ways to accomplish the same thing, but I'm sure you get the idea.)

PR #13833, opened 2023-01-04
Update Debian (20230104)

tianon:
(This does not yet include/implement https://github.com/debuerreotype/docker-debian-artifacts/pull/186)


tianon:
@yosifkit pointed out that it's technically only been ~2 weeks and I've jumped the gun a little bit :see_no_evil:

PR #13836, opened 2023-01-05
Update to JRuby commit that includes make

headius:
make is needed for several dummy gems that do not need to build an extension but still attempt to run a Makefile.

See https://github.com/jruby/docker-jruby/issues/81

### DIFF
--- a/library/jruby
+++ b/library/jruby
@@ -3,7 +3,7 @@ Maintainers: JRuby Admin <admin@jruby.org> (@jruby),
              Thomas E Enebo <tom.enebo@gmail.com> (@enebo)
 GitRepo: https://github.com/jruby/docker-jruby.git
 GitFetch: refs/heads/master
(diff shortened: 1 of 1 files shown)


headius:
Looks like a new Bundler version, released just before the holidays, no longer works on JRuby 9.2. I will look into this.

headius:
We are opting to remove the 9.2 images. JRuby 9.2 has not been officially supported for some time, and with new Bundler versions not working anymore it seems like a good time to drop it from the images.

headius:
Ok now the failures seem to be some other service going down, a service used to verify the images. I do not see how this could have been related to adding `make`, and in the previous commit these jobs were sucessful.

headius:
Repushed last commit and the down service (appears to have been rubygems.org) is now working. Ready to go!

tianon:
Makes sense, thanks for chasing it!  I'll include my standard copy-pasta here to hopefully help ease your mind on the removal:

:+1:

PR #13837, opened 2023-01-05
Tomcat 9.0.70 seems to have regression or at least behavior changes and that breaks XWiki

vmassol:


vmassol:
@tianon hi (and a happy new year to you! :)). Any idea why I get https://github.com/docker-library/official-images/actions/runs/3848087647/jobs/6555393691#step:4:47 ? It was working with `9-jre11` and it's now failing with `9.0.69-jre11`.

tianon:
That'd be because 9.0.69 is no longer supported given that 9.0.70 is released. :disappointed:

yosifkit:
Ah, the issue with tomcat `9.0.70` is https://jira.xwiki.org/browse/XWIKI-20470. Do those errors cause a loss of functionality?

@tianon, maybe it could make sense to temporarily allow the image to stay on the older tomcat release until there is a fix somewhere?  @vmassol, It wouldn't get any base image rebuilds, but at least the tomcat errors will go away.  That will mean every xwiki PR here will fail the "Naughty" test until it is updated.

If the errors don't break functionality, then I'd think that up-to-date base images would be more important than noisy logs. :shrug:

vmassol:
> Ah, the issue with tomcat 9.0.70 is https://jira.xwiki.org/browse/XWIKI-20470. Do those errors cause a loss of functionality?

@yosifkit thanks for checking this!

We have 2 issues related to Tomcat 9.0.70 (they don't fail on 9.0.69 with the exact same XWiki version and setup):
* https://jira.xwiki.org/browse/XWIKI-20470
* https://jira.xwiki.org/browse/XWIKI-20476

They both emit error logs but they also result in loss of functionality (to different degrees and I'm not 100% sure of all the consequences. They may be a problem only for some special cases that not many users will encounter, or not...).

I wanted to check if something had been reported to the Tomcat dev team but I admit that my knowledge of Bugzilla is quite low and I wasn't able to find anything (but it's very likely that I haven't used it correctly). 

> @tianon, maybe it could make sense to temporarily allow the image to stay on the older tomcat release until there is a fix somewhere? @vmassol, It wouldn't get any base image rebuilds, but at least the tomcat errors will go away. That will mean every xwiki PR here will fail the "Naughty" test until it is updated.

It would help XWiki to be able to have all the XWiki docker images rebuilt to use Tomcat 9.0.69 since right now our users using the XWiki docker images (that represents about 30% of all users) are getting errors.

I don't understand what you mean by `It wouldn't get any base image rebuilds` and `but at least the tomcat errors will go away`. How can the problem go away for the XWiki users if the XWiki images are not rebuilt? I must be misunderstanding something :) 

Thanks a lot again!

yosifkit:
Sorry for the confusion. If the xwiki images switch to `tomcat:9.0.69`, then it would only get a single build for this current Dockerfile change and no more rebuilds (without another PR with Dockerfile changes). Normally it would get rebuilt when, for example `ubuntu:22.04` is republished (every ~30 days), but now its parent image will be outside of maintained images, so it wouldn't get the periodic rebuild.

Background: 
>

Since our build system makes heavy use of Docker build cache, just rebuilding the all of the Dockerfiles won't cause any change. So we rely on periodic base image updates.

>
> \- from the same FAQ link

---

I'm fine with the change to move `xwiki` down to `tomcat:9.0.69`. I wanted to make sure you had any other information related to how that would affect the image in official images.

@vmassol Can you rebase & force push to resolve the merge conflict?

vmassol:
FYI we've now retested with Tomcat 9.0.71 and it fixes https://jira.xwiki.org/browse/XWIKI-20470

However https://jira.xwiki.org/browse/XWIKI-20476 is still happening (which is expected since Tomcat 9.0.71 has not changed anything from 9.0.70 on this).

I'm leaving this PR opened for now as we debug the issue with https://jira.xwiki.org/browse/XWIKI-20476 (which only happens on Windows so it's less severe than if it were happening on all OSes).

I'm personally on holidays and I probably won't be working on this before the 23rd, but maybe other xwikiers will ;)

Thanks

vmassol:
We don't need this anymore as Tomcat 9.0.71+  is now ok for us. Thanks

PR #13862, opened 2023-01-10
Update Debian to 20230109

tianon:


tianon:
Heh, looks like `tzdata` changes made this diff extra large: https://gist.github.com/docker-library-bot/5d0f7527cd8ffda61c668e023878dc74

yosifkit (approved):

PR #13875, opened 2023-01-12
Increase UTF_MAX for 1.9.4

vanosg:


tianon:
Three questions, hopefully easy: :sweat_smile:

1. is there some kind of checksum we could add for that download?
2. why not 8.6.13?
3. is this patch you're making something that the Tcl maintainers are aware of and also planning to fix? :eyes:

tianon:
oh, and 4. are the downloaded file and extracted directory cleaned up? (you probably don't want either of those in the final image)

vanosg:
All good questions! This code has been in the develop branch and I think we talked about some similar concepts when it was first added in to there.

1) I could calculate a checksum myself, but no 'official' checksum exists.
2) 8.6.12 is what we tested in develop. Beyond that poor excuse... I don't have a good reason
3) a) yes, Tcl maintainers are aware. b) Planning to fix.... :)   For compatibility, they actually can't fix it the way Tcl is written in 8.6. They increased this variable in 8.7, but the most recent alpha for 8.7 was released in June of '21... so we're not holding our breath.
4) Why, they are not! Thanks

tianon:
A checksum would be a great addition in the future (if nothing else to help us detect SF funny-business like https://en.wikipedia.org/wiki/SourceForge#Project_hijackings_and_bundled_malware), but not going to hold it up for that. :+1:

PR #13874, opened 2023-01-12
Update Windows images

yosifkit:

PR #13883, opened 2023-01-13
Add support for "external pins" in "diff-pr.sh"

tianon:
This uses `crane` to download some of the JSON files of the remote images to give us at least *some* amount of diffing between updates to external pins.

See https://github.com/docker-library/official-images/pull/13874#issuecomment-1381162010 for an example of the output of this. :eyes:


tianon:
Hmmmm, I think this needs changes in the GHA files too.

tianon:
Ok, I'm satisfied now. :smile:

yosifkit (approved):

PR #13894, opened 2023-01-16
Update Zookeeper

31z4:
* Upgrade to 3.5.10 and 3.6.4
* Add openjdk based tags


yosifkit:
LGTM since the images based on unsupported images are just a re-tag so that users have an escape hatch to still pull them easily. https://github.com/docker-library/official-images/pull/12949#issuecomment-1385938193

And noting that the `-openjdk` images will be re-tagged this final time, but with their unsupported base they will not get the periodic rebuilds that the rest of official images have (discussed on a recent PR also needing an older base: https://github.com/docker-library/official-images/pull/13837#issuecomment-1373958917).

PR #13910, opened 2023-01-18
Update Corretto for 2023 Q1 release

Rudometov:
Update Corretto for 2023 Q1 release (8.362.08.1, 11.0.18.10.1, 17.0.6.10.1, 19.0.2.7.1)

PR #13912, opened 2023-01-18
Update "java" tests to run javac to determine whether a JDK image is probably an actual JDK

tianon:
See https://github.com/docker-library/official-images/pull/13432 (and now https://github.com/docker-library/official-images/pull/13910 :sob:).


tianon:
(I've temporarily added the commit from #13910 so we can test that this works and fixes the `headless` case)

tianon:
Successful tests in https://github.com/docker-library/official-images/actions/runs/3952077962 :+1:

PR #13928, opened 2023-01-21
library/perl: Add Perl 5.37.8

zakame:
- https://github.com/Perl/docker-perl/issues/125
- https://github.com/Perl/docker-perl/pull/127

### DIFF
--- a/library/perl
+++ b/library/perl
@@ -1,39 +1,39 @@
 Maintainers: Peter Martini <PeterCMartini@GMail.com> (@PeterMartini),
              Zak B. Elep <zakame@cpan.org> (@zakame)
 GitRepo: https://github.com/perl/docker-perl.git
-GitCommit: f70e8ace49994efef8e90cbf730554a3e3201da7
-Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
+GitCommit: a0b67543fd56367c403a2f063dfed1a69fdda390
+Architectures: amd64, arm32v7, arm64v8, i386
 
-Tags: 5.36.0, 5.36, 5, latest, 5.36.0-bullseye, 5.36-bullseye, 5-bullseye, bullseye
+Tags: 5.36.0, 5.36, 5, latest, stable, 5.36.0-bullseye, 5.36-bullseye, 5-bullseye, bullseye, stable-bullseye
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
 Directory: 5.036.000-main-bullseye
 
-Tags: 5.36.0-buster, 5.36-buster, 5-buster, buster
-Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
+Tags: 5.36.0-buster, 5.36-buster, 5-buster, buster, stable-buster
+Architectures: amd64, arm32v7, arm64v8, i386
 Directory: 5.036.000-main-buster
 
-Tags: 5.36.0-slim, 5.36-slim, 5-slim, slim, 5.36.0-slim-bullseye, 5.36-slim-bullseye, 5-slim-bullseye, slim-bullseye
+Tags: 5.36.0-slim, 5.36-slim, 5-slim, slim, stable-slim, 5.36.0-slim-bullseye, 5.36-slim-bullseye, 5-slim-bullseye, slim-bullseye, stable-slim-bullseye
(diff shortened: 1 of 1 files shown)


library/perl
```diff
@@ -1,101 +1,133 @@
...
 Tags: 5.32.1-slim-threaded, 5.32-slim-threaded, 5.32.1-slim-threaded-bullseye, 5.32-slim-threaded-bullseye
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
 Directory: 5.032.001-slim,threaded-bullseye
 
 Tags: 5.32.1-slim-threaded-buster, 5.32-slim-threaded-buster
-Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
+Architectures: amd64, arm32v7, arm64v8, i386
 Directory: 5.032.001-slim,threaded-buster
+
+Tags: 5.37.8, 5.37, blead, devel, 5.37.8-bullseye, 5.37-bullseye, blead-bullseye, devel-bullseye
```
tianon, line 103: Wait, isn't `blead` the name used for the "master" / "main" development branch of Perl?  Is it possible this will make folks think this is an actual development build as opposed to the periodic development releases?  (Sorry, we're reaching the limits of my knowledge of how folks know/understand Perl versions and labels on them, even as a user of it myself. :sweat_smile:)

tianon: Yeah, looking at https://dev.perl.org/perl5/source.html and https://www.nntp.perl.org/group/perl.perl5.porters/2023/01/msg265547.html, it appears these _are_ commonly referred to as "development releases" but _not_ necessarily as "blead releases" right?

> The master branch, where the development takes places, is named _blead_.

vs

> We are happy to announce version 5.37.8, the ninth development release of version 37 of Perl 5.

jberger: This does seem wrong to me ... though I too am at my limit of process here. Maybe @haarg or @leonerd would know better (just two people off the top of my head)?

zakame: Yeah, digging deeper down we actually have tools like plenv/perl-build take `blead` tarballs from the git archive cf https://github.com/tokuhirom/Perl-Build/blob/b5e809342bed7b8ba51b53210cf5e1665953b34c/script/perl-build#L70

We can take off the `blead` tag for now and use `devel` only.

library/perl
```diff
@@ -1,101 +1,133 @@
 Maintainers: Peter Martini <PeterCMartini@GMail.com> (@PeterMartini),
              Zak B. Elep <zakame@cpan.org> (@zakame)
 GitRepo: https://github.com/perl/docker-perl.git
-GitCommit: f70e8ace49994efef8e90cbf730554a3e3201da7
-Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
+GitCommit: a0b67543fd56367c403a2f063dfed1a69fdda390
+Architectures: amd64, arm32v7, arm64v8, i386
 
 Tags: 5.36.0, 5.36, 5, latest, 5.36.0-bullseye, 5.36-bullseye, 5-bullseye, bullseye
```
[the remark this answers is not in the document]
tianon: Sorry, sometimes I forget that I'm working from a huge amount of assumed context that not everyone has :joy:

For background, the reason this PR made me think about adding a `stable` alias is not just that I looked up a lot of details of "what does Perl label releases as" but also because adding a `stable` alias is an extra clue to users looking at the list of tags that 5.37 might _not_ be what they want to choose, and that 5.36 is probably more what they desire (assuming they don't know exactly which version of Perl they want/need _and_ thus also likely don't know the Perl convention around odd-numbered minor releases being unstable/development "preview" releases).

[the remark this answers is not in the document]
zakame: Yeah, adding `stable` is also easy enough with our current automation, thanks for suggesting! ⚙️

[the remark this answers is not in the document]
zakame: We should now have `stable` tags in https://github.com/docker-library/official-images/pull/13928/files#diff-40ed9e38417ffb2e97e136f9de487c6cdb45a2c5c1f8667b871cb8573b12f4a1L7-R36

library/perl
leonerd: If it has a `5.x.y` version number then it isn't blead, it's a release. Depending on the parity of `x`, odd numbers are development releases, and even numbers are production releases. `blead` is the name of the main development git branch, which is always in flight.

zakame:
Thanks everyone! 💯 🎉

PR #13964, opened 2023-01-30
Upgrade stable version to 15.0

tmortagne:


yosifkit:
Just to thread forward the limitations of staying on an old `tomcat` image https://github.com/docker-library/official-images/pull/13837; the change to use `tomcat:9.0.69` is a fine temporary solution (with the caveats noted over there).

PR #13970, opened 2023-01-31
Removes pierreozoux as a maintainer.

pierreozoux:
Hey,

I'm no longer using matomo.
Thanks for your work and have a nice day!


library/matomo
```diff
@@ -1,5 +1,4 @@
 # This file is generated via https://github.com/matomo-org/docker/blob/2252146c6a2edbaf5028ed148e57e073d722afb4/generate-stackbrew-library.sh
-Maintainers: Pierre Ozoux <pierre@piwik.org> (@pierreozoux)
```
tianon, line 2: We need to keep _someone_ listed here - perhaps @J0WI is willing to be listed? :sweat_smile:
(email is not strictly required, `Name (@github)` is an accepted format and even Name is fungible)

pierreozoux:
https://github.com/matomo-org/docker/pull/301

PR #13986, opened 2023-02-01
add new image for ekuiper

Rory-Z:
Signed-off-by: Rory Z <rory-z@outlook.com>


Rory-Z:
Hello all, I don't know why the 'Naughty' check failed, any ideas ?

yosifkit:
Background: 
>

Since our build system makes heavy use of Docker build cache, just rebuilding the all of the Dockerfiles won't cause any change. So we rely on periodic base image updates to address things like CVE updates and bug fixes in OS packages:

>

While the `golang:1.18.5-alpine` is available for users to pull from Docker Hub, official images should only be based on current/supported images. This ensures that when a base image like Debian or Alpine is updated, that all dependent images are in the rebuild tree (if it is based on an image no longer in a library file it would not be subject to these periodic rebuilds).

The recommendation is to use a more generic tag like `golang:1.18-alpine` so that the dependent image automatically gets rebuilt with the patch updates. The currently active tags for `1.18*-alpine` are in the library file: 

https://github.com/docker-library/official-images/blob/23d3ac00b652aa2d81230fdb9c9491c05c029d39/library/golang#L117

Unfortunately, those are going to go away today with the release of Go `1.20` (https://github.com/docker-library/golang/pull/451), so either a `1.19*` or `1.20*` image would be the only options.

---

The same is true for the `python` image, the `3.8.12-slim-bullseye` is on Docker Hub, but it is not actively supported, so a `3.8-slim-bullseye` would keep the image more up-to-date.

https://github.com/docker-library/official-images/blob/23d3ac00b652aa2d81230fdb9c9491c05c029d39/library/python#L179

Rory-Z:
Hi @yosifkit, Thanks for for your detailed answers, I already change it

PR #13990, opened 2023-02-01
SonarQube upgrade to 9.9 (LTS)

jCOTINEAU:
Hi team, we are going to release the new SonarQube images soon, with a new arm64 support.

We just drafted the PR, could you please give us some feedback on the structure of the new `library/sonarqube` and new Dockerfiles we are going to publish?

Please note that the commit sha will be changed for the real release.

~~Do not merge it for now~~

### DIFF
--- a/library/sonarqube
+++ b/library/sonarqube
@@ -1,38 +1,29 @@
 Maintainers: Christophe Levis <christophe.levis@sonarsource.com> (@christophelevis),
              Pierre Guillot <pierre.guillot@sonarsource.com> (@pierre-guillot-gh),
              Lukasz Jarocki <lukasz.jarocki@sonarsource.com> (@lukasz-jarocki-sonarsource),
-             Wouter Admiraal <wouter.admiraal@sonarsource.com> (@wouter-admiraal-sonarsource),
              Aurelien Poscia <aurelien.poscia@sonarsource.com> (@aurelien-poscia-sonarsource),
-             Matteo Mara <matteo.mara@sonarsource.com> (@matteo-mara-sonarsource)
+             Carmine Vassallo <carmine.vassallo@sonarsource.com> (@carminevassallo),
+             Jeremy Cotineau <jeremy.cotineau@sonarsource.com> (@jCOTINEAU)
 GitRepo: https://github.com/SonarSource/docker-sonarqube.git
-GitCommit: 6e6bd428f38d5ba5b6169fd15d29ce8dbe00921f
+Architectures: amd64, arm64v8
+GitCommit: ed2071a47632c34d7639b6ad1b393ab2aee1c9a8
 
-Tags: 8.9.10-community, 8.9-community, 8-community, lts, lts-community
-Directory: 8/community
-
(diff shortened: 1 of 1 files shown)


yosifkit:
I think they look pretty good. The first bit of feedback is to add a `rm -rf /var/lib/apt/lists/*` at the end of the `RUN` lines that do a `apt-get update` to help keep the image just a little bit smaller.

---

> All official images should provide a consistent interface. A beginning user should be able to `docker run official-image bash` (or `sh`) without needing to learn about `--entrypoint`. It is also nice for advanced users to take advantage of entrypoint, so that they can `docker run official-image --arg1 --arg2` without having to specify the binary to execute.
>
> \- https://github.com/docker-library/official-images/tree/05f539b9a91128e2238f584d6dc4f0a81a8e1a9f#consistency

(this is the source of the test failures for `override cmd`)

`ENTRYPOINT ["/opt/java/openjdk/bin/java","-jar","lib/sonarqube.jar","-Dsonar.log.console=true"]`

The new `ENTRYPOINT` on a couple of the images is not written with the same interface that we require for official images. A minimal shell script would work like you had before, or just moving it to `CMD` would also work if you don't expect or desire that users do `docker run ... sonarqube -Dsome-java-opt`

tianon:
Just a couple other really minor things, nothing super major above what you've noted. :+1:

---

```diff
+    adduser --system --uid 1000 sonarqube; \
+    addgroup --system --gid 1000 sonarqube; \
+    adduser sonarqube sonarqube; \
```

The end result of this ordering is that `sonarqube` will be part of `nogroup` (65534) as well -- what you probably want is something like this:

```dockerfile
    groupadd --system --gid 1000 sonarqube; \
    useradd --system --uid 1000 --gid sonarqube sonarqube; \
```

---

```diff
+	sed -i -e 's/\r$//' elasticsearch elasticsearch-env elasticsearch-env-from-file ;
```

This one's a little bit eyebrow-raising, but not a blocker -- I think you probably want to fix this in the source distribution. :sweat_smile:

jCOTINEAU:
Hello everyone, thanks a lot for the huge feedback !

We adjusted the PR with the corresponding new commit.

If those changes looks fine to you, that PR will stay on hold before being merged next week when the new SonarQube will be released.

Thanks for your time.

yosifkit:
It looks like the `useradd` and `rm -rf` changes were missed on `sonarqube:lts-datacenter-app` and `sonarqube:lts-datacenter-search`. Everything else looks good for next week. :+1:

jCOTINEAU:
Hello everyone,

We are now ready to merge !

jCOTINEAU:
Hello @tianon, @yosifkit, thanks a lot for taking action so promptly ! 

Nonetheless we noticed that the `arm64v8` is not available on the docker hub.

This is a big issue because we planned to support arm64 with the new SonarQube release.

Have we missed something or are there some ongoing operations on your side ? If there is something going on, what would be the timeline for it ? 

Thanks again.

yosifkit:
https://github.com/docker-library/official-images/issues/14024#issuecomment-1419689359

PR #13996, opened 2023-02-01
Update buildpack-deps

tianon:
Changes:

- https://github.com/docker-library/buildpack-deps/commit/598c0e2: Remove xenial (now EOL)


tianon:
(Nice, looks like this doesn't affect any other images :partying_face:)

PR #14015, opened 2023-02-03
Update Debian to 20230202

tianon:
Notably, this drops the `riscv64` architecture from `unstable`/`experimental` -- this architecture is part of Debian Ports (which is not part of official releases of Debian proper), and unfortunately Ports is currently suffering from a lack of snapshot.debian.org updates (https://bugs.debian.org/1029744).

Hopefully, that will be resolved in the future and `riscv64` will be able to return (this is doubly sad due to the snafu with the ports archive keyring having an incorrect expiration date; https://tracker.debian.org/news/1416043/accepted-debian-ports-archive-keyring-20230201-source-into-unstable/).

### DIFF
--- a/library/debian
+++ b/library/debian
@@ -7,34 +7,31 @@ GitRepo: https://github.com/debuerreotype/docker-debian-artifacts.git
 GitCommit: ea6ede8f53deaae64dd492001410b689d127e810
 # https://github.com/debuerreotype/docker-debian-artifacts/tree/dist-amd64
 amd64-GitFetch: refs/heads/dist-amd64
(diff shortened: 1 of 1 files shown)


tianon:
The most notable change here besides the loss of `riscv64` on unstable is https://github.com/debuerreotype/debuerreotype/pull/142, or that we no longer ship `/etc/apt/sources.list` for `debian:bookworm` (next stable) but rather prefer the new deb822-style sources file in `/etc/apt/sources.list.d/debian.sources` as `debian:unstable` has been shipping for a while now (https://lists.debian.org/debian-devel-announce/2023/02/msg00000.html).

PR #14028, opened 2023-02-07
Upgrade Zookeeper to 3.8.1 and remove 3.5 (EoL)

31z4:


library/zookeeper
```diff
@@ -26,37 +11,17 @@ Architectures: amd64, arm64v8
 GitCommit: 80d975e5c07f2f1e981fd2d61c9e1dfc3f980b80
 Directory: 3.6.4
 
-Tags: 3.6.3-openjdk
-Architectures: amd64, arm64v8
-GitCommit: 060f64d28d1780373f90ac152bd2538abcf1924c
-Directory: 3.6.3
-
-Tags: 3.7.1, 3.7
```
Slach, line 34: why we remove 3.7 and 3.7.1 ?

yosifkit: I would guess it is because those images were based on the openjdk (Debian) image and it is probably to prevent a surprise "upgrade" to the `temurin` (Ubuntu) based image.

```diff
-Tags: 3.7.1, 3.7, 3.7.1-openjdk
-Architectures: amd64, arm64v8
-GitCommit: 060f64d28d1780373f90ac152bd2538abcf1924c
-Directory: 3.7.1
```

31z4: Right, `3.7` and `3.7.1` are based on completely deprecated `openjdk`. And it is now prohibited to build such images (see https://github.com/docker-library/official-images/actions/runs/4110957035/jobs/7094251503). So I had to remove these tags.

No worries, the tags won't disappear from Docker Hub (see https://github.com/docker-library/official-images#library-definition-files):

> Tags that are removed from a library file do not get removed from the Docker Hub, so that old versions can continue to be available for use, but are not maintained by upstream or the maintainer of the official image.

Slach: thanks for clarify

PR #14037, opened 2023-02-07
Update Vault for 1.12.3, 1.11.7, and 1.10.10

mladlow:
Edit: I think we did merge commits instead of squashed commits at a part of our process, which I am hoping doesn't mess anything up for this PR.


yosifkit:
> Edit: I think we did merge commits instead of squashed commits at a part of our process, which I am hoping doesn't mess anything up for this PR.

Single commit in this PR; it looks correct to me. Merge commits in https://github.com/hashicorp/docker-vault wouldn't affect us; so whatever makes sense for your process.

PR #14051, opened 2023-02-08
caddy: v2.6.3 release

hairyhenderson:
https://github.com/caddyserver/caddy/releases/tag/v2.6.3


francislavoie:
To provide context, we made a change to include `libcap` to `setcap` the Caddy binary to make it easier to run non-root, since Caddy typically binds to ports 80/443 for HTTP/HTTPS.

The `xcaddy` build tool uses `XCADDY_SETCAP=1` to be told to run setcap on the resulting binary. Having this tool do it means no additional changes are needed by users who write their own Dockerfile to make a custom build of Caddy with plugins using the `builder` image.

Issue #13239, opened 2022-09-28
[oraclelinux] 7-slim has 3 high and 1 media vuls


lowryxiao:
Got a image vuls report on oraclelinux:7-slim based on this anchore tool(grype)), the report says there're 3 High SEVERITY vuls and 1 Medium SEVERITY vuls, even the image is from the official latest version:

The anchore tool address:
https://github.com/anchore/grype

`
grype container-registry.oracle.com/os/oraclelinux:7-slim                                                                                                          
 ✔ Vulnerability DB        [no update available]
 ✔ Loaded image
 ✔ Parsed image
 ✔ Cataloged packages      [118 packages]
 ✔ Scanned image           [5 vulnerabilities]

NAME          INSTALLED          FIXED-IN                     TYPE  VULNERABILITY   SEVERITY
openssl-libs  1:1.0.2k-25.el7_9  10:1.0.2k-24.0.3.el7_9_fips  rpm   ELSA-2022-9225  High
openssl-libs  1:1.0.2k-25.el7_9  10:1.0.2k-25.el7_9_fips      rpm   ELSA-2022-9249  High
openssl-libs  1:1.0.2k-25.el7_9  10:1.0.2k-22.el7_9_fips      rpm   ELSA-2021-9528  High
openssl-libs  1:1.0.2k-25.el7_9  10:1.0.2k-23.0.1.el7_9_fips  rpm   ELSA-2022-9023  Medium
`

The sha of this image is:
`
docker images --digests                                                                                                                                        
REPOSITORY                                     TAG              DIGEST                                                                    IMAGE ID       CREATED        SIZE
container-registry.oracle.com/os/oraclelinux   7-slim           sha256:072ae4b680a4e8c87416c8d2caab36f51d7d6cd88ed190e7f8a7047cedd489d5   60807cf6683b   5 days ago     135MB
`

Already used the latest grype db:
`
grype db check                                                                                                                                               
No update available

grype db status                                                                                                                                                 
Location:  /Users/fenxiao/Library/Caches/grype/db/4
Built:     2022-09-28 08:19:43 +0000 UTC
Schema:    4
Checksum:  sha256:772fc073887b77ac9e8876d3acb48f24504a1824a50b5a3eb86749fa88d05ec6
Status:    valid
`

Would you pls help to have a check?
Thanks.

lowryxiao:
Hi @amsteinb-orcl 
Could you pls help to have a check on this issue? Just observed you released recent versions.
Thanks

[...]

PR #13081, opened 2022-09-01
[percona] update product versions

vorsel:
1. update all psmdb versions
2. update ps57 version
3. change base OS for ps-57 from centos8 to ol8
4. remove outdated versions of psmdb-36, psmdb-40
5. add new maintainer and remove old one

PR #14053, opened 2023-02-09

adivinho:


yosifkit:
> +    yum -y update libgcrypt; \

For official images, we generally recommend against updating libraries that are included in the respective base image. Since the item cannot be deleted from an earlier layer, the update can add extra size to the resulting image. In order to have the most up-to-date packages, official images instead rely on periodic base image updates. We then rebuild all dependent images.

`oraclelinux` images are fairly active ([PR history](https://github.com/docker-library/official-images/issues?q=label%3Alibrary%2Foraclelinux+is%3Aclosed)) and so already include the most up-to-date `libgcrypt`. (If the base image is too slow or refuses to update, then a version pinned package update can be acceptable as a temporary fix for things like CVEs.)

yosifkit:
On an unrelated note, I strongly recommend removing psmdb 3.6 and 4.0 in this or the next PR. They fail to build, are based on an unsupported OS distribution, and are deemed EOL on the [Percona website](https://www.percona.com/services/policies/percona-software-support-lifecycle#mongodb). Neither image has been pushed since September 2021 ([Docker hub psmdb-3.6](https://hub.docker.com/_/percona?tab=tags&page=1&name=psmdb-3.6) and [Docker hub psmdb-4.0](https://hub.docker.com/_/percona?tab=tags&page=1&name=psmdb-4.0))

Removing it here will remove it from the "Supported" section on the Hub readme (and will prevent us from spending cycles failing to rebuild it on the official build server), but the tag will still be available to users who want to `docker pull` it.  (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)

---

I also recommend that `percona:5.7.35` be either dropped or moved to a supported OS distribution (like `oraclelinux:8`) as it also does not build successfully. Claiming it is supported is a disservice to users since it cannot be rebuilt to include security updates and has not been pushed since September 2021 (https://hub.docker.com/_/percona?tab=tags&page=1&name=5.7.35-centos).

adivinho:
> `oraclelinux` images are fairly active ([PR history](https://github.com/docker-library/official-images/issues?q=label%3Alibrary%2Foraclelinux+is%3Aclosed)) and so already include the most up-to-date `libgcrypt`. (If the base image is too slow or refuses to update, then a version pinned package update can be acceptable as a temporary fix for things like CVEs.)

It's still use a version with CVE
Screenshot 2023-02-14 at 13 47 57

yosifkit:
`CVE-2021-40528` has been fixed in `oraclelinux` images since https://github.com/docker-library/official-images/pull/12744 (See also https://linux.oracle.com/errata/ELSA-2022-5311.html). It looks like your vulnerability scanner has a false positive since it has conflated the `_fips` versions of the packages with the regular. See also https://github.com/docker-library/official-images/issues/13239.

tianon:
Another approach to verify that this package is already up-to-date and at the appropriate version in the base image:

```console
$ docker run -it --rm --pull=always oraclelinux:8
8: Pulling from library/oraclelinux
Digest: sha256:0d793be7b48033dc2599d538762e564b4bb42b5e2a86c7eb0320d42ec6c15a23
Status: Image is up to date for oraclelinux:8
[root@487cbeacd0af /]# yum update
Oracle Linux 8 BaseOS Latest (x86_64)                                                  63 MB/s |  54 MB     00:00    
Oracle Linux 8 Application Stream (x86_64)                                             61 MB/s |  43 MB     00:00    
Last metadata expiration check: 0:00:09 ago on Tue Feb 14 23:45:45 2023.
Dependencies resolved.
Nothing to do.
Complete!
```

jkabat:
@adivinho we are missing latest updated Percona image for couple months, even half a year already? Would be possible to make changes so it gets approved? Thank you.

tianon:
```diff
+    dnf -y update; \
```

> For official images, we generally recommend against updating libraries that are included in the respective base image. Since the item cannot be deleted from an earlier layer, the update can add extra size to the resulting image. In order to have the most up-to-date packages, official images instead rely on periodic base image updates. We then rebuild all dependent images.
> 
> `oraclelinux` images are fairly active ([PR history](https://github.com/docker-library/official-images/issues?q=label%3Alibrary%2Foraclelinux+is%3Aclosed)) ... (If the base image is too slow or refuses to update, then a version pinned package update can be acceptable as a temporary fix for things like CVEs.)

---

```diff
+    fips-mode-setup --enable; \
```

Did you enable FIPS mode intentionally here?  (I don't think you want FIPS mode generally except/unless you're deploying to the US government)

---

```diff
+ENV PS_VERSION 8.0.32-24.1
+ENV MYSQL_SHELL_VERSION 8.0.32-1
+ENV OS_VER el9
+ENV FULL_PERCONA_VERSION "$PS_VERSION.$OS_VER"
+ENV FULL_MYSQL_SHELL_VERSION "$MYSQL_SHELL_VERSION.$OS_VER"
+ENV PS_REPO testing
```

These variables will cause a cache bust every time they change, and _each_ should generally stay as low in the `Dockerfile` as they can reasonably be defined.

---

```diff
-    dnf -y module disable mysql; \
+#    dnf -y module disable mysql; \
```

Is this commented out code intentionally kept?

---

> On an unrelated note, I strongly recommend removing psmdb 3.6 and 4.0 in this or the next PR. They fail to build, are based on an unsupported OS distribution, and are deemed EOL on the [Percona website](https://www.percona.com/services/policies/percona-software-support-lifecycle#mongodb). Neither image has been pushed since September 2021 ([Docker hub psmdb-3.6](https://hub.docker.com/_/percona?tab=tags&page=1&name=psmdb-3.6) and [Docker hub psmdb-4.0](https://hub.docker.com/_/percona?tab=tags&page=1&name=psmdb-4.0))
> 
> Removing it here will remove it from the "Supported" section on the Hub readme (and will prevent us from spending cycles failing to rebuild it on the official build server), but the tag will still be available to users who want to `docker pull` it. (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)
> 
> I also recommend that `percona:5.7.35` be either dropped or moved to a supported OS distribution (like `oraclelinux:8`) as it also does not build successfully. Claiming it is supported is a disservice to users since it cannot be rebuilt to include security updates and has not been pushed since September 2021 (https://hub.docker.com/_/percona?tab=tags&page=1&name=5.7.35-centos).
> 
> Edit: :facepalm: it looks like #13081 addresses this. It just needs one more change (and a rebase if this PR is merged first).

yosifkit:
```diff
+ENV PS_VERSION 8.0.32-24.1
+ENV MYSQL_SHELL_VERSION 8.0.32-1
+ENV OS_VER el9
+ENV FULL_PERCONA_VERSION "$PS_VERSION.$OS_VER"
+ENV FULL_MYSQL_SHELL_VERSION "$MYSQL_SHELL_VERSION.$OS_VER"
+ENV PS_REPO testing
```
These variables will cause a build cache bust every time they change.

Each should be moved down just before the layer/instruction which uses/requires/changes based on it.

---

```diff
+    dnf -y install crypto-policies-scripts; \
+    dnf -y install epel-release; \
```
These seem left behind from removing the `fips-mode-setup`?

---

Also, slightly worried that a `testing` (pre-release) version is accidentally going out instead of a `release`. If that is correct and expected, then it is fine. Is the testing channel expected for this version (`8.0.32-24.1`)?

adivinho:
> Each should be moved down just before the layer/instruction which uses/requires/changes based on it.

The block has been moved down.

> These seem left behind from removing the fips-mode-setup?

crypto-policies-scripts has been removed but epel-release is still required.

> Also, slightly worried that a testing (pre-release) version is accidentally going out instead of a release. 
> If that is correct and expected, then it is fine. 

Yes, it's expected to use "testing"

yosifkit (approved):

PR #14078, opened 2023-02-13
Updated swipl devel to 9.1.5

JanWielemaker:
Also addresses https://github.com/SWI-Prolog/docker-swipl/issues/32 (armhf/armel builds) by dropping affected plugin on these platforms.

### DIFF
--- a/library/swipl
+++ b/library/swipl
@@ -1,11 +1,11 @@
 Maintainers: Jan Wielemaker <jan@swi-prolog.org> (@JanWielemaker),
              Dave Curylo <dave@curylo.org> (@ninjarobot)
 GitRepo: https://github.com/SWI-Prolog/docker-swipl.git
-GitCommit: 366d912cf8e9c8ba3783a76acf57380d7a281883
+GitCommit: 67b156afb3feb8e0a4537bfb7671662f12526156
 Architectures: amd64, arm32v7, arm64v8
 
-Tags: latest, 9.1.4
-Directory: 9.1.4/bullseye
+Tags: latest, 9.1.5
+Directory: 9.1.5/bullseye
 
 Tags: stable, 9.0.4
 Directory: 9.0.4/bullseye


tianon:
This is good and has the intended effect, but then it fails on `prosqlite` (and then `rserve_client`) :sweat_smile: :joy: :heart: :sob:

I've grabbed the `swipl:latest` source `Dockerfile` locally and built it on `arm32v7` and can confirm the following diff fixes the build (although I don't know what the end resulting effect excluding _all_ these addons is - maybe it's fine?):

```diff
diff --git a/9.1.5/bullseye/Dockerfile b/9.1.5/bullseye/Dockerfile
index 8920a6f..b2fb8c8 100644
--- a/9.1.5/bullseye/Dockerfile
+++ b/9.1.5/bullseye/Dockerfile
@@ -62,9 +62,9 @@ RUN set -eux; \
     }; \
     dpkgArch="$(dpkg --print-architecture)"; \
     [ "$dpkgArch" = 'armhf' ] || [ "$dpkgArch" = 'armel' ] || install_addin space https://github.com/JanWielemaker/space.git 8ab230a67e2babb3e81fac043512a7de7f4593bf; \
-    install_addin prosqlite https://github.com/nicos-angelopoulos/prosqlite.git cfd2f68709f5fb61833c0e2f8e9c6546e542009c; \
+    [ "$dpkgArch" = 'armhf' ] || install_addin prosqlite https://github.com/nicos-angelopoulos/prosqlite.git cfd2f68709f5fb61833c0e2f8e9c6546e542009c; \
     [ "$dpkgArch" = 'armhf' ] || [ "$dpkgArch" = 'armel' ] || install_addin rocksdb https://github.com/JanWielemaker/rocksdb.git 634c31e928e2a5100fbcfd26c21cd32eeb6bf369; \
     [ "$dpkgArch" = 'armhf' ] || [ "$dpkgArch" = 'armel' ] ||  install_addin hdt https://github.com/JanWielemaker/hdt.git e0a0eff87fc3318434cb493690c570e1255ed30e; \
-    install_addin rserve_client https://github.com/JanWielemaker/rserve_client.git 48a46160bc2768182be757ab179c26935db41de7; \
+    [ "$dpkgArch" = 'armhf' ] || install_addin rserve_client https://github.com/JanWielemaker/rserve_client.git 48a46160bc2768182be757ab179c26935db41de7; \
     apt-get purge -y --auto-remove $BUILD_DEPS
 CMD ["swipl"]
```

Did you want to also apply similar fixes to the `swipl:stable` build?

JanWielemaker:
Thanks.   The rserve_client is also C++, so suffering from the same issues.  I thought the prosqlite was plain C.  Do you have the errors?

If this is resolved I'll probably release 9.0.5 (stable) and include the patches there.

tianon:
Oh, I misread the output -- as I pasted it here to share it with you, I noticed the `prosqlite` install actually _does_ succeed and it's just `rserve_client` that's failing. :facepalm: 

Sorry for the false alarm on that half!

JanWielemaker:
Thanks.  So if all goes right there will be armhf docker images for 9.1.5 shortly and I should do the same update for the next 9.0 release.

PR #14081, opened 2023-02-13
Update API-Firewall to v0.6.11

afr1ka:


library/api-firewall
```diff
@@ -2,8 +2,8 @@ Maintainers: Ivan Novikov <in@wallarm.com> (@d0znpp),
...
 GitRepo: https://github.com/wallarm/api-firewall-docker.git
 
-Tags: 0.6.9, latest
+Tags: 0.6.11, latest
 Architectures: amd64, arm64v8, i386
-GitCommit: c2b622e7f59546d9fb338c81fa82d405a8767485
+GitCommit: 8da89a87d29d1b6d5a52e3064722d9a24a4c39fb
 GitFetch: refs/heads/main
-Directory: 0.6.10
+Directory: 0.6.11
```
yosifkit, line 9: Oh, https://github.com/docker-library/official-images/pull/13757 bumped the version in the Dockerfile, but not the image tag so `api-firewall:0.6.9` actually contains `0.6.10` software and that there is no `0.6.10` tag.  Do you want to temporarily re-add `.9` and `.10` to fix their tags on Docker Hub?

```yaml
Tags: 0.6.9
Architectures: amd64, arm64v8, i386
GitCommit: b0579bf17de8f41b162e150b00a87b95648258ee
GitFetch: refs/heads/main
Directory: 0.6.9

Tags: 0.6.10
Architectures: amd64, arm64v8, i386
GitCommit: c2b622e7f59546d9fb338c81fa82d405a8767485
GitFetch: refs/heads/main
Directory: 0.6.10
```

afr1ka: Yes, thanks for the suggestion. I have updated the PR.

PR #14083, opened 2023-02-14
maven: 3.9.0

carlossg:
remove openjdk-18 EOL

### DIFF
--- a/library/maven
+++ b/library/maven
@@ -1,117 +1,107 @@
 Maintainers: Carlos Sanchez <carlos@apache.org> (@carlossg)
 GitRepo: https://github.com/carlossg/docker-maven.git
 
-Tags: 3.8.7-openjdk-18, 3.8-openjdk-18, 3-openjdk-18
-Architectures: amd64, arm64v8
-GitCommit: 34e7d27260ee61c6866922c523e04b53ba098337
-Directory: openjdk-18
-
-Tags: 3.8.7-openjdk-18-slim, 3.8-openjdk-18-slim, 3-openjdk-18-slim
-Architectures: amd64, arm64v8
-GitCommit: 34e7d27260ee61c6866922c523e04b53ba098337
-Directory: openjdk-18-slim
-
-Tags: 3.8.7-eclipse-temurin-11, 3.8-eclipse-temurin-11, 3-eclipse-temurin-11
+Tags: 3.9.0-eclipse-temurin-11, 3.9-eclipse-temurin-11, 3-eclipse-temurin-11
 Architectures: amd64, arm32v7, arm64v8, ppc64le, s390x
-GitCommit: 34e7d27260ee61c6866922c523e04b53ba098337
+GitCommit: 32760466401d45fabb8166b5dcf9003b07598918
 Directory: eclipse-temurin-11
 
-Tags: 3.8.7-eclipse-temurin-11-alpine, 3.8-eclipse-temurin-11-alpine, 3-eclipse-temurin-11-alpine
+Tags: 3.9.0-eclipse-temurin-11-alpine, 3.9-eclipse-temurin-11-alpine, 3-eclipse-temurin-11-alpine
 Architectures: amd64
-GitCommit: 34e7d27260ee61c6866922c523e04b53ba098337
+GitCommit: 32760466401d45fabb8166b5dcf9003b07598918
 Directory: eclipse-temurin-11-alpine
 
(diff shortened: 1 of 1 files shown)


yosifkit:
The reason for the diff job failing is because of the included `KEYS` files. See https://github.com/docker-library/faq#openpgp--gnupg-keys-and-verification -- TLDR, `COPY KEYS ...` is hard for users to verify, annoying during review (since the full exported key file shows up in our diffs), etc; I'd recommend instead pulling the keys via full fingerprint from a PGP keyserver (both so users can verify it with a fingerprint that's hopefully published on an official Maven page and so it gets fully verified during the image build)

carlossg:
thanks, I had that but I though it would be a problem to pull the keys on each build, I'll put it back

yosifkit:
```diff
+  && gpg --batch --keyserver hkps://keyserver.ubuntu.com --recv-keys \
+  6A814B1F869C2BBEAB7CB7271A2A1C94BDE89688 \
+  29BEA2A645F2D6CED7FB12E02B172E3E156466E8 \
```
Just a note that `recv-keys` can still return a successful exit code even if some keys fail to fetch, so we usually recommend to use multiple `gpg --batch --recv-keys` invocations (to keep the failure line as close to the cause as possible). The current Dockerfiles should work since the key would then be immediately used and fail there.

For whatever reason it might fail, we have multiple retries on our build system to combat flaky remote services like keyservers.

tianon:
```diff
+  && echo "${SHA}  apache-maven-${MAVEN_VERSION}-bin.tar.gz" | sha512sum -c - \
```

I'd suggest adding `*` here to put `sha512sum` into "binary" mode:

```diff
+  && echo "${SHA} *apache-maven-${MAVEN_VERSION}-bin.tar.gz" | sha512sum -c - \
```

---

```diff
+  # GNUPGHOME may fail to delete even with -rf
```

If you add `gpgconf --kill all` (in places with `gpgconf` and more one that supports `--kill`) it will succeed _more_ often, but still not a full fix (because it doesn't wait for `dirmngr` to fully die before it returns - it just sends the signal and then there's _still_ a race condition). :sob: :heart:

carlossg:
thanks! I've applied the suggestions. One question, for containers stuck with gpg 2.0 (amazoncorretto with amazon linux and  libericaopenjdk with centos 7) I have an issue as the latest binaries are signed with EDDSA keys not supported. 

I've seen some options like https://people.kernel.org/monsieuricon/run-gnupg-2-2-17-on-your-el7-system but needs some trust
Any suggestions?

https://github.com/carlossg/docker-maven/pull/333#issuecomment-1432887058

carlossg:
one option could be to use multistage builds, first to download and check signature, second to just copy the files

tianon:
Oh, given this isn't a _compiled_ project (where the distro you download/build on might affect the final result), doing `COPY --from=` is pretty reasonable if you wanted to choose one variant as the "canonical" downloading variant and `COPY --from=` in all the rest. :thinking:

We do something similar in `tomcat` for JDK vs JRE and the compiled Tomcat Native bits: https://github.com/docker-library/tomcat/blob/76cf542f30956309447f435a432181bc64b3dce6/9.0/jre8/temurin-jammy/Dockerfile#L26

We also often do this for Windows Nano Server images, where it's complicated to download things, so we build the Server Core version and then `COPY --from=` into Nano Server.

I would suggest making sure the `COPY --from=maven:x.y.z` tagging is as specific as you can stand to automate, but that should work really reasonably and still give us a good chain-of-trust. :+1:

carlossg:
is this setup valid for you?

* one canonical dockerfile (multi step) https://github.com/carlossg/docker-maven/blob/01666296749a28cd08fdc3b66faaee81632185df/eclipse-temurin-11/Dockerfile
* other dockerfiles referencing it https://github.com/carlossg/docker-maven/blob/01666296749a28cd08fdc3b66faaee81632185df/eclipse-temurin-11-focal/Dockerfile

tianon:
I guess you're using multi-stage so that the "builder" image has the same history/shape in the final output as the rest of the images too?

It's not strictly in line with https://github.com/docker-library/faq#multi-stage-builds, but if you add an explicit `Builder: buildkit` to at least that image (doesn't hurt to put globally next to `Maintainers:` if you want and just have it apply to all your images), it should be fine.  We're working towards enabling that for all images by default, but that's still in-progress. :sweat_smile:

carlossg:
I'm using multi-stage to avoid the cleanup of build time dependencies and so the resulting image is the same as the other ones, no sideeffects. But I could do it in one step just fine, as it was before.

tianon:
Naw, I think it's fine :+1:

PR #14096, opened 2023-02-15
Gradle 8.0

keeganwitt:


yosifkit:
:bowing_man: I am unsure of upstream Gradle's actively supported releases. But should this be dropping `7.x` images? Are they no longer supported now that `8.x` is out? What about `6.x`, are they still supported? Are they end of life and should be dropped instead of `7.x`?

library/gradle
```diff
@@ -4,59 +4,59 @@ GitRepo: https://github.com/keeganwitt/docker-gradle.git
 
 # Gradle 7.x
 
-Tags: 7.6.0-jdk8, 7.6-jdk8, 7-jdk8, jdk8, 7.6.0-jdk8-jammy, 7.6-jdk8-jammy, 7-jdk8-jammy, jdk8-jammy
+Tags: 8.0.0-jdk8, 8.0-jdk8, 7-jdk8, jdk8, 8.0.0-jdk8-jammy, 8.0-jdk8-jammy, 7-jdk8-jammy, jdk8-jammy
```
monosoul, line 7: I think this should be a new set of tags rather than replacement of all v7 ones. Also, you missed all "7-" tags anyway :slightly_smiling_face:

keeganwitt:
> 🙇‍♂️ I am unsure of upstream Gradle's actively supported releases. But should this be dropping `7.x` images? Are they no longer supported now that `8.x` is out? What about `6.x`, are they still supported? Are they end of life and should be dropped instead of `7.x`?

They typically don't support both. 6 was the only time I'm aware of that they ever did that. It's unclear whether that will happen to 7. If it does, I may have to add that back in a new branch.

library/gradle
monosoul: And the comment line above if you're going to drop all v7 tags 😉

keeganwitt: Yep, I sure did. Thank you.

library/gradle
```diff
@@ -4,59 +4,59 @@ GitRepo: https://github.com/keeganwitt/docker-gradle.git
```
monosoul, line 4: Call it a nitpick, but this comment is misleading now

keeganwitt:
> > 🙇‍♂️ I am unsure of upstream Gradle's actively supported releases. But should this be dropping `7.x` images? Are they no longer supported now that `8.x` is out? What about `6.x`, are they still supported? Are they end of life and should be dropped instead of `7.x`?
> 
> They typically don't support both. 6 was the only time I'm aware of that they ever did that. It's unclear whether that will happen to 7. If it does, I may have to add that back in a new branch.

Although it'd probably be more consistent to remove the 6 images at this point. When I asked they didn't have specific plans to release any more 6 updates, just whenever some need arises. 7 will likely be in the same boat.

library/gradle
```diff
@@ -2,109 +2,59 @@ Maintainers: Keegan Witt <keeganwitt@gmail.com> (@keeganwitt)
...
 
-Tags: 7.6.0-jdk11-alpine, 7.6-jdk11-alpine, 7-jdk11-alpine, jdk11-alpine
+Tags: 8.0.0-jdk11-alpine, 8.0-jdk11-alpine, 8-jdk11-alpine, jdk11-alpine
 Architectures: amd64
-GitCommit: 24b8697d05b77df7eaacdbb20b8f9cc023d659fc
+GitCommit: 189b3519cdd54bd76f7aed58f9a99ca7deec3512
 Directory: jdk11-alpine
 
-Tags: 7.6.0-jdk17, 7.6-jdk17, 7-jdk17, jdk17, 7.6.0-jdk, 7.6-jdk, 7-jdk, jdk, 7.6.0, 7.6, 7, latest, 7.6.0-jdk17-jammy, 7.6-jdk17-jammy, 7-jdk17-jammy, jdk17-jammy, 7.6.0-jdk-jammy, 7.6-jdk-jammy, 7-jdk-jammy, jdk-jammy, 7.6.0-jammy, 7.6-jammy, 7-jammy, jammy
+Tags: 8.0.0-jdk17, 8.0-jdk17, 8-jdk17, jdk17, 8.0.0-jdk, 8.0-jdk, 8-jdk, jdk, 8.0.0, 8.0, 7, latest, 8.0.0-jdk17-jammy, 8.0-jdk17-jammy, 8-jdk17-jammy, jdk17-jammy, 8.0.0-jdk-jammy, 8.0-jdk-jammy, 8-jdk-jammy, jdk-jammy, 8.0.0-jammy, 8.0-jammy, 8-jammy, jammy
```
yosifkit, line 32: One more missed tag: `8.0, 7, latest,`

keeganwitt: Thank you!

PR #14103, opened 2023-02-15
Update openjdk

tianon:
Changes:

- https://github.com/docker-library/openjdk/commit/e8f4b96: Update 20 to 20


tianon:
From https://openjdk.org/projects/jdk/20/, GA should be sometime around 2023/03/21 :eyes:

PR #14107, opened 2023-02-16
GlassFish Image is back!

dmatej:
Many users (not just [here](https://github.com/eclipse-ee4j/glassfish/issues/22598)) asked for the rescue of the official [GlassFish Docker Image](https://hub.docker.com/_/glassfish), and the time has come.

I don't have an experience with creating the official Docker Images, so please, if I am doing something incorrect, don't kill me ;-)

The repository: https://github.com/eclipse-ee4j/glassfish.docker
(note - the original OmniFish repo was donated to the Eclipse Foundation which is also the owner of Eclipse GlassFish)

# Checklist for Review

GF: OK
GF: OK, [EPL2](https://opensource.org/license/epl-2-0)
GF: OK, service, well known full [Jakarta EE](https://jakarta.ee/) platform implementation.
GF: OK, still popular after all, now growing community again.
-	[ ] does a [documentation](https://github.com/docker-library/docs/blob/master/README.md) PR exist? (should be reviewed and merged at roughly the same time so that we don't have an empty image page on the Hub for very long). 
GF: OK, improving as we have feedback from users.
OK, we followed this, dockerhub team recommendations, feedbacks and even advices from some tools.
GF: TBD, we are still waiting for this. It will be 2 years soon.
GF: OK, we use [Eclipse Temurin JDK](https://hub.docker.com/_/eclipse-temurin) as a base.
GF: OK, not from scratch.
GF: OK, we added start+stop test, other options are tested in our repo.

### DIFF
--- a/library/glassfish
+++ b/library/glassfish
@@ -3,7 +3,7 @@ Maintainers: David Matejcek <david.matejcek@omnifish.ee> (@dmatej),
              Ondro Mihalyi <ondro.mihalyi@omnifish.ee> (@ondromih)
 GitRepo: https://github.com/OmniFish-EE/docker-library-glassfish.git
 GitFetch: refs/heads/main
-GitCommit: 12efd43d7633d89cb28edf3a5b0cf3c12ac31c1f
+GitCommit: 3e22307467c6f9557077f9eb4d481e3ede8df30f
 
 Tags: 7.0.0, 7.0.0-jdk17, 7.0.0-jdk17-eclipse-temurin
 Architectures: amd64
--- a/library/glassfish
+++ b/library/glassfish
@@ -5,10 +5,6 @@ GitRepo: https://github.com/OmniFish-EE/docker-library-glassfish.git
 GitFetch: refs/heads/main
 GitCommit: ac45a85358cd69e26e241d4f10cbe4c47c62cc5f
 
-Tags: 7.0.0, 7.0.0-jdk17, 7.0.0-jdk17-eclipse-temurin
-Architectures: amd64, arm64v8
-Directory: 7.0.0
-
 Tags: 7.0.1, 7.0.1-jdk17, 7.0.1-jdk17-eclipse-temurin
 Architectures: amd64, arm64v8
 Directory: 7.0.1
--- a/library/glassfish
+++ b/library/glassfish
@@ -3,7 +3,7 @@ Maintainers: David Matejcek <david.matejcek@omnifish.ee> (@dmatej),
              Ondro Mihalyi <ondro.mihalyi@omnifish.ee> (@ondromih)
 GitRepo: https://github.com/OmniFish-EE/docker-library-glassfish.git
 GitFetch: refs/heads/main
-GitCommit: e0cdd75be537838a9c54fcec026461e03be334b6
+GitCommit: e631d3d9334fb32d17fc3155453bf2a2f70b3180
 
 Tags: 7.0.1, 7.0.1-jdk17, 7.0.1-jdk17-eclipse-temurin
 Architectures: amd64, arm64v8
@@ -29,6 +29,10 @@ Tags: 7.0.6, 7.0.6-jdk17, 7.0.6-jdk17-eclipse-temurin
 Architectures: amd64, arm64v8
 Directory: 7.0.6
 
-Tags: latest, 7.0.7, 7.0.7-jdk17, 7.0.7-jdk17-eclipse-temurin
+Tags: 7.0.7, 7.0.7-jdk17, 7.0.7-jdk17-eclipse-temurin
 Architectures: amd64, arm64v8
 Directory: 7.0.7
+
+Tags: latest, 7.0.8, 7.0.8-jdk17, 7.0.8-jdk17-eclipse-temurin
+Architectures: amd64, arm64v8
+Directory: 7.0.8
--- a/library/glassfish
+++ b/library/glassfish
@@ -3,7 +3,7 @@ Maintainers: David Matejcek <david.matejcek@omnifish.ee> (@dmatej),
              Ondro Mihalyi <ondro.mihalyi@omnifish.ee> (@ondromih)
 GitRepo: https://github.com/OmniFish-EE/docker-library-glassfish.git
 GitFetch: refs/heads/main
-GitCommit: 36560c745bacc3ac22448dd00e45288322ff8883
+GitCommit: aed9de3cf9e565b5bb71787a810554643b99e993
 
 Tags: 7.0.0, 7.0.0-jdk17, 7.0.0-jdk17-eclipse-temurin
 Architectures: amd64
@@ -17,6 +17,11 @@ Tags: 7.0.2, 7.0.2-jdk17, 7.0.2-jdk17-eclipse-temurin
 Architectures: amd64
 Directory: 7.0.2
 
-Tags: latest, 7.0.3, 7.0.3-jdk17, 7.0.3-jdk17-eclipse-temurin
+Tags: 7.0.3, 7.0.3-jdk17, 7.0.3-jdk17-eclipse-temurin
 Architectures: amd64
 Directory: 7.0.3
+
+Tags: latest, 7.0.4, 7.0.4-jdk17, 7.0.4-jdk17-eclipse-temurin
+Architectures: amd64
+Directory: 7.0.4
+
--- a/library/glassfish
+++ b/library/glassfish
@@ -3,7 +3,7 @@ Maintainers: David Matejcek <david.matejcek@omnifish.ee> (@dmatej),
              Ondro Mihalyi <ondro.mihalyi@omnifish.ee> (@ondromih)
 GitRepo: https://github.com/OmniFish-EE/docker-library-glassfish.git
 GitFetch: refs/heads/main
-GitCommit: 75f8239d93149b1d665e065b289e465d3e4de5d3
+GitCommit: e0cdd75be537838a9c54fcec026461e03be334b6
 
 Tags: 7.0.1, 7.0.1-jdk17, 7.0.1-jdk17-eclipse-temurin
 Architectures: amd64, arm64v8
@@ -25,6 +25,10 @@ Tags: 7.0.5, 7.0.5-jdk17, 7.0.5-jdk17-eclipse-temurin
 Architectures: amd64, arm64v8
 Directory: 7.0.5
 
-Tags: latest, 7.0.6, 7.0.6-jdk17, 7.0.6-jdk17-eclipse-temurin
+Tags: 7.0.6, 7.0.6-jdk17, 7.0.6-jdk17-eclipse-temurin
 Architectures: amd64, arm64v8
 Directory: 7.0.6
+
+Tags: latest, 7.0.7, 7.0.7-jdk17, 7.0.7-jdk17-eclipse-temurin
+Architectures: amd64, arm64v8
+Directory: 7.0.7
--- a/library/glassfish
+++ b/library/glassfish
@@ -3,7 +3,7 @@ Maintainers: David Matejcek <david.matejcek@omnifish.ee> (@dmatej),
              Ondro Mihalyi <ondro.mihalyi@omnifish.ee> (@ondromih)
 GitRepo: https://github.com/OmniFish-EE/docker-library-glassfish.git
 GitFetch: refs/heads/main
-GitCommit: e631d3d9334fb32d17fc3155453bf2a2f70b3180
+GitCommit: 6e52c231a5f50963b9fd91e8206157ae8b87c9e0
 
 Tags: 7.0.1, 7.0.1-jdk17, 7.0.1-jdk17-eclipse-temurin
 Architectures: amd64, arm64v8
@@ -33,6 +33,10 @@ Tags: 7.0.7, 7.0.7-jdk17, 7.0.7-jdk17-eclipse-temurin
 Architectures: amd64, arm64v8
 Directory: 7.0.7
 
-Tags: latest, 7.0.8, 7.0.8-jdk17, 7.0.8-jdk17-eclipse-temurin
+Tags: 7.0.8, 7.0.8-jdk17, 7.0.8-jdk17-eclipse-temurin
 Architectures: amd64, arm64v8
 Directory: 7.0.8
+
+Tags: latest, 7.0.9, 7.0.9-jdk17, 7.0.9-jdk17-eclipse-temurin
+Architectures: amd64, arm64v8
+Directory: 7.0.9
--- a/library/glassfish
+++ b/library/glassfish
@@ -3,7 +3,7 @@ Maintainers: David Matejcek <david.matejcek@omnifish.ee> (@dmatej),
              Ondro Mihalyi <ondro.mihalyi@omnifish.ee> (@ondromih)
 GitRepo: https://github.com/OmniFish-EE/docker-library-glassfish.git
 GitFetch: refs/heads/main
-GitCommit: 6e52c231a5f50963b9fd91e8206157ae8b87c9e0
+GitCommit: c9f6f188fce90e105286261d4977db07e6fc5c9e
 
 Tags: 7.0.1, 7.0.1-jdk17, 7.0.1-jdk17-eclipse-temurin
 Architectures: amd64, arm64v8
@@ -37,6 +37,10 @@ Tags: 7.0.8, 7.0.8-jdk17, 7.0.8-jdk17-eclipse-temurin
 Architectures: amd64, arm64v8
 Directory: 7.0.8
 
-Tags: latest, 7.0.9, 7.0.9-jdk17, 7.0.9-jdk17-eclipse-temurin
+Tags: 7.0.9, 7.0.9-jdk17, 7.0.9-jdk17-eclipse-temurin
 Architectures: amd64, arm64v8
 Directory: 7.0.9
+
+Tags: latest, 7.0.10, 7.0.10-jdk17, 7.0.10-jdk17-eclipse-temurin
+Architectures: amd64, arm64v8
+Directory: 7.0.10
--- a/library/glassfish
+++ b/library/glassfish
@@ -3,7 +3,7 @@ Maintainers: David Matejcek <david.matejcek@omnifish.ee> (@dmatej),
              Ondro Mihalyi <ondro.mihalyi@omnifish.ee> (@ondromih)
 GitRepo: https://github.com/eclipse-ee4j/glassfish.docker.git
 GitFetch: refs/heads/main
-GitCommit: 4507192e8332614e653d9a0394eb5c94955c1bfe
+GitCommit: 6aa8e66761d586e620cc08c18cb2540a0be5ff4b
 
 Tags: 7.0.8, 7.0.8-jdk17, 7.0.8-jdk17-eclipse-temurin
 Architectures: amd64, arm64v8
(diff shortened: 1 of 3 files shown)


dmatej:
Is there some example how to work with default passwords in Dockerfiles constructed for this repository?

dmatej:
@tianon Could you please approve the latest to run those tests again? Is something else yet missing/incorrect except the documentation?
https://github.com/docker-library/official-images/actions/workflows/test-pr.yml?query=event%3Apull_request+branch%3Aglassfish-is-back++

dmatej:
Ok, now it should pass all tests, at least I hope so.
- removed sys user password
- updated run-glassfish script, so now it is possible to replace the asadmin command with another. I did not like it, but after I did it I see it is now much easier to use it :-)
- added one more test, just starts the server and check logs for a known message. Maybe later we can find something more complicated, however it doesn't make sense to test deployments etc, which are already tested in some 40000 tests by the glassfish project.

Note: Tini and similar tools don't work with GF, because of the asadmin implementation. But trap running the stop-domain command works well. There are some details I have to document yet (like stop-domain and docker stop timeouts).

grooverdan:
entrypoint:

`exec "$@"` for the non-asadmin case so the container returns the exit code of the command.

Can GlassFish be explicitly started in the foreground? While some protection with trapping EXIT has been done, there's a wider range of signals to trap. Its usually better if this can be handled by the main process.

On passwords, databases official image generally have a volume in which the server verification of passwords is stored and the entrypoints either:
* general random password, output in log in parse-able format (e.g. [mariadb](https://github.com/MariaDB/mariadb-docker/blob/56ef6d9f842e1ddf50e4359625a6e5cef1748c38/docker-entrypoint.sh#L285))
* env variable to set password on first run
* env variable to allow no-auth

Which recording the password in a persistent volume, after which the password is set.

Convention on other docker official images is:
```
```

Official images generally are allowed to be run under any `docker run --user` rather than fixing it to 1000.

To improve maintainability of Dockerfile/entrypoint other repos e.g. [redis](https://github.com/docker-library/redis):
* Dockerfile.template at top directory, same with entrypoint.
* update.sh script to populate version directories
* generate-stackbrew.sh to create library/glassfish

disclaimer: I just a mariadb maintainer with no glassfish/j2ee experience

dmatej:
> exec "$@" for the non-asadmin case so the container returns the exit code of the command.

Fixed, thanks for the review!

> Can GlassFish be explicitly started in the foreground

It is some kind of a compromise. I have spent nearly a week until I reached at least this point. The `asadmin` is a shell script provided by GF used to work with the GF server. Then `asadmin start-domain` starts JVM using `exec`, which loads the domain.xml and other files and starts another JVM (and sets even different bootstrap classpath); with` --verbose` it inherits IO streams and uses them for logging.
The problem is that it doesn't process signals in combination with docker. I tried to use tiny, other scripts, and finally went with this solution, because it doesn't use any tricks with limited results and side effects, but does the most important thing - nicely stops the instance when I use docker stop, or kills the instance when I use docker kill.

I am trying to figure out some future changes in this behavior not just because of docker, but first we need to cleanup other things, make them more reliable, less fragile, clean. Since GF7 is possible to see startup+shutdown in logs, this helped me much to understand all those "traps" along the path. 

Perhaps I missed something, but then I need a help with this. For sure I will probably write more tests to prevent wrong paths I already visited ...

> On passwords, databases official image

I have already removed the original password of the user account, so the only password is for the GF admin. The context ...

* HTTPS in GlassFish makes password mandatory (and one admin user at least)
* For simplicity there is AS_ADMIN_PASSWORD ARG for the first run. 
* User can later use asadmin commands to create more accounts and passwords, delete them, change them. 
* These passwords are used for the remote access.
* It is also possible to change their storage ie. to the database, but usually nobody does that for admin accounts
* I agree that [your solution](https://github.com/MariaDB/mariadb-docker/blob/56ef6d9f842e1ddf50e4359625a6e5cef1748c38/docker-entrypoint.sh#L285) is much safer, but not much user friendly for people who want to use the image for experiments and tests (user would have to set explicit password). Maybe we will discuss it yet.

> Convention on other docker official images ... /usr/local/bin/ ... docker-entrypoint.sh

Fixed

> run under any docker run --user rather than fixing it to 1000.

Here I am not sure. 
* I had some issues with that in k8s  with different dockerfile
* I have followed the Dockerfile best practices: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user
* I tried to inspire myself by several "similar" images ...
  * Jetty uses it's own user: https://github.com/eclipse/jetty.docker/blob/master/baseDockerfile
    * Bit paradox, I would expect that Jetty should be capable to start with --user
    * crashes, PermissionDenied, whoami failure
  * Tomcat doesn't https://github.com/docker-library/tomcat/blob/master/11.0/jdk17/temurin-jammy/Dockerfile
    * Tomcat started ok with the --user
    * But I don't think this is secure ... 
        ```
    	chmod -R +rX .; 
	  chmod 777 logs temp work; 
        ```
  * TomEE doesn't: https://github.com/tomitribe/docker-tomee/blob/master/TomEE-9.0/jre17/Temurin/ubuntu/plus/Dockerfile
    * TomEE doesn't do that, but also doesn't work with the --user ...
      ``` java.io.FileNotFoundException: /usr/local/tomee/logs/catalina.2023-02-24.log (Permission denied) ```
  * OpenLiberty uses it's own user with id 1001: https://github.com/OpenLiberty/ci.docker/blob/main/releases/23.0.0.1/full/Dockerfile.ubuntu.openjdk17
  * Finally redis uses hardcoded user 999: https://github.com/docker-library/redis/blob/master/7.0/Dockerfile
    * Did I understand you well? Even if not, you forced me to learn many things, which is fine ;-)
* As a result, I tried to google more some recent articles ...
  * This is quite good: https://nickjanetakis.com/blog/running-docker-containers-as-a-non-root-user-with-a-custom-uid-and-gid , so I updated Dockerfiles to allow to set ARG for UID and GID.
  

> Dockerfile.template

Yeah, I know, but I am lazy to write these scripts now. That's also why I started with two versions of GF, I am learning how much I have to change for each version, which results in some routine. I don't plan to do that manually for 10 versions later, on the other hand - I hate unreliable scripts doing just half of the work. This part yet waits for me, probably after the merge or when I will have to add the third version soon.

Btw, MariaDb scripts look really nice and clean, thank you!

dmatej:
Locally executed tests:
```
./run.sh glassfish
testing glassfish
'utc' [1/4]...passed
'no-hard-coded-passwords' [2/4]...passed
'override-cmd' [3/4]...passed
'glassfish' [4/4]...passed
```

arturotena:
It seems that the domain name repo.maven.org is preferred over repo1.maven.org now?

[https://issues.apache.org/jira/browse/MNG-5151](https://issues.apache.org/jira/browse/MNG-5151)

[https://stackoverflow.com/a/36156652](https://stackoverflow.com/a/36156652)

arturotena:
Thank you for answering!

I agree, repo1.maven.org is fine for now.

I just want to add these comments:

> $ ping repo.maven.org
> ping: repo.maven.org: Name or service not known

The usual mitigation for the ICMP ping flood attack is disabling the ICMP functionality. The host does answer HTTPS requests: [https://repo1.maven.org/maven2/org/glassfish/main/distributions/glassfish/](https://repo1.maven.org/maven2/org/glassfish/main/distributions/glassfish/)

The cited improvement issue [MNG-5151](https://issues.apache.org/jira/browse/MNG-5151) do not give enough explanation about why Maven itself changed the hostname, except "use CNAME or repo to provide more stability".

Again, thank you for answering, and thank you for your work on this Docker image!

arjantijms:
@yosifkit wdyt? Can this be merged now?

arjantijms:
@grooverdan any idea what we can do next?

grooverdan:
#docker-library IRC channel on libera.chat at a US based time.

tianon:
Please have patience - "new image" PRs are high effort reviews, and yours is in queue.

dmatej:
> Please have patience - "new image" PRs are high effort reviews, and yours is in queue.

Could you please approve the workflow as the first step?

arjantijms:
@dmatej did you find out who 7.0.0 failed? Can it be fixed?

arjantijms:
> No, locally it passed. Sad I can't see more logs.

Hmmm, so how to continue here?

arjantijms:
@tianon @grooverdan sorry to ask again, but just wondering what the current status is of things.

dmatej:
The failure now happened because I forgot to push the source repo first, sorry. It should pass if rebuilt.

tianon:
Unfortunately, the foreground/background/process supervisor behavior (`"$@" & wait` in the current revision) is still a non-starter (and the biggest blocker here in my very short glance over the Dockerization).

However, I recently had a thought about a potential way forward -- as far as I've understood, the current behavior is because there is a "wrapper" application which is responsible for actually starting GlassFish itself, but Java does not have the primitives necessary to be an actual process supervisor (or do something like `exec` for full foregrounded process replacement).  Is it possible that the wrapper could include a "dry run" mode which outputs a valid line of shell for execution outside the wrapper application?

Some rough pseudocode for how I'd envision that working:

```bash
# outputs something like "java -jar ... --flags --to --application"
java="$(glassfish-wrapper start --dry-run --flags --to --application)"

# executes something like "exec java -jar ....."
eval "exec $java"
```

This would have the desired end result of PID1 inside the container running the actual GlassFish `java` process (and thus being able to manage signals, lifecycle, etc correctly).

dmatej:
> Oh, that's excellent news! Yes, I think removing the older versions that don't support the proper starting behavior would be a prudent path forward. 🙇 ❤️

Ok, so I will do that in the next update - we released GF 7.0.11, I still did not update this PR, so I will do that.

arjantijms:
@tianon Any updates here?

OndroMih:
Hi @tianon, as you requested, we've removed the older versions of GlassFish images that don't support the proper starting behavior. Can you please review again and let us know if there's still something to improve?

dmatej:
Hi, @tianon and @yosifkit , the original OmniFish git repo was donated to the Eclipse Foundation which is also the owner of Eclipse GlassFish. Can we move on now?

dmatej:
Another update, another ignorance. Guys, we did not give up for more than two years, meanwhile we even donated the image to the Eclipse Foundation, resolved all issues, now the image is quite mature, gains new features as requested by users, ... but nothing is happening here. What is missing to approve and merge it?

tianon:
I only reviewed 7.0.23 for now -- honestly we should maybe consider dropping to a single version and doing the backfill as a follow-up instead (the review would certainly be simpler that way).

I didn't put as much time into editorializing (or sorting) my comments here as I normally would, so apologies for the brevity/rough cut.  Hopefully you can make sense of them.

---

```diff
+FROM eclipse-temurin:17.0.14_7-jdk
```

This is too specific and means that every update to Temurin will cause GlassFish to be unsupported.

```diff
+ARG AS_ADMIN_PASSWORD=admin \
+    PATH_GF_PASSWORD_FILE_FOR_CHANGE=/password-change.txt \
+    UID=1000 \
+    GID=1000
```

This is harmless, but odd -- is it actually useful to specify `ARG`s here?  What's the use case?

```diff
+COPY docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
```

This should be moved at least below the `RUN` (but generally `COPY` should be as late as possible).

```diff
+    && apt update \
+    && apt upgrade -y \
+    && apt install -y gpg unzip \
```

These should use `apt-get` (not `apt` - that's still a WIP and throws warnings accordingly), and `upgrade` is a no-go (we prefer to see the base images updated instead).

```diff
+    && echo "Generating password file at ${AS_PASSWORD_FILE} ..." \
+    && set +x \
+    && echo "AS_ADMIN_PASSWORD=${AS_ADMIN_PASSWORD}" > "${AS_PASSWORD_FILE}" \
+    && chown glassfish:glassfish "${AS_PASSWORD_FILE}" \
+    && echo "AS_ADMIN_PASSWORD=" > "${PATH_GF_PASSWORD_FILE_FOR_CHANGE}" \
+    && echo "AS_ADMIN_NEWPASSWORD=${AS_ADMIN_PASSWORD}" >> "${PATH_GF_PASSWORD_FILE_FOR_CHANGE}" \
+    && echo "" >> "${PATH_GF_PASSWORD_FILE_FOR_CHANGE}" \
+    && unset AS_ADMIN_PASSWORD \
```

This is weird -- am I reading correctly that it's doing something with passwords at image build time?

```diff
+    && AS_START_TIMEOUT=120000 asadmin start-domain \
+    && curl  -o /dev/null http://localhost:4848 \
+    && asadmin --passwordfile ${PATH_GF_PASSWORD_FILE_FOR_CHANGE} change-admin-password \
+    && asadmin stop-domain --kill \
+    && AS_START_TIMEOUT=120000 asadmin start-domain \
+    && curl  -o /dev/null http://localhost:4848 \
```

Again, weird - why is it starting GlassFish?  Why is it preseeding a password at image build time?

```diff
+    && chmod +x /usr/local/bin/docker-entrypoint.sh \
```

This effectively copies the file a second time, so this should be applied in Git instead (so that `COPY` simply DTRT natively).

```diff
+if [ -f custom/init.sh ]; then
+  /bin/bash custom/init.sh
+fi
```

Why this instead of just having users use something like `sh -c 'custom/init.sh && exec docker-entrypoint.sh startserv`?  Is there something users can do with this extra functionality that they can't do otherwise? (or is difficult to do otherwise?)

```diff
+if [ -f custom/init.asadmin ]; then
+  asadmin --interactive=false multimode -f custom/init.asadmin
+fi
```

Again, failing to see what makes this unique, especially if the command doesn't need GlassFish already started in order to run successfully?  This is a good use case for a k8s `initContainer`, even.

```diff
+if [ "$1" == 'runembedded' ]; then
+  shift 1
+  if [[ "$SUSPEND" == true ]]
+    then 
+      JVM_OPTS="-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=9009 $JVM_OPTS"
+  elif [[ "$DEBUG" == true ]]
+    then
+      JVM_OPTS="-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=9009 $JVM_OPTS"
+  fi
+  exec java $JVM_OPTS -jar glassfish/lib/embedded/glassfish-embedded-static-shell.jar "$@"
+fi
```

Is this a normal standard "upstream" thing that I'm understanding correctly doesn't have an existing entrypoint script of some kind upstream?  Should it?  (it's a little bit weird to have a synthetic image command implemented here, although not unprecedented)

```diff
+CONTAINER_ALREADY_STARTED="CONTAINER_ALREADY_STARTED_PLACEHOLDER"
+if [ ! -f "$CONTAINER_ALREADY_STARTED" ]
+then
+    touch "$CONTAINER_ALREADY_STARTED" &&
+    rm -rf glassfish/domains/domain1/autodeploy/.autodeploystatus || true
+fi
```

This is weird - what's this about?

```diff
+on_exit () {
+    EXIT_CODE=$?
+    set +e;
+    ps -lAf;
+    asadmin stop-domain --force --kill;
+    exit $EXIT_CODE;
+}
+trap on_exit EXIT
+
+env|sort && "$@" & wait
```

As discussed previously, this is a non-starter.

dmatej:
> I only reviewed 7.0.23 for now -- honestly we should maybe consider dropping to a single version and doing the backfill as a follow-up instead (the review would certainly be simpler that way).
> 
> I didn't put as much time into editorializing (or sorting) my comments here as I normally would, so apologies for the brevity/rough cut. Hopefully you can make sense of them.
> 
> ```diff
> +FROM eclipse-temurin:17.0.14_7-jdk
> ```
> 
> This is too specific and means that every update to Temurin will cause GlassFish to be unsupported.

On the other hand I preferred to have concrete version to have some reproducibility, but yeah, I did not consider that dockerhub removes old images, or maintainers do. Ok, changed to `eclipse-temurin:17-jdk` and `eclipse-temurin:21-jdk` for GF8.

> 
> ```diff
> +ARG AS_ADMIN_PASSWORD=admin \
> +    PATH_GF_PASSWORD_FILE_FOR_CHANGE=/password-change.txt \
> +    UID=1000 \
> +    GID=1000
> ```
> 
> This is harmless, but odd -- is it actually useful to specify `ARG`s here? What's the use case?

This is a "must-have", because GlassFish checks if user does access from the local host, so as absolute minimum we have to create some dummy password, which enables `enable-secured-admin` command and usage of HTTPS. By default the domain1 doesn't have any password, but then GlassFish refuses remote access.

> 
> ```diff
> +COPY docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
> ```
> 
> This should be moved at least below the `RUN` (but generally `COPY` should be as late as possible).

DONE. 

> 
> ```diff
> +    && apt update \
> +    && apt upgrade -y \
> +    && apt install -y gpg unzip \
> ```
> 
> These should use `apt-get` (not `apt` - that's still a WIP and throws warnings accordingly), and `upgrade` is a no-go (we prefer to see the base images updated instead).

DONE. 

> 
> ```diff
> +    && echo "Generating password file at ${AS_PASSWORD_FILE} ..." \
> +    && set +x \
> +    && echo "AS_ADMIN_PASSWORD=${AS_ADMIN_PASSWORD}" > "${AS_PASSWORD_FILE}" \
> +    && chown glassfish:glassfish "${AS_PASSWORD_FILE}" \
> +    && echo "AS_ADMIN_PASSWORD=" > "${PATH_GF_PASSWORD_FILE_FOR_CHANGE}" \
> +    && echo "AS_ADMIN_NEWPASSWORD=${AS_ADMIN_PASSWORD}" >> "${PATH_GF_PASSWORD_FILE_FOR_CHANGE}" \
> +    && echo "" >> "${PATH_GF_PASSWORD_FILE_FOR_CHANGE}" \
> +    && unset AS_ADMIN_PASSWORD \
> ```
> 
> This is weird -- am I reading correctly that it's doing something with passwords at image build time?

Yes, because without that user would not be able to access the admin port.
We are thinking about some improvements in the future which might reduce this boilerplate.

> 
> ```diff
> +    && AS_START_TIMEOUT=120000 asadmin start-domain \
> +    && curl  -o /dev/null http://localhost:4848 \
> +    && asadmin --passwordfile ${PATH_GF_PASSWORD_FILE_FOR_CHANGE} change-admin-password \
> +    && asadmin stop-domain --kill \
> +    && AS_START_TIMEOUT=120000 asadmin start-domain \
> +    && curl  -o /dev/null http://localhost:4848 \
> ```
> 
> Again, weird - why is it starting GlassFish? Why is it preseeding a password at image build time?

Because following commands can be executed just on running GlassFish. It takes just few seconds, so it is not so much pain.

> 
> ```diff
> +    && chmod +x /usr/local/bin/docker-entrypoint.sh \
> ```
> 
> This effectively copies the file a second time, so this should be applied in Git instead (so that `COPY` simply DTRT natively).

Thanks, I did not know we can do that in a single command. Done.

> 
> ```diff
> +if [ -f custom/init.sh ]; then
> +  /bin/bash custom/init.sh
> +fi
> ```
> 
> Why this instead of just having users use something like `sh -c 'custom/init.sh && exec docker-entrypoint.sh startserv`? Is there something users can do with this extra functionality that they can't do otherwise? (or is difficult to do otherwise?)
> 
> ```diff
> +if [ -f custom/init.asadmin ]; then
> +  asadmin --interactive=false multimode -f custom/init.asadmin
> +fi
> ```
> 
> Again, failing to see what makes this unique, especially if the command doesn't need GlassFish already started in order to run successfully? This is a good use case for a k8s `initContainer`, even.

GlassFish has a feature called "multimode", when it can create a batch of asadmin commands which are then executed much faster than if they would be executed one by one. That is why @OndroMih created this separation.

> 
> ```diff
> +if [ "$1" == 'runembedded' ]; then
> +  shift 1
> +  if [[ "$SUSPEND" == true ]]
> +    then 
> +      JVM_OPTS="-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=9009 $JVM_OPTS"
> +  elif [[ "$DEBUG" == true ]]
> +    then
> +      JVM_OPTS="-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=9009 $JVM_OPTS"
> +  fi
> +  exec java $JVM_OPTS -jar glassfish/lib/embedded/glassfish-embedded-static-shell.jar "$@"
> +fi
> ```
> 
> Is this a normal standard "upstream" thing that I'm understanding correctly doesn't have an existing entrypoint script of some kind upstream? Should it? (it's a little bit weird to have a synthetic image command implemented here, although not unprecedented)

This was removed from this image to own, for now distributed just via GitHub Packages.

> 
> ```diff
> +CONTAINER_ALREADY_STARTED="CONTAINER_ALREADY_STARTED_PLACEHOLDER"
> +if [ ! -f "$CONTAINER_ALREADY_STARTED" ]
> +then
> +    touch "$CONTAINER_ALREADY_STARTED" &&
> +    rm -rf glassfish/domains/domain1/autodeploy/.autodeploystatus || true
> +fi
> ```
> 
> This is weird - what's this about?

* https://glassfish.org/docs/8.0.0/application-deployment-guide.html#to-deploy-an-application-or-module-automatically
* https://github.com/eclipse-ee4j/glassfish.docker/pull/2

> 
> ```diff
> +on_exit () {
> +    EXIT_CODE=$?
> +    set +e;
> +    ps -lAf;
> +    asadmin stop-domain --force --kill;
> +    exit $EXIT_CODE;
> +}
> +trap on_exit EXIT
> +
> +env|sort && "$@" & wait
> ```
> 
> As discussed previously, this is a non-starter.

This must be there as GlassFish became bit overengineered over time and this ensures proper reaction on signals independently on how the domain was started. Last years we spent a lot of time covering all possible scenarios; there is multiple options of the GF lifecycle, this joins them all to a single reliable shutdown.

If it is really required, we could remove the `ps -lAf` line, however in the past when we had crisis on production 24/7 system, this line helped a lot to diagnose what happened, more than any other log.

PR #14112, opened 2023-02-17
Update kong library

gszr:
Hello, Docker Library maintainers! 👋🏻 

Changelog for this PR:
- Remove old (< 2.8) versions from the list, so they do not get rebuilt
- Apply a new fix to our arm64 2.8.3 image

Thank you! 🚢 📦 🦍

### DIFF
--- a/library/kong
+++ b/library/kong
@@ -26,13 +26,13 @@ Directory: ubuntu
 Architectures: amd64, arm64v8
 
 Tags: 2.8.3-alpine, 2.8.3, 2.8
-GitCommit: 9538eec15456ce93d69a4a1ae9e379a28fd9040b
+GitCommit: 51604ac58ec41b3a57de95933b23dc3314552073
 GitFetch: refs/tags/2.8.3
 Directory: alpine
 Architectures: amd64, arm64v8
 
 Tags: 2.8.3-ubuntu, 2.8-ubuntu
-GitCommit: 9538eec15456ce93d69a4a1ae9e379a28fd9040b
+GitCommit: 51604ac58ec41b3a57de95933b23dc3314552073
 GitFetch: refs/tags/2.8.3
 Directory: ubuntu
 Architectures: amd64, arm64v8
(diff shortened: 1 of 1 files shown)


gszr:
Fixes https://github.com/Kong/docker-kong/issues/627

gszr:
Hey Maintainers!

I am a bit confused. The `kong:2.8.3-ubuntu` job passed (bionic dep was pulled successfully), but the Naughty is being naughty and failing 😄 Any tips?

yosifkit:
Sorry, the test is a little broken right now. Your PR is fine.

gszr:
@yosifkit
The 2.8.3 label only has an amd64 build, while the 2.8 label has both arm64 and amd64. Any clues? Nothing seems off in the library file itself.

yosifkit:
The `2.8` label can use any old image that was successful as `2.8` whereas the `2.8.3` will only get an image tagged as `2.8.3`. This helps smooth the transition of version bumps when one architecture is ready before others (rather than arches being dropped while the new version builds). The manifest list will be eventually updated once the new architecture+version image is ready.

Builds: https://doi-janky.infosiftr.net/job/multiarch/view/images/view/kong/
Staging repo where arm64v8 kong images are pushed before being combined into the manifest list under `hub.docker.com/_/kong`: https://hub.docker.com/r/arm64v8/kong

gszr:
> The `2.8` label can use the last successful old image that was tagged as `2.8` for a specific architecture whereas the `2.8.3` will only get an image tagged as `2.8.3`. This helps smooth the transition of version bumps when one architecture is ready before others (rather than arches being dropped while the new version builds). The manifest list will be eventually updated once the new architecture+version image is ready.

I did which patch version the `2.8` had before asking -duh! Thanks for the detailed reply.

> Builds: https://doi-janky.infosiftr.net/job/multiarch/view/images/view/kong/ Staging repo where arm64v8 kong images are pushed before being combined into the manifest list under `hub.docker.com/_/kong`: https://hub.docker.com/r/arm64v8/kong

Neat. I will watch the build logs. (Any plans to use "Jenkins Blue Ocean" there? :)

tianon:
Unfortunately, Jenkins Blue Ocean has been really buggy (and the layout is really off-putting) every time I try it out, so we don't plan to use it any time soon.

PR #14119, opened 2023-02-17
Hide diff comments instead of deleting them

tianon:
This is a little bit of an experiment -- hopefully this lets us stop deleting the diff comments, especially since older diff comments are often useful context. :crossed_fingers:


tianon:
Yesssss, it's so good!  :grinning:

yosifkit (approved):

PR #14124, opened 2023-02-20
XWiki 14.10.5 released

vmassol:


yosifkit:
There are no build context changes here; perhaps it is missing a `GitCommit:` change to go along with the `Tags:` bumps?

vmassol:
Indeed I forgot the commit changes... fixing

vmassol:
@yosifkit thanks for the check. It should be better now!

PR #14158, opened 2023-02-27
Replace AL2022 images with AL2023

sktomer:
Hi,

This PR includes the pilot container image for AL 2023.
AL 2023 is a rename of the formerly known OS variant AL 2022.
AL 2023 (like AL 2022) is still in tech-preview phase and is not recommended for production workloads.

Regards,
Sumit


yosifkit:
The `curl` could use a `-f`, but otherwise looks good to me.

sktomer:
We will make that change in the next release.

PR #14167, opened 2023-02-28
kong: update 2.8.3

gszr:
Similarly to https://github.com/docker-library/official-images/pull/14112, correct shas, this time for Ubuntu.

We apologize for opening an extra PR, but we only found out the Ubuntu image is affected after #14112 got merged.

### DIFF
--- a/library/kong
+++ b/library/kong
@@ -26,13 +26,13 @@ Directory: ubuntu
 Architectures: amd64, arm64v8
 
 Tags: 2.8.3-alpine, 2.8.3, 2.8
-GitCommit: 51604ac58ec41b3a57de95933b23dc3314552073
+GitCommit: 7557a360568fb70650d35724462446064ec081b3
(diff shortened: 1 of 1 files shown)


gszr:
Hello, maintainers o/

In case of failures like seen here[1], is there a way for us to get notified about the failure?

[1] https://doi-janky.infosiftr.net/job/multiarch/view/images/view/kong/job/arm64v8/job/kong/

yosifkit:
> is there a way for us to get notified about the failure?

We don't have anything automated right now. We try to review build failures periodically and reach out to the relevant maintainer, but it is not as often as we'd like it to be.

tianon:
What we do for our own repositories is have a set of GitHub Actions that do "smoke test" builds (and run the tests from this repository on the resulting images), and we run that on every commit and on a schedule every week too, which helps us catch high-level cases like this one. :bow:

PR #14188, opened 2023-03-03
Gradle 8.0.2, 7.6.1, and 6.9.4

keeganwitt:


keeganwitt:
It seems they're not done with the 6.x line. Which is not what I expected, given my previous discussion with them 🤷

PR #14196, opened 2023-03-06
Update XWiki LTS to 14.10.6 and change base back to tomcat:9-jre11

michitux:


michitux:
I've just updated all images such that they're now based again on `tomcat:9-jre11` (in addition to the update of the LTS version). The main regression that impacted XWiki has been fixed in Tomcat 9.0.71 (current version is 9.0.73) so no need to stick with 9.0.69.

tianon:
Nice!

PR #14202, opened 2023-03-07
Update Vault for 1.13.0, 1.12.4, 1.11.8, and 1.10.11

anwittin:


tianon:
I'm going to merge this for now, but the CI failure is because of the branch rename, so you'll want to update the `library/vault` file in the future to include `GitFetch: refs/heads/main` (probably up by `GitRepo:` is where I'd recommend putting it).

Issue #14136, opened 2023-02-22
Provide clickhouse/clickhouse-server image as the official image


Felixoid:
Hello dear team.

For some time we want to provide the ClickHouse DB as an official image. The image https://hub.docker.com/r/clickhouse/clickhouse-server/ is intensively tested and rebuilt regularly for multiple architectures.

It's built for every release automatically from https://github.com/ClickHouse/ClickHouse/tree/master/docker/server.

So, I'd like to start the review process. What do you think will be the appropriate next step for it?

tianon:
You'll want to read through https://github.com/docker-library/official-images#contributing-to-the-standard-library (and https://github.com/docker-library/official-images/blob/master/NEW-IMAGE-CHECKLIST.md), and then open a PR to start the review process in earnest.  If you have specific questions or problems, I'm happy to do my best to answer them.

Felixoid:
Sorry for not getting back to you sooner, but I have a question.

As I get it, the PR should contain a new file `library/clickhouse` with some content. But as well I need to create the PR to `docker-library/docs` repo. And there it looks like we need some repository with a dedicated tree for an image. Or we could use the same directory as I've shown above?

How the process of releasing new versions looks like? I think, the [ubuntu](https://github.com/docker-library/official-images/blob/master/library/ubuntu) content can be used as an example. So we have some active releases, and update it [regularly](https://github.com/ClickHouse/ClickHouse/releases). On each new release, we build and push new [tags](https://hub.docker.com/r/clickhouse/clickhouse-server/tags). I assume it will look completely different. Right?

Felixoid:
I think https://github.com/docker-library/official-images/blob/master/library/golang is a much better example, so we can use the same repository.

On the other hand, we need more directories in https://github.com/ClickHouse/ClickHouse/tree/master/docker/server for the active releases, right? It's impossible to use build arguments during the build, and every Dockerfile should be launched as just 'docker build .`?

tianon:
Yeah, `ubuntu` is definitely a bad example for your use case and `golang` is a better one :sweat_smile:

Correct on creating `library/clickhouse` and a `clickhouse` directory in `docker-library/docs` (with no `README.md` in it as that file is auto-generated from the other files).

https://github.com/docker-library/faq#an-images-source-changed-in-git-now-what might actually be a pretty helpful flow for you to get a full sense of what "maintenance" looks like (it doesn't talk about `docs` but those I think are more straightforward) :eyes:

[...]

PR #14213, opened 2023-03-08
Add a note about possible alternative Dockerfile

Felixoid:
During a discussion in #14136 I've found, that important option is missing in the readme.


tianon:
This was intentionally omitted (as we don't actually want to encourage maintainers to use it), but including it is fair.

PR #14223, opened 2023-03-09
maven: refactor build process

carlossg:
to have one image as base and others using --copy-from

Following advice from https://github.com/docker-library/official-images/pull/14083#issuecomment-1433740677

### DIFF
--- a/library/maven
+++ b/library/maven
@@ -63,7 +63,7 @@ GitCommit: ba0653e9cbde55b7806b4b99a5afcbf7ca8485d7
 Directory: eclipse-temurin-8-focal
 
 Tags: 3.9.0-ibmjava-8, 3.9.0-ibmjava, 3.9-ibmjava-8, 3.9-ibmjava, 3-ibmjava-8, 3-ibmjava, ibmjava
-Architectures: amd64, i386, ppc64le, s390x
+Architectures: amd64, ppc64le, s390x
 GitCommit: ba0653e9cbde55b7806b4b99a5afcbf7ca8485d7
 Directory: ibmjava-8
 
--- a/library/maven
+++ b/library/maven
@@ -1,107 +1,108 @@
 Maintainers: Carlos Sanchez <carlos@apache.org> (@carlossg)
+Builder: buildkit
 GitRepo: https://github.com/carlossg/docker-maven.git
 
 Tags: 3.9.0-eclipse-temurin-11, 3.9-eclipse-temurin-11, 3-eclipse-temurin-11
 Architectures: amd64, arm32v7, arm64v8, ppc64le, s390x
-GitCommit: 32760466401d45fabb8166b5dcf9003b07598918
+GitCommit: de70b3e783c1c9b2835b6be4c3826d6221bf4fca
 Directory: eclipse-temurin-11
 
 Tags: 3.9.0-eclipse-temurin-11-alpine, 3.9-eclipse-temurin-11-alpine, 3-eclipse-temurin-11-alpine
 Architectures: amd64
-GitCommit: 32760466401d45fabb8166b5dcf9003b07598918
+GitCommit: ba0653e9cbde55b7806b4b99a5afcbf7ca8485d7
 Directory: eclipse-temurin-11-alpine
 
 Tags: 3.9.0-eclipse-temurin-11-focal, 3.9-eclipse-temurin-11-focal, 3-eclipse-temurin-11-focal
 Architectures: amd64, arm32v7, arm64v8, ppc64le, s390x
-GitCommit: 32760466401d45fabb8166b5dcf9003b07598918
+GitCommit: ba0653e9cbde55b7806b4b99a5afcbf7ca8485d7
 Directory: eclipse-temurin-11-focal
 
 Tags: 3.9.0-eclipse-temurin-17, 3.9.0, 3.9.0-eclipse-temurin, 3.9-eclipse-temurin-17, 3.9, 3.9-eclipse-temurin, 3-eclipse-temurin-17, 3, latest, 3-eclipse-temurin, eclipse-temurin
 Architectures: amd64, arm32v7, arm64v8, ppc64le, s390x
-GitCommit: 32760466401d45fabb8166b5dcf9003b07598918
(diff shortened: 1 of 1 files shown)


carlossg:
need to check why it is failing as naughty is passing locally

```
❯ ./naughty-from.sh maven
 - maven:3.9.0-amazoncorretto-11 (FROM amazoncorretto:11) [arm32v7, ppc64le]
 - maven:3.9.0-amazoncorretto-17 (FROM amazoncorretto:17) [arm32v7, ppc64le]
 - maven:3.9.0-amazoncorretto-19 (FROM amazoncorretto:19) [arm32v7, ppc64le]
 - maven:3.9.0-amazoncorretto-8 (FROM amazoncorretto:8) [arm32v7, ppc64le]
 - maven:3.9.0-eclipse-temurin-11-alpine (FROM eclipse-temurin:11-jdk-alpine) [arm32v7, arm64v8, ppc64le]
 - maven:3.9.0-eclipse-temurin-17-alpine (FROM eclipse-temurin:17-jdk-alpine) [arm32v7, arm64v8, ppc64le]
 - maven:3.9.0-eclipse-temurin-19-alpine (FROM eclipse-temurin:19-jdk-alpine) [arm32v7, arm64v8, ppc64le]
 - maven:3.9.0-eclipse-temurin-8-alpine (FROM eclipse-temurin:8-jdk-alpine) [arm32v7, arm64v8, ppc64le]
 - maven:3.9.0-ibm-semeru-11-focal (FROM ibm-semeru-runtimes:open-11-jdk-focal) [arm32v7]
 - maven:3.9.0-ibm-semeru-17-focal (FROM ibm-semeru-runtimes:open-17-jdk-focal) [arm32v7]
 - maven:3.9.0-ibmjava-8 (FROM ibmjava:8-sdk) [arm32v7, arm64v8]
 - maven:3.9.0-sapmachine-11 (FROM sapmachine:11) [arm32v7]
 - maven:3.9.0-sapmachine-17 (FROM sapmachine:17) [arm32v7]
```

yosifkit:
Any output from `naughtly-from.sh` means that there is a problem. Naughty from is meant to point out when there is too many architectures on a child image compared to its parent. We can't build an  image if the `FROM` doesn't exist for that architecture. For example, `sapmachine:*` does not have an `arm32v7` image, so the `maven:*sapmachine*` images can't claim to support the architecture:

```console
 - maven:3.9.0-sapmachine-11 (FROM sapmachine:11) [arm32v7]
 - maven:3.9.0-sapmachine-17 (FROM sapmachine:17) [arm32v7]
```

https://github.com/docker-library/official-images/blob/3133b5d7a212ff845c439435b85c6bc1cbc4f6a8/library/sapmachine#L4-L10

carlossg:
Thanks, I got it now, `naughtly-from.sh` is not returning an error, just printing the bad tags. 
I've found the issue and fixed it.

The problem now is with ibmjava-8

```
 - maven:3.9.0-ibmjava-8 (FROM maven:3.9.0-eclipse-temurin-11) [i386]
 ```

IIUC ibmjava-8 has a i386 arch but the source image where I'm copying from (eclipse-temurin-11) doesn't 
In fact none of the other java images provide i386 so I might just as well ditch that arch

**Comment by carlossg** [comment]
removed ibmjava-8 i386 arch

**Comment by tianon** [comment]
In `maven:3-eclipse-temurin-19`, it appears you're missing a few bits of the `apt-get` invocation that you have elsewhere, was that intentional?  (no strong opinion, but it stood out while I reviewed)

```diff
+RUN apt-get update \
+  && apt-get install -y ca-certificates curl git
...
-  && apt-get update \
-  && apt-get install -y ca-certificates curl git gnupg dirmngr --no-install-recommends \
-  && rm -rf /var/lib/apt/lists/* \
```

vs (in the other `apt-get`-using images):

```diff
+RUN apt-get update \
+  && apt-get install -y ca-certificates curl git --no-install-recommends \
+  && rm -rf /var/lib/apt/lists/*
```

---

Even less "blocking" comments:

We'd suggest combining the two `RUN` lines that `curl` the artifacts and `gpg --batch --verify` them so that if they fail to verify, any retries will definitely redownload (otherwise we might have a bad download that then gets cached and the build just fails over and over again with no chance to get better unless/until cache is purged).

Also, `for` inside an `&&` chain has somewhat pathological behavior, so it'd probably be better to use `;` and `set -e` in that block to ensure things fail appropriately (in this case, if we fail to fetch the first key and it's the one that signed the release, the line would still fail and be retried, so it's not a big problem).

carlossg:
> In maven:3-eclipse-temurin-19, it appears you're missing a few bits of the apt-get invocation that you have elsewhere, was that intentional?

good catch

About the `for`, something like this?

```
RUN curl -fsSLO --compressed ${BASE_URL}/apache-maven-${MAVEN_VERSION}-bin.tar.gz \
  && echo "${SHA} *apache-maven-${MAVEN_VERSION}-bin.tar.gz" | sha512sum -c - \
  && curl -fsSLO --compressed ${BASE_URL}/apache-maven-${MAVEN_VERSION}-bin.tar.gz.asc \
  && export GNUPGHOME="$(mktemp -d)"; \
  set -e; for key in \
  6A814B1F869C2BBEAB7CB7271A2A1C94BDE89688 \
  29BEA2A645F2D6CED7FB12E02B172E3E156466E8 \
  ; do \
  gpg --batch --keyserver hkps://keyserver.ubuntu.com --recv-keys "$key" ; \
  done; \
  gpg --batch --verify apache-maven-${MAVEN_VERSION}-bin.tar.gz.asc apache-maven-${MAVEN_VERSION}-bin.tar.gz
RUN mkdir -p ${MAVEN_HOME} ${MAVEN_HOME}/ref \
  && tar -xzf apache-maven-${MAVEN_VERSION}-bin.tar.gz -C ${MAVEN_HOME} --strip-components=1 \
  && ln -s ${MAVEN_HOME}/bin/mvn /usr/bin/mvn
```

tianon:
Yeah, something like that is probably fine, although I'd personally put the `set -e` at the very beginning of the `RUN` to make sure it catches everything (I'd even go as far as `set -eux`, but I like me some verbose builds :joy:).

carlossg:
done

PR #14225, opened 2023-03-09
feat: release InfluxDB Enterprise 1.9.9

bnpfeife:


yosifkit:

`curl` invocations need a `-f` so that they fail appropriately.

I think this should be `rm -rf /var/lib/apt/lists/*` since the `lists` folder already exists in the base image.

To exemplify good security practices in the Dockerfile, it is better to keep the older, separate/temporary `GNUPGHOME`. A single, persistent `GNUPGHOME` means that either key could be used to validate the detached signatures. With official images' build cache, it would require more than just a key compromise and interception attack, but for regular users, they add extra risk by following the practice of a single `GNUPGHOME` for package verification.

bnpfeife:
Hi, I believe the updates address everything that you mentioned in your comment. -- I have separated out the `GNUPGHOME` directories. I do think that it would take a substantial amount of effort to compromise a singular `GNUPGHOME` directory containing multiple public keys. An attacker would need access to one of the secret keys and also have access to where the another uploads packages/archives/etc.

PR #14252, opened 2023-03-15
Workaround to make Storm workers run in JDK 11 and above

31z4:
See more details at https://github.com/31z4/storm-docker/issues/39


yosifkit:
Just a quick note that this will increase the size of the image by the size of the two `.jar` files (4.9MB x 2) since they were downloaded/created in an earlier layer.

31z4:
@yosifkit good catch! I'm not sure if this could be done more efficiently though. I assume any kind of binary patching of the JARs is not an option because it's not transparent.

PR #14272, opened 2023-03-16
New image: bun

Electroid:
# Bun

[Bun](https://bun.sh/) is a new JavaScript and TypeScript runtime.

# Checklist for Review

> I work for Oven, the company behind Bun
> MIT
> Language stack for JavaScript and TypeScript
> Yes and yes
> TBD
> Yes, though if we missed anything, let us know
> TBD
> FROM debian
-	[x] if `FROM scratch`, tarballs only exist in a single commit within the associated history?
> N/A
> TBD


Electroid:
Hi @tianon, @yosifkit, ready for review and workflow approval at your convenience.

connerlphillippi:
@tianon @yosifkit following up here - is this on your radar? It's been over 2 years. What do we need to do on our side to push this forward Thanks! :)

tianon:
Thanks for the gentle ping (and the patience, it's really appreciated :heart:) :bow:

Here's some high-level review of what's here to get things started (I imagine there's probably some updates/changes that have already happened that we'll review in the delta):

---

Tags might want to be more explicit about *which* debian instead of just debian/debian-slim (ie bookworm vs the near-release trixie or debian 12 vs 13 if you prefer) -- even if you provide generic "latest" aliases like this, it helps users during the transition period (when the "square wheel" rolls from one release to another and inevitably causes some breakage).

---

Is `debian:bullseye` vs `debian:bullseye-slim` really relevant enough to Bun users to justify maintaining separate variants? (most other images have a "slim" variant because the default / non-"slim" variant is from the batteries-included `buildpack-deps` images for aggressive header file / libraries inclusion) -- I would imagine that *just* `debian:bullseye-slim` is probably sufficient and there's not a strong reason to have the other at all unless you want to swap the default to `buildpack-deps` which would then match `node` closer (but comes with a whole different set of tradeoffs that may or may not match how users use `bun` / how modules/extensions work in the `bun` ecosystem).

---

I'm not sure a multi-stage build is really [warranted here](https://github.com/docker-library/faq#multi-stage-builds) -- if the goal is just to remove the packages used for building, that's pretty easy to do reasonably cleanly (and I'm happy to help with that, if necessary).

---

`apt-get clean` is a no-op in the `debian` (and `ubuntu`) images: https://github.com/debuerreotype/debuerreotype/blob/c9542ab785e72696eb2908a6dbc9220abbabef39/scripts/debuerreotype-minimizing-config#L87-L109

---

What's `gpg --update-trustdb` doing?  Is it really necessary?  (you probably also want to use a temporary `GNUPGHOME` and clean it up after verification, although that's not really critical)

---

`which` -> `command -v` 👀 (https://salsa.debian.org/debian/debianutils/-/commit/3a8dd10b4502f7bae8fc6973c13ce23fc9da7efb)

Although alternatively, it could be removed completely, as `bun --version` is already verifying that `bun` exists in the `PATH`.

PR #14273, opened 2023-03-17
bump to Kong 3.2.2

kikito:


yosifkit:
:thinking: :confused: The removal of `curl tar gzip ca-certificates` is being removed in the Alpine based image for `3.2.2`, so these packages will now be included in the final image (instead of being cleaned up before the end of the `RUN` via the `apk del`, https://github.com/Kong/docker-kong/pull/626). Is that intended?

I'm not certain how including more packages in the final image helps with the linked issue about a CVE for `pkg-config` (which was referencing an unmaintained image, `3.1.0` when [`3.1.1`](https://hub.docker.com/_/kong/tags?page=1&name=3.1.1) was already [available](https://github.com/docker-library/official-images/pull/13713) and has the referenced fix).

---

Boilerplate information about CVE's and rebuilds in Docker Official Images:

Background: 
>

Official Images FAQ:

> Though not every CVE is removed from the images, we take CVEs seriously and try to ensure that images contain the most up-to-date packages available within a reasonable time frame
>

Since our build system makes heavy use of Docker build cache, just rebuilding the all of the Dockerfiles won't cause any change. So we rely on periodic base image updates.

>
> \- from the same FAQ link

fffonion:
@yosifkit That was an overlook from my side, we should revert the change that the PR at kong/docker-kong that removes `--virtual .build_deps`. But keep the removal of `pcre` from dockerfile. Thank you for the review!
cc @kikito .

kikito:
We have changed our images according to the feedback. Please re-review and thank you.

library/kong
```diff
@@ -1,13 +1,26 @@
 Maintainers: Kong Docker Maintainers <team-gateway@konghq.com> (@team-gateway-bot)
 GitRepo: https://github.com/Kong/docker-kong.git
 
-Tags: 3.1.1-alpine, 3.1.1, 3.1, 3, alpine, latest
+
+Tags: 3.2.2-alpine, 3.2.2, 3.2, 3, alpine, latest
+GitCommit: 5641f8836920650fc66c6d36408daf794d730b96
+GitFetch: refs/tags/3.2.2
+File: Dockerfile.apk
+Architectures: amd64, arm64v8
```
yosifkit, line 9: This block of the library file still claims that this image (`kong:3.2.2-alpine`) should be built on `arm64v8` but the Dockerfile has changed to no longer support `arm64v8`:

```diff
diff --git a/kong_latest/Dockerfile.apk b/kong_latest/Dockerfile.apk
index cf5443f..e0aeaa1 100644
--- a/kong_latest/Dockerfile.apk
+++ b/kong_latest/Dockerfile.apk
@@ -1,13 +1,15 @@
-FROM alpine:3.16
+FROM alpine:3.17
 
 LABEL maintainer="Kong Docker Maintainers <docker@konghq.com> (@team-gateway-bot)"
 
-ARG KONG_VERSION=3.1.1
+ARG KONG_VERSION=3.2.2
 ENV KONG_VERSION $KONG_VERSION
 
-ARG KONG_AMD64_SHA="9194977814e91b6997ce4795957d9b4bcec9d220667e01a08ade7ac64704f57c"
-ARG KONG_ARM64_SHA="92d98307a3193602bbe4dbed2d099cde5404a8acd09d53f0b9a9f2bcf6038500"
+ARG KONG_SHA256="a07c3bc0307e2d8fe33acb8be5fe659f348279540eaa267bc6968005094835ef"
+
+ARG KONG_PREFIX=/usr/local/kong
+ENV KONG_PREFIX $KONG_PREFIX
 
 ARG ASSET=remote
 ARG EE_PORTS
@@ -16,28 +18,24 @@ COPY kong.apk.tar.gz /tmp/kong.apk.tar.gz
 
 RUN set -ex; \
     apk add --virtual .build-deps curl tar gzip ca-certificates; \
-    arch="$(apk --print-arch)"; \
-    case "${arch}" in \
-      x86_64) export ARCH='amd64'; KONG_SHA256=$KONG_AMD64_SHA ;; \
-      aarch64) export ARCH='arm64'; KONG_SHA256=$KONG_ARM64_SHA ;; \
-    esac; \
+    export ARCH='amd64'; \
...
```

Do you want to update the Dockerfile to re-include `arm64v8` or just drop `arm64v8` support from this version in the `library/kong` file?

kikito: We have changed the way we build kong and for now we can't build Alpine arm on the new system.
I missed that the architecture was still mentioned in `library/kong`. I have removed it. Thanks for the re-review and apologies for the hassle.

yosifkit: No worries👍, glad we caught it. 😅😌

vtlkvl:
@yosifkit Do you know why alpine arm64 arch was dropped and is it one time or applies to future releases too?

fffonion:
@vtlkvl I can answer this question from our side, the alpine arm64 removal would apply for future releases as well. More specifically, in next two versions alpine support will be completely dropped. See `Deprecations` under https://docs.konghq.com/gateway/changelog/#3210.

```
Deprecated Alpine Linux images and packages.

Kong is announcing our intent to remove support for Alpine images and packages later this year. These images and packages are available in 3.2 and will continue to be available in 3.3. We will stop building Alpine images and packages in Kong Gateway 3.4.
```

PR #14303, opened 2023-03-21
Update Nextcloud tests

J0WI:
Recent versions of Nextcloud tend to consume more time to install. This shouldn't increase the test time but hopefully reduce the flakiness.

PR #14304, opened 2023-03-21
Update Nextcloud

J0WI:


tianon:
(Just a rebase so it can run against #14303's updates)

PR #14305, opened 2023-03-21
Adding Corretto 20 images

Rudometov:


tianon:
Doh, CI failure looks like it's unintended fallout from https://github.com/docker-library/official-images/pull/14158? (AL2022 is no longer supported)

PR #14308, opened 2023-03-22
maven: 3.9.1

carlossg:
I have realized that there is a chicken and egg problem. I cannot build the other variants without the eclipse-temurin-11 and this will fail as references no longer exist.
Any good solution used in other images ?

### DIFF
--- a/library/maven
+++ b/library/maven
@@ -2,107 +2,132 @@ Maintainers: Carlos Sanchez <carlos@apache.org> (@carlossg)
 Builder: buildkit
 GitRepo: https://github.com/carlossg/docker-maven.git
 
-Tags: 3.9.0-eclipse-temurin-11, 3.9-eclipse-temurin-11, 3-eclipse-temurin-11
+Tags: 3.9.1-eclipse-temurin-11, 3.9-eclipse-temurin-11, 3-eclipse-temurin-11
 Architectures: amd64, arm32v7, arm64v8, ppc64le, s390x
-GitCommit: de70b3e783c1c9b2835b6be4c3826d6221bf4fca
+GitCommit: f99b8cd582524a58a73b23b97e9b8c17001573af
 Directory: eclipse-temurin-11
 
-Tags: 3.9.0-eclipse-temurin-11-alpine, 3.9-eclipse-temurin-11-alpine, 3-eclipse-temurin-11-alpine
+Tags: 3.9.1-eclipse-temurin-11-alpine, 3.9-eclipse-temurin-11-alpine, 3-eclipse-temurin-11-alpine
 Architectures: amd64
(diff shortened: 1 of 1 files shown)


yosifkit:
So, for updates to official images, it needs to be self-contained so that all the `FROM`'s resolve to another current official image. i.e all images version bump together and `bashbrew` can figure out build order using the `library/` files and respective Dockerfiles.

For images we directly maintain and here for PRs, the GitHub actions job also uses `bashbrew` to figure out the build order and group the images with all dependents. The GitHub action along with an example [yaml](https://github.com/docker-library/bashbrew/blob/6e7f23523161440f980bac3c26773ef666de855a/scripts/github-actions/example-ci.yml) is in the bashbrew repo, but I don't know if it would fit your needs.

You can see that it created one job on your last PR since they are all dependent on the one parent: https://github.com/docker-library/official-images/actions/runs/4439248693/jobs/7791477885.

---

Ah, it looks like you are also publishing the images via GitHub actions to GHCR and your personal docker hub account. I think in order to have parallel jobs to test/build/push images individually, you'd have to first build the parent image and then share that to the other jobs (via a registry?) or have every job build that base one if it doesn't exist remotely.

carlossg:
ok, I'm moving forward for now, need to rewrite my tests to account for the chicken and egg situation.

This PR includes

*    remove eclipse-temurin-19

*    add amazoncorretto debian-slim variants

*    add eclipse-temurin and amazoncorretto 20

yosifkit:
Rather than only downloading a key, we recommend to at least run it through `gpg` to ensure that only the expected key is used; similar to `amazoncorretto`'s al2 image: https://github.com/corretto/corretto-docker/blob/564b34f09563bf5ad704aff7db5eb9a403013cf2/11/jdk/al2/Dockerfile#L17-L20

---

🤔 I'm not certain about the `-debian-slim` tags; from my perspective, they don't quite fit with the `focal` tags. But that is up to you.

---

The `FROM debian:stable-slim` will likely cause build failures when the "square wheel" of stable moves to the next release (likely very soon). We recommend using a specific release and then testing and choosing when to move to a different release.

---

> Architectures: amd64, arm32v7, arm64v8, ppc64le, s390x

This architecture list for the `debian` based images, while correct for the base image, will not work for the Amazon Corretto apt repo as it only supports three architectures: "amd64 arm64 i386" (https://apt.corretto.aws/dists/stable/Release).

carlossg:
thanks @yosifkit I'm removing the corretto debian-slim ones for now and will work on them

Soupison7:
Can it be maintained as previously stated

yosifkit:
I think it can make sense to have a plain `-debian` tag along with `-debian-bullseye` for users to choose what their change tolerance is 👍.

PR #14347, opened 2023-03-27
Update phpMyAdmin image to include timezone setting

ibennetch:
We have added an environment variable to allow users to configure their timezone within the container.

There is no update to the underlying phpMyAdmin application, this is purely a docker image improvement.


williamdes (commented):
LGTM

PR #14371, opened 2023-03-30
Upgrade SonarQube to 10.0

carminevassallo:
Dear team,

we just released SonarQube 10.0. Please merge this PR to have it released as Docker image.

### DIFF
--- a/library/sonarqube
+++ b/library/sonarqube
@@ -1,29 +1,45 @@
-Maintainers: Christophe Levis <christophe.levis@sonarsource.com> (@christophelevis),
-             Pierre Guillot <pierre.guillot@sonarsource.com> (@pierre-guillot-gh),
-             Lukasz Jarocki <lukasz.jarocki@sonarsource.com> (@lukasz-jarocki-sonarsource),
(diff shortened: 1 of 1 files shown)


yosifkit:
```diff
+LABEL org.opencontainers.image.url=https://github.com/SonarSource/sonar-scanner-cli-docker
```
Seems fine to add a label:+1:. Should this be pointing to `https://github.com/SonarSource/docker-sonarqube` instead?

One caveat of using a label is that they are inherited by images `FROM` this one; we hope to be able to provide this information in annotations automatically for official images, but that is still a todo.

tianon:

(for further context/consideration, not because it's necessarily a blocker)

carminevassallo:
> Seems fine to add a label👍. Should this be pointing to `https://github.com/SonarSource/docker-sonarqube` instead?

Thanks for spotting this guys! We are gonna fix it shortly...

carminevassallo:
> Thanks for spotting this guys! We are gonna fix it shortly...

and that's fixed. Please let us know if you have other comments, otherwise feel free to merge :)

PR #14397, opened 2023-04-03
Add Java-20-based clojure images & remove broken boot variants

cap10morgan:
New upstream Java release. We're deprecating the boot variants in alpine variants and in all images from Java 20 forward. It hasn't had a release since 2019 and is breaking in more and more builds.


cap10morgan:
https://github.com/docker-library/official-images/pull/14397/commits/8c906fb213f3f95a01cc6d8f2a6f65eddd942bdc removes the JDK 19 variants that are now unsupported. It also brings in a tools-deps upstream version bump that had just landed in our repo too. Let me know if you'd prefer that I split that change out into a separate commit / PR.

tianon:
One PR is great :smile:

PR #14422, opened 2023-04-09
haxe 4.3.0

andyli:


tianon:
```diff
+		libpcre2-* \
```

While technically harmless at runtime, this _does_ add a bunch of packages to the image that aren't actually necessary or used -- it appears the end result of building this only specifically requires `libpcre2-8-0` (which is already installed in the base image), as verified by `ldd` on the end result of building `/usr/local/bin/haxe` :sweat_smile:

It also appears you don't need `libpcre3` at all anymore, during either build (or at runtime)?

(None of this is officially Blocking, and we're happy to merge as-is if you'd prefer.)

andyli:
Thank you for the through review, much appreciated!

I've just replaced `libpcre2-*` with `libpcre2-8-0`.

> It also appears you don't need libpcre3 at all anymore, during either build (or at runtime)?

Sadly libpcre3 is still used by neko (regexp.ndll), which is used by haxelib.

```
# ldd /usr/local/lib/neko/regexp.ndll 
        linux-vdso.so.1 (0x00007ffe5c5dc000)
        libneko.so.2 => /usr/local/lib/libneko.so.2 (0x00007f08de6d7000)
        libpcre.so.3 => /lib/x86_64-linux-gnu/libpcre.so.3 (0x00007f08de664000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f08de48f000)
        libgc.so.1 => /usr/lib/x86_64-linux-gnu/libgc.so.1 (0x00007f08de417000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f08de411000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f08de2cd000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f08de2a9000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f08de705000)
```

PR #14442, opened 2023-04-13
maven: make eclipse-temurin the latest

carlossg:


yosifkit:
> fatal: couldn't find remote ref refs/heads/master

It looks like you switched to a `main` branch on your `docker-maven` repo; you'll need to add a `GitFetch: refs/heads/main` up in the global block (next to the `Builder` and `GitRepo`).

While we can technically pull the commits without knowing the branch, having the branch name lets us verify that the commit is coming from the correct repo and not just a fork or in-progress branch.

carlossg:
sorry about that

library/maven
```diff
@@ -17,7 +18,7 @@ Architectures: amd64, arm32v7, arm64v8, ppc64le, s390x
 GitCommit: 8f9adaf3d78df6ed956b0a717d51876f71f0fdfe
 Directory: eclipse-temurin-11-focal
 
-Tags: 3.9.1-eclipse-temurin-17, 3.9.1, 3.9.1-eclipse-temurin, 3.9-eclipse-temurin-17, 3.9, 3.9-eclipse-temurin, 3-eclipse-temurin-17, 3, latest, 3-eclipse-temurin, eclipse-temurin
+Tags: 3.9.1-eclipse-temurin-17, 3.9.1, 3.9-eclipse-temurin-17, 3.9, 3-eclipse-temurin-17, 3, latest
```
yosifkit, line 21: Just to double check, these other generic tags were left on jdk17 on purpose or are they supposed to move as well?

`3.9.1, 3.9, 3, latest`

carlossg: no, that was an oversight, I've fixed my script

PR #14449, opened 2023-04-14
maven: add amazoncorretto debian images

carlossg:


carlossg:
just noticed arches are wrong

yosifkit (approved):

PR #14457, opened 2023-04-17
Included additional architectures for OrientDB

tglman:


tianon:
For future reference, force pushing to the PR branch will update the PR (and thus cause less repository churn/spam for maintainers and watchers).

PR #14487, opened 2023-04-21
Update to R 4.3.0 released this morning

eddelbuettel:
Standard update of r-base to the new upstream release 4.3.0 made this morning using the updated Debian binaries. As before, with Debian in a freeze, we point to experimental (where I uploaded the distro .deb files earlier today) and a 'convenience PPA' under my account at GitHub.

No other changes.


eddelbuettel:
Thank you!  

BTW turns out I was half asleep and by accident uploaded the r-base file into unstable instead of experimental. But it is all good as far as the container here is concerned because I did double them up in my PPA. And Debian unstable does of course remain visible too.

PR #14499, opened 2023-04-24
Official Convertigo 8.2.0 release

nicolas-albert:
Thx!

### DIFF
--- a/library/convertigo
+++ b/library/convertigo
@@ -1,7 +1,7 @@
 Maintainers: Nicolas Albert <nicolasa@convertigo.com> (@nicolas-albert), Olivier Picciotto <olivier.picciotto@convertigo.com> (@opicciotto)
 GitRepo: https://github.com/convertigo/convertigo
-GitCommit: 070f8a67e61a7a9bc701ddadae0a0a0fba9470a3
+GitCommit: a2fb2dabcd979887118458aba535ff467d7f891d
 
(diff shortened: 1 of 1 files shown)


yosifkit:
Just a small warning about the entrypoint script; the `$( ... )` captures the output of the subshell and since it isn't then assigned anywhere, it would attempt to execute it (if there were output). I'd recommend a non-capturing subshell `( ... )` instead.

```diff
- $(TMPSED=`...` && \
-    echo ...)
+ (TMPSED=`...` && \
+    echo ...)
# alternatively
+ (TMPSED="$(...)" && \
+    echo ...)
```

I don't think this is blocking, so if you want us to go ahead and merge, we can.

nicolas-albert:
> Just a small warning about the entrypoint script; the `$( ... )` captures the output of the subshell and since it isn't then assigned anywhere, it would attempt to execute it (if there were output). I'd recommend a non-capturing subshell `( ... )` instead.
> 
> ```diff
> - $(TMPSED=`...` && \
> -    echo ...)
> + (TMPSED=`...` && \
> +    echo ...)
> # alternatively
> + (TMPSED="$(...)" && \
> +    echo ...)
> ```
> 
> I don't think this is blocking, so if you want us to go ahead and merge, we can.

I take notes for the next minor release, thank you so much for the advice.

You can do the merge.

PR #14507, opened 2023-04-26
eclipse-temurin: april PSU updates

gdams:
Two things to note:

1. JDK8 arm32v7 has been delayed and we can't wait any longer so we'll have to break users here
2. JDK20 we've made the decision to hold off publishing ppc64le until we've found a fix for a test case. This means breaking a small subset of users relying on the latest tag


tianon:
Those are tough tradeoffs - thank you for proactively communicating them. :bow: :heart:

PR #14583, opened 2023-05-04
Releasing Neo4j 4.4.20

ericsporre:
Thank you!


yosifkit:
We don't usually recommend adding a `HEALTHCHECK` directive for a [few reasons](https://github.com/docker-library/faq/tree/4f51e1b995017ab10dab85fed304bce60fb7fa37#healthcheck) but this isn't a blocking change if it is important to your users.

ericsporre:
Image updated with the health check removed.

PR #14595, opened 2023-05-05
Update cassandra

yosifkit:
`3.0.29` is probably imminent: https://github.com/apache/cassandra/releases/tag/3.0.29-tentative

Changes:

- https://github.com/docker-library/cassandra/commit/7f0ca5f: Update 3.11 to 3.11.15


tianon:
Aha, https://lists.apache.org/thread/dhcwk8h9krld6fk1t5l34ztbttmmozb9 -- we probably won't get a 3.0.29 for somewhere on the order of days, so we should move ahead with this. :+1:

> A vote of this test build will be initiated within the next couple of days.

PR #14607, opened 2023-05-09
alpine: bump alpine 3.18.0 as latest, drop EOL 3.14

ncopa:

PR #14617, opened 2023-05-10
traefik: remove v1

ldez:
Related to #14607

We stopped to support Traefik v1 for 2 years now so that we can drop it.


mmatur:
LGTM

PR #14625, opened 2023-05-11
MariaDB 2023 Q2 release 1

grooverdan:
Happy new release.

Adds LTS tags on the latest only.

11.0/11.1(?) bumps coming in about a week.

Adds environment variables for replication.


grooverdan:
Thanks @tianon

PR #14627, opened 2023-05-11
 Update bonita, add 2023.1

DumitruCorini:
This adds the version of bonita: 2023.1-u0

Removed previous version of 7.12

### DIFF
--- a/library/bonita
+++ b/library/bonita
@@ -12,11 +12,11 @@ GitFetch: refs/heads/docker/2021.2
 GitCommit: a1d9ee5e31d38958aa553cc7f9d465f1151d902f
 
 Tags: 2022.1-u0, 2022.1, 7.14.0, 7.14
-GitFetch: refs/heads/2022.1
+GitFetch: refs/heads/master
 GitCommit: 694bf79347add872f8c6a4c0a7f5c3ef12c31dc8
 
 Tags: 2022.2-u0, 2022.2, 7.15.0, 7.15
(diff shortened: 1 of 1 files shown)


library/bonita
```diff
@@ -8,18 +8,18 @@ Architectures: amd64, arm64v8, ppc64le
...
 Tags: 2022.1-u0, 2022.1, 7.14.0, 7.14
-GitFetch: refs/heads/master
+GitFetch: refs/heads/2022.1
 GitCommit: 694bf79347add872f8c6a4c0a7f5c3ef12c31dc8
 
-Tags: 2022.2-u0, 2022.2, 7.15.0, 7.15, latest
-GitFetch: refs/heads/master
+Tags: 2022.2-u0, 2022.2, 7.15.0, 7.15
+GitFetch: refs/heads/2022.2
 GitCommit: 21c0e51634e836feaf910e8e3d6e95de200e1814
```
tianon, line 21: Any particular reason to keep these versions listed here?  Are multiple versions of Bonita actively supported at once?  Removing tags here will remove them from the "Supported" section on the Hub readme (and will prevent us from spending cycles rebuilding them on the official build servers), but the tags will still be available to users who want them.  (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)

DumitruCorini: Hi. The four latest versions are supported currently.

PR #14651, opened 2023-05-12
Update php

tianon:
Changes:

- https://github.com/docker-library/php/commit/0a68eaa2: Update 8.1-rc
- https://github.com/docker-library/php/commit/cb2f0def: Update 8.1 to 8.1.19


tianon:
There _should_ be an 8.2 as well, not sure why there isn't :confused:

tianon:
Oh, we already got 8.2.6 :sweat_smile:

PR #14667, opened 2023-05-16
erlang/otp 26.0

getong:


tianon:
```diff
-FROM alpine:3.17
+FROM alpine:3.16
```

Oops, looks like Erlang 26 has an older version of Alpine than Erlang 25 had been updated to - was this intentional?  (Should that be updated to Alpine 3.17 or even 3.18?)

getong:
@tianon 
update to alpine 3.18

PR #14670, opened 2023-05-16
Update bash

yosifkit:
Changes:

- https://github.com/tianon/docker-bash/commit/3785a84: Update devel to 20230515, commit e44e3d50dee26d12927baca188ad2a44a0135173

### DIFF
--- a/library/bash
+++ b/library/bash
@@ -3,9 +3,9 @@
 Maintainers: Tianon Gravi <admwiggin@gmail.com> (@tianon)
 GitRepo: https://github.com/tianon/docker-bash.git
 
-Tags: devel-20230504, devel, devel-20230504-alpine3.18, devel-alpine3.18
+Tags: devel-20230515, devel, devel-20230515-alpine3.18, devel-alpine3.18
 Architectures: amd64, arm32v6, arm32v7, arm64v8, i386, ppc64le, s390x
-GitCommit: e73ff20af7431345d27345c3fb59d2979051815a
+GitCommit: 3785a84f3afd1b0c95c8a3f1a1e9c9bb382b6a70
 Directory: devel
 
 Tags: 5.2.15, 5.2, 5, latest, 5.2.15-alpine3.18, 5.2-alpine3.18, 5-alpine3.18, alpine3.18


tianon:
omg, this new stuff looks interesting: (this is "man to markdown" via pandoc)

> There is an alternate form of command substitution:
> 
> > **${***C* *command***;** **}**
> 
> which executes *command* in the current execution environment. This
> means that side effects of *command* take effect immediately in the
> current execution environment and persist in the current environment
> after the command completes (e.g., the **exit** builtin will exit the
> shell).
> 
> The character *C* following the open brace must be a space, tab,
> newline, **(**, or **\|**, and the close brace must be in a position
> where a reserved word may appear (i.e., preceded by a command terminator
> such as semicolon). **Bash** allows the close brace to be joined to the
> remaining characters in the word without being followed by a shell
> metacharacter as a reserved word would usually require.
> 
> This type of command substitution superficially resembles executing an
> unnamed shell function: local variables are created as when a shell
> function is executing, and the **return** builtin forces *command* to
> complete; however, the rest of the execution environment, including the
> positional parameters, is shared with the caller.
> 
> If the first character following the open brace is a **(**, *command* is
> executed in a subshell, and *command* must be terminated by a **)**.
> This is similar to the **(** compound command (see **Compound Commands**
> above). If the first character is a **\|**, the construct expands to the
> value of the **REPLY** shell variable after *command* executes, without
> removing any trailing newlines, and the standard output of *command*
> remains the same as in the calling shell. **Bash** creates **REPLY** as
> an initially-unset local variable when *command* executes, and restores
> **REPLY** to the value it had before the command substitution after
> *command* completes, as with any local variable.

tianon:
:eyes:

```console
bash-5.2# echo "${| REPLY=$'foo\nbar\n'; REPLY+=$'\nfoo\n\n'; }"
foo
bar

foo


```

tianon:
(I think the use case I'm most excited for here is something like a shell function that calculates something expensive and maintains a cache, and with this we can invoke that _and_ capture the output while still letting it maintain the global cache state it needs :eyes:)

PR #14687, opened 2023-05-18
Add images for Rocky Linux 9.2 and 8.8

NeilHanlon:
NOTE - We currently are not providing 9.2 for ppc64le due to an upstream
bug[1], so the difference there is expected for now. We will add the
ppc64le image once a fix is found for the bug.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=2203919

### DIFF
--- a/library/rockylinux
+++ b/library/rockylinux
@@ -2,38 +2,34 @@ Maintainers: The Rocky Linux Foundation <infrastructure@rockylinux.org> (@rocky-
              Neil Hanlon <neil@resf.org> (@neilhanlon)
 GitRepo: https://github.com/rocky-linux/sig-cloud-instance-images.git
 
-Tags: 9.1.20230215, 9.1, 9
-GitFetch: refs/heads/Rocky-9.1.20230215-Base-x86_64
-GitCommit: c554918c3fb7432f70802459ee942fead8121cae
-arm64v8-GitFetch: refs/heads/Rocky-9.1.20230215-Base-aarch64
-arm64v8-GitCommit: a44de1cb9192969dacd996beb22c152020908532
-s390x-GitFetch: refs/heads/Rocky-9.1.20230215-Base-s390x
-s390x-GitCommit: 3bd8af8923fd7b0611f1599e347897a6cb9c156b
-ppc64le-GitFetch: refs/heads/Rocky-9.1.20230215-Base-ppc64le
-ppc64le-GitCommit: 0aee39d093db7f0732854729a0e46bc8092bf607
-Architectures: amd64, arm64v8, ppc64le, s390x
+Tags: 9.2.20230513, 9.2, 9
+GitFetch: refs/heads/Rocky-9.2.20230513-Base-x86_64
+GitCommit: 5534796e428e5ac942d820bc96a36a5d0404a8a4
+arm64v8-GitFetch: refs/heads/Rocky-9.2.20230513-Base-aarch64
+arm64v8-GitCommit: 9dc2a99760aa37f4fde4cd35a8fe12ce3991e16c
+s390x-GitFetch: refs/heads/Rocky-9.2.20230513-Base-s390x
+s390x-GitCommit: 1d9cb06268ba4a479b814a2e191a5fa42048872c
+Architectures: amd64, arm64v8, s390x
 
-Tags: 9.1.20230215-minimal, 9.1-minimal, 9-minimal
-GitFetch: refs/heads/Rocky-9.1.20230215-Minimal-x86_64
-GitCommit: 957dce1794c27b8fc1666f7e283db2ba7ef5b182
-arm64v8-GitFetch: refs/heads/Rocky-9.1.20230215-Minimal-aarch64
-arm64v8-GitCommit: 499b1eeb324a768228963568a9b3afdaf5da74dc
-s390x-GitFetch: refs/heads/Rocky-9.1.20230215-Minimal-s390x
-s390x-GitCommit: db6eddbb6b6cd097f625a014dcd50a9f152720d2
-ppc64le-GitFetch: refs/heads/Rocky-9.1.20230215-Minimal-ppc64le
-ppc64le-GitCommit: 4e11f6cf490d560b37bc6b9dffff204d2bc6e5b3
-Architectures: amd64, arm64v8, ppc64le, s390x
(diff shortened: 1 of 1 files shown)


yosifkit:
😕❤️ I am unsure what this new `-ubi` variant is for. It seems to be basically the same as the regular image but with `systemd` installed. The difference between this and the user just doing `FROM rockylinux:9.2` and `RUN dnf install -y systemd` seems to be fairly minimal, so I am unsure what the use case is.

```diff
$ diff -u \
    <(docker run -i --rm rockylinux:9.2 bash -c "dnf install -y -qq systemd &> /dev/null && rpm -qa" | sort) \
    <(docker run -i --rm rockylinux:9.2-ubi rpm -qa | sort)
--- /dev/fd/63  2023-05-18 13:15:10.579202298 -0700
+++ /dev/fd/62  2023-05-18 13:15:10.579202298 -0700
@@ -3,8 +3,6 @@
 audit-libs-3.0.7-103.el9.x86_64
 basesystem-11-13.el9.noarch
 bash-5.1.8-6.el9_1.x86_64
-binutils-2.35.2-37.el9.x86_64
-binutils-gold-2.35.2-37.el9.x86_64
 bzip2-libs-1.0.8-8.el9.x86_64
 ca-certificates-2022.2.54-90.2.el9.noarch
 coreutils-single-8.32-34.el9.x86_64
@@ -17,9 +15,9 @@
 dbus-1.12.20-7.el9_1.x86_64
 dbus-broker-28-7.el9.x86_64
 dbus-common-1.12.20-7.el9_1.noarch
+dejavu-sans-fonts-2.37-18.el9.noarch
 dnf-4.14.0-5.el9_2.noarch
 dnf-data-4.14.0-5.el9_2.noarch
-elfutils-debuginfod-client-0.188-3.el9.x86_64
 elfutils-default-yama-scope-0.188-3.el9.noarch
 elfutils-libelf-0.188-3.el9.x86_64
 elfutils-libs-0.188-3.el9.x86_64
@@ -27,7 +25,9 @@
 file-libs-5.39-12.el9.x86_64
 filesystem-3.16-2.el9.x86_64
 findutils-4.8.0-5.el9.x86_64
+fonts-filesystem-2.0.5-7.el9.1.noarch
 gawk-5.1.0-6.el9.x86_64
+gdb-gdbserver-10.2-10.el9.x86_64
 gdbm-libs-1.19-4.el9.x86_64
 glib2-2.68.4-6.el9.x86_64
 glibc-2.34-60.el9.x86_64
@@ -37,16 +37,16 @@
 gnupg2-2.3.3-2.el9_0.x86_64
 gnutls-3.7.6-20.el9_2.x86_64
 gpgme-1.15.1-6.el9.x86_64
-gpg-pubkey-350d275d-6279464b
 grep-3.6-5.el9.x86_64
 gzip-1.12-1.el9.x86_64
-hostname-3.23-6.el9.x86_64
 ima-evm-utils-1.4-4.el9.x86_64
 json-c-0.14-11.el9.x86_64
 keyutils-libs-1.6.3-1.el9.x86_64
 kmod-libs-28-7.el9.x86_64
 krb5-libs-1.20.1-8.el9.x86_64
-less-590-1.el9_0.x86_64
+langpacks-core-en-3.0-16.el9.noarch
+langpacks-core-font-en-3.0-16.el9.noarch
+langpacks-en-3.0-16.el9.noarch
 libacl-2.3.1-3.el9.x86_64
 libarchive-3.5.3-4.el9.x86_64
 libassuan-2.5.5-3.el9.x86_64
@@ -85,8 +85,6 @@
 libstdc++-11.3.1-4.3.el9.x86_64
 libtasn1-4.16.0-8.el9_1.x86_64
 libunistring-0.9.10-15.el9.x86_64
-libusbx-1.0.26-1.el9.x86_64
-libuser-0.63-12.el9.x86_64
 libutempter-1.2.1-6.el9.x86_64
 libuuid-2.37.4-11.el9_2.x86_64
 libverto-0.3.2-3.el9.x86_64
@@ -107,7 +105,6 @@
 p11-kit-0.24.1-2.el9.x86_64
 p11-kit-trust-0.24.1-2.el9.x86_64
 pam-1.5.1-14.el9.x86_64
-passwd-0.80-12.el9.x86_64
 pcre2-10.40-2.el9.x86_64
 pcre2-syntax-10.40-2.el9.noarch
 pcre-8.44-3.el9.3.x86_64
@@ -142,10 +139,10 @@
 tar-1.34-6.el9_1.x86_64
 tpm2-tss-3.0.3-8.el9.x86_64
 tzdata-2023c-1.el9.noarch
-usermode-1.114-4.el9.x86_64
 util-linux-2.37.4-11.el9_2.x86_64
 util-linux-core-2.37.4-11.el9_2.x86_64
 vim-minimal-8.2.2637-20.el9_1.x86_64
+which-2.21-28.el9.x86_64
 xz-libs-5.2.5-8.el9_0.x86_64
 yum-4.14.0-5.el9_2.noarch
 zlib-1.2.11-39.el9.x86_64
```

NeilHanlon:
The UBI Variant is one which matches what our upstream creates, however it does not have to be in the library images if not desired. I included it as we have recently moved it out of an "experimental" state.

We will publish it under our own namespace even if not accepted to the library. I apologize for not communicating with you all before.

yosifkit:
Ok, so it seems like it is `redhat/ubi9-init`, but without configuring `systemd` as pid 1. Is that correct?

I think our decision when we added the `redhat/ubi*` pins is the closest to this (https://github.com/docker-library/official-images/pull/13908). Basically, we did not add `redhat/ubi*-init` because it didn't fit with the "one concern per container" that we keep in Official Images.

I think that if the purpose of the `rockylinux:*-ubi` image is for running `systemd`, then I don't think it fits our current requirements, but if it has a different purpose then maybe?  CC @tianon, does that seem reasonable?

tianon:
Agreed with your assessment :sweat_smile:

(would also add that `-ubi` feels a little bit confusing given there are four different "UBI" types and the "minimal" image here already matches "ubi-minimal" as I've understood it :see_no_evil:)

netsandbox:
Would be great if this would go further to have the 9.2 version available in the library images.

NeilHanlon:
hi @yosifkit , @tianon  - I've removed the ubi containers and added our 8.8 containers, as well.

PR #14719, opened 2023-05-24
Updated mainline nginx to 1.25.0 and njs to 0.7.12.

thresheek:
Also, extended envsubst-based templating to process stream (TCP/UDP) servers.


library/nginx
```diff
@@ -3,29 +3,29 @@
 Maintainers: NGINX Docker Maintainers <docker-maint@nginx.com> (@nginxinc)
 GitRepo: https://github.com/nginxinc/docker-nginx.git
 
-Tags: 1.23.4, mainline, 1, 1.23, latest, 1.23.4-bullseye, mainline-bullseye, 1-bullseye, 1.23-bullseye, bullseye
```
jnoordsij, line 6: The tags seem to be off, `1.25.0` is also being tagged as `1.23` on all variants.

thresheek: Ah indeed!

thresheek: Fixed in an amended / force-pushed commit.

PR #14726, opened 2023-05-24
Official Convertigo 8.2.1 release

nicolas-albert:
Thx!


yosifkit:
> WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

The `apt update/install` should be swapped back to `apt-get`.

https://manpages.debian.org/bullseye/apt/apt.8.en.html#SCRIPT_USAGE_AND_DIFFERENCES_FROM_OTHER_APT_TOOLS

PR #14728, opened 2023-05-24
eclipse-temurin: bump alpine to 3.18

gdams:


yosifkit:
Just a note that the move to Alpine 3.18 might cause some user headaches; especially around `gpg` verification. Hopefully this can help you spot the issue to help users.

I noticed that the `tomee` image rebuilds triggered from this change were all running into build timeouts on Official Images builders. They were stuck on a fairly innocuous line and `docker top` shows that aside from the shell, the only running things are `gpg` and the new `keyboxd` (a new part of the most recent gpg).

```console
...
root                493961              493909              2                   22:40               ?                   00:00:00            gpg --batch --verify tomee.tar.gz.asc apache-tomee-8.0.15-plume.tar.gz
root                493963              493909              0                   22:41               ?                   00:00:00            keyboxd --homedir /root/.gnupg --daemon
```

After some digging, I found that `keyboxd` is waiting for a lock file (it didn't get stopped cleanly in an earlier `RUN` layer).

```console
/usr/local/tomee # /usr/libexec/keyboxd --no-detach --server
keyboxd[7]: waiting for lock (held by 12) ...
keyboxd[7]: waiting for lock (held by 12) ...
```

Two workarounds:
- disable `keyboxd` by creating an empty `GNUPGHOME` directory (default `~/.gnupg/`)
    > Since version 2.4.1 the keyboxd will be used by default for a fresh install; i.e. if a ~/.gnupg directory did not yet exist.
    >
    > \- https://github.com/gpg/gnupg/blob/2783b786a9314894d76807b00789df09fe70d1d1/README#L131-L132
- cleanly shutdown `gpg` related daemons before the end of a `RUN`
   - `gpgconf --kill all` should work, but might need some amount of wait time to ensure they stop

PR #14735, opened 2023-05-26
Update Nextcloud

J0WI:


yosifkit:
(I assume the problem with Alpine 3.17+ is the upgrade to OpenSSL 3)

Are you asking if we can temporarily add back `php:8.1-fpm-alpine3.16` to the PHP images (until about October)? 🤔 I think it makes sense given the best alternative for the `nextcloud` image is to downgrade to PHP 8.0. 😢 CC @tianon

tianon:
Ok, fair enough: https://github.com/docker-library/php/pull/1412 :eyes:

tianon:
https://github.com/docker-library/official-images/pull/14751 :eyes:

PR #14761, opened 2023-05-31
Update Matomo

J0WI:


tianon:
@tsteur, just to make sure, you're OK with being listed here, right? :innocent:
(I don't see confirmation in https://github.com/matomo-org/docker/pull/301; apologies if I just missed it somewhere!)

tsteur:
@J0WI I haven't done any contributions to that project and am not too familiar with docker. Any chance we could mention someone else?

tsteur:
@J0WI Thanks for letting us know.  The Matomo Docker project https://github.com/matomo-org/docker is a community driven project and the Matomo core team hasn't been contributing to it yet. It seems that the main community contributors are @J0WI and @pierreozoux and @tianon - would it be maybe possible to list one of you in the file?

pierreozoux:
not me :upside_down_face:  I don't use matomo anymore and trying to not receive these notifications anymore :)

Maybe contact matomo to ask them?

Have a nice day everybody!

tianon:
@matomo-org here isn't ideal, but it's good enough for now (and I think accurately reflects the situation)

FWIW, we _do_ use this for things like notifying image maintainers when a major base image is deprecated and being removed, and that doesn't work for a whole org like this

PR #14766, opened 2023-05-31
23.0.0.5 release

wraschke:


wraschke:
@tianon I'm not so experienced at doing this process. Any insights into the failure at https://github.com/docker-library/official-images/actions/runs/5138363695/jobs/9247611788?pr=14766?

```
```

yosifkit:
That just means that the `diff` is extra long, so it doesn't fit in a comment. Here it is in a gist:

Diff: https://gist.github.com/docker-library-bot/43dd8a7e66d33ee2f633659e3eab9419

tianon:
That's a long diff, so we're still trying to parse it all, but wanted to give some initial feedback:

---

There are quite a few bits here that seem to reference CRIU, but that seems kind of weird inside a container (and it doesn't seem to be included in what's installed)?  Is CRIU embedded in websphere/open liberty directly or something like that?  Does it actually work with the default container permissions?  Some of the script splits like `pidplus.sh` and `checkpoint.sh` and `restore-server.sh` also don't seem like they need to be dedicated scripts unless users are expected to run them directly. :thinking:

---

```diff
+    && wget -O /usr/bin/dumb-init https://github.com/Yelp/dumb-init/releases/download/v1.2.5/dumb-init_1.2.5_x86_64 \
+    && chmod +x /usr/bin/dumb-init \
```

This is not sufficient -- it only supports one architecture (will fail on arm64v8, ppc64le, and s390x), and it needs some kind of download verification (https://github.com/docker-library/official-images#image-build), _or_ it needs to switch to installing a distro package instead.

---

```diff
+MIT LICENSE
+
+The Program includes some or all of the following that IBM obtained
+under the MIT License:
+
+DUMB-INIT (Copyright: 2015 Yelp, Inc.
+2016-2019 Dmitry Smirnov <onlyjob@debian.org>)
```

I'm not 100% sure, but I don't think this is accurate?  It looks like it came from https://sources.debian.org/src/dumb-init/1.2.5-1/debian/copyright/#L10-L14, but that _only_ applies to `debian/` in the source directory, and this isn't even installing from the Debian packages (it's downloading a single binary directly off GitHub that's published by the `dumb-init` project, not Debian).

tjwatson:
> There are quite a few bits here that seem to reference CRIU, but that seems kind of weird inside a container (and it doesn't seem to be included in what's installed)? Is CRIU embedded in websphere/open liberty directly or something like that? Does it actually work with the default container permissions? Some of the script splits like `pidplus.sh` and `checkpoint.sh` and `restore-server.sh` also don't seem like they need to be dedicated scripts unless users are expected to run them directly.

`checkpoint.sh` is intended to be called on its own.  See https://openliberty.io/blog/2023/02/10/instant-on-beta-update.html for an example of using `Run checkpoint.sh applications` is shown.  `pidplus.sh` is an internal detail to bump up pids before doing a checkpoint to avoid a checkpoint of PID 1.  `restore-server.sh` was separated from `docker-server.sh` to place any criu specific logic in a separate script.

> This is not sufficient -- it only supports one architecture (will fail on arm64v8, ppc64le, and s390x), and it needs some kind of download verification (https://github.com/docker-library/official-images#image-build), _or_ it needs to switch to installing a distro package instead.

You are correct. We need to multi-arch this.  The Semeru runtime initially only supports x86-64 for CRIUSupport, but we should update our `Dockerfile` to download the correct version.

> I'm not 100% sure, but I don't think this is accurate? It looks like it came from https://sources.debian.org/src/dumb-init/1.2.5-1/debian/copyright/#L10-L14, but that _only_ applies to `debian/` in the source directory, and this isn't even installing from the Debian packages (it's downloading a single binary directly off GitHub that's published by the `dumb-init` project, not Debian).

Got this from a licensing tool that scanned the repository.  I do think you are correct though and we likely just need what is reported from https://github.com/Yelp/dumb-init/blob/master/LICENSE

tjwatson:
> Does it actually work with the default container permissions?

For restore the `CHECKPOINT_RESTORE` and `SETPCAP` capabilities are required.  To do that you need the fix for https://github.com/moby/moby/pull/42011 so that you can `--add-cap` for `CHECKPOINT_RESTORE` with `docker`.  That is version 23+ I think.  Otherwise Podman can be used.  It also works with AWS EKS and Azure AKS as long as you grant a security context to grant `CHECKPOINT_RESTORE` and `SETPCAP` capabilities.  See https://openliberty.io/blog/2023/02/20/aws-instant-on.html

wraschke:
@tianon @yosifkit do you still have questions about this pull request?

yosifkit:
```diff
 COPY fixes/ /opt/ol/fixes/
 
+# Install dumb-init
+RUN set -eux; \
+    ARCH="$(uname -m)"; \
+    case "${ARCH}" in \
+       aarch64|arm64) \
+         DUMB_INIT_URL='https://github.com/Yelp/dumb-init/releases/download/v1.2.5/dumb-init_1.2.5_aarch64'; \
+         ;; \
+       amd64|x86_64) \
+         DUMB_INIT_URL='https://github.com/Yelp/dumb-init/releases/download/v1.2.5/dumb-init_1.2.5_x86_64'; \
+         ;; \
+       ppc64el|ppc64le) \
+         DUMB_INIT_URL='https://github.com/Yelp/dumb-init/releases/download/v1.2.5/dumb-init_1.2.5_ppc64le'; \
+         ;; \
+       s390x) \
+         DUMB_INIT_URL='https://github.com/Yelp/dumb-init/releases/download/v1.2.5/dumb-init_1.2.5_s390x'; \
+         ;;\
+       *) \
+         echo "Unsupported arch: ${ARCH}"; \
+         exit 1; \
+         ;; \
+    esac; \
+    curl -LfsSo /usr/bin/dumb-init ${DUMB_INIT_URL}; \
+    chmod +x /usr/bin/dumb-init;
+
 # Install Open Liberty
```

The `dumb-init` downloads need to be verified (https://github.com/docker-library/official-images/tree/f8193b6f16c7fcd6161821f638fdf791aa19c8a2#image-build).  It seems that they don't publish signatures, so it'll probably need the values from their `sha256sums` embedded into the Dockerfile (like [golang](https://github.com/docker-library/golang/blob/713124c10e228b178e92ecf949e86c16bc3f3057/1.20/bullseye/Dockerfile#L25-L74)).

wraschke:
@yosifkit do you have an outlook on when this can be merged? Thank you.

yosifkit:
Updated diff for 4c0610ebac8798a70c7a75804f86afedc29e2f2b: https://gist.github.com/docker-library-bot/c63119212ac48cead5efdc5ff755973c

yosifkit (approved):

PR #14767, opened 2023-05-31
Update CirrOS to 0.6.2, 0.5.3

tianon:
- https://github.com/cirros-dev/cirros/releases/tag/0.6.2
- https://github.com/cirros-dev/cirros/releases/tag/0.5.3

(Not sure how I missed "arm image is ARMv7 now to follow armhf distributions" in https://github.com/cirros-dev/cirros/releases/tag/0.6.0, but that's fixed for 0.6.x now too!)

### DIFF
--- a/library/cirros
+++ b/library/cirros
@@ -3,11 +3,41 @@
 Maintainers: Tianon Gravi <admwiggin@gmail.com> (@tianon)
 GitRepo: https://github.com/tianon/docker-brew-cirros.git
 GitFetch: refs/heads/dist
-GitCommit: 84bfc9d0e4cc631d043410d602bf5f8876ed6575
-Architectures: amd64, arm32v5, arm64v8, ppc64le
+GitCommit: e8833253f108046f977fbcecd7f02170bd20d357
+Architectures: amd64, arm32v7, arm64v8, ppc64le
 amd64-Directory: arches/amd64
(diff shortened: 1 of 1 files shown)


library/cirros
```diff
@@ -3,11 +3,20 @@
...
+GitCommit: e8833253f108046f977fbcecd7f02170bd20d357
+Architectures: amd64, arm32v7, arm64v8, ppc64le
+amd64-Directory: arches/amd64
+arm32v7-Directory: arches/arm32v7
+arm64v8-Directory: arches/arm64v8
+ppc64le-Directory: arches/ppc64le
+
+Tags: 0.6.2, 0.6, 0, latest
+
+Tags: 0.5.3, 0.5
```
tianon, line 15: Reviewer's note: this is in kind of a weird format for now because I expect this 0.5.x entry to be temporary (I don't know whether CirrOS 0.5.x is going to be actively supported still :sweat_smile:)

tianon:
Alright, I've now fixed everything "properly" (updated branches for every "wrong" release and branches for every release I've made so they can all be discovered more easily since they're all relatively small)

yosifkit:
Just a note that `0.6.1` and `0.6.0` are temporary to fix their incorrect architecture. They were previously published as `arm32v5` when they contain `arm32v7` binaries.

PR #14769, opened 2023-06-01
Remove CirrOS backfilled versions

tianon:
The builds aren't _quite_ done/pushed yet, but I don't want to forget to remove these. :sweat_smile:

(follow up to https://github.com/docker-library/official-images/pull/14767)


tianon:
Nice, looks like we're good to go: https://hub.docker.com/_/cirros/tags :partying_face:

PR #14774, opened 2023-06-02
Bump rust to 1.70.0

sfackler:


SimenB:
Any idea when the amd build for alpine will be ready?

https://hub.docker.com/layers/library/rust/1.70.0-alpine/images/sha256-9958c163bc23d9c1c0e99442f529a77e763777229d6978e2b7d34d8cfe0ffff4?context=explore

Seems stuck

neoeinstein:
I also noticed that `1.70.0` doesn't seem to have an AMD64 image available as yet either, though there has been a publish of the AMD64 for `latest` (Same with `alpine` having an ARM64 image, but no AMD64 image.)

Arnavion:
There's only one amd64 builder and the `tomee` image build has a bug that causes it to block the builder queue for hours on end. Sometimes the admins get around to aborting it but this time they haven't.

https://doi-janky.infosiftr.net/job/multiarch/job/amd64/job/tomee/

https://github.com/tomitribe/docker-tomee/issues/80

tianon:
Sorry, this has been a persistent problem that we are working on, but it's unfortunately not something with a trivial solution.  The current instance of it is resolved though. :+1:

Arnavion:
Yes, the images were published yesterday, so this is resolved. I also see that you've disabled the tomee build so I guess this won't happen again.

PR #14832, opened 2023-06-12
Update Debian, especially for Bookworm/12

tianon:


tianon:
It looks like merged-/usr is determined to wreak havoc on this PR's diff :sweat_smile: (still nuts even with `--break-rewrites` enabled :sob:)

tianon:
Diff: https://gist.github.com/docker-library-bot/f0a28a61b5d06d5d83a28188e8631a96

yosifkit (approved):

PR #14857, opened 2023-06-14
Update golang

tianon:
Changes:

- https://github.com/docker-library/golang/commit/db757a0: Merge pull request https://github.com/docker-library/golang/pull/456 from sitano/ivan_golang_bookworm
- https://github.com/docker-library/golang/commit/cd60713: Add debian:bookworm distro for 1.20 and 1.19, drop buster

### DIFF
--- a/library/golang
+++ b/library/golang
@@ -5,16 +5,16 @@ Maintainers: Tianon Gravi <admwiggin@gmail.com> (@tianon),
              Johan Euphrosine <proppy@google.com> (@proppy)
 GitRepo: https://github.com/docker-library/golang.git
 
-Tags: 1.20.5-bullseye, 1.20-bullseye, 1-bullseye, bullseye
+Tags: 1.20.5-bookworm, 1.20-bookworm, 1-bookworm, bookworm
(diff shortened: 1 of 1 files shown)


aybabtme:
FYI the change of base image is causing problems to folks who build binaries with `golang:1.20.5` and run them in a distro like `debian:bullseye`. https://twitter.com/AntoineGrondin/status/1669082690862321665

e.g. typical error strings (for those searching for the cause of those error messages with Google):

```
/lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by $binary_name)
/lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by $binary_name)
```

yosifkit:
They can also specify the Debian version of the Golang images (only the two most recent are updated/supported): `golang:1.20.5-bullseye` or `golang:1.20.5-bookworm` (and `golang:1.20-bullseye`, etc).

tianon:
If you don't need `cgo` in your binary, it should also work to specify `CGO_ENABLED=0` explicitly.

PR #14865, opened 2023-06-15
Releasing Neo4j 5.9.0

jennyowen:
Just so you know, within the next few releases I'll be adding new neo4j images based on `redhat/ubi8-minimal`. From what I understand from the documentation, that's one of the few non-official base images we can use, right?

Anyway, thanks for your hard work as usual :heart:

### DIFF
--- a/library/neo4j
+++ b/library/neo4j
@@ -1,24 +1,18 @@
-# Old point releases are included in this library file so that users
-# are not forced to upgrade to the latest point releases in order to
-# get security fixes to the base image. See
-# https://github.com/docker-library/official-images/issues/1864 for a
-# discussion of this point.
-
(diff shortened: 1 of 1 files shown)


yosifkit:
> Just so you know, within the next few releases I'll be adding new neo4j images based on `redhat/ubi8-minimal`. From what I understand from the documentation, that's one of the few non-official base images we can use, right?

Yes, that is fine. We track the external images for updates and trigger rebuilds like we do with other base images (like when [these PRs](https://github.com/docker-library/official-images/pulls?q=is%3Apr+label%3Aexternal%2Fredhat%2Fubi8-minimal+is%3Aclosed) merged).  It is also one of the few images that need to be `:latest` in the `FROM` (since the version is in the repo name).

PR #14868, opened 2023-06-15
Update Vault for 1.13.3, 1.12.7, 1.11.11

cinthiaconti:


tianon:
CI failure looks to be due to https://github.com/docker-library/official-images/pull/14607#issuecomment-1540855036 :eyes:

(it'd also be good to squash your commits, but we can do that on merge if we need to :sweat_smile:)

PR #14879, opened 2023-06-17
Update to R 4.3.1 released this morning

eddelbuettel:
Standard update of r-base to the new upstream release 4.3.1 made this morning using the updated Debian binaries. Given that Debian 12 released (yay!!) this went into `unstable` so we are back to the normal process without requiring `experimental`.

No code changes.


eddelbuettel:
Thanks!

PR #14883, opened 2023-06-18
archlinux: Release 20230618.0.159075

archlinux-github:
This is an automated release [1].

[1] https://gitlab.archlinux.org/archlinux/archlinux-docker/-/blob/master/.gitlab-ci.yml

---

Maintainers: @SantiagoTorres @shibumi @hashworks

PR #14891, opened 2023-06-20
archlinux: Release 20230620.0.159493

archlinux-github:

---

### DIFF
--- a/library/archlinux
+++ b/library/archlinux
@@ -5,13 +5,13 @@ Maintainers: Santiago Torres-Arias <santiago@archlinux.org> (@SantiagoTorres),
              Justin Kromlinger <hashworks@archlinux.org> (@hashworks)
 GitRepo: https://gitlab.archlinux.org/archlinux/archlinux-docker.git
 
-Tags: latest, base, base-20230611.0.157136
-GitCommit: 8c107a2ed9ee79927da495f556a81a6999221bab
-GitFetch: refs/tags/v20230611.0.157136
+Tags: latest, base, base-20230620.0.159493
+GitCommit: fcf2d47851e478c046c2bf56dcdb5fb449c4d5bd
+GitFetch: refs/tags/v20230620.0.159493
 File: Dockerfile.base
 
-Tags: base-devel, base-devel-20230611.0.157136
-GitCommit: 8c107a2ed9ee79927da495f556a81a6999221bab
-GitFetch: refs/tags/v20230611.0.157136
+Tags: base-devel, base-devel-20230620.0.159493
+GitCommit: fcf2d47851e478c046c2bf56dcdb5fb449c4d5bd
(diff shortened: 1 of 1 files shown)


tianon:
(https://github.com/docker-library/official-images/pull/14883#issuecomment-1599660346)

```diff
+# Note: Only official DockerHub images are allowed, see https://github.com/docker-library/official-images/actions/runs/5321247374/jobs/9636110751?pr=14890#step:4:13
```

Just a warning (in case you aren't aware), GitHub Actions logs are time-limited and don't last nearly as long as I wish they did, so this link is going to be useless at some point in the near future. :sweat:

Maybe https://gitlab.archlinux.org/archlinux/archlinux-docker/-/commit/daa67d18579024947d69a45e6d028c5adb1c8c23 and/or https://github.com/docker-library/official-images/pull/8999#issuecomment-720622797 would be more permanent places to link? :sweat_smile: :heart:

(You might also be interested in https://github.com/docker-library/official-images/pull/13769, which added support for importing an OCI image "as is" sans a `Dockerfile`, intended for use by base images, but it does require committing the bits to Git which as I recall was part of the reason for switching to this "download via `Dockerfile` which is `FROM alpine`" method in the first place :see_no_evil:)

yosifkit:
Apologies for the delay, I think we were looking for at least some acknowledgement or feedback of previous comments.

Closing in favor of https://github.com/docker-library/official-images/pull/14932. In the future, can we get the PR branch updated (like a force push) instead a new PR?

hashworks:
> Maybe https://gitlab.archlinux.org/archlinux/archlinux-docker/-/commit/daa67d18579024947d69a45e6d028c5adb1c8c23 and/or [#8999 (comment)](https://github.com/docker-library/official-images/pull/8999#issuecomment-720622797) would be more permanent places to link? :sweat_smile: :heart:

Yeah, you're right :) [Changed](https://gitlab.archlinux.org/archlinux/archlinux-docker/-/commit/301942f9e5995770cb5e4dedb4fe9166afa4806d).

> In the future, can we get the PR branch updated (like a force push) instead a new PR?

Hm. We would need to adjust the source branch name (currently `archlinux:release/20230620.0.159493`) and look for existing open MRs. [We'll take a look](https://gitlab.archlinux.org/archlinux/archlinux-docker/-/issues/81).

Alternatively we could close the old MR automatically using the bot user?

tianon:
The reason we prefer not to have new PRs is because it causes unnecessary spam in our inboxes (for folks using threaded inboxes, it creates two chains instead of just one), so closing the old PR automatically would help, but wouldn't fully alleviate the problem (especially if, like this one, there's some discussion on the older PR that might be otherwise missed/lost).

If you just force push to the same branch name every time, it'll "just work" (although the PR title would need to be updated instead of a new PR opened, but you could also generalize the title instead if you prefer).

PR #14911, opened 2023-06-21
Update golang

tianon:
Changes:

- https://github.com/docker-library/golang/commit/72bc141: Update supported architectures
- https://github.com/docker-library/golang/commit/e788e6d: Merge pull request https://github.com/docker-library/golang/pull/469 from jnoordsij/add-1.21-rc
- https://github.com/docker-library/golang/commit/1ab26fd: Add 1.21-rc variant


tianon:
(don't get excited about those `riscv64` additions - those are to record that we _could_ support `riscv64` and that upstream now has builds for it, but our base images don't support it yet)

PR #14915, opened 2023-06-22
emqx v5.1.0

id:


library/emqx
```diff
@@ -8,8 +8,14 @@ GitCommit: 129bff4ea692ede8157aa3165a8c278bf982f47d
 Directory: 4.4
```
tianon, line 8: I don't know if we've talked about it with you recently, but it seems like 4.4.x might be EOL now? :eyes:

id: @tianon not yet, but soon. 4.4.0 was released on February 18 2022, EOL is in 18 months == August 18 2023 if my math is correct.

id: We expect one more 4.4.x release, after that we will remove this tag.

tianon: Ah, makes sense; not a problem at all, thanks for double checking and sharing your plan! :heart:

PR #14925, opened 2023-06-23
Adding AL2 native rpm images

lutkerd:
This adds new tags to the existing AL2 images which customers will be able to use in the future to continue using the same images. New AL2-native images are being added that use the AL2 targeted RPMs rather than the Generic Linux RPMs. At some point in the future the default AL2 images will be changed to use the native RPMs.

The old Alpine 3.14 images are also being removed.


yosifkit:
When would a user pick `al2-generic` over `al2-native`? Is there a performance penalty or other incompatibility? Would they be confused by the names and think `native` applies to CPU architecture? The `al2-native` packages seem very different to install, or maybe that is just the diff doing a strange comparison?

Perhaps adding a section describing the differences and which to choose to the Docker Hub docs page would be helpful (https://github.com/docker-library/docs/)?

lutkerd:
We will be publishing user documentation as well, but wanted to get these out so customer can evaluate them for compatibility.

PR #14940, opened 2023-06-26
Remove latest tag from official Vault images

mladlow:
Per https://github.com/docker-library/docs/pull/2283#issuecomment-1449158801, could I also request that the latest tag be deleted please?


tianon:
As requested:

```console
$ docker pull vault:latest
Error response from daemon: manifest for vault:latest not found: manifest unknown: manifest unknown
```

:+1: :heart:

drewfish:
I can see how this is a good idea, but FYI this was a surprise and broke my CI builds.

The `latest` tag appears to documented as officially supported:
Screen Shot 2023-06-26 at 15 16 22

drewfish:
Ack... I just saw this notice... I will do as it suggets.

> DEPRECATION NOTICE
Upcoming in Vault 1.14, we will stop publishing official Dockerhub images and publish only our Verified Publisher images. Users of Docker images should pull from [hashicorp/vault](https://hub.docker.com/r/hashicorp/vault) instead of [vault](https://hub.docker.com/_/vault). Verified Publisher images can be found at https://hub.docker.com/r/hashicorp/vault.

mladlow:
Hey @drewfish I'm sorry this change broke your CI but glad that you can make the update. Thanks for adding the instructions here! I bet they will be helpful for future readers. Thanks @tianon for all your help over the years ❤️

tianon:
For sure!  We'll be here, ready to help if you ever decide you want to resurrect this image. :heart:

PR #14934, opened 2023-06-25
[ros] point buster and foxy to snapshots repository

mikaelarguedas:
ROS Foxy is now EOL : https://discourse.ros.org/t/new-packages-for-foxy-fitzroy-2023-06-20/32039
Debian buster has been EOl for a while.

This PR make them build from the ROS snapshots repository before disabling the builds

PR #14943, opened 2023-06-27
[ros] retire foxy and buster based images

mikaelarguedas:
follow-up of https://github.com/docker-library/official-images/pull/14934
Retire images now that the final version has been built


tianon:
The proactive follow-up PR is really, really appreciated! :smile::heart:

PR #14955, opened 2023-06-29
Updates for release 23.0.0.6

wraschke:
These are updates needed to publish WebSphere and Open Liberty 23.0.0.6 containers. Thank you.

### DIFF
--- a/library/open-liberty
+++ b/library/open-liberty
@@ -2,7 +2,7 @@ Maintainers: Chris Potter <crpotter@us.ibm.com> (@crpotter),
              Wendy Raschke <wraschke@us.ibm.com> (@wraschke)
 GitRepo: https://github.com/OpenLiberty/ci.docker.git
 GitFetch: refs/heads/main
-GitCommit: 5539b65a4b2733610d7cf99bc2f301a71cf0e309
+GitCommit: 30700d72a85a7afc758d779691062ea7ec80dc34
 Architectures: amd64, i386, ppc64le, s390x
 
 Tags: beta
--- a/library/open-liberty
+++ b/library/open-liberty
@@ -2,7 +2,7 @@ Maintainers: Chris Potter <crpotter@us.ibm.com> (@crpotter),
              Wendy Raschke <wraschke@us.ibm.com> (@wraschke)
 GitRepo: https://github.com/OpenLiberty/ci.docker.git
 GitFetch: refs/heads/main
-GitCommit: d75f1a08566ee919548e70ab53efb8b9572135b2
+GitCommit: 5539b65a4b2733610d7cf99bc2f301a71cf0e309
 Architectures: amd64, i386, ppc64le, s390x
 
 Tags: beta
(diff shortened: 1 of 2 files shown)


wraschke:
Okay, cool, now the only check I see failing is one that I've seen in past PRs:

```
```

Can this PR be merged soon?

wraschke:
@yosifkit can you please give an outlook on when this PR can be merged?

yosifkit:
Apologies for the delay, since the Docker official images are curated to ensure their quality, reviews on large changes often get pushed back in favor of the shorter/quicker changes. Large changes require larger contiguous blocks of time for us to understand and provide relevant feedback and that can tax our small team.

Generally, changes to files in the build context should be kept minimal to help keep our build context diffs easy to read so that we can continue to review quickly. If it becomes a large file that changes every release, it may make sense to include it in the upstream release (or released as a fully maintained project/example that can be downloaded by the image).

Here is some feedback that I have so far:

We try hard to avoid multi-stage builds in almost every instance. Since this image does a multi-stage build to basically prevent some `rm`'s and an `apt-get purge --auto-remove ...`, this is unfortunately not an acceptable use of a multi-stage build for official-images. (see also https://github.com/docker-library/faq/#multi-stage-builds).

`RUN apt-get install -y --no-install-recommends openssl`  This has no `apt-get update` and so would fail, but the package is already installed in the base.

`USER root`: this only applies to the `-ubi` images provided by `ibm-semuru-runtimes` in the IBM registry (https://github.com/ibmruntimes/semeru-containers/pull/57), so I am unsure why it was also applied to the Ubuntu based images. If they wanted to add the same `USER` to the Ubuntu-based images in Docker official images, we'd likely push back a little because of this same breakage for users.

mbroz2:
@yosifkit thanks for the feedback and the explanation regarding the process.  We'll do our best to keep the changes to a minimum moving forward to help keep the diffs small.  These past releases have been a bit of an abnormality due to new functionality being included, as well as some changes with regards to how we build the images.

Regarding the multi-stage builds, the reason we switched to them is to allow a local copy of the runtime (or repository) to be used at build time (the `COPY resources/ /tmp/` step), rather then needing to download one.  This change was done for several reasons, but the main benefits of this approach apply to the UBI based images.  Our primary reason for including this change in the Ubuntu based images was to avoid diverging the UBI and Ubuntu based Dockerfiles, though users building their own Liberty Ubuntu based images using these Dockerfiles could leverage some of the benefits as well.  We were unaware that this could cause problems for building the official images, and if necessary, can revert back to a single stage build.

leochr:
@yosifkit thank you for the review and feedback.

Regarding `USER root`, it's good to know you will likely push back on Semeru image changing the user, but if it's alright we would like to explicitly switch to root since some of the instructions in our Dockerfiles should be run as root (we later switch to non-root user). It'll make that explicit/clear to anyone looking at our Dockerfiles and also keep the Dockerfiles aligned with UBI files.

mbroz2:
@yosifkit based on the two comments above, how would you suggest we proceed regarding the multi-stage build and `USER root`?

yosifkit:
We'd rather have them as single stage builds, but the `USER root` is harmless.

wraschke:
@yosifkit can you take another look? Thank you.

leochr:
@yosifkit Thank you for the review. Please merge this PR to publish 23.0.0.6 images. We'll investigate the failures in #15098 and consider the above comment about LABEL, ENV

PR #14960, opened 2023-06-30
Update phpmyadmin to latest upstream changes

williamdes:
Fixes: https://github.com/phpmyadmin/docker/issues/416

### DIFF
--- a/library/phpmyadmin
+++ b/library/phpmyadmin
@@ -1,19 +1,19 @@
-# This file is generated via https://github.com/phpmyadmin/docker/blob/785cd48c4892a4748c0c629099c15b42570949d3/generate-stackbrew-library.sh
+# This file is generated via https://github.com/phpmyadmin/docker/blob/692333aa2d7ffd02f8763862295f78b4c61e6fed/generate-stackbrew-library.sh
 Maintainers: Isaac Bennetch <bennetch@gmail.com> (@ibennetch),
              William Desportes <williamdes@wdes.fr> (@williamdes)
 GitRepo: https://github.com/phpmyadmin/docker.git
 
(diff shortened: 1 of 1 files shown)


tianon:
```diff
+        | awk '{print $1} {system("realpath " $1)}' \
```

Doing `realpath` for the `dpkg-query` bug isn't entirely correct, and there are pathological cases where it might not actually work correctly. :see_no_evil:

What I'd suggest instead is what I proposed in https://github.com/docker-library/python/pull/822#discussion_r1227154811 (and what we used in the PHP image; https://github.com/docker-library/php/blob/0bae31aa4aa0afffdc394072bda779fa652f74fc/Dockerfile-linux.template#L425), which strips any possible `/(usr/)?` prefix so it's still matching the longest possible path:

```dockerfile
		| awk '/=>/ { so = $(NF-1); if (index(so, "/usr/local/") == 1) { next }; gsub("^/(usr/)?", "", so); print so }' \
```

tianon:
You might also enjoy https://github.com/docker-library/wordpress/blob/dd5c229a1900cab9543b180945b8a80754936eac/Dockerfile.template#L125-L128, but I feel a lot less strongly about that change - if you're happy with what you've got there, it's totally fine. :smile:

williamdes:
Hi @ederuiter 
Could you have a look about this feedbacks, let me know if you do not have time so I allocate some time to fix this on the docker repo :)

Thank you @tianon !

ederuiter:
✅ I will have a look at it tomorrow
But with these levels of workarounds in other containers, I think this also should be addressed upstream somehow .. but so far no reaction on that issue: https://github.com/debuerreotype/docker-debian-artifacts/issues/199

tianon:
Nice, thanks for updating!

PR #14971, opened 2023-07-03
[kong] add arm architecture to alpine images

NextBasket-Petyo:


tianon:
Unfortunately, https://github.com/Kong/docker-kong/blob/507775715a22e419d1c7e42028f7c3b5d9cf0418/Dockerfile.apk does not support arm64 (so this won't actually work as-is and requires more upstream changes).

PR #14972, opened 2023-07-03
Update Debian (20230703)

tianon:


tianon:
The diff is really long thanks to a change in `tzdata` combined with a change in `ncurses` :sweat_smile:

```
   * Drop /usr/share/zoneinfo/posix (identical to /usr/share/zoneinfo)
     (LP: #2008076)
   * Split right/* timezones into separate tzdata-legacy package (LP: #2008076)
```

```
   * Move the ncurses-base terminfo files to /usr/share/terminfo
     (Closes: #1028202, #622286).
```

tianon:
Diff: https://github.com/docker-library/official-images/actions/runs/5447591339/attempts/1#summary-14747220021

PR #14983, opened 2023-07-04
library/perl: Add Perl 5.38.0

zakame:
- https://github.com/Perl/docker-perl/pull/136
- https://www.nntp.perl.org/group/perl.perl5.porters/2023/07/msg266602.html

This also enables builds of 5.36.1 and 5.38.0 on Debian bookworm, and removes 5.32.1.


zakame:
Thanks @tianon ! 🎉

PR #14984, opened 2023-07-04
Fedora Update 20230704

cverna:


tianon:
`dnf5` (with lots of removed Python!) :eyes:

tianon:
Pretty dramatic image size decrease, too! (~192MiB down to ~128MiB)

PR #14992, opened 2023-07-06
emqx v4.4.19 and v5.1.1

id:


yosifkit:
Just an FYI that the subshell can be avoided by dropping the echo.
```diff
-    echo "$(sha256sum $pkg)"; \
+    sha256sum $pkg; \
```

PR #15026, opened 2023-07-13
Make 17-debian available

suzaku:
Currently, there's no tag pushed for https://github.com/corretto/corretto-docker/blob/b3911c583ac27a9b5801eba2373a9dd8fe10a3a0/17/jdk/debian/Dockerfile
With this change, we can access the debian variant.


yosifkit:
Adding this would be up to the `amazoncorretto` image maintainers (tagged in the bot comment). But I'd guess that their stance has not changed since https://github.com/corretto/corretto-docker/issues/106#issuecomment-1195988867.

lutkerd:
@yosifkit, Correct this is not something we plan to publish. We'll be updating the repo after the upcoming release for clarify supported vs example Dockerfiles.

PR #15039, opened 2023-07-17
chore: release InfluxDB 1.11.1

bnpfeife:


yosifkit:
Dockefile changes look good to me.

Did you want to keep these removed tags? (It seems like they are currently tied to the version of the non-enterprise tags, which only have `1.8` and not `1.9`, `1.10`, or `1.11`)
```diff
-influxdb:data
-influxdb:data-alpine
-influxdb:meta
-influxdb:meta-alpine
```

yosifkit:
Yeah, they'll still point at the same images as the `1.8-data*` and `1.8-meta*` images and remain unchanged. If that's fine we can merge this.

yosifkit:
Merging to get the update out. Feel free to add back the plain `data` and/or `meta` tags when necessary.

PR #15043, opened 2023-07-17
SapMachine: Add missing JRE images

RealCLanger:
Before my last PR, testing for the added images failed. Should hopefully work now.


RealCLanger:
Ready from my end.

yosifkit:
Just some food for thought, not a blocker for official images: with `ubuntu` now in the tags what happens when the image is updated to Ubuntu `24.04` or `26.04` (just a natural update or maybe `22.04` end of life). We've seen in the past that users can run into issues at major OS updates. One way we've tried to combat this in images we maintain is by providing more specific release tags that can let them escape to an older version (even if we only support one OS release). For example, we have Debian's `bookworm` and `bullseye` releases in the [redis](https://hub.docker.com/_/redis) tags, but only build/support the most recent.

PR #15056, opened 2023-07-19
Releasing Neo4j 5.10.0

jennyowen:
I also have some exciting redhat versions of Neo4j 5.10.0 waiting to be released, but I'll include those in a separate PR after this one so that they don't block the usual stuff from being released.
Thanks!


yosifkit:
A little debugging help in case you have users that suddenly cannot run things in the Debian Bookworm based images; the answer is that it is `libseccomp` related. See also https://github.com/docker-library/python/issues/837#issuecomment-1599640563.

jennyowen:
thanks for the information @yosifkit !
Maybe I'll just make life easier for myself and bump the base image back down to bullseye.

PR #15062, opened 2023-07-20
releasing Neo4j 5.10.0 redhat images

jennyowen:
ok here goes... the long promised redhat neo4j images :cold_sweat:

### DIFF
--- a/library/neo4j
+++ b/library/neo4j
@@ -3,25 +3,24 @@ Maintainers: Jenny Owen <jenny.owen@neo4j.com> (@jennyowen),
              Eric Sporre <eric.sporre@neo4j.com> (@ericsporre)
 GitRepo: https://github.com/neo4j/docker-neo4j-publish.git
 
-
-Tags: 5.10.0, 5.10.0-community, 5.10.0-debian, 5.10.0-community-debian, 5.10, 5.10-community, 5.10-debian, 5.10-community-debian, 5, 5-community, 5-debian, 5-community-debian, community, latest, community-debian, debian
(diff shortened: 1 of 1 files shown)


yosifkit:
Just a few little points of feedback:
 - In the RUN line that installs `tini` and `gosu`, the `set -e` (and optionally the `-ux` flags of the `set`) needs to be first in the `RUN` line. In the off chance that the `microdnf install` exits with a non-zero exit code, it would currently blunder forward and so could create an incorrect image.
     ```diff
     -RUN microdnf install -y \
     ...
     -    set -eux; \
     +RUN set -eux; \
     +    microdnf install -y \
     ...
     ```
 - Similar feedback on the other `RUN` that downloads and installs Neo4j is that it needs to start with a `set -e;` in order for the `;` format of the shell code to exit on a failure (the `u` & `x` flags could also be added if desired).
 - as far as `debian` in the tags, I have similar feedback as https://github.com/docker-library/official-images/pull/15043#issuecomment-1639043630; i.e. would it help users to make it more specific, like `bullseye`, so that users are aware when a major OS update happens?

yosifkit (approved):
```dockerfile
RUN ...\
    gpg --batch --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys \
        595E85A6B1B4779EA4DAAEC70B588DFF0527A9B7 \
        B42F6819007F00F88E364FD4036A9C25BF357DD4; \
```

`--recv-keys` won't fail if it happens to not fetch one of the keys, so we usually recommend putting each key in its own `gpg --recv-keys`. However, I think it's ok here since each key is then used via each of the `--verify` commands, so it would catch any network problems in key fetching.  So, LGTM. cc @tianon

tianon:
A minor future improvement could be to embed sha256 hashes for the two binaries and checking their signatures separately (so the keys match each project correctly), but not a strong blocker. :+1:

PR #13093, opened 2022-09-02
[Swift] Update to Swift 5.6.3

shahmishal:

PR #15071, opened 2023-07-21
[Swift] Add support for Windows Docker image

shahmishal:


shahmishal:
```
Run set -- 'swift:5.8.1-windowsservercore-ltsc2022'
  set -- 'swift:5.8.1-windowsservercore-ltsc2022'
  if [ -s ./.test/config.sh ]; then set -- --config ~/oi/test/config.sh --config ./.test/config.sh "$@"; fi
  ./test/run.sh "$@"
  shell: C:\Program Files\Git\bin\bash.EXE -Eeuo pipefail -x {0}
  env:
    GIT_LFS_SKIP_SMUDGE: 1
    MSYS: winsymlinks:nativestrict
    BASHBREW_VERSION: v0.1.8
    BASHBREW_SCRIPTS: D:\a\_actions\docker-library\bashbrew\v0.1.8/scripts
    BASHBREW_BUILDKIT_SYNTAX: docker/dockerfile:1@sha256:39b85bbfa7536a5feceb7372a0817649ecb2724562a38360f4d6a7782a409b14
+ set -- swift:5.8.1-windowsservercore-ltsc2022
+ '[' -s ./.test/config.sh ']'
+ ./test/run.sh swift:5.8.1-windowsservercore-ltsc2022
testing swift:5.8.1-windowsservercore-ltsc2022
	'override-cmd' [1/2]...passed
<unknown>:0: error: could not load the swift standard library
	'swift-hello-world' [2/2]...failed
```

cc: @compnerd

tianon:
This feels familiar :eyes: (https://github.com/docker-library/official-images/pull/13093)

grooverdan:
https://github.com/docker-library/official-images/tree/master/test/tests/swift-hello-world

compnerd:
@grooverdan, @tianon thanks!  I did poke a bit more at this - the problem is that this is running under MinGW (bash) and that is not supported.  That tends to break pretty often and is not supported.  Can we run this as a PowerShell script somehow?

yosifkit:
I have a fix for the test to make it work on windows. The solution is to just use `swiftc` and then run the result. I am happy to add the commit to this branch but figured I'd get review done first. The change in testing would also apply to the non-windows versions so it is OS agnostic. Here is what I came up with:

<details>
<summary>Diff:</summary>

```diff
--- a/test/tests/run-swift-in-container.sh
+++ b/test/tests/run-swift-in-container.sh
@@ -4,4 +4,15 @@ set -e
 testDir="$(readlink -f "$(dirname "$BASH_SOURCE")")"
 runDir="$(dirname "$(readlink -f "$BASH_SOURCE")")"

-source "$runDir/run-in-container.sh" "$testDir" "$1" swift ./container.swift
+image="$1"
+
+testName="$(basename "$testDir")"
+newImage="$("$runDir/image-name.sh" "librarytest/$testName" "$image")"
+"$runDir/docker-build.sh" "$testDir" "$newImage" <<EOD
+FROM $image
+COPY dir/container.swift /
+RUN swiftc /container.swift -o container
+CMD [ "/container" ]
+EOD
+
+docker run --rm "$newImage"
```

</details>

Dockerfile review:

The installers for `vs_buildtools`, `git`, `swift` and `python` should be verified in some way: https://github.com/docker-library/official-images/tree/15490443c295a651575af2e18a78003be2f1747c#image-build.

Python `3.10.*` no longer receives updates to the binary builds, so this version is already out of date. I don't have a great solution other than updating to Python `3.11`. Have you considered just using `FROM python:3.11-windowsservercore-ltsc2022` instead?

I am also very intrigued by the size disparity between this python install layer and the one in the `python` images. The size here is 10x bigger. I did some investigations and even copied the python Dockerfile and it was still huge. I discovered that if python is installed before `vs_buildtools`, then it is the expected ~90MB instead of 990MB.  I also noticed the swift layer gets smaller also, but then the VS layer gains almost as much as the swift layer saved.

More space can be saved by cleaning up the `$TEMP` directory in each `RUN` layer. I also recommend adding something like the [python image](https://github.com/docker-library/python/blob/15678510e9b41b3ad7e81da4569780139fed1101/3.11/windows/windowsservercore-ltsc2022/Dockerfile#L35-L39) so that any installer errors can print a log on failure to help debug later issues.

compnerd:
Unfortunately, the python version is pinned and we cannot update that easily :-( (the lldb instance relies on the python matching since python has no stable ABI).  I didn't know about the python base image, if that shaves size, I'm happy to swap to it.

yosifkit:
>Unfortunately, the python version is pinned and we cannot update that easily :-( (the lldb instance relies on the python matching since python has no stable ABI). I didn't know about the python base image, if that shaves size, I'm happy to swap to it.

Well, since you need a specific, pinned version python, then the images we maintain won't help since we don't maintain old releases (and `3.10` doesn't have binary releases for windows anymore, so that combination is dropped altogether). I would suggest changing the order of installation to get some space saving though. Installing `vs_buildtools` last should save about 900MB.

compnerd:
That is slightly ludicrous.  Is there an explanation for _why_?  I'm not against it, just completely flabbergasted at the results.  It absolutely is worth the effort to do (unfortunately for older releases of Swift this would not be possible since there is a dependency on install ordering).

compnerd:
@tianon @yosifkit @grooverdan - anything that I can do to help push this along?

yosifkit:
Apologies for the delay.  The Dockerfiles need a few updates.

The installers for `vs_buildtools`, `git`, `swift` and `python` should be verified in some way (at least a sha256sum embedded in the Dockerfile): https://github.com/docker-library/official-images/tree/15490443c295a651575af2e18a78003be2f1747c#image-build.

To get the size saving, I'd suggest installing `python` before `vs_buildtools`. It seems to be coming from `ProgramData/Package Cache` when doing a diff of files between the two layers that install python.

The `$TEMP` directory should be cleaned out in each `RUN` layer. I also recommend adding something like the [python image](https://github.com/docker-library/python/blob/15678510e9b41b3ad7e81da4569780139fed1101/3.11/windows/windowsservercore-ltsc2022/Dockerfile#L35-L39) so that any installer errors can print a log on failure to help debug later issues.

For [cacheability](https://github.com/docker-library/official-images/tree/4af34920e4098cf510d4e4cf2e64b5844e2a5486#cacheability) optimization we suggest putting all `ARG`/`ENV` lines just before the `RUN` in which they are used to maximize build cache if only a later item changed. For example, if swift is last and the only item is a swift version bump, then only the last layers need to build.

---

This last part has me very confused. If I use the same python install arguments you have, but running it via powershell (same as the `python` image), it exits with a failure and so fails to install. I could only solve it by removing the `Include_launcher=0` argument. So, I am a little hesitant to trust the `start /w` of these installations from `CMD`, since it didn't capture this python installation failure (this is unrelated to the size issue as I was using powershell in both cases).

compnerd:
@yosifkit do you have examples for Windows that do that with sha?

LaurentGoderre:
@compnerd here is an example: https://github.com/docker-library/pypy/blob/master/Dockerfile-windows-servercore.template#L97-L101

compnerd:
Thank you!  I ended up with something similar (and tweaking the order as well for the space savings).

compnerd:
@LaurentGoderre @yosifkit - seems that we cannot really do the hash checking on the VS installer as the bootstrapper seems to change.

LaurentGoderre:
@compnerd isn't a CI creating the final installer? Would it change from that?

compnerd:
@LaurentGoderre no, the VS installer is a dependency that we are installing that is not under our control and only has a "latest" version.

LaurentGoderre:
@compnerd is it this step? https://github.com/apple/swift-docker

compnerd:
@LaurentGoderre I think that I might have a version that works for now? Let me upload the version so I can indicate the step.

compnerd:
@LaurentGoderre https://github.com/apple/swift-docker/pull/357/files#diff-0013beb8eefbd6f33dc8dc6e78fa8b0638ea382f744ff311b14d3189fd5c6a2aR46-R54

compnerd:
@LaurentGoderre, @yosifkit:

- added verification for the installers (sha256sum)
- installing python before vs build tools
- cleaning out $TEMP logs
- moved the `ARG` to near site of use
- migrated the rules to powershell

Tested on Windows 22H2 (22621.2428).

compnerd:
@LaurentGoderre - seems that the bootstrapper has changed between yesterday and today. What do you recommend as the best way to deal with this?

SHA256 yesterday: 96e08166034c5110cdb8361ba55f8a2010d0bad714b311a7c8f1c7c8f7c03bce
SHA256 today: d4e08524cb0e5bd061a24f507928d1cfb91dce192c5e12ed964b8343fc4cdedd

compnerd:
@LaurentGoderre @yosifkit - seems that there is an infrastructure failure here? Could you please advise on how to get this resolved?

compnerd:
The windows failure is also happening elsewhere: https://github.com/docker-library/official-images/actions/runs/6617790120/job/18015651374?pr=15602 (from #15602)

yosifkit:
> seems that there is an infrastructure failure here? Could you please advise on how to get this resolved?

>The windows failure is also happening elsewhere

Currently working to get the fix to the GitHub action released.

LaurentGoderre:
Almost there. Now there's a problem with only one of the Windows variant.

https://github.com/docker-library/official-images/actions/runs/6656340547/job/18088715252?pr=15071#step:9:17

LaurentGoderre:
@compnerd yep, talked to him about it

compnerd:
@LaurentGoderre what's the best way forward for this? Do we want to just push that change into this PR, have a separate PR? Or will @yosifkit push that?

yosifkit:
I can add it as an extra commit to this PR once we're ready. I'm just working on reviewing it once again.

yosifkit:
I haven't really realized this before (and I think some of our own images might have this bug), but it looks like just running `Start-Process` is not enough to correctly fail a build if the started process fails (even with `$ErrorActionPreference = 'Stop` set). So, the recommendation is to implement a check of the exit code like the `python` windows [images do](https://github.com/docker-library/python/blob/2d31ccc9f8487908ded7944a54b8e923eff9ad1f/Dockerfile-windows.template#L17-L33), summarized and explained below:

```powershell
# https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.management/start-process
$exitCode = (Start-Process executable-to-run -Wait -NoNewWindow -PassThru \
		-ArgumentList @( \
			'arg1', \
			'arg2', \
			'argn' \
		) \
	).ExitCode; \
	if ($exitCode -ne 0) { \
		Write-Host ('Message to print when the executable-to-run exited non-zero: {0}' -f $exitCode); \
# not required, but attempting to print some log files from TEMP can be helpful in debugging failing installs.
# this example might miss printing the correct log file if there are multiple with the same LastWrite time
		Get-ChildItem $env:TEMP | Sort-Object -Descending -Property LastWriteTime | Select-Object -First 1 | Get-Content; \
# be sure to fail the build so that a partially installed dependency doesn't get committed to a layer
		exit $exitCode; \
	} \
```

compnerd:
Okay, redid some of the image setup; found that Microsoft has a pre-setup python image and I opted to use that and add the additional checking around the other installations.

compnerd:
@yosifkit, the changes that you requested are now completed. The python installation ordering is important, and using the Microsoft provided python bundled image results in a smaller image overall (~13.2G vs ~14.8G).

compnerd:
@yosifkit @tianon @LaurentGoderre - sorry to keep bugging you guys about this, but I really would like to get this PR resolved.

yosifkit:
From the failing "Naughty" test:
> error: unknown Windows image: mcr.microsoft.com/windows-cssc/python:3.9-servercore-ltsc2022
> Error: Process completed with exit code 1.

And then from the [Official Images Readme](https://github.com/docker-library/official-images/tree/0ef4a0297c5109b51e1d6fdc328d319853fc0898#repeatability):
> No official images can be derived from, or depend on, non-official images (allowing the non-image [scratch](https://hub.docker.com/_/scratch/) and the intentionally limited exceptions pinned in [.external-pins](https://github.com/docker-library/official-images/blob/master/.external-pins) -- see also [.external-pins/list.sh](https://github.com/docker-library/official-images/blob/master/.external-pins/list.sh)).

Unfortunately, the `mcr.microsoft.com/windows-cssc/python:*` images are not part of the allowed exceptions.

compnerd:
@yosifkit - is there a way to get the Dockerfile for that image?  It would be nice to do something similar - otherwise, we can ignore the error and just run the python installer I suppose.

whalelines:
Depending on how the image was created, you can sometimes extract details about the Dockerfile by inspecting the image or interrogating the image's provenance, if provided.

https://docs.docker.com/engine/reference/commandline/inspect/
https://docs.docker.com/build/attestations/slsa-provenance/

shahmishal:
I updated the SHA to pick up changes from @compnerd

shahmishal:
```
+ ./test/run.sh swift:5.9.1-windowsservercore-ltsc2022
testing swift:5.9.1-windowsservercore-ltsc2022
	'override-cmd' [1/2]...passed
<unknown>:0: error: could not load the swift standard library
	'swift-hello-world' [2/2]...failed
```

compnerd:
> I can add it as an extra commit to this PR once we're ready. I'm just working on reviewing it once again.

I think that we are at that point!

LaurentGoderre:
It seem problematic that the Sha changes all the time. It would mean that if the Windows base image gets a security patch, a rebuild of this will fail because the Sha will have changed

compnerd:
@LaurentGoderre - for now, it just records that the SHA for the VS build tools bootstrapper is mismatched, so it should rebuild. I don't see any way to control that as there is no pinned URL that Microsoft has.

yosifkit:
Just a couple review points:
- minor typo in the python installation block:
    ```diff
    -    Write-Host -NoNewLine ('Verifying SHA256 ({0}) ... ' -f ${env:PY39_SH256}); \
    +    Write-Host -NoNewLine ('Verifying SHA256 ({0}) ... ' -f ${env:PY39_SHA256}); \
    ```
- The python installation got moved after `vs_buildtools`, so the image is bigger again:
````console
$ docker images
REPOSITORY                             TAG                 IMAGE ID       CREATED          SIZE
swift                                  test-reorder        47f92e93d96f   2 minutes ago    12.9GB
swift                                  test-new            d3215fc02273   17 hours ago     13.8GB
$ docker history swift:test-reorder
IMAGE          CREATED             CREATED BY                                      SIZE      COMMENT
47f92e93d96f   About an hour ago   powershell -Command $ErrorActionPreference =…   41kB
f14b5afaf41b   About an hour ago   |8 GIT=https://github.com/git-for-windows/gi…   3.88GB
e3c9d85e859b   About an hour ago   powershell -Command $ErrorActionPreference =…   41kB
428846da6b5b   About an hour ago   powershell -Command $ErrorActionPreference =…   41kB
5be92c0cb314   2 hours ago         |6 GIT=https://github.com/git-for-windows/gi…   4.49GB
4ff06a63cd2c   2 hours ago         powershell -Command $ErrorActionPreference =…   41kB
25fc6976ebd4   2 hours ago         powershell -Command $ErrorActionPreference =…   41kB
66f2d9d9e559   2 hours ago         |4 GIT=https://github.com/git-for-windows/gi…   91.6MB
a1f55ed4f134   2 hours ago         powershell -Command $ErrorActionPreference =…   41kB
3e1a580e48e0   2 hours ago         powershell -Command $ErrorActionPreference =…   41kB
682acaf68a0c   21 hours ago        |2 GIT=https://github.com/git-for-windows/gi…   379MB
5ae0ef8cdb2a   21 hours ago        powershell -Command $ErrorActionPreference =…   41kB
1c6169758db4   21 hours ago        powershell -Command $ErrorActionPreference =…   41kB
81addb6469a1   21 hours ago        powershell -Command $ErrorActionPreference =…   41kB
565570542ca6   8 days ago          powershell -Command $ErrorActionPreference =…   41kB
72fdef526f1c   2 months ago        cmd /S /C #(nop)  LABEL description=Docker C…   41kB
e5bd6c080618   2 months ago        cmd /S /C #(nop)  LABEL maintainer=Swift Inf…   41kB
4742812abba1   3 months ago        Install update 10.0.20348.1906                  961MB
<missing>      4 months ago        Apply image 10.0.20348.1787                     3.06GB
$ docker history swift:test-new
IMAGE          CREATED        CREATED BY                                      SIZE      COMMENT
d3215fc02273   19 hours ago   powershell -Command $ErrorActionPreference =…   41kB
fcdb5e22f47d   19 hours ago   |8 GIT=https://github.com/git-for-windows/gi…   1.04GB
8c30f588be56   19 hours ago   powershell -Command $ErrorActionPreference =…   41kB
29c7881674e7   19 hours ago   powershell -Command $ErrorActionPreference =…   41kB
e69bc25b3a53   19 hours ago   |6 GIT=https://github.com/git-for-windows/gi…   3.87GB
eca397835e43   20 hours ago   powershell -Command $ErrorActionPreference =…   41kB
3d137b648862   20 hours ago   powershell -Command $ErrorActionPreference =…   41kB
fecd9c6100fa   20 hours ago   |4 GIT=https://github.com/git-for-windows/gi…   4.48GB
1209e34793e8   21 hours ago   powershell -Command $ErrorActionPreference =…   41kB
6249d947eaa2   21 hours ago   powershell -Command $ErrorActionPreference =…   41kB
682acaf68a0c   21 hours ago   |2 GIT=https://github.com/git-for-windows/gi…   379MB
5ae0ef8cdb2a   21 hours ago   powershell -Command $ErrorActionPreference =…   41kB
1c6169758db4   21 hours ago   powershell -Command $ErrorActionPreference =…   41kB
81addb6469a1   21 hours ago   powershell -Command $ErrorActionPreference =…   41kB
565570542ca6   8 days ago     powershell -Command $ErrorActionPreference =…   41kB
72fdef526f1c   2 months ago   cmd /S /C #(nop)  LABEL description=Docker C…   41kB
e5bd6c080618   2 months ago   cmd /S /C #(nop)  LABEL maintainer=Swift Inf…   41kB
4742812abba1   3 months ago   Install update 10.0.20348.1906                  961MB
<missing>      4 months ago   Apply image 10.0.20348.1787                     3.06GB
```

compnerd:
Swapped the order but I don't see a difference locally (its 13.1G either way).

yosifkit:
> Swapped the order but I don't see a difference locally (its 13.1G either way).

That's super strange. I re-pulled the windows base image to get it more up-to-date and tried again and still they are very different.  I'm building on a Windows 11 host using Powershell and Docker Desktop.
```console
docker images
REPOSITORY                             TAG            IMAGE ID       CREATED          SIZE
swift                                  test-reorder   b5bfe34d3d86   54 minutes ago   13GB
swift                                  test-new       bb65ec56ae86   2 hours ago      13.9GB
```

compnerd:
@yosifkit interesting.  I think it might have something to do with a pretty old installation of Windows in my case.  Can you help get the test changes applied so that we can get this merged?

yosifkit:
Squash & rebased, updated swift test, added updated swift-docker 5.9 Windows commit to library file: https://github.com/apple/swift-docker/commit/3fc609343f771e31c17140ed90d6cb2023e6e701

shahmishal:
🎉

yosifkit (approved):

PR #15079, opened 2023-07-24
Add NGINX Agent as an official library image

dhurley:
Hi,

This PR is to start the process of adding [NGINX Agent](https://github.com/nginx/agent) docker images under the umbrella of Docker Library.

NGINX Agent is a companion daemon for your NGINX Open Source instance. It provides gRPC and REST interfaces for configuration management and metrics collection from the NGINX process and operating system. We use the official NGINX image as our base image and install the NGINX Agent alongside it.

To help fill out a review checklist:

- I am an employee of F5, Inc. / NGINX Inc., and I have [commit rights](https://github.com/nginx/agent/commits?author=dhurley) to the repo. Our devteam leaders are OK with the enrolment to the Official Library.
- NGINX Agent is published under [Apache 2.0](https://github.com/nginx/agent/blob/main/LICENSE) source code license.
- Docs PR https://github.com/docker-library/docs/pull/2353

# Checklist for Review

### DIFF
--- a/library/nginx-agent
+++ b/library/nginx-agent
@@ -1,0 +1,14 @@
+Maintainers: NGINX Docker Maintainers <docker-maint@nginx.com> (@nginxinc)
+GitRepo: https://github.com/nginx/agent.git
+
+Tags: 2.27.0-nginx-mainline-alpine, nginx-mainline-alpine, latest
+Architectures: amd64, arm64v8
+GitCommit: 2f22eb8f373701cffa8a4dde78877e2df1d69840
+Directory: scripts/docker/official/nginx-oss-with-nginx-agent/alpine
+File: Dockerfile.mainline
+
+Tags: 2.27.0-nginx-stable-alpine, nginx-stable-alpine
+Architectures: amd64, arm64v8
+GitCommit: 2f22eb8f373701cffa8a4dde78877e2df1d69840
+Directory: scripts/docker/official/nginx-oss-with-nginx-agent/alpine
+File: Dockerfile.stable


tianon:
Thanks for the contribution!  I haven't had a chance to dig into the details yet (or do a proper review of the Dockerization/software), but a bit of early feedback would be that I'm not sure this makes sense as a standalone image given that it's only really useful for NGINX itself.  As far as I can tell in my limited look at it, it seems to be primarily a kind of log-based polyfill for some amount of NGINX Plus functionality?  (In other words, it does not appear to be "standalone" or "generally useful" software by itself, and only does something in the context of NGINX, unlike NGINX or even NGINX Unit.)

Additionally, having an image that runs some bit of software with another as a sidecar (and some fragile shell code to manage both as if it were an init system) is not going to meet our requirements (see https://github.com/docker-library/official-images#init for more).

I don't know if you've discussed with the maintainer of the NGINX image (cc @thresheek), but this might make more sense as an alternate tag of that repository (since they're strongly tied together), and would have a better chance of being accepted if it were running alongside a separate NGINX container instead. :bow:

dhurley:
> Thanks for the contribution! I haven't had a chance to dig into the details yet (or do a proper review of the Dockerization/software), but a bit of early feedback would be that I'm not sure this makes sense as a standalone image given that it's only really useful for NGINX itself. As far as I can tell in my limited look at it, it seems to be primarily a kind of log-based polyfill for some amount of NGINX Plus functionality? (In other words, it does not appear to be "standalone" or "generally useful" software by itself, and only does something in the context of NGINX, unlike NGINX or even NGINX Unit.)
> 
> Additionally, having an image that runs some bit of software with another as a sidecar (and some fragile shell code to manage both as if it were an init system) is not going to meet our requirements (see https://github.com/docker-library/official-images#init for more).
> 
> I don't know if you've discussed with the maintainer of the NGINX image (cc @thresheek), but this might make more sense as an alternate tag of that repository (since they're strongly tied together), and would have a better chance of being accepted if it were running alongside a separate NGINX container instead. 🙇

Hi @tianon, thanks for the feedback. I just have a question in relation to your suggestion of possibly adding an alternative tag to the already existing NGINX image. If we were to add an additional tag to the NGINX image that would provide an image that would have NGINX and NGINX Agent running alongside one another in the one container, would that kind of image be accepted?

We did explore the idea of having NGINX Agent running in a separate container to NGINX but we ran into difficulty with that based on how the NGINX Agent interacts with NGINX. Thats why we decided to go with just the one container that had both NGINX and NGINX Agent.

tianon:
Unfortunately not, as there really isn't a great way to reliably manage both processes (especially when it comes to the failure modes).

yosifkit:

Unfortunately, our strict "one concern" per container means NGINX Agent doesn't quite fit into the mold. If Agent were a process that managed NGINX server as a child process or vise versa rather than co-processes managed by a shell script it would be easier to accept. Also, it'd need to be a variant of the `nginx:*` images since it is only useful in the context of NGINX server.

PR #15087, opened 2023-07-24
SapMachine: Add distroless images and add release specific ubuntu tags

RealCLanger:


tianon:
(see https://github.com/docker-library/official-images#repeatability, especially "No official images can be derived from, or depend on, non-official images")

RealCLanger:
> (see https://github.com/docker-library/official-images#repeatability, especially "No official images can be derived from, or depend on, non-official images")

OK, understood.

However, the message in the failing naughty check is misleading, see here: https://github.com/docker-library/official-images/actions/runs/5650086573/job/15305765066

It says `Invalid 'FROM' + 'Architectures' combinations detected:`. So I was looking for problems with the architecture definitions first. The [link](https://github.com/docker-library/official-images#multiple-architectures ) that is mentioned also doesn't inform that non-official repositories are blocked.

PR #15088, opened 2023-07-25
SapMachine: New versions and release specific ubuntu tags

RealCLanger:
This PR adds release specific Ubuntu tags to enable support for multiple Ubuntu versions.

Furthermore it updates to the recent SapMachine versions.

### DIFF
--- a/library/sapmachine
+++ b/library/sapmachine
@@ -1,62 +1,42 @@
 Maintainers: Christoph Langer <sapmachine@sap.com> (@realclanger), Christian Halstrick <sapmachine@sap.com> (@chalstrick)
 GitRepo: https://github.com/SAP/SapMachine-infrastructure.git
 
-Tags: jre-headless-ubuntu-11, jre-headless-ubuntu-11.0.20
+Tags: 11-jdk-headless-ubuntu, 11-jdk-headless-ubuntu-jammy, 11-jdk-headless-ubuntu-22.04, 11.0.20.1-jdk-headless-ubuntu, 11.0.20.1-jdk-headless-ubuntu-jammy, 11.0.20.1-jdk-headless-ubuntu-22.04
 Architectures: amd64, arm64v8, ppc64le
-GitCommit: 03a69c14ccf4cdb173bb67823d779770bcb44d72
-Directory: dockerfiles/official/11/ubuntu/jre-headless
-
-Tags: jre-ubuntu-11, jre-ubuntu-11.0.20
-Architectures: amd64, arm64v8, ppc64le
-GitCommit: 03a69c14ccf4cdb173bb67823d779770bcb44d72
-Directory: dockerfiles/official/11/ubuntu/jre
-
-Tags: jdk-headless-ubuntu-11, jdk-headless-ubuntu-11.0.20
-Architectures: amd64, arm64v8, ppc64le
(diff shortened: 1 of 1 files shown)


yosifkit:
🤔 With these new tags that have the Ubuntu version number next to the OpenJDK version, like `jdk-ubuntu-22.04-20`, it might make it a little harder for users to see the OpenJDK version at a glance or misinterpret which item the version number applies to. For example, they may pull `sapmachine:jdk-ubuntu-20` and expect to get Ubuntu 20.04.

Would it make sense to move the OpenJDK version to be earlier in the tag? Something like this:
 - `jdk-ubuntu-22.04-20` -> `20-jdk-ubuntu-22.04`
 - `jdk-headless-ubuntu-lts` -> `lts-jdk-headless-ubuntu`

Official images tend to have the packaged software version first in the tag (at least the other openjdk builds: [amazoncorretto](https://github.com/docker-library/official-images/blob/master/library/amazoncorretto), [eclipse-temurin](https://github.com/docker-library/official-images/blob/master/library/eclipse-temurin), [ibm-semeru-runtimes](https://github.com/docker-library/official-images/blob/master/library/ibm-semeru-runtimes)), but it is up to you either way.

yosifkit:
Friendly ping, @RealCLanger. Do you have any thoughts about my comment, or should we merge as-is?

LaurentGoderre:
I agree with @yosifkit that it would be preferable to have a clearer separation between the software and os versions to avoid confusion

RealCLanger:
> Friendly ping, @RealCLanger. Do you have any thoughts about my comment, or should we merge as-is?

Yes, these are good suggestions. Will update the PR later today. Sorry for taking so long to reply but I somehow dropped the hammer on that one for a while, due to other business. 😉

RealCLanger:
OK, here's the new proposal. It also updates to the latest SapMachine versions. I force-pushed, hope that's ok in terms of reviewing. Please check and if ok, merge the PR asap. Thanks.

RealCLanger:
This PR does not include the SapMachine 11 and 17 JRE images because of an issue in testing. However, once the new style tags from this PR are available in Docker library, I'll add the JREs again. Then the test will work, because it then finds the container with the right SapMachine version for its testing.

yosifkit (approved):
LGTM ❤️ 

If our rough `java-*` tests are only failing on `jre` images that is a failing on us. So, feel free to include them even if they fail; we understand that our `java-*` tests have some [rough edges](https://github.com/docker-library/official-images/blob/6ab25cec0e1bf12bf3d03a5e2e22cbdbf08d333a/test/tests/run-java-in-container.sh#L9C1-L9C1) 🙈😢.

RealCLanger:
> LGTM ❤️
> 
> If our rough `java-*` tests are only failing on `jre` images that is a failing on us. So, feel free to include them even if they fail; we understand that our `java-*` tests have some [rough edges](https://github.com/docker-library/official-images/blob/6ab25cec0e1bf12bf3d03a5e2e22cbdbf08d333a/test/tests/run-java-in-container.sh#L9C1-L9C1) 🙈😢.

No problem. It's just on the initial submit. I've opened #15405 to add the missing pieces. Let's see whether tests are green now.

Thanks for merging this.

PR #15104, opened 2023-07-26
arangodb 3.11.2

KVS85:
Updated ArangoDB 3.11 to 3.11.2.

### DIFF
--- a/library/arangodb
+++ b/library/arangodb
@@ -12,7 +12,7 @@ Architectures: amd64, arm64v8
 GitCommit: bdd1fadac33809b6344b6ce1adceef73717e576e
 Directory: alpine/3.10.9
 
-Tags: 3.11, 3.11.1, latest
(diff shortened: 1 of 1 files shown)


yosifkit:
😕 There doesn't seem to be any Dockerfile changes in the new version (i.e. missing the version bump).  It looks like the script failed to bump it: https://github.com/arangodb/arangodb-docker/blob/0b588f0287b8992f21807f11f9af039252588a6c/alpine/3.11.2/Dockerfile

KVS85:
> 😕 There doesn't seem to be any Dockerfile changes in the new version (i.e. missing the version bump). It looks like the script failed to bump it: https://github.com/arangodb/arangodb-docker/blob/0b588f0287b8992f21807f11f9af039252588a6c/alpine/3.11.2/Dockerfile

Should be fixed now.

PR #14867, opened 2023-06-15
Update Ubuntu 2023-06-15

ROCKsBot:
Update Ubuntu to the latest published OCI tarballs as of 2023-06-15.

PR #15134, opened 2023-08-03
Add Couchbase Server EE 7.1.5

ceejatec:


yosifkit:
The test failure is unrelated to the new version, so I will merge.

---

Test failure is because of the removal of Ubuntu 18.04/Bionic since it is EOL:

> Please either drop the affected tags from the supported list (by removing them from your `library/` file) or update them to a supported [Ubuntu] release. 
>
> \- https://github.com/docker-library/official-images/pull/14867#issuecomment-1593769545

ceejatec:
Thanks @yosifkit . Regarding our community-6.6.0 image - Ubuntu 18.04 is the newest OS that we supported for that version, so I guess we need to just remove it. I can pull it out of our official-images/library/couchbase file. What exactly does that "mean" to remove it, though? Eg., if I look here: https://hub.docker.com/_/couchbase/tags?page=1&name=comm I see quite a number of our older versions still there.

yosifkit:
Removing it here will remove it from the "Supported" section on the Hub readme, but the tag will still be available to users who want to pull it.  (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)

ceejatec:
@yosifkit That's a little awkward, since we DO still support that version. The official-images README doesn't seem to cover that case.

PR #15143, opened 2023-08-03
Add 20-al2023-generic for amazoncorretto

rgithubli:
### Notes

Add 20-al2023-generic for amazoncorretto. The Dockerfile is: https://github.com/corretto/corretto-docker/blob/main/20/jdk/al2023-generic/Dockerfile


rgithubli:
This action didn't pass: https://github.com/docker-library/official-images/actions/runs/5755332213/job/15602546549 due to:

```
/bin/sh: can't create /etc/apk/repositories: Interrupted system call
```

The related Dockerfile is not part of our Dockerfile change: [link](https://github.com/corretto/corretto-docker/blob/main/11/jdk/alpine/3.17/Dockerfile). It seems a transient failure due to the host issue. I can't re-run this workflow. Tried build at local, can pass.

PR #15148, opened 2023-08-04
[nats] Release v2.9.21

bruth:
Details can be found [here](https://github.com/nats-io/nats-server/releases/tag/v2.9.21)

### DIFF
--- a/library/nats
+++ b/library/nats
@@ -5,7 +5,7 @@ Maintainers: Derek Collison <derek@synadia.com> (@derekcollison),
              Phil Pennock <pdp@synadia.com> (@philpennock)
 GitRepo: https://github.com/nats-io/nats-docker.git
 GitFetch: refs/heads/main
-GitCommit: 0ba84109d8b45c54685f639bbc582141adceda1a
+GitCommit: 36c73754f3d7b2b2e69a0c4680a48636929012cc
 
 Tags: 2.9.21-alpine3.18, 2.9-alpine3.18, 2-alpine3.18, alpine3.18, 2.9.21-alpine, 2.9-alpine, 2-alpine, alpine
 Architectures: amd64, arm32v6, arm32v7, arm64v8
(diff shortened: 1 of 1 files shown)


yosifkit:
> +	apk add --upgrade libcrypto3 libssl3; \

We recommend against using package upgrades (`apt-get upgrade`/`apk upgrade`/`yum upgrade`/`yum update`) for official-images. When package upgrades are applied in a dependent image, it duplicates content of the base image, making the image larger than necessary. It also only delays the inevitable "there are outdated packages". The Official Images build pipeline makes heavy use of `docker build` cache, so we make periodic base image updates to then fully rebuild all dependent images (e.g. the [Debian image updates](https://github.com/docker-library/official-images/pulls?q=is%3Apr+label%3Alibrary%2Fdebian)).

>

bruth:
@yosifkit The reason we added this line was due to some vulnerabilities found in the current Alpine image. Upgrading the packages resolves these for scanners. See this comment for context: https://github.com/nats-io/nats-docker/pull/115#discussion_r1284568228

If it is still a no-go, then I will remove the line and update the PR.

bruth:
> Many Official Images are maintained by the community or their respective upstream projects, like Ubuntu, Alpine, and Oracle Linux, and are subject to their own maintenance schedule. These refreshed base images also means that any other image in the Official Images program that is FROM them will also be rebuilt (as described in [the project README.md file](https://github.com/docker-library/official-images#library-definition-files)).

Re-reading, so once the alpine image is updated, the NATS images will automatically be rebuilt with the latest patch (since we built on 3.18)?

bruth:
@yosifkit Let us know what the preferred path is here, so we can remediate and get the images up on Docker Hub.

yosifkit:
> Re-reading, so once the alpine image is updated, the NATS images will automatically be rebuilt with the latest patch (since we built on 3.18)?

Correct, which will happen with https://github.com/docker-library/official-images/pull/15156. So, please drop the `apk upgrade`.

bruth:
@yosifkit Updated. Referenced commit: https://github.com/nats-io/nats-docker/commit/36c73754f3d7b2b2e69a0c4680a48636929012cc

PR #15161, opened 2023-08-08
Add images for Jetty 12.0

lachlan-roberts:
Add new images after the release of Jetty `12.0.0`


lachlan-roberts:
I think the reason the these tests are failing is because the 12.0 images do not add any `deploy` or `jsp` modules. We don't do this because there are 4 options (`core`, `ee8`, `ee9`, `ee10`) and we want to leave the decision to the user of which one to use.

I will look at the test script and see if I can add the `ee10-deploy` and `ee10-jsp` modules for the 12.0 images.

lachlan-roberts:
fixed

yosifkit:
```bash
echo "exec " $JAVA $JAVA_OPTIONS "$@" $JAVA_OPTIONS $JETTY_PROPERTIES > $JETTY_BASE/jetty.exec
. $JETTY_BASE/jetty.exec
```

I've seen something similar in a few projects and I've never quite understood why. Is this so that they can use variables within `JAVA_OPTIONS` or `JETTY_PROPERTIES`?  e.g., `docker run -e JAVA_OPTIONS='$JAVA-something'`. I don't see a functional difference between this and just an `eval` except that the eval doesn't create a file.  The current `exec $JAVA $JAVA_OPTIONS "$@" $JAVA_OPTIONS $JETTY_PROPERTIES` is the safest.

I will also note that quoting is lost for "$@" when run through echo, so any spaces within an argument will definitely be wrong and now be two arguments.

lachlan-roberts:
@yosifkit this is for cases like this `ENV JETTY_PROPERTIES "testProp=\"foo bar\""`. The bash variables will do weird things with escaping quotes if used directly with the `exec $JAVA $JAVA_OPTIONS "$@" $JAVA_OPTIONS $JETTY_PROPERTIES`, and we did not find a way around this. This also fails if used with the eval.

Although the tmp file is not ideal it was the only workaround which seemed to work for all cases.

Note: if you attempt to reproduce this, you will need to build Jetty from the current 10.0.x branch (there is no release out with an additional fix required). And then copy over the jetty-home into `/usr/local/jetty` in the docker container. It works with `jetty.exec` tmp file but not direct `exec` or `eval`.

yosifkit:
Ah, I see why an `eval` wouldn't work, it needs to be a single argument, otherwise it concatenates all of them together itself. So this should work and even keep args correct for `$@` (since it gets resolved during the `eval` and not earlier when the string is created):

```bash
eval "exec $JAVA $JAVA_OPTIONS \"\$@\" $JAVA_OPTIONS $JETTY_PROPERTIES"

```

lachlan-roberts:
@yosifkit thanks, I have updated the `docker-entrypoint.sh` file with your suggestion.
It seems to work and is cleaner than using the `jetty.exec` file.

gregw:
@yosifkit just a heads up that your requested change broke some users.  See https://github.com/eclipse/jetty.docker/issues/153
Our current plan is to revert to something we know works.   If you have the shell-foo to fix it better than that, then please comment on that issue.

PR #15177, opened 2023-08-11
Update wordpress

yosifkit:
Changes:

- https://github.com/docker-library/wordpress/commit/6e0d5ac: Update cli to 2.8.1
- https://github.com/docker-library/wordpress/commit/141b55c: Update cli to 2.8.0


yosifkit:
Doh, this is an `update.sh` scraping failure of "update" to the previous CLI, then back to the current.

PR #15190, opened 2023-08-13
Replaced 9.0 with 9.1

scriptmonkey:


yosifkit:
Now that https://github.com/docker-library/official-images/pull/15192 is merged, I'm going to give the build servers a little bit to rebuild the `tomee` `9.0` versions from `eclipse-temurin` so that we don't leave the old version in a possibly broken state (https://github.com/adoptium/containers/issues/415).  Then I'll merge this to get the new version out.

PR #15202, opened 2023-08-15
Update Debian (to 20230814)

tianon:
The most notable change here is that our riscv64 builds are now pointing at the main Debian archive proper (instead of Debian Ports, as prior builds did).  See also https://github.com/docker-library/official-images/pull/15095 (and https://lists.debian.org/debian-riscv/2023/07/msg00053.html).

This is the result of a large effort to get an exciting percentage of the archive rebuilt successfully: https://buildd.debian.org/stats/graph-week-big.png (over 60% now! something like 19.5k riscv64-specific binary packages!)

You can find additional validation I performed to convince myself it was a sane time to make this change at https://gist.github.com/tianon/8352581726068fbc0422297d9ec223e6 (testing `buildpack-deps` packages, and finding the list of "missing" dependencies impressively low).

Given this is unstable *and* a brand new officially supported architecture, I opted to make the transition now even though we're not yet 100% at parity with the Ports builds, especially given that this is the only way we'll see security updates (or any updates of any kind) in these builds. 👍

The good news is that this should mean that in the future, we'll eventually see trixie/testing builds for riscv64 too! 👀

### DIFF
--- a/library/debian
+++ b/library/debian
@@ -7,34 +7,34 @@ GitRepo: https://github.com/debuerreotype/docker-debian-artifacts.git
 GitCommit: ea6ede8f53deaae64dd492001410b689d127e810
 # https://github.com/debuerreotype/docker-debian-artifacts/tree/dist-amd64
 amd64-GitFetch: refs/heads/dist-amd64
-amd64-GitCommit: cdaecbc02940a220749a9b57f5c2444c21919e74
+amd64-GitCommit: e5a19296731859d07765b6ae79aca17954f82b17
 # https://github.com/debuerreotype/docker-debian-artifacts/tree/dist-arm32v5
 arm32v5-GitFetch: refs/heads/dist-arm32v5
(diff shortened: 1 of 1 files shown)


tianon:
https://github.com/debuerreotype/docker-debian-artifacts/compare/6a374511b8ae28b473f8171ad63f3d29605a99f6..8277f4704543e0278445101c2c40515e791f34f6 is a fun comparison, especially the `debian.sources` diffs:

```diff
 Types: deb
-# http://snapshot.debian.org/archive/debian-ports/20230725T000000Z
-URIs: http://deb.debian.org/debian-ports
-Suites: unstable unreleased
+# http://snapshot.debian.org/archive/debian/20230814T000000Z
+URIs: http://deb.debian.org/debian
+Suites: unstable
 Components: main
-Signed-By: /usr/share/keyrings/debian-ports-archive-keyring.gpg
+Signed-By: /usr/share/keyrings/debian-archive-keyring.gpg
```

PR #15203, opened 2023-08-15
Updated swipl devel to 9.1.14

JanWielemaker:
Updates add-ons rocksdb and hdt as well.  These are now based on the
much more portable new C++ interface and will probably compile on the
arm platforms.   I do not have the facilities to verify the build on these
platforms.

### DIFF
--- a/library/swipl
+++ b/library/swipl
@@ -1,11 +1,11 @@
 Maintainers: Jan Wielemaker <jan@swi-prolog.org> (@JanWielemaker),
              Dave Curylo <dave@curylo.org> (@ninjarobot)
 GitRepo: https://github.com/SWI-Prolog/docker-swipl.git
-GitCommit: e3851850186cded44b986da8095df36f5d87f0df
+GitCommit: 367d4601846856eaa2468bef93715919512a1a69
 Architectures: amd64, arm32v7, arm64v8
 
-Tags: latest, 9.1.13
-Directory: 9.1.13/bullseye
+Tags: latest, 9.1.14
+Directory: 9.1.14/bullseye
 
 Tags: stable, 9.0.4
 Directory: 9.0.4/bullseye


yosifkit:
Here is what I get when attempting to build `arm32v7` via `qemu` on Docker Desktop on Windows (WSL):

```console
$ DOCKER_BUILDKIT=0 docker build --platform linux/arm/v7 .
...
ERROR: db/memtable.cc: In member function ‘void rocksdb::MemTable::UpdateOldestKeyTime()’:
ERROR: db/memtable.cc:256:1: error: insn does not satisfy its constraints:
ERROR:   256 | }
ERROR:       | ^
% (
% insn
%  297
%  296
%  298
%  14
%
% (
% set
%
% (
% reg
% :CC_DNE
%  100
ERROR:  cc)
ERROR:         (compare:CC_DNE (ior:SI (ne:SI (reg:SI 0 r0 [166])
ERROR:                     (const_int -1 [0xffffffffffffffff]))
ERROR:                 (ne:SI (reg:SI 1 r1 [+4 ])
ERROR:                     (const_int -1 [0xffffffffffffffff])))
ERROR:             (const_int 0 [0]))) "/usr/include/c++/10/bits/atomic_base.h":509:36 337 {*cmp_ior}
ERROR:      (expr_list:REG_DEAD (reg:SI 1 r1 [+4 ])
ERROR:         (expr_list:REG_DEAD (reg:SI 0 r0 [166])
ERROR:             (nil))))
ERROR: during RTL pass: cprop_hardreg
ERROR: db/memtable.cc:256:1: internal compiler error: in extract_constrain_insn, at recog.c:2195
ERROR: mmap: Permission denied
ERROR: Please submit a full bug report,
ERROR: with preprocessed source if appropriate.
ERROR: See <file:///usr/share/doc/gcc-10/README.Bugs> for instructions.
```
I don't know if this is related to using qemu user-mode emulation, so I'll attempt to run it on an arm chip and report back.

JanWielemaker:
Thanks for the feedback.  Bad luck.   This is not the interface, but rocksdb itself.   A search for `rocksdb docker "arm32v7"` doesn't look promising.   I'm afraid we will have to disable the rocksdb binding again.     Will you do so, or should I make a new PR?   Seems I should also be able to test this using QEMU.   Is there guide how to get Docker running with QEMU platforms on Ubuntu (AMD64)?

yosifkit:
> Will you do so, or should I make a new PR?

I can't adjust your Dockerfile in `SWI-Prolog/docker-swipl` directly.  If you could do that and then adjust this PR to refer to the new commit (`library/swipl` file) we should be good to go with the version bump.

>Seems I should also be able to test this using QEMU. Is there guide how to get Docker running with QEMU platforms on Ubuntu (AMD64)?

I did a quick test with https://wiki.debian.org/QemuUserEmulation in Debian and it seems it is as easy as `apt install binfmt-support qemu-user-static` (I think Ubuntu should work the same):

```console
$ sudo apt install binfmt-support qemu-user-static
...
$ docker run -it --rm --platform linux/arm64/v8 debian:bullseye
Unable to find image 'debian:bullseye' locally
bullseye: Pulling from library/debian
e837d9f05c62: Pull complete
Digest: sha256:1beb7cf458bdfe71b5220cb2069eb45e3fc7eb77a1ccfb169eaebf5f6c4809ab
Status: Downloaded newer image for debian:bullseye
root@9e338fb0fec6:/# uname -a
Linux 9e338fb0fec6 6.1.0-11-amd64 #1 SMP PREEMPT_DYNAMIC 6.1.38-4 (2023-08-08) aarch64 GNU/Linux
root@9e338fb0fec6:/# dpkg --print-architecture
arm64
```

The biggest part when using them is that you can't yet have "one" image tagged as `debian:bullseye-slim` for multiple architectures. So, when you `docker run` (or build) with a `--platform` and then run without `--platform` on the same image name, it will be using the last platform that was pulled.  Also note that we've seen times where `qemu` would fail to run or compile something but it works fine on the CPU.

The [containerd image store](https://docs.docker.com/storage/containerd/) does have support for multiple architectures under one tag, but it is still experimental.

JanWielemaker:
Thanks.    Don't have time for Qemu today.    I've updated the repo and will create a new PR.    Hope HDT works out well.

PR #15212, opened 2023-08-16
Add Zookeeper 3.9.0

31z4:


library/zookeeper
```diff
@@ -10,6 +10,10 @@ Tags: 3.8.2, 3.8
 GitCommit: 952b2f523f4f9a829d77114ddbe08ab9249d820b
 Directory: 3.8.2
 
-Tags: 3.8.2-temurin, 3.8-temurin, latest
+Tags: 3.8.2-temurin, 3.8-temurin
 GitCommit: 952b2f523f4f9a829d77114ddbe08ab9249d820b
 Directory: 3.8.2
+
+Tags: 3.9.0, latest
```
yosifkit, line 17: I'll go ahead and merge, but did you also want `3.9.0-temurin, 3.9-temurin` tags too?

🤔 It seems that the `3.8.2, 3.8` section uses the same source commit and directory as `3.8.2-temurin, 3.8-temurin`, is that correct? The sections could be combined, but the `bashbrew` tool detects the duplication and combines them anyway.

31z4: No I didn't wan't `temurin` tags. I follow the plan described here: https://github.com/docker-library/official-images/pull/12949#issuecomment-1383656252

> Yes, the plan is to move the plain tags during corresponding version upgrades. For example, currently 3.5.9 and 3.5 are based on openjdk. Once 3.5.10 is released 3.5.10 and 3.5 becomes temurin

Good catch about `3.9`. Thanks!

31z4:
It seems that `zookeeper:3.9.0` is missing some architectures (e.g., `linux/amd64`)

image

@yosifkit could you please help investigate?

adam-bielasty-yb:
That merge broke our setup.

We are getting infos like that:
```
[35mzookeeper_1          |[0m 2023-08-18 08:28:47,579 [myid:] - INFO  [NIOWorkerThread-4:o.a.z.s.NIOServerCnxn@339] - Unable to read additional data from client, it probably closed the socket: address = /172.18.0.4:45754, session = 0x1000001bc950000
```
which then it is causing errors to connect to the nodes.

Nothing changes on our side apart from getting the latest ZooKeeper image **3.9.0**. All worked fine on **3.8.2**.
Do you know what might be the issue?

yosifkit:
> It seems that `zookeeper:3.9.0` is missing some architectures (e.g., `linux/amd64`)
>
> @yosifkit could you please help investigate?

Sorry for the delay. Some architecture builders were a more backed up than others with the merge of both Debian and Ubuntu images, so they had many child image rebuilds to get through.

PR #15251, opened 2023-08-23
fix arguments for terminating commands in Jetty

lachlan-roberts:
Fix issue with `docker-entrypoint.sh` which causes issues during startup for several users. 

See https://github.com/eclipse/jetty.docker/issues/153


yosifkit:
Won't this just surface another issue with a different set of users? i.e. the quoting of `"$@"` getting lost when run through echo.  Perhaps reverting back to the `exec` from the official images PR before the last (https://github.com/docker-library/official-images/pull/14550) and add the extra `$JAVA_OPTIONS`?

```bash
exec $JAVA $JAVA_OPTIONS "$@" $JETTY_PROPERTIES
```

lachlan-roberts:
Closing this for now. I am going to focus on getting proper tests for all the use cases we want to support before making any further changes to this.

PR #15259, opened 2023-08-24
Bump Rust to 1.72.0

sfackler:


sfackler:
Oh, should we not be building Trixie images yet?

yosifkit:
We haven't added Trixie to `buildpack-deps` just yet. I can't remember how early we do that since Trixie is still equivalent to `unstable`.

kristof-mattei:
See https://github.com/docker-library/buildpack-deps/pull/149

sfackler:
Updated to remove the trixie images.

PR #15321, opened 2023-09-06
Update ghost

yosifkit:
Changes:

- https://github.com/docker-library/ghost/commit/12e4dad: Merge pull request https://github.com/docker-library/ghost/pull/390 from infosiftr/node-bump
- https://github.com/docker-library/ghost/commit/a023271: Update to 5.61.3, ghost-cli 1.24.2
- https://github.com/docker-library/ghost/commit/e7e76fa: Bump node to 18 since 16 is EOL later this month

### DIFF
--- a/library/ghost
+++ b/library/ghost
@@ -5,12 +5,12 @@ Maintainers: Tianon Gravi <admwiggin@gmail.com> (@tianon),
              Austin Burdine <austin@acburdine.me> (@acburdine)
 GitRepo: https://github.com/docker-library/ghost.git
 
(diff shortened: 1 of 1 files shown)


yosifkit:
Just to be clear, we need to bump the node version within the next couple weeks. I chose to include it in this minor bump so that users can revert to the previous minor if it breaks their setup. If we didn't include it at a specific patch, a user might re-pull the same tag and unknowingly get the new node version with no easy way to revert (without knowing the full sha256 image digest).

PR #8100, opened 2020-06-01
sonar-scanner-cli 4.5

michal-duda-sonarsource:
SonarScanner CLI is the official scanner used to run code analysis on SonarQube and SonarCloud. 

Please don't merge this PR yet even if everything is ok- we still have to release version 4.4 of the scanner, so we will have to update the version of the binary that is downloaded in the image.

Docs PR: https://github.com/docker-library/docs/pull/1728

PR #15326, opened 2023-09-07
[Draft DO-NOT-MERGE] Add sonar-scanner-cli as official docker image

jCOTINEAU:
Hello everyone, we would like to add a new official image for the tool called sonar-scanner-cli.

This is the tool used to interact with SonarQube product (which is already an official image).

Here is the related [doc PR](https://github.com/docker-library/docs/pull/2366)

Those PR will stay in draft while gathering your feedback on the [ongoing Dockerfile here](https://github.com/SonarSource/sonar-scanner-cli-docker/blob/master/5/Dockerfile)

Thanks a lot in advance

### DIFF
--- a/library/sonar-scanner-cli
+++ b/library/sonar-scanner-cli
@@ -1,0 +1,10 @@
+Maintainers: Carmine Vassallo <carmine.vassallo@sonarsource.com> (@carminevassallo),
+             Jeremy Cotineau <jeremy.cotineau@sonarsource.com> (@jCOTINEAU),
+             Davi Koscianski-vidal <davi.koscianski-vidal@sonarsource.com> (@davividal)
+GitRepo: https://github.com/SonarSource/sonar-scanner-cli-docker
+Architectures: amd64, arm64v8
+GitCommit: dc6c344d32f09e905665d2463e3a0de9c7f8c948
+
+Tags: 5.0.1.3006, 5.0.1, 5.0, 5, latest
+Architectures: amd64, arm64v8
+Directory: 5


yosifkit:
Has something changed since https://github.com/docker-library/official-images/pull/8100? 

Rewording/expanding [my comment](https://github.com/docker-library/official-images/pull/6843#issuecomment-554158766) linked from `#8100`:
My initial reaction is that it doesn't sound like the image would be "generally useful". My reasoning for that is that this image is tied to manage a specific service. It is not a generic service that you would plug into another system.  All the other images here are a self-contained service, language stack, or base OS. This seems more like the WP-CLI images that are just a tag under the `wordpress` images (`wordpress:cli`) or just a tool that's included in the server image like `mysqladmin`.

jCOTINEAU:
Hello @yosifkit, please may you excuse me, i was not aware this was already attempted that way at sonar.

I definitely understand the reasoning, so just to double check, f we specify it as a subtag of sonarqube, this will be accepted and considered a valid use case ?

Thanks a lot for your kind answers, I will draft this.

jCOTINEAU:
Hello again @yosifkit, brainstormed a bit on that with the team and we have few other questions.

One major change compared to #8100 is that the scanner is used both with SonarQube and SonarCloud (our SaaS solution)

Having the scanner as a subpart of the SonarQube image would be confusing to users who are looking to use it against SonarCloud.

The situation is a bit tricky as you mentioned 

> All the other images here are a self-contained service, language stack, or base OS

For SonarQube itself it is not entirely true, as the product does need this scanner-cli in order to be used.

Do we have other options to look for? We were thinking maybe having a Sonar image instead of SonarQube? 

With something like

- Sonar:sonarqube
- Sonar:scanner-cli 

But I guess it would be too tight to the company? even if those are opensource products.

Also lso the refactor might be tricky as it would be considered a new image/docker-hub page etc ? 

Please let us know what you think about the situation.

Huge thanks in advance.

whalelines:
What is the motivation for having this tool as a Docker Official Image (DOI)? Making the [sonarsource organization](https://hub.docker.com/u/sonarsource) a [Docker Verified Publisher](https://docs.docker.com/trusted-content/dvp-program/) might be a better fit.

jCOTINEAU:
Hello @whalelines, two main reason.

First was to make sure all our images are distributed the same way, so it is clear and easy to understand where to pick them for our users.
Second would be to benefits from the auto-rebuilt feature of the official image and all the positive "marketing" that comes with it.

In the meantime, being a verified publisher is a discussion we are having internally.

PR #15331, opened 2023-09-08
Removing Neo4j 4.1.13 entries

glindroth:
As promised earlier to avoid unnecessary updates


yosifkit:
👍 images are up on Docker Hub: https://hub.docker.com/_/neo4j/tags?page=1&name=4.1.13

PR #15332, opened 2023-09-08
Update maintainers and update alpine image to 3.18

achimklittich:


yosifkit:
While this will update the maintainer, this will not change the images.  Images only rebuild with a change to the context (the Dockerfile and scripts) or if the base image is updated.  I think the simplest fix for the `teamspeak` image is to move to a more recent version of Alpine (like 3.18). Alpine 3.13 has been EOL since November 2022 (https://alpinelinux.org/releases/, https://github.com/docker-library/official-images/pull/13577).

achimklittich:
Thank you for the suggestion with the alpine image. We have now updated the alpine image and pushed it into the respective branch.

LaurentGoderre (approved):

yosifkit (approved):

PR #15351, opened 2023-09-11
Update JRuby 9.3 to 9.3.11.0

headius:


headius:
Oops, forgot sha update at first.

PR #15362, opened 2023-09-14
Odoo: Build for arm64 too

samip5:
I believe this should be just a matter of including them here.

### DIFF
--- a/library/odoo
+++ b/library/odoo
@@ -3,14 +3,14 @@ GitRepo: https://github.com/odoo/docker
 GitCommit: 61148a86eed7fb5452dc7705b479988e1f49f9a6
 
 Tags: 16.0, 16, latest
-Architectures: amd64
+Architectures: amd64,arm64
 Directory: 16.0
 
 Tags: 15.0, 15
-Architectures: amd64
+Architectures: amd64,arm64
 Directory: 15.0
 
 Tags: 14.0, 14
-Architectures: amd64
+Architectures: amd64,arm64
 Directory: 14.0
 
--- a/library/odoo
+++ b/library/odoo
@@ -3,14 +3,14 @@ GitRepo: https://github.com/odoo/docker
 GitCommit: 61148a86eed7fb5452dc7705b479988e1f49f9a6
 
 Tags: 16.0, 16, latest
-Architectures: amd64,arm64
+Architectures: amd64,arm64v8
 Directory: 16.0
 
 Tags: 15.0, 15
-Architectures: amd64,arm64
+Architectures: amd64,arm64v8
 Directory: 15.0
 
 Tags: 14.0, 14
-Architectures: amd64,arm64
+Architectures: amd64,arm64v8
 Directory: 14.0
 


LaurentGoderre (approved):

derfl0:
I don't want to break your enthusiasm @samip5 but running an amd64 packet on arm64 will not work:
https://github.com/odoo/docker/blob/master/16.0/Dockerfile#L36

samip5:
> I don't want to break your enthusiasm @samip5 but running an amd64 packet on arm64 will not work: https://github.com/odoo/docker/blob/master/16.0/Dockerfile#L36

Yeah, just noticed that and I also noticed that https://github.com/odoo/docker/pull/410#issuecomment-1433025790 comment which means @d-fence might not want to use the version that's available for arm64 :(

yosifkit:
Without the Odoo Dockerfiles being updated to support building on multiple architectures, this PR will not work.  As for https://github.com/odoo/docker/pull/464, that is only part of a fix since it would then fail to build on `amd64`. It is up to the Odoo image maintainer, @d-fence, but it would likely need something like this section in the `golang` Dockerfile that uses `dpkg` to get the userspace architecture and then chooses the architecture-specific item to download:

https://github.com/docker-library/golang/blob/18778673d05af968094fe71d45dfd7fbf1856b65/1.21/bookworm/Dockerfile#L26-L79

---

A simple way to test architectures that you do not have access to is to use Docker Desktop (and it sets up qemu): `docker build --platform=linux/arm64 16.0/`. Alternatively, installing `binfmt-support` and `qemu-user-static` in a Debian host OS running Docker and you can do the same `docker build --platform`.  We have seen issues when using qemu user static to emulate other architectures, so while it isn't a full replacement for actual hardware 😢, it is still helpful if you don't have them.

ludovic-gasc:
The support of ARM for Odoo is now available if you want to test: https://hub.docker.com/search?q=odoo&image_filter=official

rfc2119:
@ludovic-gasc Any idea how are the images built for arm64 ? I don't think the current Dockerfile is multiarch

tianon:
The current `Dockerfile` _is_ multiarch (and built for multiarch):

```json
$ crane manifest odoo | jq '.manifests[].platform | select(.os != "unknown")'
{
  "architecture": "amd64",
  "os": "linux"
}
{
  "architecture": "arm64",
  "os": "linux",
  "variant": "v8"
}
{
  "architecture": "ppc64le",
  "os": "linux"
}
```

rfc2119:
Yes, but I don't see how this line will fit for both `arm64` and `amd64` packages from the [nightly archive](http://nightly.odoo.com/17.0/nightly/deb/)
https://github.com/odoo/docker/blob/b6411dfc5391faaa18a0e12951adcbc3469c2a30/17.0/Dockerfile#L76

samip5:
> Yes, but I don't see how this line will fit for both `arm64` and `amd64` packages from the [nightly archive](http://nightly.odoo.com/17.0/nightly/deb/) https://github.com/odoo/docker/blob/b6411dfc5391faaa18a0e12951adcbc3469c2a30/17.0/Dockerfile#L76

It's the way the Docker's build system uses targetarch ARG, and builds it according to that, so you'll not see arch specific instructions in most cases.

rfc2119:
> That deb package also is not arch specific as the thing packaged is a Python application.

TIL. Interesting.

PR #15369, opened 2023-09-14
Add Couchbase Server 7.2.1 (EE and CE)

ceejatec:


ceejatec:
@yosifkit Continuing the discussion started on #15134 - It's not clear to me what the right answer is for us regarding Couchbase Server 6.6.0, which we do still definitely support although we don't provide binaries for distros newer than Ubuntu 18.04. Is it really necessary that we pull it from the list of Supported Versions on Docker Hub?

yosifkit (approved):

yosifkit:
Oh, I think I understand our disconnect a little now. Our definition/interpretation of "supported" is different. For official images we think that it means that the image will receive updates for the software itself as well as the underlying container OS. But from https://www.couchbase.com/support-policy/enterprise-software/, it seems that for Couchbase versions under "Limited Support", it means that you'll provide limited help in using it but not any updates.

It seems generally at odds to keep items "active" in Official Images that aren't going to receive updates since one of our main tenets is to ["Actively rebuild for updates and security fixes"](https://github.com/docker-library-bot/official-images/tree/b1496084f0077e2eaaab273d585d3bc4ea249d57#what-are-official-images). The image will be available to pull from Docker Hub regardless of its status. For the future, perhaps we can find a way to customize the (Docker Hub) docs page so that Couchbase and others can have a different set of supported tags than is in the respective `library/` file or maybe we allow renaming the "Supported tags and respective `Dockerfile` links" section to something else ("updated tags", "active tags", ?). If you have any suggestions, we'd like to make something that will work for you.

Seeing that the planned EOL for 6.6 is just October 2023, it's probably fine to stay until then.

PR #15380, opened 2023-09-18
emqx 5.2.1

id:


yosifkit (dismissed):

yosifkit:
https://github.com/emqx/emqx-docker/blob/6e4c675c3929161db508ca56eec5576ce231170a/5.2/Dockerfile#L33C40-L33C40

Removing `curl` in a later `RUN` line than it was installed means that, even though inaccessible in the final image, it will still take up space in the image.

id:
@yosifkit I understand. The objective here was to silence vulnerability scanners.

yosifkit:
For official images we rely on OS security teams to update OS packages in order to silence vulnerability scanners. If the package truly is not needed, it should be installed and removed in the same `RUN`. I have opened a PR on the `emqx/emqx-docker` repo.

---

Assuming it is one of the curl CVEs listed by some scanners, it seems that none of them really apply to Debian `curl` packages (and scanners ought to mark these as already "fixed" in their results):

- https://security-tracker.debian.org/tracker/CVE-2023-28320: `Debian binary package builds are not affected`
- https://security-tracker.debian.org/tracker/CVE-2021-22923: `Metalink support not enabled in Debian builds.`
- https://security-tracker.debian.org/tracker/CVE-2021-22922: `Metalink support not enabled in Debian builds.`

id:
@yosifkit user reported this one https://security-tracker.debian.org/tracker/CVE-2023-23914

yosifkit (approved):

PR #15382, opened 2023-09-18
[oraclelinux] Updating 8, 8-slim ,8-slim-fips for correct images.

amsteinb-orcl:
oraclelinux:8 aarch64 image was incorrect.  This will correct the issue.


amsteinb-orcl:
Need to rework changes.

PR #15387, opened 2023-09-18
Update php

yosifkit:
Changes:

- https://github.com/docker-library/php/commit/09e2bdf8: Update 8.1-rc to 8.1.24RC1


LaurentGoderre:
Why is it a new section here instead of replacing `8.1.23`?

yosifkit:
> Why is it a new section here instead of replacing `8.1.23`?

This is just a release candidate. We want the regular release to still be active until `8.1.24` is no longer an RC.

PR #15399, opened 2023-09-20
[varnish] fix uid:gid in 7.4

gquintard:
I forgot that 7.3 introduced predictable uid:gid, and forgot to move a few things around, leading 7.3 to miss some setup code. This fixes it


gquintard:
ah, 7.4.1 got released during the night, so let push that in too

yosifkit (approved):

PR #15405, opened 2023-09-20
SapMachine: Add missing JRE images

RealCLanger:
This adds the missing JRE images. Hopefully tests are all green now.

### DIFF
--- a/library/sapmachine
+++ b/library/sapmachine
@@ -1,6 +1,16 @@
 Maintainers: Christoph Langer <sapmachine@sap.com> (@realclanger), Christian Halstrick <sapmachine@sap.com> (@chalstrick)
 GitRepo: https://github.com/SAP/SapMachine-infrastructure.git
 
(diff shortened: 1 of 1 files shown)


RealCLanger:
Hm, tests still don't work - need to check what happens...

yosifkit:
The Official Images build queue is still very full (https://doi-janky.infosiftr.net/) since all Debian images have been working though rebuilds since https://github.com/docker-library/official-images/pull/15395 was merged. So, it may be a little while longer before the `sapmachine:*-jdk` images are available. I'm happy to merge with failing builds, or we can wait for the jdk image builds (https://doi-janky.infosiftr.net/job/multiarch/view/images/view/sapmachine/).

RealCLanger:
The required base images should be available now. Let's see if tests will be green now.

RealCLanger:
Looks good now. :)

PR #15410, opened 2023-09-21
Update tomcat

yosifkit:
Changes:

- https://github.com/docker-library/tomcat/commit/a9121f8: Update 10.1
- https://github.com/docker-library/tomcat/commit/588b898: Update 10.1


yosifkit:
Looks like a content flip flop. No actual change.

PR #15423, opened 2023-09-22
Switch to Docker Scout sbom indexer

yosifkit:
This updates to the Docker Scout SBOM indexer (`docker/scout-sbom-indexer`) to provide better/more complete SBOMs. The generated SBOMs for the `docker` images are so large and different that I don't have a good way to compare them.

~There aren't any release tags on https://hub.docker.com/r/docker/scout-sbom-indexer/tags yet, so this just uses the newest one with the most architectures (as of today). It will unfortunately not get automatically updated since the tag seems to be a commit hash or similar.~ There are some releases now: https://hub.docker.com/r/docker/scout-sbom-indexer/tags. ~I've moved to the `latest` tag because I don't see a tag to follow the major release.~ Switched to `1` release tags to follow the major release.

---

Should we also remove the `docker/buildkit-syft-scanner` pin? edit: removed


LaurentGoderre (approved):

whalelines (commented):
Looks good, probably no reason to retain the pointer to the syft sbom generator.

.bin/bashbrew-buildkit-env-setup.sh
```diff
@@ -62,7 +62,9 @@ _bashbrew_buildkit_env_setup() {
 			vars="$(_jq_setenv <<<"$vars" BUILDX_BUILDER "$buildxBuilder")"
 
 			local sbomTag
-			sbomTag="$(grep <<<"$externalPins" -m1 '^docker/buildkit-syft-scanner:')"
+			# https://hub.docker.com/r/docker/scout-sbom-indexer/tags
+			# TODO: update to a pinned major version tag when available
```
whalelines, line 66: This comment is no longer relevant, correct?

whalelines (commented):
LGTM, just need to remove an outdated comment. Thanks!

.bin/bashbrew-buildkit-env-setup.sh
whalelines: After talking with @yosifkit , there is a preference to follow a "stable" tag, e.g., `1` rather than `latest`.

whalelines (approved):
LGTM

LaurentGoderre (approved):

LaurentGoderre (approved):

tianon:
@cdupuis I think a few words about what makes this different, why Docker has two, why this one is better, etc would be really useful/important here 😅

Maybe they would make a good description on https://hub.docker.com/r/docker/scout-sbom-indexer too?

whalelines:
@cdupuis and I came up with this:

The differences between Scout- and syft-generated SBOMs largely focus on features Scout leverages to present vulnerability information to the user:

#   The Scout SBOM scanner puts more effort into mapping packages with layers so when Scout is presenting the layer/vulnerability mapping it is more accurate. In other words, both syft and Scout will produce similar package lists, but Scout is better able to map the packages, and hence vulnerabilities, to specific layers.
#   The Scout SBOM scanner incorporates Go toolchain information into a layer when it finds a Go binary. While this can result in some false positives, we use it as the default as it allows more downstream flexibility.
#   The Scout SBOM scanner combines approaches used by both syft's and trivy's SBOM scanners, resulting in slightly richer metadata than either single approach provides.

tianon:
Following up again, can we copy that description into Docker Hub as the image description on https://hub.docker.com/r/docker/scout-sbom-indexer (or something like it)?

This is probably a question for @cdupuis and the scout team / indexer image maintainers more than our team :see_no_evil:

whalelines:
Done: https://hub.docker.com/r/docker/scout-sbom-indexer

PR #15433, opened 2023-09-26
CrateDB new hotfix `5.2.9 ` release for older minor version.

matriv:


whalelines:
Why does the [bashbrew diff](https://github.com/docker-library/official-images/pull/15433#issuecomment-1735383925) seem to indicate the latest tag is being replaced by the new 4.2 builds?

whalelines:
@matriv , it's not a question for you; more a question around the diff generation.

yosifkit:
> Why does the [bashbrew diff](https://github.com/docker-library/official-images/pull/15433#issuecomment-1735383925) seem to indicate the latest tag is being replaced by the new 4.2 builds?

The diff generator mostly ignores names and tries to find a close match so that a fresh image entry in the library file isn't just 100% new since it is likely very similar to other images already in the library file. This way we can review just what is different rather than a whole Dockerfile. There are times where it makes interesting comparison choices.

PR #15455, opened 2023-09-30
Add Andrei Dziahel as haskell maintainer

AlistairB:
cc @develop7

And remove myself. I don't really have time to assist anymore and lots of activity without me which is good.


yosifkit (approved):

yosifkit:
Thanks @AlistairB! ❤️ 

And welcome @develop7 🎉

PR #15468, opened 2023-10-03
Update to Fluentd v1.16.2

ashie:
Update to the latest stable version.


yosifkit (approved):

ashie:
Thanks!

tianon:
Looks like we're missing a few updates (1.16.5 being the latest in the 1.16 series and the newer 1.17.0) -- any chance of an image update?  How can we help with getting `library/fluentd` updated more often so that https://hub.docker.com/_/fluentd stays up-to-date? :eyes:

PR #15470, opened 2023-10-03
elixir 1.15.6

getong:


LaurentGoderre:
Is skipping from 4 to 6 (and skipping over 5 intentional?

getong:
> Is skipping from 4 to 6 (and skipping over 5 intentional?

No. I am busy with something, and cause something ignored by accident.

LaurentGoderre:
@getong I don't think it's you, it seems the upstream repo did the skipping

epinault:
@LaurentGoderre I did the upstream MR because nothing existed (neither .5 or .6 ) so I created the .6 cause I would like to get the latest . does it matter the .5 is not released?

yosifkit (approved):

PR #15473, opened 2023-10-04
Kong 2.8.4

curiositycasualty:


LaurentGoderre:
@yosifkit the alpine version looks good but the Ubuntu version doesn't seem to be mapping to the right version

```diff
diff --git a/kong_3.1-ubuntu/Dockerfile b/kong_2.8-ubuntu/Dockerfile
similarity index 68%
copy from kong_3.1-ubuntu/Dockerfile
copy to kong_2.8-ubuntu/Dockerfile
index bf4d572..6d3d2de 100644
--- a/kong_3.1-ubuntu/Dockerfile
+++ b/kong_2.8-ubuntu/Dockerfile
@@ -1,6 +1,4 @@
-FROM ubuntu:focal
-
-LABEL maintainer="Kong Docker Maintainers <docker@konghq.com> (@team-gateway-bot)"
+FROM ubuntu:jammy
 
 ARG ASSET=ce
 ENV ASSET $ASSET
@@ -9,32 +7,32 @@ ARG EE_PORTS
 
 COPY kong.deb /tmp/kong.deb
 
-ARG KONG_VERSION=3.1.1
+ARG KONG_VERSION=2.8.4
 ENV KONG_VERSION $KONG_VERSION
```

curiositycasualty:
@LaurentGoderre @yosifkit Is there something else that I need to do to get this over the finish line?

yosifkit:
> the Ubuntu version doesn't seem to be mapping to the right version

That's just because there wasn't a `2.8` version on ubuntu in the previous set of images, so the diff finds the "closest" match in order to minimize additions (rather than just the whole file as "new" when it is likely to be similar to other Dockerfiles).  It looks fine to me.

yosifkit (approved):

PR #15474, opened 2023-10-04
Update RabbitMQ with SBOM attestations

whalelines:
Provide attestations for software built within the Docker build context.

### DIFF
--- a/library/rabbitmq
+++ b/library/rabbitmq
@@ -7,7 +7,7 @@ Builder: buildkit
 
 Tags: 3.13.0-beta.6, 3.13-rc
 Architectures: amd64, arm32v7, arm64v8, ppc64le, s390x
-GitCommit: ced055dcc013310157788d8e5a710f231bb7cdba
+GitCommit: 6e5870068747ebef1d00134c52bd9a8e41cb2b4b
 Directory: 3.13-rc/ubuntu
 
 Tags: 3.13.0-beta.6-management, 3.13-rc-management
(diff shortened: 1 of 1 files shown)


LaurentGoderre (dismissed):

yosifkit (approved):
Accepting as experimental extra SBOM information. I'd love to have something that is less "manual". Ideally container SBOM scanners could automatically discover these non-os packages possibly through a combination of Dockerfile and container filesystem analysis.

library/rabbitmq
```diff
@@ -7,7 +7,7 @@ Builder: buildkit
 
 Tags: 3.13.0-beta.6, 3.13-rc
 Architectures: amd64, arm32v7, arm64v8, ppc64le, s390x
-GitCommit: ced055dcc013310157788d8e5a710f231bb7cdba
+GitCommit: b2387a802850aa877b8feefdd679e12536072db9
```
yosifkit, line 10: These should be pointing at https://github.com/docker-library/rabbitmq/commit/6e5870068747ebef1d00134c52bd9a8e41cb2b4b as that is what `generate-stackbrew-library.sh` will end up with.

yosifkit (changes requested):

yosifkit (approved):

yosifkit:
Repushed to match the `library/rabbitmq` file to the output of `generate-stackbrew-library.sh` from https://github.com/docker-library/rabbitmq/.

PR #15501, opened 2023-10-06
[nats] Release v2.10.2

bruth:
Details can be found [here](https://github.com/nats-io/nats-server/releases/tag/v2.10.2)


bruth:
@yosifkit Could we get a merge if all looks good? Thanks.

yosifkit (approved):

PR #15508, opened 2023-10-09
Update Debian (esp. for 12.2 and 11.8)

tianon:


yosifkit:
I normally would merge this as soon as possible to get rebuilds of dependent official images started, but I think we should wait until Wednesday. Since the base Debian images don't contain `curl`, we can wait to merge this until the [`curl` CVE fixes scheduled for Wednesday Oct 11](https://github.com/curl/curl/discussions/12026) are available in Debian packages. This way all of the dependent Official-Images get the update when they are rebuilt. Otherwise, we'd need to rebuild again on Wednesday and we don't want that much unnecessary image churn for both Docker Hub and the image users.

yosifkit (approved):

yosifkit:
Curl fixes are available; merging to get dependent image rebuilds started.

https://security-tracker.debian.org/tracker/CVE-2023-38545
https://security-tracker.debian.org/tracker/CVE-2023-38546

PR #15510, opened 2023-10-10
update rocket.chat

sampaiodiego:


library/rocket.chat
```diff
@@ -1,10 +1,14 @@
...
-GitCommit: 42a6ffd7ceb1064ab4baf38e5733d41faab0519d
+Tags: 6.4.1, 6.4, 6, latest
+GitCommit: b73be519a9f4b22e58905885790c17dbcca19bc7
+Directory: 6.4
+
+Tags: 6.3.10, 6.3
+GitCommit: b73be519a9f4b22e58905885790c17dbcca19bc7
 Directory: 6.3
 
 Tags: 6.2.12, 6.2
```
yosifkit, line 14: Any particular reason to keep these older versions (`5.4` and `6.2`) listed here? They seem to have reached end of life (https://docs.rocket.chat/resources/rocket.chats-support-structure/enterprise-support-and-version-durability.

yosifkit (approved):
I'd suggest dropping the end of life versions in this or a follow up PR.

library/rocket.chat
yosifkit: Apologies, I think I was waiting for an update to this PR.  Should I merge this PR to get the new version in or did you want to update it to drop the EOL version?

sampaiodiego: my bad @yosifkit .. I've updated it now removing 5.4 .. thx

yosifkit: No worries. I was making sure that you weren't waiting on me to merge in order to do it in a separate PR (either was fine by me).

yosifkit (approved):

PR #15536, opened 2023-10-13
[nats] Releases v2.10.3 and v2.9.23

wallyqs:
Details can be found here:

https://github.com/nats-io/nats-server/releases/tag/v2.10.3 
https://github.com/nats-io/nats-server/releases/tag/v2.9.23

### DIFF
--- a/library/nats
+++ b/library/nats
@@ -4,47 +4,47 @@ Maintainers: Derek Collison <derek@synadia.com> (@derekcollison),
              Neil Twigg <neil@synadia.com> (@neilalexander),
              Phil Pennock <pdp@synadia.com> (@philpennock)
 GitRepo: https://github.com/nats-io/nats-docker.git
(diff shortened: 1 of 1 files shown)


wallyqs:
@yosifkit would be great to get a merge on this one if all looks good, thanks!

derekcollison:
This is for a Golang CVE that was released that could affect our users and customers. Thanks.

library/nats
```diff
@@ -28,23 +28,23 @@ Architectures: windows-amd64
 Directory: 2.10.x/nanoserver-1809
```
yosifkit, line 28: This `2.10*nanoserver-1809` entry should also be getting the `2.10.3` bump in its tags.

wallyqs: Thanks @yosifkit, just updated.

wallyqs:
Thanks @yosifkit !

PR #15544, opened 2023-10-14
Upgrade Zookeeper to 3.7.2, 3.8.3 and 3.9.1

31z4:
* I removed the last `temurin` tag according to the plan described here: https://github.com/docker-library/official-images/pull/12949#issuecomment-1383656252
* I introduced `jre` tags to help users migrate to JRE 17. JRE 11 is EoL so `3.7.2`, `3.8.3` and `3.9.1` will be the last tags to support it.

### DIFF
--- a/library/zookeeper
+++ b/library/zookeeper
@@ -2,14 +2,14 @@ Maintainers: Elisey Zanko <elisey.zanko@gmail.com> (@31z4)
 GitRepo: https://github.com/31z4/zookeeper-docker.git
 Architectures: amd64, arm64v8, s390x, ppc64le
 
-Tags: 3.7.1-temurin, 3.7-temurin
-GitCommit: 5cf119d9c5d61024fdba66f7be707413513a8b0d
-Directory: 3.7.1
+Tags: 3.7.2-temurin, 3.7-temurin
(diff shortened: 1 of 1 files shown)


yosifkit (approved):

library/zookeeper
```diff
@@ -2,14 +2,26 @@ Maintainers: Elisey Zanko <elisey.zanko@gmail.com> (@31z4)
...
+
+Tags: 3.7.2-jre-17, 3.7-jre-17
+GitCommit: 5076660820c73f3b119cbdd1267c25a1e29cbbf4
+Directory: 3.7.2
+
+Tags: 3.8.3-jre-17, 3.8-jre-17
+GitCommit: 5076660820c73f3b119cbdd1267c25a1e29cbbf4
+Directory: 3.8.3
+
+Tags: 3.9.1-jre-17, 3.9-jre-17, latest
```
yosifkit, line 25: I noticed that `latest` moved to jre17, but not the plain tags like `3.9`, `3.9.1` or `3.8.3`. Did you want to move those also?

31z4: No I wanted to move plain tag to jet17 with the next version upgrade.

PR #15553, opened 2023-10-16
Updated swipl devel to 9.1.17

JanWielemaker:


yosifkit:
There doesn't seem to be a difference in the Dockerfile.  Ah, it still has the old version number: https://github.com/SWI-Prolog/docker-swipl/blob/201cb3bb2b0c9440573eda2bcdbdc739e2061a4e/9.1.17/bullseye/Dockerfile#L31

PR #15568, opened 2023-10-18
Update notary with attestation

LaurentGoderre:


whalelines:
The Test PR GHA failure is complaining about not being able to find `golang:1.19-alpine3.16`, but that image exists, https://hub.docker.com/_/golang/tags?page=1&name=1.19-alpine3.16.

Who can restart the GHA to retry?

LaurentGoderre:
@whalelines I think this is a problem of the tag existing but not being updated anymore (not supported).I do feel like upgrading to a supported version of Go goes beyond the scope of this change

tianon:
Is this `notary:builder` image intended for end-user consumption?  What's the use case?

LaurentGoderre:
@tianon it can be but this is mostly to prevent duplicate builds of the same thing

LaurentGoderre:
I could rework the upstream to use one Dockerfile with conditions with templating

tianon:
Yeah, solving the "maintaining the same thing twice" problem sounds like a better fit for templating :bow: :heart:

whalelines (approved):
LGTM

Should we wait to merge this until the signing code has been updated?

yosifkit:
How can we get the `notary` images to use more up to date base images than `golang:1.19-alpine3.16` and `alpine:3.16`? Both of them are past end of life in their respective upstream releases: Go 1.19 on 06 Sep 2023 and Alpine 3.16 on 23 May 2024. The `golang:1.19-alpine3.16` image was dropped from official images on May 11, 2023 (https://github.com/docker-library/official-images/pull/14638 / https://github.com/docker-library/golang/pull/460)

tianon:
I've opened https://github.com/docker/notary-official-images/pull/38, but it might be worth considering how we can stay on top of it better (perhaps we need to get more folks with merge access/maintainership on that packaging repository so it doesn't always end up on Jonny's plate?)

tianon:
(we should also update `Maintainers:` here, because that's surely not accurate)

yosifkit:
The notary docker images upstream was archived on Jun 4, 2026 (https://github.com/docker/notary-official-images) and the [upstream project](https://github.com/notaryproject/notary) was archived on Jul 30, 2025, so this can be closed.

PR #15590, opened 2023-10-20
Fix python tests in 2.7 by converting pip option to an env var

LaurentGoderre:


test/tests/run-sh-in-container.sh
```diff
@@ -9,11 +9,19 @@ runDir="$(dirname "$runDir")"
...
 		;;
 
 	*)
 		[ -f "$testDir/container.sh" ]
-		source "$runDir/run-in-container.sh" "$testDir" "$1" sh ./container.sh
+		source "$runDir/run-in-container.sh" \
+			--docker-arg --env=PIP_DISABLE_PIP_VERSION_CHECK=1 \
+			--docker-arg --env=PIP_NO_PYTHON_VERSION_WARNING=1 \
+			--docker-arg --env=PIP_ROOT_USER_ACTION=ignore \
+			"$testDir" "$1" sh ./container.sh
```
yosifkit, line 25: `python-hy` is the only `run-sh-in-contianer` test that does a `pip install`, we should leave the env vars in that specific test rather than this general helper since there are other tests using `run-sh-in-contianer` that the vars don't apply to.

LaurentGoderre: I realize now it's a symlink!! I thought I was modifying just the python version

PR #15606, opened 2023-10-24
Odoo: update 14.0-16.0 to release 20231024

d-fence:
Hello,

Here are the latest updates of Odoo supported versions.
Also, this adds support of ppc64le and arm64 for  Odoo 16.0.

Thanks

### DIFF
--- a/library/odoo
+++ b/library/odoo
@@ -1,9 +1,9 @@
 Maintainers: Christophe Monniez <moc@odoo.com> (@d-fence)
 GitRepo: https://github.com/odoo/docker
-GitCommit: d04d885eea6b8f60ada727fcc498f5a8fd44da51
+GitCommit: 3864188c94c33dace7d4b5f6767e00a4328fe08c
 
 Tags: 16.0, 16, latest
-Architectures: amd64
+Architectures: amd64, arm64v8, ppc64le
 Directory: 16.0
 
 Tags: 15.0, 15


samip5:
I think the architecture fields will also need to be modified to indicate the support for ARM64 and ppc64le.

library/odoo
```diff
@@ -1,6 +1,6 @@
 Maintainers: Christophe Monniez <moc@odoo.com> (@d-fence)
 GitRepo: https://github.com/odoo/docker
-GitCommit: d04d885eea6b8f60ada727fcc498f5a8fd44da51
+GitCommit: 8a93c61c028c306c32610cdef4cfcb606edd4ff0
 
 Tags: 16.0, 16, latest
 Architectures: amd64
```
yosifkit, line 6: We don't yet use `ARG TARGETARCH` in official-images. We recommend using `apk --print-arch` or `dpkg --print-architecture` as the most reliable options to retrieve the userspace architecture of the image.

You'll also need to expand this `Architectures:` value in the library file for the image tags that support building on other architectures:

```diff
-Architectures: amd64
+Architectures: amd64, arm64v8, ppc64le
```

d-fence: Hello. I changed to use `dpkg --print-architecture` when TARGETARCH is not set.
I think that it should work in both cases.

LaurentGoderre (dismissed):

library/odoo
yosifkit: `TARGETARCH` isn't interchangeable with `dpkg --print-architecture`. The `dpkg` values convey more information when run on some architectures (32bit arm architectures for example).

d-fence: I agree but I think that it should work as we only want to set `TARGETARCH` when it's not already set and we only build for `amd64, arm64v8, ppc64le`. I tested with buildx and it worked as expected.

d-fence:
Hello.
@yosifkit Do I need to change anything ?

ludovic-gasc:
hi @d-fence and @yosifkit ,

Just in case, I have SSH access on the amd64 and ppc64le servers, and I have free time to be able to test this Dockerfile build, if necessary.
Don't hesitate to contact me if I can help.

Have a nice weekend!

yosifkit:
Ah, `TARGETARCH` will be different than `dpkg --print-architecture` on ppc64le too since it is the `goarch` value of `ppc64le`, but `dpkg` will be `ppc64el`.

ludovic-gasc:
Hi @yosifkit ,

Good catch, I confirm your analysis on an IBM Power server:

```
# dpkg --print-architecture
ppc64el
```

If I understand correctly, it seems simply to replace `"ppc64le")` by `"ppc64le"|"ppc64el")` in the Dockerfile.
@d-fence do you want a pull request for this change?

FYI, I have just followed this documentation where `TARGETARCH` is defined: https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/

Maybe not a good idea to mix at the same time `TARGETARCH` and `dpkg --print-architecture` ?
Do we have a risk of a side effect?

d-fence:
@yosifkit I changed to support both cases and while at it, I updated to the latest Odoo releases of 2023106

library/odoo
```diff
@@ -1,9 +1,9 @@
 Maintainers: Christophe Monniez <moc@odoo.com> (@d-fence)
 GitRepo: https://github.com/odoo/docker
-GitCommit: d04d885eea6b8f60ada727fcc498f5a8fd44da51
+GitCommit: 9b5e2cf27bcc23da4096c40b7369d0d43a4d90c6
```
yosifkit, line 3: The Dockerfile diff looks the same. 😕 

```diff
+    "ppc64le") WKHTMLTOPDF_ARCH=ppc64el && WKHTMLTOPDF_SHA=7ed8f6dcedf5345a3dd4eeb58dc89704d862f9cd  ;; \
```

Ah, I think the `GitCommit` needs to be updated to https://github.com/odoo/docker/commit/3864188c94c33dace7d4b5f6767e00a4328fe08c?

d-fence: 🤦🏻 omg, you are right.
Changed now.

yosifkit (approved):

ludovic-gasc:
Hi @LaurentGoderre ,

I permit to ping you because you already approve this pull request in the past, and moreover, tomorrow and during 3 days, Odoo is organizing an awesome event in Brussels, Belgium, the Odoo experience: https://www.odoo.com/event/odoo-experience-2023-3735/page/oxp23-introduction

It would be awesome in case of you have time this week to validate this pull request, to be able to say during the event that the ARM and IBM Power Odoo Docker images are officially available as official Docker images.

Thanks a lot!

ludovic-gasc:
hi @yosifkit , thank you a lot for the merge, it works pretty well on RedHat OpenShift and IBM Power.

Have a nice week

PR #15639, opened 2023-10-31
Update Debian

tianon:


whalelines:
This update does not appear to address the critical zlib vulnerability, https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1054290 , whose investigation appears to be ongoing. Should we wait for that to get sorted or proceed with this update?

yosifkit:
> This update does not appear to address the critical zlib vulnerability, https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1054290 , whose investigation appears to be ongoing. Should we wait for that to get sorted or proceed with this update?

Given that it's still being looked at and that the timeline is unknown, I think we can merge without waiting for a resolution there.

yosifkit (approved):

PR #15656, opened 2023-11-02
Releasing Neo4j 4.4.27

jennyowen:
Hi there, here's the newest Neo4j 4.4 release.

For our next release, I've been asked to replace `gosu` with `su-exec`. However, since su-exec does not provide a binary or a signature, I will have to update the Dockerfile to do a multi-part build. Is that allowed in official Dockerfiles?

The change would be to add this (or similar) to the top of the current Dockerfiles:

```
FROM debian:bullseye-slim as suexec-builder
# get su-exec source and checkout a known good version (current HEAD). Verify files contents are expected.
RUN apt update -q \
    && apt install -y gcc git make \
    && git clone https://github.com/ncopa/su-exec.git \
    && cd su-exec \
    && git checkout 4c3bb42b093f14da70d8ab924b487ccfbb1397af \
    && echo d6c40440609a23483f12eb6295b5191e94baf08298a856bab6e15b10c3b82891 su-exec.c | sha256sum -c \
    && echo 2a87af245eb125aca9305a0b1025525ac80825590800f047419dc57bba36b334 Makefile | sha256sum -c \
    && make \


FROM debian:bullseye-slim
COPY --from=suexec-builder /su-exec/su-exec /usr/bin/su-exec
# the rest of the existing dockerfile
```

I've done my best to make the builder stage repeatable and secure.
Are these proposed change likely to get rejected?

Thanks for all your help :slightly_smiling_face:

### DIFF
--- a/library/neo4j
+++ b/library/neo4j
@@ -25,12 +25,12 @@ Directory: 5.13.0/ubi8/enterprise
 
 
 
-Tags: 4.4.26, 4.4.26-community, 4.4, 4.4-community
+Tags: 4.4.27, 4.4.27-community, 4.4, 4.4-community
 Architectures: amd64, arm64v8
-GitCommit: 06d08eefe166a90662ea228cfbddce3438bd2732
-Directory: 4.4.26/bullseye/community
+GitCommit: 6c15231abd6294f0a736b4a46f59d258dea7f672
+Directory: 4.4.27/bullseye/community
 
-Tags: 4.4.26-enterprise, 4.4-enterprise
+Tags: 4.4.27-enterprise, 4.4-enterprise
 Architectures: amd64, arm64v8
-GitCommit: 06d08eefe166a90662ea228cfbddce3438bd2732
-Directory: 4.4.26/bullseye/enterprise
+GitCommit: 6c15231abd6294f0a736b4a46f59d258dea7f672
+Directory: 4.4.27/bullseye/enterprise


whalelines:
@jennyowen wrote:
> I will have to update the Dockerfile to do a multi-part build. Is that allowed in official Dockerfiles?

Yes, multi-stage builds are allowed. See, for example, https://github.com/docker-library/rabbitmq/blob/master/3.12/ubuntu/Dockerfile .

yosifkit:
>I will have to update the Dockerfile to do a multi-part build. Is that allowed in official Dockerfiles?

I'd say generally yes, but I usually recommend that if it can be done without multi-stage it is probably better/clearer. Multi-stage makes `docker history` much less friendly for users inspecting a pulled image.

In this particular case, I'd deeply question why there is a need to replace `gosu` with `su-exec`. While they are mostly compatible, `su-exec` has been very slow at making a release with some of those edge cases (likely that is why you are grabbing the current HEAD commit). And if perceived "vulnerabilities" are the issue, then I think the [newest release of gosu](https://github.com/tianon/gosu/releases/tag/1.17) might quell some of those and I'd also point to https://github.com/tianon/gosu/blob/master/SECURITY.md that explains that most/all of the current CVEs shown by tools are in code unreachable from `gosu`. (Tianon is even working on providing a VEX file to programmatically explain to vuln tools which CVEs don't apply)

yosifkit (approved):

jennyowen:
Thank you for the detailed reply @yosifkit.
Yeah I would also rather not include `su-exec` in the image, it doesn't seem to get much maintenance.

The request to change comes from our support team who have to deal with frequent questions about why docker scans are showing vulnerability XYZ from `gosu`. Tianon's security advisory page is very helpful in that regard, but there are still users (particularly banks) who are very wary of _any_ vulnerabilities existing. If he can get some automated CVE mitigation in place that would be **great** and I could switch it back.

PR #15668, opened 2023-11-06
add dotted environment variables test

gdams:
This should help us catch bugs like https://github.com/adoptium/containers/issues/415 in the future.

This could be made an OpenJDK-specific test but I figured there is likely no harm running it for all images

### DIFF
--- a/test/tests/run-java-in-container.sh
+++ b/test/tests/run-java-in-container.sh
@@ -52,7 +52,7 @@ WORKDIR /container
 EOD
 
 # if testDir contains dotted-environment-variables we need to add an environment variable to the docker run command
-if [ $testDir == *"dotted-environment-variables"* ]; then
+if [[ "$testDir" == *"dotted-environment-variables"* ]]; then
   docker run --rm -e "variable.with.a.dot=a.dotted.value" "$newImage" java -cp . container
 else
   docker run --rm "$newImage" java -cp . container


LaurentGoderre (approved):

test/tests/run-java-in-container.sh
```diff
@@ -51,4 +51,9 @@ COPY --from=jdk /container /container
 WORKDIR /container
 EOD
 
-docker run --rm "$newImage" java -cp . container
+# if testDir contains dotted-environment-variables we need to add an environment variable to the docker run command
+if [ $testDir == *"dotted-environment-variables"* ]; then
+  docker run --rm -e "variable.with.a.dot=a.dotted.value" "$newImage" java -cp . container
+else
+  docker run --rm "$newImage" java -cp . container
+fi
```
tianon, line 59: This one-test-specific conditional feels a bit icky in an otherwise generic script :disappointed:

Our test framework wasn't really built with this kind of thing in mind, and it's really showing its age :sob:

whalelines: Seems like the double square brackets are needed:

```sh
$ bash --version
GNU bash, version 5.2.15(1)-release (aarch64-apple-darwin22.1.0)
Copyright (C) 2022 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
$ bash -c 'testDir=a-dotted-environment-variables-dir; if [ $testDir == *"dotted-environment-variables"* ]; then echo true; else echo false; fi;'
false
$ bash -c 'testDir=a-dotted-environment-variables-dir; if [[ $testDir == *"dotted-environment-variables"* ]]; then echo true; else echo false; fi;'
true
```

PR #15686, opened 2023-11-09
bump Kong to 3.5.0

hanshuebner:


library/kong
```diff
@@ -1,7 +1,13 @@
 Maintainers: Kong Docker Maintainers <team-gateway@konghq.com> (@team-gateway-bot)
 GitRepo: https://github.com/Kong/docker-kong.git
 
-Tags: 3.4.2-ubuntu, 3.4-ubuntu, 3.4.2, 3.4, 3, latest, ubuntu
+Tags: 3.5.0-ubuntu, 3.5-ubuntu, 3.5.0, 3.5, 3, latest, ubuntu
+GitCommit: e4ba2e351f3da34727fd016409a2669004b3fce0
+GitFetch: refs/tags/3.5.0
+Directory: ubuntu
+Architectures: amd64, arm64v8
+
```
yosifkit, line 9: I was just double-checking old versions and it looks like `3.0.x` has reached "end of full support", but not yet "end of sunset support".  Do you want the `3.0.x` images to stay in the library file and continue to be rebuilt for base OS updates or do you want to remove them?

Removing it here will remove it from the "Supported" section on the Hub readme, but the tags on Docker Hub will still be available to users who want it.  (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)

Just something to consider, we are fine if you'd like to keep them until end of sunset support.

yosifkit (approved):

library/kong
AndyZhang0707: Thanks @yosifkit for bringing this up. We will discuss this internally and update the thread later.

AndyZhang0707: Hi @yosifkit , we've discussed this internally and the conclusion is to remove 3.0.x from the "Supported" section while make it still available to users who want it as you suggested.

Thanks for your help and it's really appreciated.

tianon: Want to make a small follow-up PR doing that, or would you rather we open one? :eyes:

AndyZhang0707: I've created a PR to remove Kong 3.0. #15878. Could you please help and merge it? Thanks in advance.

PR #15715, opened 2023-11-14
Update bonita, add 2023.2

rbioteau:
* Add 2023.2 version
* Update maintainers
* Update fetch reference branch
* Remove old unsupported version

### DIFF
--- a/library/bonita
+++ b/library/bonita
@@ -2,23 +2,23 @@ Maintainers: Danila Mazour <danila.mazour@bonitasoft.org> (@danila-m),
              Emmanuel Duchastenier <emmanuel.duchastenier@bonitasoft.org> (@educhastenier),
              Pascal Garcia <pascal.garcia@bonitasoft.org> (@passga),
              Anthony Birembaut <anthony.birembaut@bonitasoft.org> (@abirembaut),
-             Dumitru Corini <dumitru.corini@bonitasoft.org> (@DumitruCorini)
+             Romain Bioteau <romain.bioteau@bonitasoft.org> (@rbioteau)
 Architectures: amd64, arm64v8, ppc64le
 GitRepo: https://github.com/bonitasoft/bonita-distrib.git
 Directory: docker
 
-Tags: 2021.2-u0, 2021.2, 7.13.0, 7.13
-GitFetch: refs/heads/docker/2021.2
-GitCommit: a1d9ee5e31d38958aa553cc7f9d465f1151d902f
-
 Tags: 2022.1-u0, 2022.1, 7.14.0, 7.14
-GitFetch: refs/heads/master
-GitCommit: 694bf79347add872f8c6a4c0a7f5c3ef12c31dc8
+GitFetch: refs/heads/docker/2022.1
+GitCommit: 4cdeb1c385b981e7074ce19cc685c08028d7149d
 
 Tags: 2022.2-u0, 2022.2, 7.15.0, 7.15
-GitFetch: refs/heads/master
-GitCommit: 21c0e51634e836feaf910e8e3d6e95de200e1814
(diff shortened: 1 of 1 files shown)


yosifkit:

---

I'm confused by the addition of `gosu` as it doesn't appear to be used anywhere because `su-exec` is used instead and installed from Alpine packages via `apk`. `gosu` version `1.13` is also a bit out of date (https://github.com/tianon/gosu/releases).

---

```dockerfile
VOLUME ["/opt/bonita/conf/logs"]
```

Does the application expect to modify the directory in a persistent way that needs to stick around from invocation to invocation? (Given the "permanence" of a `VOLUME` declaration, we recommend keeping them to only the directories the application stores active state in, and even more specifically active state that's actually necessary to keep if the container is completely destroyed and recreated, such as MySQL's `/var/lib/mysql` directory.)  I recommend documenting directories like this that users may want to save beyond container lifecycle. (But if you really want users to get an anonymous automatic volume on when they don't specify one because the logs are important, then that is also fine)

rbioteau:
@yosifkit Thanks for the feedback !

> Using `uname -m` or `arch` for platform detection is very strongly discouraged, as it retrieves the kernel's architecture (from the host), when you really need to retrieve the userspace architecture of the image. `apk --print-arch` or `dpkg --print-architecture` are going to be the most reliable options.
> 
> I'm confused by the addition of `gosu` as it doesn't appear to be used anywhere because `su-exec` is used instead and installed from Alpine packages via `apk`. `gosu` version `1.13` is also a bit out of date (https://github.com/tianon/gosu/releases).

Indeed, `gosu` should be removed.

> ```dockerfile
> VOLUME ["/opt/bonita/conf/logs"]
> ```
> 
> Does the application expect to modify the directory in a persistent way that needs to stick around from invocation to invocation? (Given the "permanence" of a `VOLUME` declaration, we recommend keeping them to only the directories the application stores active state in, and even more specifically active state that's actually necessary to keep if the container is completely destroyed and recreated, such as MySQL's `/var/lib/mysql` directory.) I recommend documenting directories like this that users may want to save beyond container lifecycle. (But if you really want users to get an anonymous automatic volume on when they don't specify one because the logs are important, then that is also fine)

We used `VOLUME` as a way to explicit the "mountable" volumes. Regarding our use case, we should be able to remove it safely and document the mounting options.

All these changes will be applied in our development branch, is it ok on your side to approve the PR without these changes ?

We also have found an inconsistency in an older tag. Is it ok to change the git reference for this older tag without updating the version ? Will the image be rebuilt and overwritten, or is it against the publication policy ?

Thanks
Romain

yosifkit:
> We also have found an inconsistency in an older tag. Is it ok to change the git reference for this older tag without updating the version ? Will the image be rebuilt and overwritten, or is it against the publication policy ?

Updating an already published image is perfectly fine. One of the main tenets of Docker Official Images is to rebuild for updates and security fixes. All of the official images listed in the `library` files are subject to rebuild when their base image is updated I.E., if the `ubuntu` images are updated, then all official images from `ubuntu` will be rebuilt upon those new images.

yosifkit:
> All these changes will be applied in our development branch, is it ok on your side to approve the PR without these changes ?

For the Volume, that is up to you if you want to change it. For the unneeded `gosu` + `arch`, I'd strongly encourage their removal and updating this PR in order to keep the official images at their best.

rbioteau:
@yosifkit I removed `gosu` and related `arch` usage. For the `VOLUME` we will make additional testing before updating the images.
We are currently performing our release rollout and the merge of the PR is kind of blocking. 
Do you think you can merge it asap ?

Thanks a lot !

yosifkit (approved):

PR #15716, opened 2023-11-14
Updates to WebSphere and Open Liberty for release 23.0.0.11

wraschke:


leochr:
Not sure why the `Munge PR / Diff Comment` action failed. Seems like a connection issue, perhaps intermittent. I am not sure how to retrigger the action.
```
ERROR: failed to authorize: failed to fetch oauth token: Post "https://auth.docker.io/token": read tcp 10.1.0.115:53506->3.216.34.172:443: read: connection reset by peer
```

@yosifkit please help when you get a chance. Thank you

yosifkit:
Using `uname -m` or `arch` for platform detection is very strongly discouraged since it retrieves the kernel's architecture (from the host), when you really need to retrieve the userspace architecture of the image. `apk --print-arch` or `dpkg --print-architecture` are going to be the most reliable options.

leochr:
@yosifkit Thanks for the review. This particular change (with `uname -m`) is related to Checkpoint/restore (CRIU) functionality in Liberty and we need the host architecture in this case. Checkpoint should be done based on what the host is running.

PR #15724, opened 2023-11-15
Update Eggdrop develop branch to needtls patchlevel

vanosg:


yosifkit:
Image label names should adhere to the image spec: https://github.com/opencontainers/image-spec/blob/v1.0.1/annotations.md. So, probably `LABEL org.opencontainers.image.authors` is the closest match for `MAINTAINER`?

vanosg:
Thanks for the pointer, info is updated

PR #15731, opened 2023-11-15
emqx v5.3.1

id:


yosifkit:
`--allow-unauthenticated` shouldn't be added to `apt-get`; that could reduce the security of the image.

id:
Fixed.

yosifkit (approved):

PR #15736, opened 2023-11-16
Releasing Neo4j 4.4.28 and 5.14.0

glindroth:


yosifkit:
🙇‍♂️  Apologies, I should have linked to https://github.com/docker-library/faq/#multi-stage-builds on your [earlier query](https://github.com/docker-library/official-images/pull/15656#issue-1974505172).

We try hard to avoid multi-stage builds in almost every instance. Since this does a multi-stage build to basically prevent an `rm` and an `apt-get purge --auto-remove ...`, we would rather that it be just a regular non-multistage build. It will help make it clearer for users viewing `docker history` where things came from.

jennyowen:
Hi @yosifkit no problem :slightly_smiling_face: I've had another crack at it.

jennyowen:
I guess the changes were not so great? I did a little more cleaning up so I hope the Dockerfiles are ok now?

yosifkit:
My feedback is that the images are a little bloated (~200MB bigger) and likely need to uninstall `gcc`, `git`, and `make` (and their deps) in the `RUN` layer in which they are installed (added to the `apt-get purge --auto-remove`). And that the `apt` cli usages should be using `apt-get` (I think they have been in there a while).

`microdnf` is sadly not great for cleaning up a package and its deps since its only goal is installing them: https://github.com/infosiftr/tomcat/blob/b3ffb23f4a7b59d5dc41c01330390efca56ad6ce/9.0/jdk22/openjdk-oraclelinux8/Dockerfile#L29-L44 https://github.com/infosiftr/tomcat/blob/b3ffb23f4a7b59d5dc41c01330390efca56ad6ce/9.0/jdk22/openjdk-oraclelinux8/Dockerfile#L140-L141

yosifkit (approved):

jennyowen:
thanks @yosifkit !

yosifkit:
Oh, thanks for the notice; looks like a job failure: https://doi-janky.infosiftr.net/job/docs/. Looking into it.

PR #15751, opened 2023-11-19
archlinux: Release 20231119.0.193111

archlinux-github:

---

### DIFF
--- a/library/archlinux
+++ b/library/archlinux
@@ -5,13 +5,18 @@ Maintainers: Santiago Torres-Arias <santiago@archlinux.org> (@SantiagoTorres),
              Justin Kromlinger <hashworks@archlinux.org> (@hashworks)
 GitRepo: https://gitlab.archlinux.org/archlinux/archlinux-docker.git
 
-Tags: latest, base, base-20231112.0.191179
-GitCommit: 9e3e5066b444021f47836449f6a9d89d0cf5861a
-GitFetch: refs/tags/v20231112.0.191179
+Tags: latest, base, base-20231119.0.193111
+GitCommit: 9e52830b269204a8f28868aa138b51692d2015a1
+GitFetch: refs/tags/v20231119.0.193111
 File: Dockerfile.base
 
-Tags: base-devel, base-devel-20231112.0.191179
-GitCommit: 9e3e5066b444021f47836449f6a9d89d0cf5861a
-GitFetch: refs/tags/v20231112.0.191179
+Tags: base-devel, base-devel-20231119.0.193111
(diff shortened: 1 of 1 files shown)


hashworks:
This adds a new type: `multilib-devel`, which includes the respective [meta package](https://archlinux.org/packages/multilib/x86_64/multilib-devel/).

tianon:
Am I reading correctly in https://gitlab.archlinux.org/archlinux/archlinux-docker/-/blob/337a19c996695f92aac17e2002b534dee732ab0a/scripts/make-rootfs.sh#L29 that all the images include the `base` group, and that `base-devel` and `multilib-devel` are included on top of `base` in all cases?

Something you could consider is continuing to maintain `archlinux:latest`/`archlinux:base` as you do today, but swapping the other two to instead be a short `Dockerfile` that's `FROM archlinux:latest` (or `archlinux:base`) and installs the relevant metapackages instead, which would decrease the number of artifacts you have to produce directly (and give users who use more than one of them a slight space savings since they'd then all share that bottom layer).

(The limitations on `archlinux` images being `FROM archlinux` only apply to an image being `FROM` itself -- being from a different image in the same repo is just fine and how many of our images are maintained. :eyes: :+1:)

hashworks:
You mean like:
```
FROM archlinux:base

RUN pacman -S base-devel
```
?

We explicitly don't do that, since it would break the reproducibility - which is a requirement for the `official-images` repo, AFAIK? There is no guaranty that the files in the repository won't change later on.

If that wouldn't be a problem we wouldn't need the rootfs at all.

hashworks:
Or do you mean: Don't offer `base-devel` and `multilib-devel` in `official-images` at all? This would break existing pipelines.

dvzrv:
Since this morning the official container image is broken (due to the keyring being out of-date for too long).
Please act now so that a new image can be released!

aminvakil:
Having this problem also here, there is not an appropriate way to fix this right now if base image is broken.

hashworks:
I'm sorry for any inconvenience this has caused, the delay here is on me. Until we resolve this, please use the daily images (f.e. `docker.io/archlinux/archlinux:base`).

As mentioned on IRC, it would be great if we could close this MR and merge a new one with the current process, so we can fix stuff for of end-users for now. I've already [thought about a better approach on our side](https://gitlab.archlinux.org/archlinux/archlinux-docker/-/issues/100), but it will take a while to implement the required changes for all registries.

tianon:
Yep, agreed, let's fix this ❤️  I'll close this and wait for your new PR 👍

tianon:
Done! https://github.com/docker-library/official-images/pull/15984 👍

PR #15754, opened 2023-11-20
New Docker image for Silverpeas 6.3.2

mmoqui:
Update silverpeas Docker library to take into account Silverpeas 6.3.2


whalelines:
Why does the download location of wildfly change in one of the images but not the other?

mmoqui:
> Why does the download location of wildfly change in one of the images but not the other?

Because it is an error. The Docker image for Silverpeas 6.2.3 should be kept at  version 6.3.2-b1 instead of 6.3.2. I'm going to fix it. Thanks to have reporting that.

mmoqui:
For information, the download location of Wildfly is now the same between the two images

PR #15764, opened 2023-11-24
Update Nextcloud

J0WI:


LaurentGoderre (approved):

yosifkit (approved):

weits:
Why was stable and production tags removed? Isn't it some kind of glitch?

yosifkit:
> Why were stable and production tags removed? Isn't it some kind of glitch?

As far as I know they are detected automatically. So, it is probably the same reason as previously: https://github.com/docker-library/official-images/pull/15063#issue-1813864614.

PR #15780, opened 2023-11-28
Upstream "doi-buildkit.patch" from meta-scripts

tianon:
Finally biting the bullet and upstreaming https://github.com/docker-library/meta-scripts/blob/8182b4ae6742735211feeb2b03c8e5f45401548f/doi-buildkit.patch (so we can stop patching in our build jobs)


whalelines (approved):
LGTM

Are there any plans to upstream the changes in tianon/buildkit into moby/buildkit?

yosifkit (approved):

tianon:
That's a good idea, but probably not a high priority for either team -- we need more architectures than they support (currently that set is mostly just `mips64le`, which they did have previously but probably dropped due to compilation errors) and https://github.com/tianon/dockerfiles/blob/014f1215f2cbfd7ae544e4d8715e97204378a092/buildkit/mips64le.patch, which I don't think is necessarily the 100% _correct_ patch, but it's correct enough for our needs.

In general, it's a very, very vanilla build of BuildKit: https://github.com/tianon/dockerfiles/blob/014f1215f2cbfd7ae544e4d8715e97204378a092/buildkit/Dockerfile

Can also see that via attestations on the image on Hub:
- https://explore.ggcr.dev/?image=tianon/buildkit:0.12
- https://explore.ggcr.dev/?blob=tianon%2Fbuildkit%40sha256%3A5ccc7f886c92bb456a21954dc410fd23fc5ee2d046baaf9c6508eb645556f954&jq=.predicate.metadata%5B%22https%3A%2F%2Fmobyproject.org%2Fbuildkit%40v1%23metadata%22%5D.source.infos%5B0%5D.data&jq=base64+-d&mt=application%2Fvnd.in-toto%2Bjson&render=raw&size=17499

PR #15787, opened 2023-11-29
Remove unsupported flink-1.15

Myasuka:
Since flink-1.18 has been released, we could remove the unsupported flink-1.15 now.


yosifkit (approved):

yosifkit:
In the future, please use one PR for `flink` changes rather than three. Luckily, none of them (https://github.com/docker-library/official-images/pull/15786, https://github.com/docker-library/official-images/pull/15787, https://github.com/docker-library/official-images/pull/15788) conflicted.

PR #15788, opened 2023-11-29
Update to Flink 1.16.3

1996fanrui:


yosifkit:
In the future, please use one PR for `flink` changes rather than three. Luckily, none of them (https://github.com/docker-library/official-images/pull/15786, https://github.com/docker-library/official-images/pull/15787, https://github.com/docker-library/official-images/pull/15788) conflicted.

yosifkit (approved):

Issue #15753, opened 2023-11-20
Elasticsearch arm images missing for 8.11.1 and 7.17.15


alpar-t:
example:  
  https://hub.docker.com/layers/library/elasticsearch/8.11.1/images/sha256-31fee9107fc01bbe3808e8f8bd1a0375b1c185c5d4045c4d92871ab13be4f26a?context=explore
  https://hub.docker.com/layers/library/elasticsearch/7.17.15/images/sha256-93e81ebea43a24a9d1007aa9415af8291ec6744ad6d5317b505c53f5907d47b7?context=explore

We did have them for older versions, example 8.11.0:
  https://hub.docker.com/layers/library/elasticsearch/8.11.0/images/sha256-16c4a2652c83a992e245ca3b88c01bfe518cbbcd6a14dca2eb9a92856ad6b560?context=explore
  
  Originally reported in https://github.com/elastic/elasticsearch/issues/102332#issuecomment-1817234340

LaurentGoderre:
This also affected Kibana 8.11.1 & 7.17.15 and logstash 8.11.1

tianon:
There's something weird happening with our builder that we're not 100% sure what the root cause is (we're still investigating):

```console
[...]

PR #15808, opened 2023-11-30
Update ELK to "proper" images

tianon:
~~This is a concrete test of what I was suggesting in https://github.com/docker-library/official-images/issues/15753#issuecomment-1834442877 (not intended for merge as-is)~~

This updates them to point to the original source repository so they build directly instead of being a special case, and updates the maintainers appropriately.

See https://github.com/docker-library/official-images/issues/15753#issuecomment-1834442877 and following conversation :+1:

### DIFF
--- a/library/elasticsearch
+++ b/library/elasticsearch
@@ -1,15 +1,14 @@
-# this file is generated via https://github.com/docker-library/elasticsearch/blob/1c213e964445c73ff44e2ed1ff18e28995800d8e/generate-stackbrew-library.sh
-
-Maintainers: Tianon Gravi <admwiggin@gmail.com> (@tianon),
-             Joseph Ferguson <yosifkit@gmail.com> (@yosifkit)
-GitRepo: https://github.com/docker-library/elasticsearch.git
+Maintainers: Mark Vieira (@mark-vieira)
+GitRepo: https://github.com/elastic/dockerfiles.git
+Directory: elasticsearch
+Builder: buildkit
 
 Tags: 8.11.1
 Architectures: amd64, arm64v8
-GitCommit: 84a62adaf958d51f39376ed636a7d59f2e92ca69
-Directory: 8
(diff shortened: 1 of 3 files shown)


tianon:
A GitHub team is OK, but we'll want at least one individual listed too because someone outside the org/team unfortunately can't @-mention a team on GitHub (so neither us nor our bot would be able to do so :disappointed:).

mark-vieira:
> A GitHub team is OK, but we'll want at least one individual listed too because someone outside the org/team unfortunately can't @-mention a team on GitHub (so neither us nor our bot would be able to do so 😞).

Grr. Does the contact have to be a github alias? Could it be an email mailing list or something? Just trying to avoid the obvious problem of folks moving teams, leaving the org, etc.

mark-vieira:
> (Pulling that thought thread a little is probably going to lead to some questions that might be answered by https://github.com/docker-library/faq#can-i-use-a-bot-to-make-my-image-update-prs so I'll just drop that here in case it does 😄)

@alpar-t this might be something to consider going forward.

mark-vieira:
> Email is an optional component of the format in the file, but most of our pings happen on GitHub (like you can see in the bot comment above on this PR) 😞

In the interest of getting our arm images available sooner rather than later I'm good with assigning myself for now. @watson, and @jsvd, thoughts?

watson:
I'm not the correct maintainer for Kibana. I'm currently reaching out internally to find the right person. However, if you want to merge this to unblock the ARM images, go ahead. We can always update this maintainer once I figure out who it is.

watson:
After checking, I think I'm probably the best candidate to put down as the maintainer for Kibana after all 👍

jsvd:
Sign me up. In the future I'd like for the first point of contact to be a mailing list and potentially a bot, to avoid the issues raised by @mark-vieira.

yosifkit (approved):

PR #15809, opened 2023-12-01
[ros] one-off rebuild of snapshot images to have new GPG key

mikaelarguedas:
The ROS Snapshots APT repository key [expired](https://github.com/osrf/docker_images/issues/697) causing users of snapshots images unable to install pinned versions of their ROS packages.

This PR reenables snapshot images build for a one-off build before disabling them again
To ensure rebuild / cache busting, two run statements have been swapped.

I understand these base images are EOL for a while, but as multiple users and companies still rely on them on production systems we would love to be fix this issue via a one-shot build

### DIFF
--- a/library/ros
+++ b/library/ros
@@ -2,6 +2,233 @@ Maintainers: Tully Foote <tfoote+buildfarm@osrfoundation.org> (@tfoote)
 GitRepo: https://github.com/osrf/docker_images.git
 
 ################################################################################
+# Release: indigo
+
+########################################
+# Distro: ubuntu:trusty
+
+Tags: indigo-ros-core, indigo-ros-core-trusty
+Architectures: amd64, arm32v7
+GitCommit: 1516dcfd1a92ffd2b944c007763a84abe40aad7a
+Directory: ros/indigo/ubuntu/trusty/ros-core
+
+Tags: indigo-ros-base, indigo-ros-base-trusty, indigo
+Architectures: amd64, arm32v7
+GitCommit: 20061b005b245d1b7e23626afd0ea2c39de9db49
+Directory: ros/indigo/ubuntu/trusty/ros-base
+
+Tags: indigo-robot, indigo-robot-trusty
+Architectures: amd64, arm32v7
+GitCommit: 20061b005b245d1b7e23626afd0ea2c39de9db49
+Directory: ros/indigo/ubuntu/trusty/robot
+
+Tags: indigo-perception, indigo-perception-trusty
+Architectures: amd64, arm32v7
+GitCommit: 20061b005b245d1b7e23626afd0ea2c39de9db49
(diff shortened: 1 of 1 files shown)


tfoote (commented):
Thanks for forcing these rebuilds.

dambrosio:
@mikaelarguedas thanks for this. Any update on this? Specifically interested in the `foxy-ros-base-focal` as this is the base image for the ros1 bridge. Appreciate it!

yosifkit:
I think generally, a one-off rebuilding of EOL `ros` images seems fine (assuming that they can currently build on whatever base OS they are on).

From the other part of the failing "Naughty" test, you'll need to add a `GitFetch` in each item in the `library/ros` file to point to the branch in `osrf/docker_images` that contains the commit (like `GitFetch: refs/heads/branchname`).

yosifkit:
A tiny nitpick (mostly because it increases the chance of build failure, but also because we try to make sure images don't change files from previous layers), `dashing` and `eloquent` images install the same gpg key in `core` and `bridge` (which is from `base` which is from `core`).

---

I don't think we've mentioned this before, but we now recommend using `gpg` directly instead of `apt-key` since `apt-key` is [deprecated](https://manpages.debian.org/testing/apt/apt-key.8.en.html#DESCRIPTION). This also allows you to scope the usage of the key to a specific apt repo, like the [`mysql`](https://github.com/docker-library/mysql/blob/dc35beded3a41c4c62dfaa7f86ea5721c02a2e8b/8.0/Dockerfile.debian#L55-L68) image.

mikaelarguedas:
Thanks for the pointer!

There is discussion to migrate to a non deprecated approach in the future https://github.com/osrf/docker_images/issues/621#issuecomment-1134779223 but this may take a bit of time.

In the meantime to move this across the finish line for the EOL images I see different approaches:
- is there a path to get this PR to build once as-is disable the builds afterwards?
- copy logic from the mysql image and hardcode it once in the EOL images to get them to build once
- move back from fetching packages from the main apt repository (packages.ros.org) in the ros1-bridge instead of the archiving repopsitory (snpashots.ros.org)

If 1 is possible that'd be awesome, otherwise I'll explore 2 or 3 and update this PR accordingly

yosifkit (approved):
> is there a path to get this PR to build once as-is disable the builds afterwards?

 LGTM

tianon:
Just to be clear, the method to "disable" the builds after they succeed is
a new PR that removes them from this file again.

mikaelarguedas:
>  > is there a path to get this PR to build once as-is disable the builds afterwards?

> LGTM

Awesome thanks!

> Just to be clear, the method to "disable" the builds after they succeed is a new PR that removes them from this file again.

:+1: https://github.com/docker-library/official-images/pull/15871

PR #15826, opened 2023-12-04
Update julia

yosifkit:
Changes:

- https://github.com/docker-library/julia/commit/8852e4c: Update 1.10-rc to 1.10.0-rc2


tianon:
We should probably figure out those test failures eventually, but they're _just_ one of the tests failing to download from google.com on Windows Server 1809 specifically (LTSC 2022 is fine), so not a super high priority to dig into. :joy:

PR #15828, opened 2023-12-04
Improve diff-pr build order comparison

yosifkit:
Use the first tag (aka least specific) before and after sorting by build order


tianon:
Noting for posterity that https://github.com/docker-library/official-images/pull/15823 is the one @yosifkit noticed this bug on :eyes:

tianon (approved):

PR #15830, opened 2023-12-05
Update Jetty versions to 12.0.4

lachlan-roberts:
## Update Jetty Versions

- `12.0.3` -> `12.0.4`
- adds variants for JDK21.


tianon:
FWIW, if you wanted to cut down on the number of transitory build failures, it might be worth adding `gpgconf --kill all` between your use of `gpg` and before removing `$GNUPGHOME` :eyes:

PR #15836, opened 2023-12-05
[percona] update ps-57 version

adivinho:


yosifkit:
`ADD` on a remote URL is allowed in official images. The item should be downloaded and verified ([as much as possible](https://github.com/docker-library/official-images/tree/15cd81789be99cf4562fd84ee4257e9903bb1d2d#image-build)) and prepared (like permissions, or extraction from an archive) within a single `RUN` line.  (This is why the `Diff Comment` GHA job is failing, since our tooling around official images cannot handle the remote URL for `ADD`)

adivinho:
@yosifkit This part has been reworked.

yosifkit (approved):
🤔 I am uncertain what users might think about automatic homephoning without the users' consent. I wouldn't want it on images that I personally use. I understand that telemetry is super useful, even in open source projects (like [golang](https://research.swtch.com/telemetry-intro) plans to add), so I think this is generally reasonable.

Can you do a follow up PR to update the Docker Hub documentation so that users will know that this is added and how to opt out?  (https://github.com/docker-library/docs/)

tianon:
> Note that at the start of the discussion the design was opt-out, but the final design, as a result of that discussion and other conversations, is opt-in. Thanks to the community for the constructive feedback.

(from https://research.swtch.com/telemetry -- might be a useful/interesting data point for you :eyes:)

PR #15846, opened 2023-12-06
Initial commitment of clickhouse official image

Felixoid:
Hello, dear official library team. Here's our attempt to create an official image `clickhouse` in the scope of https://github.com/docker-library/official-images/issues/14136 and https://github.com/ClickHouse/ClickHouse/issues/31473

Both the docs and official images PRs are created simultaneously.

https://github.com/docker-library/docs/pull/2397

# Checklist for Review

-	[x] does a [documentation](https://github.com/docker-library/docs/blob/master/README.md) PR exist? (should be reviewed and merged at roughly the same time so that we don't have an empty image page on the Hub for very long) https://github.com/docker-library/docs/pull/2397


Felixoid:
Interesting [case](https://github.com/docker-library/official-images/actions/runs/7115508102/job/19372251987?pr=15846#step:7:424), I am addressing it in our master, which will regenerate the LDF

tianon:
Unfortunately having the `Dockerfile` in the root of the upstream repository (while useful for local development) doesn't work well as the target for official-images. Since the `Dockerfile` uses `COPY . ...`, it essentially means the entire upstream source code is now part of the Docker build context, and thus becomes part of the reviewed artifacts (see `diff-pr.sh` for how we generate our diff; it is basically taking all docker build contexts from the starting `library/foo` file and diffing them to the new contexts from the PR).

This is why the "Diff Comment" CI job is failing :see_no_evil:

tianon:
Apologies, I misread the PR -- I think that might actually be failing because it's backfilling too many versions :sweat_smile:

tianon:
Are all of these major.minor combinations still actively supported? (By which I mean that if they had a problem, there would be a new release/update?)

Felixoid:
Hi Tianon, thanks for the review!

Indeed, the failure of `Diff Comment` is about the POST body.

> Are all of these major.minor combinations still actively supported? (By which I mean that if they had a problem, there would be a new release/update?)

No, the file for currently supported versions is the following:

```
 The file is generated by https://github.com/ClickHouse/ClickHouse/blob/e2f425011c589c5e6d8b1d4f67ccd7a8730158a4/tests/ci/official_docker.py

Maintainers: Misha f. Shiryaev <felixoid@clickhouse.com> (@Felixoid),
             Max Kainov <max.kainov@clickhouse.com> (@mkaynov),
             Alexander Sapin <alesapin@clickhouse.com> (@alesapin)
GitRepo: https://github.com/ClickHouse/ClickHouse.git
GitFetch: refs/heads/docker-library
GitCommit: 43208ec354e4873f59508c45b81204b9f7bf439e

Tags: alpine, 23-alpine, 23.11-alpine, 23.11.1-alpine, 23.11.1.2711-alpine
Architectures: amd64, arm64v8
Directory: docker/official/server/23.11.1.2711
File: Dockerfile.alpine

Tags: latest, focal, 23, 23-focal, 23.11, 23.11-focal, 23.11.1, 23.11.1-focal, 23.11.1.2711, 23.11.1.2711-focal
Architectures: amd64, arm64v8
Directory: docker/official/server/23.11.1.2711
File: Dockerfile.ubuntu

Tags: 23.10-alpine, 23.10.5-alpine, 23.10.5.20-alpine
Architectures: amd64, arm64v8
Directory: docker/official/server/23.10.5.20
File: Dockerfile.alpine

Tags: 23.10, 23.10-focal, 23.10.5, 23.10.5-focal, 23.10.5.20, 23.10.5.20-focal
Architectures: amd64, arm64v8
Directory: docker/official/server/23.10.5.20
File: Dockerfile.ubuntu

Tags: 23.9-alpine, 23.9.6-alpine, 23.9.6.20-alpine
Architectures: amd64, arm64v8
Directory: docker/official/server/23.9.6.20
File: Dockerfile.alpine

Tags: 23.9, 23.9-focal, 23.9.6, 23.9.6-focal, 23.9.6.20, 23.9.6.20-focal
Architectures: amd64, arm64v8
Directory: docker/official/server/23.9.6.20
File: Dockerfile.ubuntu

Tags: alpine-lts, 23.8-alpine, 23.8.8-alpine, 23.8.8.20-alpine
Architectures: amd64, arm64v8
Directory: docker/official/server/23.8.8.20
File: Dockerfile.alpine

Tags: lts, focal-lts, 23.8, 23.8-focal, 23.8.8, 23.8.8-focal, 23.8.8.20, 23.8.8.20-focal
Architectures: amd64, arm64v8
Directory: docker/official/server/23.8.8.20
File: Dockerfile.ubuntu

Tags: 23.3-alpine, 23.3.18-alpine, 23.3.18.15-alpine
Architectures: amd64, arm64v8
Directory: docker/official/server/23.3.18.15
File: Dockerfile.alpine

Tags: 23.3, 23.3-focal, 23.3.18, 23.3.18-focal, 23.3.18.15, 23.3.18.15-focal
Architectures: amd64, arm64v8
Directory: docker/official/server/23.3.18.15
File: Dockerfile.ubuntu
```

I've generated all the versions that could be built at the moment, so following the next sentence from [README](https://github.com/docker-library/official-images#library-definition-files):

> When a new repository is proposed, it is common to include some older unsupported versions in the initial pull request with the agreement to remove them right after acceptance.

It's pretty easy to rebuild the file to any version, so I am very open to advice on what to use as the minimal one.

tianon:
Ah, let's start with just 23.11 (unless there's a significant difference in the Dockerization for older versions) and see if that gets the diff small enough.  If we start there and the delta between that and other versions is small, then future diffs will be small too because they'll all diff against the merged version instead of all versions being considered brand new code.

tianon:
For a small bit of initial Dockerization review, you'll want to take a look at https://github.com/docker-library/official-images/issues/10794 (Alpine + glibc is going to be a non-starter).

Additionally, having `/bin/sh` or `/bin/bash` as PID1 while the server is running (speaking of the `while true` loop I see in the entrypoint which is effectively acting as a poor-man's process supervisor) is going to be a very steep uphill battle to convince us it's stable and reliable, especially in the face of receiving signals and error handling / process death edge cases.

Beyond that, please be patient -- new image review requires much higher maintainer bandwidth than update review does, so we do have a backlog.

Felixoid:
> Ah, let's start with just 23.11

Yep, all versions have the same content now

> For a small bit of initial Dockerization review, you'll want to take a look at #10794 (Alpine + glibc is going to be a non-starter).

I think it's related to the part `COPY --from=glibc-donor`? So, it looks like we won't be able to build alpine images. Looks fair to me, we can take them off the discussion

> Additionally, having /bin/sh or /bin/bash as PID1 while the server is running (speaking of the while true loop I see in the entrypoint which is effectively acting as a poor-man's process supervisor) is going to be a very steep uphill battle to convince us it's stable and reliable, especially in the face of receiving signals and error handling / process death edge cases.

I feel you are referencing https://github.com/ClickHouse/ClickHouse/blob/master/docker/server/entrypoint.sh#L132-L139. It's the bringing the clickhouse-server up for the [initial setup](https://github.com/ClickHouse/ClickHouse/blob/master/docker/server/entrypoint.sh#L132-L139) of DB. And it has the fuse to check that the server is actually [finished](https://github.com/ClickHouse/ClickHouse/blob/master/docker/server/entrypoint.sh#L170-L173) after it. The real server works as the only process in the container after [exec](https://github.com/ClickHouse/ClickHouse/blob/master/docker/server/entrypoint.sh#L196) command

> Beyond that, please be patient -- new image review requires much higher maintainer bandwidth than update review does, so we do have a backlog.

Thank you! It's very nice of you to have a transparent process :pray:

Felixoid:
Happy New Year, dear colleagues!

Is there anything we could do from our site to boost the review, maybe?

Felixoid:
Hello, dear @tianon @yosifkit. It's already two months since the latest update. Any feedback?

Felixoid:
Colleagues? It's already three months since the last review round. Is it even possible to add the image to the library?

melvynator:
@tianon Is it possible to have an estimate of when this PR can be reviewed?

whalelines:
We cannot give an estimate, but we will review this PR as soon as we can. Reviewing new DOI submissions is quite time intensive and we have to balance that commitment with ensuring the normal flow of updates to existing DOI.

It can help to speed things up if you ensure your PR adheres to our contribution guidelines, https://github.com/docker-library/official-images?tab=readme-ov-file#contributing-to-the-standard-library .

We apologize for the delay and truly appreciate your patience through this process.

yosifkit:
Here is a first pass review. There might be other required changes, but I wanted to share these to lessen the wall of text.

---

```dockerfile
&& if [ -n "${deb_location_url}" ]; then \
...
RUN if [ -n "${single_binary_location_url}" ]; then \
```

- Since these would unused by the Docker Official Image, these essentially empty layers should probably live elsewhere so as to simplify the Dockerfile for DOI review and for users reviewing the image.

---

- `DEBIAN_FRONTEND` is usually not necessary. In newer versions of Ubuntu and Debian, the problem is completely fixed. (https://github.com/moby/moby/issues/4032#issuecomment-2008339218) If there is a specific install that fails without it, it should only be applied for that install, like `apt-get update; DEBIAN_FRONTEND=noninteractive apt-get install -y ...`.

---

```dockerfile
apt-get upgrade -yq
```

- We recommend against using blanket package upgrades (`apt-get upgrade`/`apk upgrade`/`yum upgrade`/`yum update`) for official-images. When package upgrades are applied in a dependent image, it duplicates content of the base image, making the image larger than necessary. It also only delays the inevitable "there are outdated packages". The Official Images build pipeline only rebuilds on a update to the Dockerfile or a base image update, so we make periodic base image updates to then fully rebuild all dependent images (e.g., the [Debian](https://github.com/docker-library/official-images/pulls?q=is%3Apr+label%3Alibrary%2Fdebian) and [Ubuntu](https://github.com/docker-library/official-images/pulls?q=is%3Apr+label%3Alibrary%2Fubuntu) images are updated a least every 30 day).

	>

---

```dockerfile
ARG TARGETARCH
```

- We do not recommend relying on `TARGETARCH` - if it is needed, then use `dpkg -- print-architecture` instead, but it looks like that is part of the "unused for Docker Official Images" block

---

- `gpg` invocations should always include `--batch` (which effectively puts it into "API mode" instead of "UI mode").

---

- `gpg --recv-keys` should use a full fingerprint (not merely short or long).

---

- `RUN chmod +x ...` is going to duplicate the file added by `COPY`, so `+x` should be applied in Git instead.

---

Somewhat harmless, but good to improve:

- The `ca-certificates` installed with `apt-transport-https` is unnecessary since it is installed on an earlier `RUN` with `wget`. 

- Running `apt-get clean` will be a no-op in all the Debian and Ubuntu base images -- it happens automatically (https://github.com/debuerreotype/debuerreotype/blob/d29dd5e030525d9a5d9bd925030d1c11a163380c/scripts/debuerreotype-minimizing-config#L85-L109).

- Having an `ARG` for swapping the Ubuntu mirror in `sources.list` is little odd.

Felixoid:
Thank you @yosifkit, I'll improve it in our repository to address the points.

Felixoid:
So, here is what I've come up with https://github.com/ClickHouse/ClickHouse/pull/63400/files

---

I'll try to isolate our specific layers so they won't be in the official Dockerfile. The TARGETARCH will also be filtered.

---

We use ubuntu:20.04. `ARG DEBIAN_FRONTEND=noninteractive` is build-specific, so it doesn't affect the final image. We follow the one-before-the-last [comment](https://github.com/moby/moby/issues/4032#issuecomment-192327844) in the issue you gave.

Since the image we use is affected, I'd like to leave it as is if there's no strict requirement to remove it.

---

`apt-get upgrade -yq` will go away with no regrets.

---

`gpg` parts will be fixed

---

`COPY --chmod=0755` should be a valid replacement. Not sure since when it's available, but should work for both us and the official build system, right?

---

- Nice catch of `ca-certificates`, I thought I cleaned everything already
- Didn't know about the `apt-get clean`. But I am not sure I got your link. Is it happen automatically on every build step? Because we run `apt update,` the repositories indices are downloaded in the very same step. I think it can be a noop because of `rm -rf /var/lib/apt/lists/* /var/cache/debconf /tmp/*`.
- It relates to our build system to easily switch on another repository. If the building process moves somewhere, we can switch it by a single argument instead of editing many places. I hope, it really doesn't harm.

LaurentGoderre:
Would it be possible to have an orphan branch that doesn't have all the rest of the code. Cloning even the docker branch is several hundred of meg and takes a while to download.

Felixoid:
Hi Laurent. Do you mean for the review or regularly?

LaurentGoderre:
I mean regularly not to slow down the builds

Felixoid:
Doesn't the build system use `--depth=0`? Despite the topic, it's a good practice to use it when the history is unnecessary.

With the argument the download is "only" 80M and 16 seconds

```
> git clone --depth=1 git@github.com:ClickHouse/ClickHouse.git ClickHouse
Cloning into 'ClickHouse'...
remote: Enumerating objects: 26569, done.
remote: Counting objects: 100% (26569/26569), done.
remote: Compressing objects: 100% (21855/21855), done.
remote: Total 26569 (delta 3428), reused 12207 (delta 2301), pack-reused 0
Receiving objects: 100% (26569/26569), 77.09 MiB | 12.43 MiB/s, done.
Resolving deltas: 100% (3428/3428), done.
Updating files: 100% (27655/27655), done.
```

I am not sure how safe will be an orphan branch in the repo. And having another repo for the docker-exclusive staff adds another layer of complexity to manage

LaurentGoderre:
ooh no., if there is a separate repo then we don't need the orphan branch

Felixoid:
I've found how to improve `bashbrew` to fetch only necessary commits. It improves the build time by an order of magnitude for `library/clickhouse` https://github.com/docker-library/bashbrew/pull/95

Felixoid:
Kindly hope for the next review round

tbragin:
Docker Team - please let us know if there are any updates, and thank you in advance!

whalelines:
Performing another review on the current changes is still in our queue. Unfortunately, our queue is quite long at present and reviews of new DOI are quite time consuming. We will get to this as soon as we possibly can. Thanks for being patient!

tianon:
# Blockers

- https://github.com/docker-library/official-images/pull/15846#issuecomment-2095975051

  As noted previously, we don't necessarily directly control the clone process, so updating `bashbrew` is not a solution and this is a hard-blocker.  My tests with BuildKit against the current monorepo took three full minutes to error out, failing to find the commit in the end.

  I would suggest a separate repository -- it can auto-sync from the monorepo if that is what you prefer (I'd suggest `git filter-branch` to accomplish that in a way that preserves the commits, but that's up-to-you also).

- what does `if ! clickhouse local -q "SELECT ''" > /dev/null 2>&1; then` do?  is the conditional necessary?  (it appears to always run this block and thus the conditional is not necessary and should be removed)

- with the appropriate GnuPG dance in place, is `apt-get install --allow-unauthenticated` really still necessary?

- the `CLICKHOUSE_DOCKER_RESTART_ON_EXIT` functionality is the `while true` referenced above

# Questions

- setting `LANG` *and* `LANGUAGE` *and* `LC_ALL` is weird -- is that really necessary?

- setting `TZ` explicitly is odd (we actually have an explicit test we run against all images to verify they're in UTC already, so images don't need to be explicit like that)

- many parts of the entrypoint script rely on `clickhouse` being in `PATH` but then others use `/usr/bin/clickhouse` explicitly -- is that intentional?

- `apt-get autoremove --purge -yq libksba8` 🤔 why this package? what pulled it in that is no longer required? (also the autoremove after that purge is a no-op because this command will remove anything else dangling too)

# Recommendations

- UID/GID 101 is really low, with a high chance of future collisions (with distro updates, package installs, etc)

  we normally suggest something higher than 1000, with over 2000 being less likely to collide with users' systems

- `CLICKHOUSE_UID`/`CLICKHOUSE_GID` is extremely unusual -- for users who need a specific UID/GID combination, we recommend using `--user` in the runtime/orchestration system (especially combined with `--security-opt no-new-privileges`) so that they *also* get the tighter security benefits associated with that

- it's definitely worth considering adding `set -u`, especially as all the variable references already appear to handle it appropriately

- I realize you probably need to have the entrypoint script, but I will tell you that our own experience maintaining scripts like it for PostgreSQL, MySQL, MongoDB, etc has been that it's extremely fragile, so if there's any way you could get away with avoiding it, you'd be doing yourself a favor (in environments like Kubernetes, initialization behavior is typically better handled by an explicit `initContainer`, for example).

  - related to that, https://github.com/docker-library/postgres/issues/159 is a particularly nasty failure mode that's worth being aware of / thinking about (`--restart=always` where the initialization fails, but only partway through, then the container restarts and comes right up but with an only halfway initialized database and no indication that there is anything wrong unless they look at the restart count on the container, which isn't exactly shown anywhere obvious)

- `apt-transport-https` is a transitional package (even in Ubuntu 20.04), and can be removed 👀

tianon:
Nice, thank you!

Regarding `RUN if ! clickhouse local -q "SELECT ''" > /dev/null 2>&1; then`, what if we swapped it to something like `if true; then` so it still matches really obviously for you, and is then obviously a no-op for us?  Would that be an acceptable middle ground?

The last thing I see while re-reviewing (that I thought I'd included in my last round but I guess I missed!) was this block:

```bash
        mkdir="clickhouse su ""${USER}:${GROUP}"" mkdir"
    fi
    if ! $mkdir -p "$dir"; then
```

100% not a blocker, but if you wanted this to quote `$USER:$GROUP` properly in the actual `mkdir` invocation you'll want to use `eval`, something like this instead:

```bash
        mkdir='clickhouse su "${USER}:${GROUP}" mkdir'
    fi
    mkdir+=' -p "$dir"'
    if ! eval "$mkdir"; then
```

or, if you prefer, Bash arrays (which `shellcheck` might be less loud about):

```bash
        mkdir=( mkdir )
...
        mkdir=( clickhouse su "${USER}:${GROUP}" mkdir )
    fi
    if ! "${mkdir[@]}" -p "$dir"; then
```

Thank you for your patience overall on this -- I'll take a look at the docs PR now. :eyes:

Felixoid:
> Regarding `RUN if ! clickhouse local -q "SELECT ''" > /dev/null 2>&1; then`, what if we swapped it to something like `if true; then` so it still matches really obviously for you, and is then obviously a no-op for us? Would that be an acceptable middle ground?

Hmm, what about I change it to the other way around? `clickhouse local -q "SELECT ''" && exit 0 || : ; \`. So, it will work for us, and the RUN block won't be executed. And it looks a bit less dirty.

`mkdir=( clickhouse su "${USER}:${GROUP}" mkdir )` looks like an excellent bashism to me. I'll add it to the review-related PR.

tianon:
OpenShift runs the image with a totally random UID (`--user`) -- it looks like the use case for that was to run the image _as root explicitly_ to support folks in user namespaces where root is mapped to their user on the host, but that's also arguably not a configuration I'd personally recommend (UID 0 in the kernel is still a special case in a lot of places/ways, so non-0 is suggested even in user namespaces).  Perhaps a "yes I really want to run the image as root and not switch to the clickhouse user" variable would be sufficient?

Re `clickhouse local`, I can live with that change, thanks! :heart:

Felixoid:
> Perhaps a "yes I really want to run the image as root and not switch to the clickhouse user" variable would be sufficient?

So, should we replace `CLICKHOUSE_UID/CLICKHOUSE_GID` with `CLICKHOUSE_RUN_AS_USER`? That sounds fair; it's a pretty solid thing to implement. Otherwise, people should use `--user` argument to set the container UID/GID.

tianon:
Yeah, that sounds a lot cleaner IMO (and pushes users to the more secure behavior) :+1: :rocket:

tianon (approved):
Still reviewing https://github.com/docker-library/docs/pull/2397 (and we need to merge both together), but this looks solid IMO now :heart:

Felixoid:
Should we add more images from the supported branches in another PR?

tianon:
Yep, let's do anything else in a follow-up; don't jinx it :joy: :heart:

(from our side, I just want @yosifkit to give this another set of eyes before we pull the trigger)

yosifkit:
```bash
&& chmod ugo+Xrw -R /var/lib/clickhouse /var/log/clickhouse-server /etc/clickhouse-server /etc/clickhouse-client
```

To minimize duplication across layers, `chmod`'s should be done in the layer that creates the file/folder. Some of these folders are empty or new this layer, but others (`/etc/clickhouse-*`) have files that come from an earlier layer. Not a blocker this time, but I'd recommend an improvement to `chmod` them in the `RUN` layer in which they are created.

yosifkit (approved):

Felixoid:
I can't say how grateful we are for this merge!

Seeing it alive is so satisfying

```
> docker pull clickhouse
Using default tag: latest
latest: Pulling from library/clickhouse
6414378b6477: Already exists
178d6bab72f0: Pull complete
a239d72431be: Pull complete
d9916e2f35d3: Pull complete
9762342e3969: Pull complete
3683edf9dd01: Pull complete
30f5422ae8a5: Pull complete
c5d92e4d9909: Pull complete
Digest: sha256:b7b9951c19ff195b1e6b832ef8679a46605a96ee5ec36cf8d94f199c30d810b8
Status: Downloaded newer image for clickhouse:latest
docker.io/library/clickhouse:latest
```

I am preparing the follow-up to push historical + supported releases

Felixoid:
Maybe the thing worth to leave solved here is the `chmod`

I've tested it, and the layer has ~100kB of data. It's not nice.

On the other hand, we have a lot of different [installation steps](https://github.com/ClickHouse/ClickHouse/blob/e50bbc433e5c57c96bbf71e22b900a28eb5be6c5/docker/server/Dockerfile.ubuntu#L34-L88) that aren't in the DOI Dockerfile.

Do you consider the following change as enough compromise to keep the line `&& chmod ugo+Xrw -R /var/lib/clickhouse /var/log/clickhouse-server /etc/clickhouse-server /etc/clickhouse-client` as is?

```diff
diff --git a/docker/server/Dockerfile.ubuntu b/docker/server/Dockerfile.ubuntu
index 0fe9a409ee4..e6bde845c4e 100644
--- a/docker/server/Dockerfile.ubuntu
+++ b/docker/server/Dockerfile.ubuntu
@@ -113,7 +113,9 @@ RUN clickhouse local -q 'SELECT 1' >/dev/null 2>&1 && exit 0 || : \
         /var/lib/apt/lists/* \
         /var/cache/debconf \
         /tmp/* \
-    && apt-get autoremove --purge -yq dirmngr gnupg2
+    && apt-get autoremove --purge -yq dirmngr gnupg2 \
+    && chmod ugo+Xrw -R /etc/clickhouse-server /etc/clickhouse-client
+# The last chmod is here to make the next one is No-op in docker official library Dockerfile
```

I checked the `docker history`, and with this change the layer has 0B of diff

```
> docker history --no-trunc clickhouse-server:latest | grep chmod
<missing>                                                                 5 seconds ago   RUN |10 DEBIAN_FRONTEND=noninteractive apt_archive=http://archive.ubuntu.com REPO_CHANNEL=stable REPOSITORY=deb [signed-by=/usr/share/keyrings/clickhouse-keyring.gpg] https://packages.clickhouse.com/deb stable main VERSION=24.10.1.2812 PACKAGES=clickhouse-client clickhouse-server clickhouse-common-static deb_location_url= DIRECT_DOWNLOAD_URLS= single_binary_location_url= TARGETARCH=amd64 /bin/sh -c clickhouse-local -q 'SELECT * FROM system.build_options'     && mkdir -p /var/lib/clickhouse /var/log/clickhouse-server /etc/clickhouse-server /etc/clickhouse-client     && chmod ugo+Xrw -R /var/lib/clickhouse /var/log/clickhouse-server /etc/clickhouse-server /etc/clickhouse-client # buildkit                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           0B        buildkit.dockerfile.v0
<missing>                                                                 5 seconds ago   RUN |10 DEBIAN_FRONTEND=noninteractive apt_archive=http://archive.ubuntu.com REPO_CHANNEL=stable REPOSITORY=deb [signed-by=/usr/share/keyrings/clickhouse-keyring.gpg] https://packages.clickhouse.com/deb stable main VERSION=24.10.1.2812 PACKAGES=clickhouse-client clickhouse-server clickhouse-common-static deb_location_url= DIRECT_DOWNLOAD_URLS= single_binary_location_url= TARGETARCH=amd64 /bin/sh -c clickhouse local -q 'SELECT 1' >/dev/null 2>&1 && exit 0 || :     ; apt-get update     && apt-get install --yes --no-install-recommends         dirmngr         gnupg2     && mkdir -p /etc/apt/sources.list.d     && GNUPGHOME=$(mktemp -d)     && GNUPGHOME="$GNUPGHOME" gpg --batch --no-default-keyring         --keyring /usr/share/keyrings/clickhouse-keyring.gpg         --keyserver hkp://keyserver.ubuntu.com:80         --recv-keys 3a9ea1193a97b548be1457d48919f6bd2b48d754     && rm -rf "$GNUPGHOME"     && chmod +r /usr/share/keyrings/clickhouse-keyring.gpg     && echo "${REPOSITORY}" > /etc/apt/sources.list.d/clickhouse.list    && echo "installing from repository: ${REPOSITORY}"     && apt-get update     && for package in ${PACKAGES}; do         packages="${packages} ${package}=${VERSION}"     ; done     && apt-get install --yes --no-install-recommends ${packages} || exit 1     && rm -rf         /var/lib/apt/lists/*         /var/cache/debconf         /tmp/*     && apt-get autoremove --purge -yq dirmngr gnupg2     && chmod ugo+Xrw -R /etc/clickhouse-server /etc/clickhouse-client # buildkit   517MB     buildkit.dockerfile.v0
```

PR #15861, opened 2023-12-08
[varnish] drop alpine tag

gquintard:
see https://github.com/docker-library/official-images/pull/15852#issuecomment-1845912882

### DIFF
--- a/library/varnish
+++ b/library/varnish
@@ -1,4 +1,4 @@
-# this file was generated using https://github.com/varnish/docker-varnish/blob/0c295b528f28a98650fb2580eab6d34b30b165c4/populate.sh
+# this file was generated using https://github.com/varnish/docker-varnish/blob/5644d31e8fa5bc78e764e8604ffa17c47528f5e3/populate.sh
 Maintainers: Guillaume Quintard <guillaume.quintard@gmail.com> (@gquintard)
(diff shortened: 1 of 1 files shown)


whalelines:
Seems you are updating the base images used in the alpine tag, not dropping it, no?

gquintard:
@whalelines , sorry, I don't understand what you mean, we are doing this:

```
-FROM alpine:3.15
+FROM alpine
```

which is dropping the tag of the base images

tianon:
I would still recommend pinning that to a specific base image -- sorry, I think my message must've been unclear.  I meant either remove the `varnish` tag that uses that image or update it, and updating it is totally valid.  I'd suggest pinning simply so that you don't get unexpected build failures when the "square wheel" moves and Alpine updates (so as of yesterday, that'd be `FROM alpine:3.19`).

whalelines:
@gquintard , I read the title of this PR as removing the alpine tag from the varnish image, not removing the tag on the alpine base image, which, as @tianon mentions, is not a good idea.

gquintard:
ah, thanks @whalelines , I see how my title was confusing, sorry about that.

I've updated the PR to use `3.19` now.

By the way, I'm stealing the square wheel image, it's brilliant.

PR #15871, opened 2023-12-08
[ros] retire EOL images builds

mikaelarguedas:
follow-up of https://github.com/docker-library/official-images/pull/15809

Waiting for images rebuild to submit this PR as ready for merging


tianon:
(up to you whether you'd like us to merge this with or without investigating that first)

mikaelarguedas:
I couldn't find the history for those builds. Did the dashing arm32 builds pass before ?

---
I opened a ticket for the armhf missing dashing packages https://github.com/osrf/infrastructure/issues/18 but I think we can merge as is

Thanks!

tianon:
Looking at https://hub.docker.com/r/arm32v7/ros/tags?name=dashing, it seems they built successfully ~3 years ago, FWIW :sweat_smile:

(We definitely don't keep build logs that long :see_no_evil:)

PR #15873, opened 2023-12-09
haxe: add alpine3.19, remove alpine3.15

andyli:
Re. https://github.com/docker-library/official-images/pull/15852

### DIFF
--- a/library/haxe
+++ b/library/haxe
@@ -26,7 +26,12 @@ GitCommit: 52a5f2360f84687908f3d80be326c810875c9be6
 Directory: 4.3/windowsservercore-1809
 Constraints: windowsservercore-1809
 
-Tags: 4.3.3-alpine3.18, 4.3-alpine3.18, 4.3.3-alpine, 4.3-alpine
+Tags: 4.3.3-alpine3.19, 4.3-alpine3.19, 4.3.3-alpine, 4.3-alpine
+Architectures: amd64, arm64v8
+GitCommit: b7f47ba7340373d40202dfb77621de1ed3ba3677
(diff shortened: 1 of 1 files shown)


tianon:
Thank you!

yosifkit:
Just noticed that the `3.1` Windows images have been left around and don't have a corresponding Linux variant on their shared tags. This means that the `3.1.3` and `3.1` tags only have Windows images (so they can't be pulled on Linux). This isn't a problem but might be unexpected for users trying to use the old image (they can still pull the `haxe:3.1*-stretch` images). The `-stretch` images were dropped in https://github.com/docker-library/official-images/pull/12914 when `debian:stretch` reached end of life.

Is `3.1.3` (or any release other than the most recent) still a supported release? I.E., would there be minor bugfix version bump, like `3.1.4` or `4.2.6`, at some point?

We recommend removing versions that are no longer actively supported. Removing it here will remove it from the "Supported" section on the Hub readme (and will prevent us from spending cycles rebuilding it on the official build server), but the tag will still be available to users who want it.

andyli:
There are no clearly-defined supported haxe versions. By past experience, it's rare to see bug fix releases to non-latest versions.

There are `x.x_bugfix` branches in the compiler repo, but are only used when the next "major" release is in development. For example, currently the team started working on 5.x on the main branch, a `4.3_bugfix` branch was created for making 4.3.x bugfix releases before 5.0.0 is released.  Once 5.0.0 is released, it is very likely no more 4.3.x releaes will be made.

I guess dropping 3.1, and probably 4.0 and 4.1 too, from docker hub versions is fine to do.

PR #15878, opened 2023-12-11
Remove Kong 3.0 due to its EOL

AndyZhang0707:
Kong 3.0 is EOL thus remove it from the "supported" list.

This is a PR to follow up a discussion in #15686


tianon:
Nice, thank you! :+1:

PR #15881, opened 2023-12-11
Fix github action for too long diffs

Felixoid:
In #15846 the issue with `too long body` appeared. It prevents the service comment from being posted.

Here's the fix to truncate too-long diff.


tianon:
Unfortunately this is somewhat intentional -- the generated diff is what we actually review, so if it's truncated, we have to generate it via other means and the failing CI is our signal to do so.  We also use it as a rough "code smell" (because a "diff too long" in a Dockerization is usually a sign that something else has gone wrong; new image PRs with lots of versions are a bit of an exceptional case).

PR #15894, opened 2023-12-12
Update ELK to 8.11.3, 7.17.16, backfill missed 8.11.2

tianon:
Follow up to https://github.com/docker-library/official-images/pull/15808

@mark-vieira @watson @jsvd (@alpar-t) this is a good example of what a "bump" PR would gemerally look like (and this one is an extra fun example because it includes a back-fill for the missed 8.11.2 version -- after that's built successfully once, we'll want to remove it from the list so it doesn't show up on Hub in the "supported tags" list)

### DIFF
--- a/library/elasticsearch
+++ b/library/elasticsearch
@@ -3,12 +3,17 @@ GitRepo: https://github.com/elastic/dockerfiles.git
 Directory: elasticsearch
 Builder: buildkit
 
-Tags: 8.11.1
+Tags: 8.11.3
 Architectures: amd64, arm64v8
 GitFetch: refs/heads/8.11
-GitCommit: 0dbf0d4297a2bf48d0cd980935431c121396885d
+GitCommit: 1484679916b52a402f0d02ea290d60f6e1eca39a
 
-Tags: 7.17.15
+Tags: 8.11.2
+Architectures: amd64, arm64v8
+GitFetch: refs/heads/8.11
+GitCommit: 58aa6df8045922c7c6ac92055a120365e158dd20
+
+Tags: 7.17.16
 Architectures: amd64, arm64v8
 GitFetch: refs/heads/7.17
-GitCommit: 63269da8548948c20f3e1650cebbed1ae5eee90e
+GitCommit: f3f1c976063b3549ed75e18aa99d06fdf51f8857
--- a/library/kibana
+++ b/library/kibana
@@ -3,12 +3,17 @@ GitRepo: https://github.com/elastic/dockerfiles.git
 Directory: kibana
 Builder: buildkit
 
-Tags: 8.11.1
+Tags: 8.11.3
 Architectures: amd64, arm64v8
 GitFetch: refs/heads/8.11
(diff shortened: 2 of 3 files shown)


mark-vieira:
Do we need to open one of these then on every release? I thought there was already some automation in place here @alpar-t?

watson:
>  after that's built successfully once, we'll want to remove it from the list so it doesn't show up on Hub in the "supported tags" list)

What list are you referring to?

whalelines:
@watson wrote:
> What list are you referring to?

The "Supported tags" list in the repository overview pages on Docker Hub.

-   https://hub.docker.com/_/elasticsearch
-   https://hub.docker.com/_/logstash
-   https://hub.docker.com/_/kibana

When this PR gets merged and deployed, both 8.11.3 _and_ 8.11.2 will be listed there, since they both appear each repository's bashbrew file. Since only 8.11.3 is current and 8.11.2 is only included to backfill it since that version update was previously missed, once 8.11.2 is built and deployed to Hub, another PR should be created removing that entry from the bashbrew files, which will also remove it from the repository overviews on Docker Hub. This will help reduce confusion for the image users.

-   https://github.com/docker-library/official-images/pull/15894/files#diff-613e109b2d9819643f54e909f063fcea7d3b874624b8041735e23b7b6b58454fR11-R15
-   https://github.com/docker-library/official-images/pull/15894/files#diff-2de8d3da84f693b929961ece8bc096f28b8d7b8be2b0a93fa3fbbd89d5dd2143R11-R15
-    https://github.com/docker-library/official-images/pull/15894/files#diff-5ee7c96aaa83c26a51808d146b9f5a39649d7a32ea1b24581ca213129b38026bR11-R15

alpar-t:
> Do we need to open one of these then on every release? I thought there was already some automation in place here @alpar-t?

I'm not aware of a change that would require us to do so. AFAIK we automatically raise a PR such as https://github.com/elastic/dockerfiles/pull/162 and create a tag such as https://github.com/elastic/dockerfiles/tree/v8.11.2 my understanding is that automation from the docker team takes over from there. 

Is any of this changing now ?

tianon:
Yes, the main change in https://github.com/docker-library/official-images/pull/15808 is that we'll need PRs here for each release (which we needed before too, but we handled those because they were going through our janky https://github.com/docker-library/elasticsearch and friends repositories).  However, as I've written these files right now, they're pulling the references from the branches instead of the tags on that Elastic GitHub repository -- I'm not sure if that makes your life any easier (not having to create those tags anymore if you don't want to, which I understand has been a bit of a hassle in the past)?

alpar-t:
Thanks for the clarification! I'll make sure to update the release process on our side.

tianon:
:heart:

I think we're good to go :eyes:

I'll leave the backfilled 8.11.2 for now in case you want to test some automation (so there's a reasonable change that can be made) or practice making the update PR. :+1: :heart:

mark-vieira:
Thanks @tianon!

PR #15929, opened 2023-12-18
Updated swipl devel to 9.1.21

JanWielemaker:
Also updated the base image from Debian bullseye to Debian bookworm.

The repo also contains the updated bullseye Dockerfile.   Should we add that as an alternative tag?   What is the best practice wrt. updating the Debian version of the base image?   If this is not according to your best practices, please either edit the commit or reject and point me at what to do.


JanWielemaker:
The tests fails because library(space) does not exist.    That is correct.  This library is, at least for now, disabled for bookworm because libgeos-dev is broken on bookworm.    I think this is acceptable (I think usage is low, but I could of course be wrong). 

I'll submit new patch with a better quick test.

PR #15930, opened 2023-12-18
Updated swipl devel to 9.1.21

JanWielemaker:
Also updated the base image from Debian bullseye to Debian bookworm and makes test.sh + test.pl test all plugins and dynamically adjust to the available ones.

2nd try because test.sh failed due to dropping the `space` plugin.   This is dropped because
Bookworm `libgeos-dev` is broken.    I do not expect this to affect many people (but one never knows).

The repo also contains the Bullseye Dockerfile for 9.1.21.    I do not know the best practices for updating the Debian base image.    Please adjust as you seem fit or reject, pointing how this should be done.


JanWielemaker:
Oops, still failing.   Now I am lost as we see the error below, but I can't find any references to a test `swipl-modules` in my repo.   Is there some test script that is stored and maintained elsewhere?

```
ERROR: -g test: source_sink `library(space/space)' does not exist
	'swipl-modules' [4/4]...failed
```

tianon:
Yep, that'd be https://github.com/docker-library/official-images/blob/master/test/tests/swipl-modules/container.pl :eyes:

(Please feel free to just force push to this PR -- that'll re-run the CI :+1:)

tianon:
Nice, thank you! :heart:

PR #15936, opened 2023-12-19
[ros][gazebo] remove apt-key use in favor of recommended gpg usage

mikaelarguedas:
Following recommendations from https://github.com/docker-library/official-images/pull/15809#issuecomment-1846053036

### DIFF
--- a/library/gazebo
+++ b/library/gazebo
@@ -9,7 +9,7 @@ GitRepo: https://github.com/osrf/docker_images.git
 
 Tags: gzserver11, gzserver11-focal
 Architectures: amd64
-GitCommit: 532c4f1af8c9dadb54b0c9769543c67c40c0c84f
+GitCommit: 27cc0b68263bbbb10bb58dd814efc0a6b0a01ec7
 Directory: gazebo/11/ubuntu/focal/gzserver11
 
 Tags: libgazebo11, libgazebo11-focal, latest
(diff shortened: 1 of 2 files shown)


yosifkit:
I think putting the key into `/usr/share/keyrings/` is a little odd, since that is where a `deb` package would put self-managed keys. Otherwise, looks good to me.

tianon:
https://manpages.debian.org/unstable/apt/apt-key.8.en.html#:~:text=Since%20APT%202.4%2C%20/etc/apt/keyrings%20is%20provided%20as%20the%20recommended%20location%20for%20keys%20not%20managed%20by%20packages. :eyes:

> Since APT 2.4, /etc/apt/keyrings is provided as the recommended location for keys not managed by packages.

(in other words, `/etc/apt/keyrings` is the officially APT-recommended location for non-package-maintained keyrings)

mikaelarguedas:
@yosifkit @tianon I agree with your reading of the docs.
I did is this way to match what was in the installation instructions of the projects. From https://github.com/osrf/docker_images/pull/705#discussion_r1431154106 I understand that the fact that it is managed by hand is temporary and will be handled by a specific package in the future that's why the location is already the "final one".

If this inconsistency is problematic I can modify this to install to `/etc/apt` for now and do a separate PR once the management of the keys by a package is implemented

yosifkit:
Thanks @mikaelarguedas, given your explanation `/usr/share/keyrings/` is fine.

mikaelarguedas:
> Thanks mikaelarguedas, given your explanation /usr/share/keyrings/ is fine.

Sounds good :+1:
Rebased and took this out of draft state to be ready for review / merge

tianon:
It's all good!  We're thrilled when the answer to our questions ends up being "we already thought about, discussed, and made an intentional choice on exactly that point" :smile: :heart:

PR #15938, opened 2023-12-19
add a new registry pre-release

milosgajdos:
Not sure this is the right way to deal with a new major pre-release but I followed https://github.com/docker-library/official-images/pull/11718 so let see 😄

### DIFF
--- a/library/registry
+++ b/library/registry
@@ -1,10 +1,10 @@
 Maintainers: Milos Gajdos (@milosgajdos),
              Sebastiaan van Stijn (@thajeztah)
 GitRepo: https://github.com/docker/distribution-library-image.git
+GitFetch: refs/heads/master
 
 Tags: 3.0.0-alpha.1
 Architectures: amd64, arm64v8, arm32v6, arm32v7, ppc64le, s390x
-GitFetch: refs/heads/master
 GitCommit: 6356e8dbe1969f7340ca25a9874faa26e6e6f71d
 
 Tags: 2.8.3, 2.8, 2, latest
--- a/library/registry
+++ b/library/registry
@@ -1,8 +1,11 @@
 Maintainers: Milos Gajdos (@milosgajdos),
              Sebastiaan van Stijn (@thajeztah)
 GitRepo: https://github.com/docker/distribution-library-image.git
+
+Tags: 3.0.0-alpha.1
+Architectures: amd64, arm64v8, arm32v6, arm32v7, ppc64le, s390x
 GitFetch: refs/heads/master
-GitCommit: 39dd72feaab7066334829d6945c54bc51a0aee98
+GitCommit: 6356e8dbe1969f7340ca25a9874faa26e6e6f71d
 
 Tags: 2.8.3, 2.8, 2, latest
 Architectures: amd64, arm64v8, arm32v6, arm32v7, ppc64le, s390x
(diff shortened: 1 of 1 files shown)


milosgajdos:
Ok, this is breaking `v2` builds so yeah maybe @tianon can give us a hand 😄

library/registry
```diff
@@ -1,8 +1,11 @@
 Maintainers: Milos Gajdos (@milosgajdos),
              Sebastiaan van Stijn (@thajeztah)
 GitRepo: https://github.com/docker/distribution-library-image.git
+
+Tags: 3.0.0-alpha.1
+Architectures: amd64, arm64v8, arm32v6, arm32v7, ppc64le, s390x
 GitFetch: refs/heads/master
```
tianon, line 7: This should leave `GitFetch:` up in the global block (so it applies to all entries in the file):

```suggestion
GitFetch: refs/heads/master

Tags: 3.0.0-alpha.1
Architectures: amd64, arm64v8, arm32v6, arm32v7, ppc64le, s390x
```

library/registry
```diff
@@ -1,8 +1,11 @@
...
-GitCommit: 39dd72feaab7066334829d6945c54bc51a0aee98
+GitCommit: 6356e8dbe1969f7340ca25a9874faa26e6e6f71d
 
 Tags: 2.8.3, 2.8, 2, latest
 Architectures: amd64, arm64v8, arm32v6, arm32v7, ppc64le, s390x
```
tianon, line 11: And then, since you want to still include 2.x (for now), it'll still need an appropriate `GitCommit:` value (that points at the 2.x `Dockerfile`): :heart:

```suggestion
Architectures: amd64, arm64v8, arm32v6, arm32v7, ppc64le, s390x
GitCommit: 39dd72feaab7066334829d6945c54bc51a0aee98
```

tianon (commented):
Ok, left some "GitHub Suggestions" that hopefully make it clear how I'd suggest fixing the file to match your intent :bow:

Do you think there will be any more 2.x releases in the future, or is it 3.x only from here out?  (If you think there _will_ be 2.x releases in the future, you might want to restructure the repo so you have a `2/Dockerfile` and `3/Dockerfile` so it's easier for you to maintain both concurrently later, but if not, then you're fine and the next release will just drop the alpha and go back down to one entry in this file since 2.x will effectively be EOL :+1:)

library/registry
```diff
@@ -2,7 +2,11 @@ Maintainers: Milos Gajdos (@milosgajdos),
              Sebastiaan van Stijn (@thajeztah)
 GitRepo: https://github.com/docker/distribution-library-image.git
 GitFetch: refs/heads/master
-GitCommit: 39dd72feaab7066334829d6945c54bc51a0aee98
+
+Tags: 3.0.0-alpha.1
```
milosgajdos, line 6: @tianon should this has `latest` tag instead of keeping it with the `v2` release? I think it should 🤔 cc @thaJeztah

tianon: Depends on what you want users to "use by default" if they don't know any better -- putting that on an Alpha seems a bit ambitious to me, but you're in a better position to know whether 3.0.0-alpha.1 is suitable for general use than I am. :smile:

tianon: (in other words, "latest" actually means "default", not literally "the newest one")

milosgajdos: Let me chat to @thaJeztah first 😄 my thinking is: if we don't mark it as `latest` it will have less of a chance to get properly tested; folks can always pick an explicit `v2` tag if they wish to do so. BRB, don't merge yet please 😄

thaJeztah: We should probably wait until GA, I don’t think we usually ship pre-releases as “latest”.

Pre-releases are for testing, but should be opt-in

milosgajdos: ok, @tianon, we're sticking to your original suggestion 😄 `latest` = latest stable which is still `2.x`. Merge away please 🙇‍♂️

PR #15943, opened 2023-12-19
Update docker

yosifkit:
Changes:

- https://github.com/docker-library/docker/commit/3a04248: (temporarily) Revert "Add a "docker" group (with GID 2375)"
- https://github.com/docker-library/docker/commit/8f1087c: Update 25-rc to 25.0.0-beta.3
- https://github.com/docker-library/docker/commit/9238e26: Merge pull request https://github.com/docker-library/docker/pull/462 from infosiftr/docker-group

### DIFF
--- a/library/docker
+++ b/library/docker
@@ -5,45 +5,45 @@ Maintainers: Tianon Gravi <tianon@dockerproject.org> (@tianon),
 GitRepo: https://github.com/docker-library/docker.git
 Builder: buildkit
 
-Tags: 25.0.0-beta.2-cli, 25-rc-cli, rc-cli, 25.0.0-beta.2-cli-alpine3.19
+Tags: 25.0.0-beta.3-cli, 25-rc-cli, rc-cli, 25.0.0-beta.3-cli-alpine3.19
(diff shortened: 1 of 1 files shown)


tianon:
Hmm, https://github.com/docker-library/docker/pull/462 is a good change, but I do worry about merging it _right_ before our holiday break when we won't even see any breakage until 2024.

@thaJeztah do you have an opinion on whether we should just delay releasing the beta 3 image until 2024 or whether we should temporarily revert the `docker` group to get the pre-release out?

thaJeztah:
I'm good with temporarily reverting the `docker` group; it was not critical in any way. It'd be nice to have the beta out (which should not affect any production use), allowing people to give it a try

tianon:
Ok, I've applied https://github.com/docker-library/docker/commit/3a0424834732f20142291cdd1d392e6d860af2ad for now so we can get this in.  Since you're planning at least one RC before GA in (hopefully) Jan, I'll leave that as-is for now, and will plan to revert that revert after the holiday season.

thaJeztah:
Thank you both!! And Happy Holidays!

PR #15949, opened 2023-12-20
[oraclelinux] Updating 8, 8-slim and 8-slim-fips for ELSA-2023-7877

amsteinb-orcl:
This update incorporates fixes for CVE-2023-3446, CVE-2023-3817, CVE-2023-5678

https://linux.oracle.com/errata/ELSA-2023-7877.html

### DIFF
--- a/library/oraclelinux
+++ b/library/oraclelinux
@@ -4,10 +4,10 @@ GitCommit: 5c8a1c296acd6e90487cd261d16cf85fd6bcb73f
 GitFetch: refs/heads/master
 # https://github.com/oracle/container-images/tree/dist-amd64
 amd64-GitFetch: refs/heads/dist-amd64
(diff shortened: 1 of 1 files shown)


tianon:
Are all those added binaries expected (unfortunate/unexpected/uncontrollable dependency changes for example) or a glitch in the minimization? (`modprobe`, LVM related tools, etc)

amsteinb-orcl:
I will need to investigate why the new packages were pulled in.  The easy answer would be "the dependencies changed", but I would have to confirm that.

amsteinb-orcl:
I found that indeed the systemd package added several dependencies which would account for the additional packages to be added.

tianon:
Ok, thanks for double checking! :heart:

(I know minifying these as much as possible was a pretty herculean/concerted effort, so I wanted to make sure regressions in that are intentional/unavoidable or at least known/noted :heart:)

PR #15967, opened 2023-12-23
elixir 1.16.0

getong:


tianon:
Given that 1.11 and older are all now out of even security support, they should probably be dropped from the file. :eyes:

PR #15969, opened 2023-12-25
erlang/otp 26.2.1

getong:


tianon:
Now that they're out of security support, we should remove versions 23 and lower from the file :eyes:

PR #15984, opened 2024-01-01
archlinux: Release 20240101.0.204074

archlinux-github:

---

### DIFF
--- a/library/archlinux
+++ b/library/archlinux
@@ -5,13 +5,18 @@ Maintainers: Santiago Torres-Arias <santiago@archlinux.org> (@SantiagoTorres),
              Justin Kromlinger <hashworks@archlinux.org> (@hashworks)
 GitRepo: https://gitlab.archlinux.org/archlinux/archlinux-docker.git
 
-Tags: latest, base, base-20231112.0.191179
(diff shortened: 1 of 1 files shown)


tianon:
Merging per https://github.com/docker-library/official-images/pull/15751 discussion 👍

aminvakil:
@hashworks @tianon Thank you for doing this, I can also confirm issue has been fixed.
https://github.com/aminvakil/docker-archlinux/commit/dbd06a2

tianon:
We're probably due for another update sometime soon, right? :sweat_smile:

hashworks:
Sadly I hadn't found time to adjust the deployment to make base-devel and multilib extend from base. I can re-enable the old variant anytime if you think that's okay.

tianon:
Ah yeah, that's fair -- I guess having them up-to-date again is better than making sure they're perfect.  I'm looking through the old discussions and not actually understanding why this still included the new `multilib-devel` (maybe we just didn't notice in our review of it?), but perhaps that's not actually that important.  Please do feel free to re-enable :heart:

tianon:
(as a reminder for myself, that improvement is tracked in https://gitlab.archlinux.org/archlinux/archlinux-docker/-/issues/100 :rocket:)

PR #16022, opened 2024-01-08
Rename CODE-OF-CONDUCT.md to CODE-OF-Canceled.md

Umbrella-Pad:
Echo svg-xaml.wpa


yosifkit:
This seems like spam.

PR #16035, opened 2024-01-10
Transfer library maintenance to the Apache Storm Project

rzo1:
The Apache Storm project will be taking over maintenance of the official `storm` image as per https://lists.apache.org/thread/34kkzbcfr4qxcxmclgqmjc08o9s28gjz from the previous maintainer @31z4

The mail goes to the project's developer list responsible for releases and administration. I also added @jnioche and myself from the Storm PMC.

In the long run, we want to publish the images from the Storm PMC under `apache/storm` on DockerHub. However, (for now) we just want to take control over the publishing of this official image, which wasn't maintained for the last few months.

Transferring the repo (instead of forking) to the ASF organization on GitHub would be much more overhead compared to loosing the few issues / prs on the original repo. I hope, that this is ok because we want to switch to `apache/storm` on DockerHub in the near future anyway.

ASF Infra told me, that a redirect from `storm` to `apache/storm` on DockerHub isn't feasable / possible at the moment, so perhaps best to get control here, put up an image for 2.6.0 and than update the README on DockerHub and start distributing under `apache/storm`. Thoughts are welcome, though.

### DIFF
--- a/library/storm
+++ b/library/storm
@@ -1,5 +1,7 @@
-Maintainers: Elisey Zanko <elisey.zanko@gmail.com> (@31z4)
-GitRepo: https://github.com/31z4/storm-docker.git
+Maintainers: The Apache Storm Project <dev@storm.apache.org> (@asfbot),
+             Julien Nioche (@jnioche),
+             Richard Zowalla (@rzo1)
+GitRepo: https://github.com/apache/storm-docker.git
 Architectures: amd64, arm64v8
 
 Tags: 1.2.4-temurin, 1.2-temurin


yosifkit (approved):

rzo1:
Thanks for the reviews and links to the repository for the documentation (readme). We will have a look and adjust it accordingly. Also we should get rid of the EOL versions soon + provide updated images for 2.6.0

> It is a little strange to take over the image maintenance to plan to deprecate it, but it is understandable. 

To be honest, there are ongoing discussions inside the ASF how to deal with the branding for other distribution sources like Dockerhub and if there should be policies to force projects to do `apache/xy`. But this is still a very early thing, so I think, that it is (for now) fine to use the official distribution way (via `storm`). It is also a good way to maintain backward compat for now.

rzo1:
> I guess this also means https://issues.apache.org/jira/browse/STORM-1836 can finally close? 😄 ❤️

Yes ;-)

rzo1:
Shall we add 2.6.0 in a separate PR or should we add it to this PR too (so people get the latest version)? @jnioche will prepare a PR for the docs update to point to the correct repo, etc.

tianon:
Let's get this in and do a follow-up with the latest version :+1: :heart:

(Feel free to pass on Britney's email to any of the other folks in the ASF having discussions about this -- she'd be really happy to discuss it, especially in the context of DVP vs DOI, moving between them, and branding around all that, including whether/how Docker Hub could make changes to improve that for DOI, for example :bow:)

PR #16040, opened 2024-01-11
Update ELK Images

navyau09:


tianon:
Nice (and welcome)!

This should also drop `8.11.3` from the files (since 8.11.3 won't be updated anymore, only 8.11.4 is "supported" -- if you're hand-editing these, that's easiest to accomplish by modifying the existing 8.x or 7.x entry instead of adding a new one when updating :+1:).

yosifkit (approved):

PR #12406, opened 2022-05-11
Apache Solr - Use docker-library/official-images for PRs (#3)

docker-solr-builder:
Apache Solr:

Use docker-library/official-images for PRs (#3)

Also add mike as a maintainer

Relevant Maintainers: @HoustonPutman @janhoy @dsmiley @madrob

This is an automated PR based on a change made in the https://github.com/apache/solr-docker repository's `main` branch.

PR #16065, opened 2024-01-16
Replace EOL versions (1.2.4, 2.4.0, 2.5.0) with current release of Apache Storm (2.6.0)

rzo1:
This PR removes the EOL versions of Apache Storm and replaces them with the current release, which is 2.6.0.


tianon:
See https://github.com/docker-library/official-images/pull/12406#issuecomment-1124393213, especially:

> If you are going to download a keys file, the line needs to somehow use the expected fingerprint ids to verify that it is only using the expected keys. This will also provide a natural `docker build` cache bust of the line being changed when a new key is needed.

What I would suggest is some automation to download that file and embed all the key IDs from it directly in the `Dockerfile` (and then using something like `gpg --export` with that list to make sure it keeps only those keys, or a loop with `--recv-keys` to ensure they're all fetched from a keyserver).

See https://github.com/docker-library/cassandra/blob/a1e2446ab30903f6a21e1c06831a45ab763f2719/Dockerfile.template#L72-L99 for an example of what I mean. :bow:

tianon:
(also, any use of `curl` for downloading should include at least `-f`/`--fail` to ensure it fails appropriately if the download does)

rzo1:
Thx for review. I will adjust it similar to how it is done for TomEE atm: https://github.com/tomitribe/docker-tomee/tree/master

rzo1:
@tianon Should be better now.

tianon (approved):

tianon:
Thank you!

PR #16082, opened 2024-01-18
Fix BuildKit buikld for GitRepo without .git suffix

LaurentGoderre:


tianon:
I'm still of the opinion that this is a bug in either our tooling or BuildKit + Docker CLI/Engine, since `git clone` works just fine without the `.git` suffix. :sweat_smile:

tianon:
(in other words, I think https://github.com/docker-library/meta-scripts/pull/15 is a very sane short and medium -term solution)

LaurentGoderre:
It was still valuable as a backup

library/ubuntu
```diff
@@ -2,7 +2,7 @@
 # see also https://wiki.ubuntu.com/Releases#Current
 
 Maintainers: Tomáš Virtus <tomas.virtus@canonical.com> (@woky), Cristóvão Cordeiro <cristovao.cordeiro@canonical.com> (@cjdcordeiro)
-GitRepo: https://git.launchpad.net/cloud-images/+oci/ubuntu-base
+GitRepo: https://git.launchpad.net/cloud-images/+oci/ubuntu-base.git
```
tianon, line 5: This one is actually the example URL I gave to the Moby folks internally -- Launchpad canonically does not use `.git` and does not support it (hence the CI failures on this PR :smile:)

tianon:
https://github.com/docker-library/meta-scripts/pull/15 was successful :tada: :heart:

PR #16096, opened 2024-01-19
Update Joomla!

Llewellynvdm:
Changes:

- joomla-docker/docker-joomla@6407e1b: Updated all the images with the permission fix.
- joomla-docker/docker-joomla@d90c4f6: Fix [joomla-docker/docker-joomla#174](https://github.com/joomla-docker/docker-joomla/issues/174) to correct the `configuration.php` ownership and permissions.

### DIFF
--- a/library/joomla
+++ b/library/joomla
@@ -6,90 +6,90 @@ GitRepo: https://github.com/joomla-docker/docker-joomla.git
 
 Tags: 5.0.2-php8.1-apache, 5.0-php8.1-apache, 5-php8.1-apache
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
-GitCommit: 6407e1befb107d065ba1e6c73c07145220aaa787
+GitCommit: a416792606b1f49dfbe9ad2155c7df972ade09d3
 Directory: 5.0/php8.1/apache
 
 Tags: 5.0.2-php8.1-fpm-alpine, 5.0-php8.1-fpm-alpine, 5-php8.1-fpm-alpine
 Architectures: amd64, arm32v6, arm32v7, arm64v8, i386, ppc64le, s390x
-GitCommit: 6407e1befb107d065ba1e6c73c07145220aaa787
+GitCommit: a416792606b1f49dfbe9ad2155c7df972ade09d3
 Directory: 5.0/php8.1/fpm-alpine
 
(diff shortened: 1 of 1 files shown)


tianon:
You'll probably want at _least_ an "are we root?" check around this, like you've got further up in the entrypoint:

```bash
                if [ "$uid" = '0' ] && [ "$(stat -c '%u:%g' .)" = '0:0' ]; then
                        chown "$user:$group" .
                fi
```

You might also want to consider what you'll do if the `chmod` or `chown` fails, since that'll be a common scenario for setups using SELinux (in our experience).

I'm also not sure how we didn't catch it before (I blame long scripts :sweat_smile: :innocent:), but your `$?` check will probably never trigger since the script uses `set -e` -- I would suggest moving whatever command you're testing straight into the `if`, like this:

```bash
if php ...; then
```

Llewellynvdm:
Thank you for your insightful feedback, especially regarding the checks and error handling in the script. I understand your concern about the `$?` check potentially not triggering due to the `set -e` configuration. However, from my testing, it seems that `[ $? -eq 0 ]` does reliably trigger as expected. I believe this might be due to the specific context in which it's used, but I'll revisit this to ensure there's no oversight.

Regarding the file ownership issue, I agree that directly addressing the root cause is more effective. Instead of changing file permissions post-creation, running the PHP command as the intended user seems to be a cleaner solution:

```bash
sudo -u "$user" php installation/joomla.php install "${installJoomlaArgs[@]}"
```

This approach should prevent the configuration file from being created with root ownership in the first place. Of course, I'll be mindful of any security implications and compatibility issues this change might introduce, especially in diverse Docker environments.

I plan to conduct thorough testing on this and would appreciate any additional insights or concerns you might have, particularly regarding compatibility with SELinux or other specific Docker setups. Your experience in these matters is extremely valuable.

Looking forward to your thoughts on this approach.

Llewellynvdm:
Hmm the sudo option is not working for me... painful. I will continue down the initial path with the suggested changes.

Llewellynvdm:
Okay here is what seems to work as expected:
```
if [ "$uid" = '0' ] && [ "$(stat -c '%u:%g' configuration.php)" != "$user:$group" ]; then
    chown "$user:$group" configuration.php
    chmod 444 configuration.php
fi
```
Unless I am still missing something.... I will push out this update, let me know.

yosifkit (approved):
It all looks good to me!

PR #16099, opened 2024-01-19
Update bashbrew to 0.1.10

tianon:
https://github.com/docker-library/bashbrew/releases/tag/v0.1.10


tianon:
With this, we _could_ update `diff-pr.sh` to use `bashbrew context` but I didn't want to get too far ahead of myself (and that's not exactly an urgent change and it's one that's potentially disruptive and also complicates future plans for the `bashbrew context` command).

tianon:
Doh, I need https://github.com/docker-library/bashbrew/pull/90 too :facepalm:

PR #16105, opened 2024-01-22
releasing Neo4j 5.16.0

jennyowen:
We'd like to start releasing the redhat images based on ubi9 rather than ubi8 in the near future. Do you have an established best practice for doing that kind of deprecation?

The best way I can think to do it is to release the two images concurrently for a short while, and have a deprecation notice in the ubi8 image saying to update to ubi9. Would you be ok with me adding a notice like that in the image?

Thanks as always :slightly_smiling_face:

### DIFF
--- a/library/neo4j
+++ b/library/neo4j
@@ -4,25 +4,25 @@ Maintainers: Jenny Owen <jenny.owen@neo4j.com> (@jennyowen),
 GitRepo: https://github.com/neo4j/docker-neo4j-publish.git
 
 
-Tags: 5.15.0-community-bullseye, 5.15-community-bullseye, 5-community-bullseye, 5.15.0-community, 5.15-community, 5-community, 5.15.0-bullseye, 5.15-bullseye, 5-bullseye, 5.15.0, 5.15, 5, community-bullseye, community, bullseye, latest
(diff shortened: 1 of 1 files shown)


yosifkit:
>We'd like to start releasing the redhat images based on ubi9 rather than ubi8 in the near future. Do you have an established best practice for doing that kind of deprecation?
>
>The best way I can think to do it is to release the two images concurrently for a short while, and have a deprecation notice in the ubi8 image saying to update to ubi9. Would you be ok with me adding a notice like that in the image?

I think most of our practice revolves around providing tags that make it clear when it switches (like the `-ubi8` tag that you already have). We often maintain two releases of a distro at a time, but it is up to you how long you want to do that. I think a warning in the image could be good. Just make sure it doesn't print when running non-`neoj4` commands or it'll not pass the simple tests we run here (like [this](https://github.com/docker-library/official-images/actions/runs/7614090797/job/20735589788?pr=16105#step:9:14)).

tianon (approved):

tianon:
(I'll merge this for now -- I'll let you do any ubi9 or warnings in a follow-up :heart:)

jennyowen:
Thanks @yosifkit for the thoughtful response. It'll probably be next month at least by the time I'm ready to release any such changes.

> Just make sure it doesn't print when running non-neo4j commands or it'll not pass the simple tests we run here

Good to know, I would probably have made that mistake!

PR #16093, opened 2024-01-19
eclipse-temurin: Bump JDK11, JDK17 and JDK21

plebcity:
Bumped to 21.0.2, 17.0.10 and 11.0.22. 8u401 isn't available yet but we can merge this already so people can get the latest bug fixes, especially for JDK 21.

PR #16119, opened 2024-01-24
eclipse-temurin: January 2024 CPU updates

andrew-m-leonard:
January 2024 CPU updates


andrew-m-leonard:
Dup of: https://github.com/docker-library/official-images/pull/16093

PR #16141, opened 2024-01-26
Add Zulu Images from Azul systems

azul-publisher:
Documentation MR: https://github.com/docker-library/docs/pull/2320

### DIFF
--- a/library/azul-zulu
+++ b/library/azul-zulu
@@ -1,4 +1,4 @@
-Maintainers: Eduard Makedonskii <emakedonskii@azul.com> (@it-mak), Anton Kozlov <akozlov@azul.com> (@AntonKozlov), Sergey Grinev <sergey.grinev@azul.com> (@sgrinev)
+Maintainers: Anton Kozlov <akozlov@azul.com> (@AntonKozlov), Sergey Grinev <sergey.grinev@azul.com> (@sgrinev)
 GitRepo: https://github.com/zulu-openjdk/zulu-openjdk
 
 
@@ -96,4 +96,3 @@ Tags: 17-distroless, 17.0.8-distroless, 17.0.8-17.44.15-distroless
 Architectures: amd64
 GitCommit: a84ae34861c5e41fbb2f7c62d11abe3a18071c9f
 Directory: distroless/17-latest
-
--- a/library/azul-zulu-openjdk
+++ b/library/azul-zulu-openjdk
@@ -1,0 +1,99 @@
+Maintainers: Eduard Makedonskii <emakedonskii@azul.com> (@it-mak), Anton Kozlov <akozlov@azul.com> (@AntonKozlov), Sergey Grinev <sergey.grinev@azul.com> (@sgrinev)
+GitRepo: https://github.com/zulu-openjdk/zulu-openjdk
+
+
+#-----ubuntu-----
+
+Tags: 8-ubuntu, 8u382-ubuntu, 8u382-8.72.0.17-ubuntu
+Architectures: amd64, arm64v8
+GitCommit: 63034147c1ec9aab2f3b6aeced311c500f4586d9
+Directory: ubuntu/8-latest
+
+Tags: 11-ubuntu, 11.0.20-ubuntu, 11.0.20-11.66.15-ubuntu
+Architectures: amd64, arm64v8
+GitCommit: 63034147c1ec9aab2f3b6aeced311c500f4586d9
+Directory: ubuntu/11-latest
+
+Tags: 17-ubuntu, 17.0.8-ubuntu, 17.0.8-17.44.15-ubuntu
+Architectures: amd64, arm64v8
+GitCommit: 63034147c1ec9aab2f3b6aeced311c500f4586d9
+Directory: ubuntu/17-latest
+
+Tags: 20-ubuntu, 20.0.2-ubuntu, 20.0.2-20.32.11-ubuntu
+Architectures: amd64, arm64v8
+GitCommit: 63034147c1ec9aab2f3b6aeced311c500f4586d9
+Directory: ubuntu/20-latest
+
+#-----alpine-----
+
+Tags: 8-alpine, 8u382-alpine, 8u382-8.72.0.17-alpine
+Architectures: amd64, arm64v8
+GitCommit: 63034147c1ec9aab2f3b6aeced311c500f4586d9
+Directory: alpine/8-latest
+
+Tags: 11-alpine, 11.0.20-alpine, 11.0.20-11.66.15-alpine
+Architectures: amd64, arm64v8
+GitCommit: 63034147c1ec9aab2f3b6aeced311c500f4586d9
+Directory: alpine/11-latest
+
+Tags: 17-alpine, 17.0.8-alpine, 17.0.8-17.44.15-alpine
+Architectures: amd64, arm64v8
+GitCommit: 63034147c1ec9aab2f3b6aeced311c500f4586d9
+Directory: alpine/17-latest
+
+Tags: 20-alpine, 20.0.2-alpine, 20.0.2-20.32.11-alpine
+Architectures: amd64, arm64v8
+GitCommit: 63034147c1ec9aab2f3b6aeced311c500f4586d9
+Directory: alpine/20-latest
+
+#-----debian-----
+
+Tags: 8-debian, 8u382-debian, 8u382-8.72.0.17-debian
+Architectures: amd64, arm64v8
+GitCommit: 63034147c1ec9aab2f3b6aeced311c500f4586d9
+Directory: debian/8-latest
+
+Tags: 11-debian, 11.0.20-debian, 11.0.20-11.66.15-debian
+Architectures: amd64, arm64v8
+GitCommit: 63034147c1ec9aab2f3b6aeced311c500f4586d9
+Directory: debian/11-latest
+
+Tags: 17-debian, 17.0.8-debian, 17.0.8-17.44.15-debian
+Architectures: amd64, arm64v8
+GitCommit: 63034147c1ec9aab2f3b6aeced311c500f4586d9
+Directory: debian/17-latest
+
+Tags: 20-debian, 20.0.2-debian, 20.0.2-20.32.11-debian
+Architectures: amd64, arm64v8
+GitCommit: 63034147c1ec9aab2f3b6aeced311c500f4586d9
+Directory: debian/20-latest
+
+#-----centos-----
+
+Tags: 8-centos, 8u382-centos, 8u382-8.72.0.17-centos
+Architectures: amd64, arm64v8
+GitCommit: a84ae34861c5e41fbb2f7c62d11abe3a18071c9f
+Directory: centos/8-latest
+
+Tags: 11-centos, 11.0.20-centos, 11.0.20-11.66.15-centos
+Architectures: amd64, arm64v8
+GitCommit: a84ae34861c5e41fbb2f7c62d11abe3a18071c9f
+Directory: centos/11-latest
+
+Tags: 17-centos, 17.0.8-centos, 17.0.8-17.44.15-centos
+Architectures: amd64, arm64v8
+GitCommit: a84ae34861c5e41fbb2f7c62d11abe3a18071c9f
+Directory: centos/17-latest
+
+Tags: 20-centos, 20.0.2-centos, 20.0.2-20.32.11-centos
+Architectures: amd64, arm64v8
+GitCommit: a84ae34861c5e41fbb2f7c62d11abe3a18071c9f
+Directory: centos/20-latest
+
+#-----distroless----
+
+Tags: 17-distroless, 17.0.8-distroless, 17.0.8-17.44.15-distroless
+Architectures: amd64
+GitCommit: a84ae34861c5e41fbb2f7c62d11abe3a18071c9f
+Directory: distroless/17-latest
+
--- a/test/config.sh
+++ b/test/config.sh
@@ -20,6 +20,7 @@ testAlias+=(
 	[eclipse-temurin]='openjdk'
 	[sapmachine]='openjdk'
 	[ibm-semeru-runtimes]='openjdk'
+	[azul-zulu-openjdk]='openjdk'
 
 	[jruby]='ruby'
 	[pypy]='python'
@@ -301,6 +302,13 @@ globalExcludeTests+=(
 	[openjdk:alpine_java-uimanager-font]=1
 	[openjdk:slim_java-uimanager-font]=1
 	[openjdk:nanoserver_java-uimanager-font]=1
+
+	# skip tests for distroless
+	[azul-zulu-openjdk:distroless_utc]=1
+	[azul-zulu-openjdk:distroless_no-hard-coded-passwords]=1
+	[azul-zulu-openjdk:distroless_java-hello-world]=1
+	[azul-zulu-openjdk:distroless_java-uimanager-font]=1
+	[azul-zulu-openjdk:distroless_java-ca-certificates]=1
 
 	# the Swift slim images are not expected to be able to run the swift-hello-world test because it involves compiling Swift code. The slim images are for running an already built binary.
 	# https://github.com/docker-library/official-images/pull/6302#issuecomment-512181863
--- a/test/config.sh
+++ b/test/config.sh
@@ -20,7 +20,7 @@ testAlias+=(
 	[eclipse-temurin]='openjdk'
 	[sapmachine]='openjdk'
 	[ibm-semeru-runtimes]='openjdk'
-	[azul-zulu-openjdk]='openjdk'
+	[azul-zulu]='openjdk'
 
 	[jruby]='ruby'
 	[pypy]='python'
@@ -304,11 +304,11 @@ globalExcludeTests+=(
 	[openjdk:nanoserver_java-uimanager-font]=1
 
 	# skip tests for distroless
-	[azul-zulu-openjdk:distroless_utc]=1
-	[azul-zulu-openjdk:distroless_no-hard-coded-passwords]=1
-	[azul-zulu-openjdk:distroless_java-hello-world]=1
-	[azul-zulu-openjdk:distroless_java-uimanager-font]=1
-	[azul-zulu-openjdk:distroless_java-ca-certificates]=1
+	[azul-zulu:distroless_utc]=1
+	[azul-zulu:distroless_no-hard-coded-passwords]=1
+	[azul-zulu:distroless_java-hello-world]=1
+	[azul-zulu:distroless_java-uimanager-font]=1
+	[azul-zulu:distroless_java-ca-certificates]=1
 
 	# the Swift slim images are not expected to be able to run the swift-hello-world test because it involves compiling Swift code. The slim images are for running an already built binary.
 	# https://github.com/docker-library/official-images/pull/6302#issuecomment-512181863


gdams (commented):
LGTM

geertjanw:
@tianon @yosifkit — can you help get this further and approved?

library/azul-zulu
```diff
@@ -0,0 +1,98 @@
+Maintainers: azul_openjdk@azul.com
```
LaurentGoderre, line 1: Build is failing because it is expecting a name. Since this seems like a shared mailbox you can do something like this: https://github.com/docker-library/official-images/blob/master/library/amazoncorretto#L1

tianon: ... but please do include at least one human GitHub handle (or the handle of a bot whose email the whole team will receive that we can @-reference from non-Azul orgs)

geertjanw:
@LaurentGoderre @tianon @yosifkit -- good to go now?

geertjanw:
Hi @LaurentGoderre @tianon @yosifkit -- we believe everything should be good now, can you confirm?

LaurentGoderre:
Not a blocker but typically there is a tag for each variant that default to a specific version (i.e. `ubuntu`, `debian`) so someone can do `docker run azul-zulu:ubuntu`

LaurentGoderre (approved):

geertjanw:
Can we gently prod @tianon and @yosifkit here? Would love to get further with this.

geertjanw:
Hi @tianon @yosifkit -- we believe everything should be good now, can you confirm?

geertjanw:
Hi @tianon @yosifkit -- we believe everything should be good now, can you confirm?

gdams (commented):
Changes still LGTM. To add a bit more context Azul Zulu is another OpenJDK vendor. Their binaries are shipped under GNU General Public License, version 2 with the Classpath Exception.

geertjanw:
Hello @tianon and @yosifkit and anyone else -- can we get further with this, we'd really love to wrap this up. :-)

gdams (commented):
For what it's worth (I'm not a maintainer of this repo or affiliated with Azul) This PR should be good to merge @yosifkit @tianon would it be possible for a review?

geertjanw:
@yosifkit @tianon -- what can we do to move this forward?

geertjanw:
Hi @tianon and @yosifkit, we love Docker and what you guys do for the community, please can you take a look at our work here, we'd love to be included in the official Docker images.

tianon:
Thank you for your patience :heart:

I'm guessing there've been some updates to these in the meantime (or that they're due for some updates), but here's some review of what's currently proposed as a starting point. :+1:

---

We typically discourage this sort of (all predominantly glibc-based) base OS diversity within a single image because for most users it's not functionally all that different or useful (and usually just makes choosing an appropriate base more difficult), and instead encourage maintainers to focus on the "upstream-preferred" base distribution, especially since Docker makes distribution differences matter quite a bit less -- are there concrete functional differences/advantages to users choosing one of these over the other besides their own personal preferences?

For example, are the binaries installed in the Debian and Ubuntu images different enough that one isn't compatible with the runtime environment of the other?

(See https://github.com/docker-library/docs/tree/939fc2b95219f744303c9ddea67f2cc353e255c5/eclipse-temurin#using-a-different-base-image for an example of what I mean, and if you scroll up to the tag listing, a good example of how out-of-control the supported tags get even _with_ limiting it like this. :see_no_evil:)

---

Using the term "distroless" for something that isn't part of or officially related to [the distroless project](https://github.com/GoogleContainerTools/distroless) feels somewhat inappropriate; maybe instead document how to get to an image like this from another variant instead of officially supporting it here for now?

---

Just "debian", "centos", "alpine", "ubuntu" are not terribly specific aliases -- would recommend getting more specific so users can pin/better self-manage when there are issues with base image updates (for example "bookworm" or "debian12" so that when the "square wheel" rolls and Debian Trixie/13 is released you can decide whether to support multiple versions _and_ users can manage the breakage they might experience from the rolling wheel by having an easier tag to target for pinning to the older version).

---

Related to the above, the `centos` images are effectively EOL (upstream is only supporting/actively maintaining CentOS Stream images for several years now, and not in DOI so not as an acceptable base here).

---

```diff
+ENV LANG en_US.UTF-8
+ENV LANGUAGE en_US:en
+ENV LC_ALL en_US.UTF-8
```

Does this really need to be so explicit/verbose?  For example, would it be sufficient to _only_ set `LANG`?  For added context, this makes it harder for users who want to change the value to something else, and there isn't currently a way in the container runtime to *unset* them. :see_no_evil:

---

```diff
+ENV TZ=Etc/UTC
```

This is already the default timezone in all official images (and we even have a test that validates it), so this is unnecessary. :+1:

---

```diff
+    apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 0xB1998361219BD9C9 && \
```

This command (`apt-key`) is deprecated and removed in newer releases -- this should be `gpg --batch --recv-keys` and `deb [ signed-by=/etc/keyrings/... ] ...` or `Signed-By:` in the newer deb822 `sources.list` format.

Also, the key fingerprint should be specified by the full fingerprint (`27BC0C8CB3D81623F59BDADCB1998361219BD9C9`), so it can be fully validated by the tooling during fetching of it.

See https://github.com/docker-library/mysql/blob/6978e12b0d46abf24015045bd22a0cf11b19c150/8.0/Dockerfile.debian#L55-L71 for an example, in case that's helpful.

---

```diff
+    dpkg -i zulu-repo_${ZULU_REPO_VER}_all.deb && \
```

This should probably be `apt-get install -y ./zulu-repo_${ZULU_REPO_VER}_all.deb` instead. :eyes:

---

```diff
+    echo "Package: zulu11-*\nPin: version 11.0.20-*\nPin-Priority: 1001" > /etc/apt/preferences && \
+    apt-get -qq -y --no-install-recommends install zulu11-jdk=11.0.20-* && \
```

If you ever end up using "debian revisions" for rebuilds that don't change the "upstream version" (11.0.20-1 vs 11.0.20-2) you might want to update this to be more explicit (ie no `-*` wildcard) so that there's a natural place to cache-bust (because our build system otherwise won't know to rebuild to pick up your packaging changes).

---

Any particular reason you need `software-properties-common` installed?  (It's mostly a set of helpers for creating text files in `/etc/apt/sources.list.d`, so probably isn't necessary here as you/`zulu-repo` could just create the file directly instead. :sweat_smile:)

---

If you end up keeping an RPM-based distro (after removing `centos`), you'll need to update from `rpm --import https://repos.azulsystems.com/RPM-GPG-KEY-azulsystems` to something that downloads and verifies that key by fingerprint or checksum before importing it. :eyes:

---

Also, to be explicit, I'm happy to elaborate on or further discuss any of these. :heart:

bydga:
Hello,

I would like to revive this PR. We’ve carefully gone through your previous feedback and have made several updates and improvements to the images and their tagging, including:

- Changed the repository containing the Dockerfile definitions to https://github.com/AzulSystems/azul-zulu-images so it no longer conflicts with the existing (and hopefully soon-to-be-deprecated) non-official images.
- Dropped support for the previously unclear/ambiguous “distroless” variant.
- Reduced supported distros to three (Alpine / Debian / Rocky Linux), representing one per major packaging system (apk / apt / yum).
- Removed asterisks for specific version installations.
- Simplified the Dockerfiles by removing unnecessary commands, environment variables, and package installations.


Could you please review the PR again? I am personally taking over this initiative on Azul’s side and would be happy to provide clarification, discuss changes, or make further improvements to get this finalized 🙂

bydga:
also mentioning @tianon for the review 🙏

bydga:
I'm registering those failures, it looks like some discrepancy in the build system used (our Dockerfiles  work on `BuildKit` or i.e. using `buildah` but they fail on "regular" `docker build` (DOCKER_BUILDKIT=0). We will update the dockerfiles asap 👍

bydga:
Hopefully fixed. We removed the heredoc in rocky distro that was causing the trouble 👍

bydga:
Hmm, seems like your [test](https://github.com/docker-library/official-images/blob/master/test/tests/java-uimanager-font/container.java) specifically asks for `fontconfig` and rocky fails on that. I'm not 100% convinced such java baseimage we are producing should cover such usecase, but I guess there's no way around that, so we will add the package.. 🙂

bydga:
I am going through the run report again - fixing test errors regarding the missing fontconfig is easy, but looking into the naughty-check, i think we have a problem:

https://github.com/docker-library/official-images/actions/runs/20232677034/job/58131567056?pr=16141

(there's also issue with unable to fetch master- it is not a problem, we are adding proper `GitFetch: refs/heads/main` to the library file)

but this:
```
FROM rockylinux/rockylinux:10.1) -- completely unsupported base!
```
is probably a dealbreaker. You forbid to use any non-official image as a base, but the  official [rockylinux image](https://hub.docker.com/_/rockylinux) is outdated and they are even saying to use this `rockylinux/rockylinux` alternative. So how to proceed with this? What are our options here?

tianon:
Re: `fontconfig`, I think that's really negotiable - we added it to our tests because it's part of a standard library and we had users actually wanting/needing it, but I think _not_ having it is also pretty reasonable (and there should be a way to exclude that test - if it's not already easy to do so, I'm happy to help work on making it easy to do so / isolating it better).

Re: Rocky Linux, yes, it's unfortunately in a rough spot within DOI right now, but I spoke with them just yesterday about getting that updated (and helping them through some of the logistics), so hopefully that situation will improve soon.

As for reviewing this PR, we can work on getting through it all, but I would honestly recommend starting with just one or two variants of just one or two versions and working upwards incrementally, but that's not a strict requirement (it just might delay the review because it means we have to review everything all at once, where future PRs could benefit from reduced diffs given their delta from the first proposed variants might be smaller and will show in the diff better/easier).

bydga:
thanks @tianon for the reply.

Re `fontconfig`: No problem, if alpine and debian pass the test anyways (with the os-default packages), it doesn't make sense for rocky to behave differently. Plus adding one package on our side is not a big deal 👍 

To keep this initial PR as straightforward as possible, we’ll **reduce scope** and:
- drop **Alpine** and **Rocky Linux** variants for now
- keep **Debian only** (which we want as the default anyway)

So for this PR, the review scope would be:
- Java majors: **8, 11, 17, 21, 25**
- Variants: **jdk / jre**, **headless / headful**

All Dockerfiles follow the same pattern (installing from our apt repo) and differ only in the exact package name, e.g.:
`apt install zulu25-jre` vs `apt install zulu21-jdk-headless`.

When you mentioned “variants”, did you mean those multiple distros, or something else specifically you’d like us to reduce further?
If needed, we can narrow it down even more for this first PR.

bydga:
Cheers @tianon,
I'm just checking: are you OK with reducing the scope as I mentioned? It is one distro (Debian) and a couple of java versions/variants (8,11,25, jre/jdk/headless).

The Dockerfile for all variants is "the same", it just differs in the [apt package](https://github.com/AzulSystems/azul-zulu-images/blob/main/25/jre/debian/Dockerfile#L11) being installed from the repo.

All GH action checks passed, so hopefully we are good to go 🙂 Let me know if you have some additional concerns 🙏 🙂

bydga:
Hello, any progress on this one, please? 🙏 Yesterday, there was a new Java release and we would like to finish this DOI initiative so we can be up-to-date with the official images 🙂

tianon:
I was thinking maybe even as far as a single Java version here, since they're all the same (and then the follow-up will be really obvious), but perhaps that's too far -- given they're all the same, I think this is probably digestible enough as-is.

Here's some new review notes: :heart:

---

```diff
+Tags: 25.30-25.0.1-jdk-debian13.2, 25, 25-jdk, 25-jdk-debian13.2
```

It looks like I've caused an overcorrection in the opposite direction :sweat_smile:

Including `debian` as a plain alias here does seem fine, but I would suggest _also_ including something like `debian13` -- going as specific as "Debian 13.2" is actually a really bad idea as Debian does minor releases often, and you'll have to proactively keep up (and the minor release version numbers are not really all that relevant to end-users).

Here's what I'd suggest instead (hopefully the implication for all tag groups is obvious here):

```yaml
Tags: 25.30-25.0.1-jdk-debian13, 25-jdk-debian13, 25-jdk-debian, 25-jdk, 25
```

If I were maintaining this, I'd probably do even more aliases than that, but that's a decent place to start -- this makes it easier to upgrade to debian14 eventually (even if you still support debian13 for some period, and gives users an easy and obvious escape hatch to downgrade).

---

```diff
+FROM debian:13.2
```

Similar to the above, this is more specific than I'd recommend being -- we'll proactive rebuild the images when the base image has updates, and this will mean that once Debian 13.3 is released, you no longer get to partake in those rebuilds until you send us an update, so I'd suggest doing `FROM debian:13` or even `FROM debian:13-slim` instead.

---

```diff
+ARG REPO_HOST=repos.azul.com
```

This seems kind of odd, TBH - is it actually in use in some of your workflows?

---

```diff
+ENV LANG='en_US.UTF-8'
```

Just to make sure you're aware, this locale does not exist by default in Debian:

```console
$ docker run -it --rm --pull=always debian:13-slim locale -a
13-slim: Pulling from library/debian
Digest: sha256:77ba0164de17b88dd0bf6cdc8f65569e6e5fa6cd256562998b62553134a00ef0
Status: Downloaded newer image for debian:13-slim
C
C.utf8
POSIX
```

(I don't know whether the Zulu packages create it, but nothing in your `Dockerfile` appears to otherwise.)

---

```diff
+RUN apt update && \
```

As noted by the warning this spits out, `apt` is not suitable/intended for programmatic use yet (so this should be `apt-get` instead):

```console
$ docker run -i --rm --pull=always debian:13-slim apt update > /dev/null
13-slim: Pulling from library/debian
Digest: sha256:77ba0164de17b88dd0bf6cdc8f65569e6e5fa6cd256562998b62553134a00ef0
Status: Image is up to date for debian:13-slim

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

```

---

```diff
+    curl -s https://repos.azul.com/azul-repo.key | gpg --dearmor -o /usr/share/keyrings/azul.gpg && \
```

This should be doing much more explicit verification, preferably something that verifies the full PGP fingerprint (`27BC 0C8C B3D8 1623 F59B  DADC B199 8361 219B D9C9`), but a solid checksum like sha256 would be acceptable as well.  Doing this correctly with GnuPG is a bit of an esoteric exercise, so I'm happy to provide example snippets if that's helpful or necessary.

---

```diff
+    apt-get -qq -y autoremove && \
+    apt-get -qq clean && \
```

The first of these is already implied / executed by the `purge --auto-remove` command above it (so it's a no-op) and the latter is a no-op thanks to configuration built into the Debian base image: https://github.com/debuerreotype/debuerreotype/blob/d9bef2e25eae9ab5e83d7c1b56b044f4f3c87cce/scripts/debuerreotype-minimizing-config#L87-L111

---

```diff
+    rm -rf /var/lib/apt/lists/* && \
```

It's not strictly necessary, but you might be interested to note that in Debian 13+, `apt(-get)` now has a `dist-clean` command that does this operation but in a better way (such that it preserves important cryptographic information), so if you don't plan to support Debian 12 in a later update, I'd suggest updating this to `apt-get dist-clean` instead.

---

```diff
+    rm -f /etc/apt/sources.list.d/zulu.list
```

This seems kind of strange - why would you want to remove the APT configuration from the image?  Does it cause problems to leave it in?

---

I would suggest adding `CMD ["jshell"]` to the JDK images (so that users who want to just run some quick Java tests/code can `docker run -it --rm azul-zulu:25-jdk` and drop straight into a useful shell for doing so).

---

Digging into the package metadata for the packages being installed here, it appears that they're mostly additive (`zulu11-jdk` depends on `zulu11-jre` and `zulu11-jdk-headless`, `zulu11-jre` depends on `zulu11-jre-headless`, etc).  I don't know how common it will be for users to pull and use both headless and non-headless variants at once, but I _do_ imagine it will be common for users to pull both JDK and JRE of the same version/variant, so it's worth considering making these dependent images instead of standalone as they are now.

In other words, you could have `azul-zulu:21-jre-headless` which is exactly as it is here, but then `azul-zulu:21-jdk-headless` is `FROM azul-zulu:21-jre-headless` + install the JDK package on top, etc, so that the images get layered in the same way the packages do.  This would help end-users get more value/benefit from the layering that container images are built from.

---

It might be worth ending each of these "install packages" `RUN` lines with an explicit "does it work?" smoke test like `java --version` (for images we maintain, we find that doing this kind of thing helps us catch minor issues slightly earlier).

---

Putting it all together, here's an explicit example for `azul-zulu:25-jre-headless`:

```dockerfile
FROM debian:13-slim

ENV LANG='C.UTF-8'
RUN set -eux && \
    apt-get update && \
    apt-get install -y --no-install-recommends gnupg ca-certificates curl && \
    GNUPGHOME="$(mktemp -d)" && \
    export GNUPGHOME && \
    curl -fsS https://repos.azul.com/azul-repo.key | gpg --batch --import && \
    gpg --batch --export --armor '27BC 0C8C B3D8 1623 F59B  DADC B199 8361 219B D9C9' > /usr/share/keyrings/azul.pgp.asc && \
    gpgconf --kill all && \
    rm -rf "$GNUPGHOME" && \
    echo 'deb [signed-by=/usr/share/keyrings/azul.pgp.asc] https://repos.azul.com/zulu/deb stable main' | tee /etc/apt/sources.list.d/zulu.list && \
    printf 'Package: zulu25-*\nPin: version 25.0.1-1\nPin-Priority: 1001\n' > /etc/apt/preferences && \
    apt-get update && \
    apt-get -y --no-install-recommends install zulu25-jre-headless && \
    apt-get -y purge --auto-remove gnupg curl && \
    apt-get dist-clean && \
    java --version

ENV JAVA_HOME=/usr/lib/jvm/zulu25
```

Then, `azul-zulu:25-jdk-headless` would be this, instead:

```dockerfile
FROM azul-zulu:25-jre-headless

RUN set -eux && \
    apt-get update && \
    apt-get -y --no-install-recommends install zulu25-jdk-headless && \
    apt-get dist-clean && \
    javac --version

CMD ["jshell"]
```

I'd then probably hang each non-headless variant off each of those (`25-jdk` is `FROM azul-zulu:25-jdk-headless`, etc) since I imagine the "non-headless" bits on top are the smallest layer:

```dockerfile
FROM azul-zulu:25-jdk-headless

RUN set -eux && \
    apt-get update && \
    apt-get -y --no-install-recommends install zulu25-jdk && \
    apt-get dist-clean && \
    javac --version # this is mostly redundant - maybe there's a better simple one-liner smoke test for non-headless?
```

Here's the best way to visualize the benefit of this layered approach:

```console
$ docker images azul-zulu
REPOSITORY:TAG              IMAGE ID       CREATED          SIZE
azul-zulu:25-jdk            97abc1eb21a7   26 seconds ago   455MB
azul-zulu:25-jdk-headless   6ff4e1320cbc   4 minutes ago    447MB
azul-zulu:25-jre-headless   8238f72b1a5b   12 minutes ago   342MB

$ docker history azul-zulu:25-jdk
IMAGE          CREATED          CREATED BY                                      SIZE      COMMENT
97abc1eb21a7   42 seconds ago   /bin/sh -c set -eux &&     apt-get update &&…   7.9MB     
6ff4e1320cbc   5 minutes ago    /bin/sh -c #(nop)  CMD ["jshell"]               0B        
7d956485cb7b   5 minutes ago    /bin/sh -c set -eux &&     apt-get update &&…   105MB     
8238f72b1a5b   12 minutes ago   /bin/sh -c #(nop)  ENV JAVA_HOME=/usr/lib/jv…   0B        
c47cc40e48e1   12 minutes ago   /bin/sh -c set -eux &&     apt-get update &&…   264MB     
fa2e84e7d680   15 minutes ago   /bin/sh -c #(nop)  ENV LANG=C.UTF-8             0B        
d10e1548d774   10 days ago      # debian.sh --arch 'amd64' out/ 'trixie' '@1…   78.6MB    debuerreotype 0.17
```

(looking at that 8MB layer on top has me wondering if all the extra "headless" vs "non-headless" complexity is really worth it given the size of the JRE at 264MB and JDK addition at 105MB but that's really up to you)

bydga:
Hey,
thanks @tianon for the detailed feedback! We’ve gone through all your points and (hopefully) addressed everything 🙂
About the specific tags (like xxx-debian13.2) - yeah, that felt a bit weird to us too 😅 but after those changes it looks good from our side now.

Regarding the image layering you mentioned: honestly, the complexity feels pretty high for us. It would introduce build ordering and dependencies between images/tags, and we don’t really see how that could be expressed in the library metadata file. Unless there’s something that actually parses the Dockerfiles, looks at the `FROM` directives, and builds a DAG from that? Also, we’re not aware of any other Java vendors taking this approach 🤔

As for the `ARG REPO_HOST`: that’s there on purpose. It’s part of our internal workflow where we build and test images ahead of the official Java release using private Linux repos. Being able to inject those repos lets us prebuild and validate images early. If that’s OK from your side, we’d like to keep it - having the same Dockerfile for internal and official builds makes a lot of sense for us.

Thanks again for taking the time to review this!

tianon:
> Regarding the image layering you mentioned: honestly, the complexity feels pretty high for us. It would introduce build ordering and dependencies between images/tags, and we don’t really see how that could be expressed in the library metadata file. Unless there’s something that actually parses the Dockerfiles, looks at the `FROM` directives, and builds a DAG from that? Also, we’re not aware of any other Java vendors taking this approach 🤔

For the DAG part of this question, yes, our system is designed to build a DAG by parsing the `FROM` instructions, so build ordering is automatic (across the entire program).  It kind of has to be, so that when `debian:13-slim` is updated, we can know that we need to rebuild `azul-zulu` images too. :smile:

The way I might describe why no other Java vendor is taking this approach is because your packaging discipline is much more advanced than any other Java vendor (at least, that I've looked at).  You've already done the hard work upstream of splitting the JDK into separate component parts as discrete distribution packages, so I would see this complex layering as reflecting that effort and discipline down into the Docker images too, so they can benefit from the same improved splitting.

> (looking at that 8MB layer on top has me wondering if all the extra "headless" vs "non-headless" complexity is really worth it given the size of the JRE at 264MB and JDK addition at 105MB but that's really up to you)

copying this comment just so we keep track of it because it's probably relevant to think about whether you decide to go with my complex layering suggestion or not :+1:

bydga:
We discussed it internally and we actually like the idea.
However, we have two blockers right now:
- our infra/pipelines aren’t ready for this approach yet and will need some changes to suport the relations between the images
- We need internal alignment with the product team before committing to this structure. Also, the product might have some insights on how our users are actually using the images.

So can we please get this MR approved as-is to unblock the progress - as the initiative started two years ago? 😅 

We’ll prepare the necessary infra and product alignment internally and follow up with improvements in subsequent PRs once that’s in place.

yosifkit:
I think this all looks pretty good. My only feedback is to remove `CMD ["jshell"]` from the 8 jdk and 8 jdk-headless images, since it doesn't exist there.
```console
$ docker build -t azul-zulu:8-jdk 8/jdk/debian
$ docker run -it --rm azul-zulu:8-jdk
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: exec: "jshell": executable file not found in $PATH
```

bydga:
@yosifkit Oh, thanks a lot for catching that 👍 It totally slipped through our tests, we always run "something" in the container, which overrides the image’s default CMD, so we never noticed. We’ll improve our internal tests to cover this properly.

Is there anything else you’d like us to adjust, or are we good to go?

geertjanw:
Many thanks for all the interaction, feedback, and comments, @yosifkit and @tianon. Let us know what more we can do, what the status is from your end for this work, we'd of course love to move forward and promote this.

yosifkit:
I left a few comments on the [docs PR](https://github.com/docker-library/docs/pull/2320#pullrequestreview-3828667424), but otherwise this looks good to me.

tianon:
> > (looking at that 8MB layer on top has me wondering if all the extra "headless" vs "non-headless" complexity is really worth it given the size of the JRE at 264MB and JDK addition at 105MB but that's really up to you)
> 
> copying this comment just so we keep track of it because it's probably relevant to think about whether you decide to go with my complex layering suggestion or not 👍

I think this is the last thing I see in the `Dockerfile`s -- in short, are all the "headless" variants really worth supporting separately in container image granularity when the size difference is roughly only 8MB?

Is there some other benefit that I'm missing besides size that's dramatic enough to be worth the added maintenance, review, build, and image selection load for end users?

geertjanw:
> > > (looking at that 8MB layer on top has me wondering if all the extra "headless" vs "non-headless" complexity is really worth it given the size of the JRE at 264MB and JDK addition at 105MB but that's really up to you)
> > 
> > 
> > copying this comment just so we keep track of it because it's probably relevant to think about whether you decide to go with my complex layering suggestion or not 👍
> 
> I think this is the last thing I see in the `Dockerfile`s -- in short, are all the "headless" variants really worth supporting separately in container image granularity when the size difference is roughly only 8MB?
> 
> Is there some other benefit that I'm missing besides size that's dramatic enough to be worth the added maintenance, review, build, and image selection load for end users?

@tianon, thanks for this comment — is this something that needs to be solved prior to the merge, or is this an aspect to consider right after that? Ideally we’d get the show on the road and do all kinds of improvements afterwards.

tianon:
It's something you should absolutely decide/justify prior to the merge (because removing supported things after merge is _much_ more complicated).

ppetrosh:
> It's something you should absolutely decide/justify prior to the merge (because removing supported things after merge is _much_ more complicated).

Yes, that makes sense. Thank you, @tianon.
We are not going to remove these configurations after the merge. Instead, we will optimize the JRE package itself to make it smaller — that's WIP and shouldn't affect the Docker files, so it shouldn't affect the merge. It'll be optimized at the level of Linux package dependencies (layered Linux packages). I hope that makes sense.
So we should be good to go in its current state.

geertjanw:
> > It's something you should absolutely decide/justify prior to the merge (because removing supported things after merge is _much_ more complicated).
> 
> Yes, that makes sense. Thank you, @tianon. We are not going to remove these configurations after the merge. Instead, we will optimize the JRE package itself to make it smaller — that's WIP and shouldn't affect the Docker files, so it shouldn't affect the merge. It'll be optimized at the level of Linux package dependencies (layered Linux packages). I hope that makes sense. So we should be good to go in its current state.

Does @ppetrosh's response above provide what's needed here, @tianon?

bydga:
Folks, thanks for the cooperation, it was a valuable experience 👍 🙂 As our product will be doing some adjustments to the linux repos layering, we will then incorporate that into the docker layers as well - per your suggestion. As new Java 26 was released yesterday, we will soon prepare another MR to update the tags, hopefully it will be a faster change 😅

PR #16154, opened 2024-01-30
Add Nextcloud AIO

szaimen:
The official Nextcloud installation method. Provides easy deployment and maintenance with most features included in this one Nextcloud instance.

- docs PR: https://github.com/docker-library/docs/pull/2412


szaimen:
Seems like the build infrastructure is too old fo multistage builds and same for `COPY --chmod=775` ???

PR #16167, opened 2024-02-01
Updated swipl devel&stable to 9.3.0 and 9.2.0

JanWielemaker:


tianon (approved):

JanWielemaker:
It seems something went wrong.  According to https://hub.docker.com/_/swipl/, the update was effectuated, put pulling `swipl:stable` gets 9.0.4 (the previous stable) and `swipl:9.2.0` is not found.    The 9.2.0 Dockerfile looks good.   It builds fine for me.    Same happened to the _latest_ update from 9.1.19 to 9.3.0.

Any clue?   Did I make a mistake or is something wrong at your end?

tianon:
It seems to be getting a very weird (and extremely verbose -- so long I can't even post it here :sob:) build failure on `amd64` (at least): https://doi-janky.infosiftr.net/job/multiarch/job/amd64/job/swipl/233/execution/node/105/log/?consoleFull

JanWielemaker:
Thanks for sharing.   So, it seems the build for rocksdb is failing.   What I don't  get is how you get different results than I do.   I'd think the build, starting from bookworm-slim on amd64 is completely predictable, no?    But my log says (where yours bails on `utilities/blob_db/blob_db.o`

```
...
#6 151.6   CC       utilities/blob_db/blob_db.o
#6 151.7   CC       utilities/blob_db/blob_db_impl.o
...
```

I pulled debian:bookworm-slim, which appeared to have been updated since my build test.  The result is the same, i.e., it builds flawlessly.    I verified the installed package versions for gcc and g++.  They match.   Both checkout the same git hash of RocksDB.    They call `make` with the same arguments.    How can one complain about possibly uninitialized variables somewhere deep in the gcc header files (weird in itself) and the other not.

Do you have any hint where to look?    If not, I see two ways forward

  - Try the bullseye version
  - Change the RocksDB dependency to compile RocksDB with less aggressive g++ flags
 
Keen to hear your opinion.

yosifkit:
> Try the bullseye version
> Change the RocksDB dependency to compile RocksDB with less aggressive g++ flags

Either of these might work. The images currently just build on an AWS `m5.2xlarge`. If it helps, `/proc/cpuinfo` shows `Intel(R) Xeon(R) Platinum 8259CL CPU @ 2.50GHz`. (I am happy to dig up more details if that is useful.)

---

An alternative solution is for us to move `swipl` to our new build queue. For `amd64` it uses our same Jenkins instance to trigger image builds in GitHub Actions (https://github.com/docker-library/meta). Nothing will change here in the official-images repo and the images will still end up on Docker Hub. https://github.com/docker-library/meta/pull/32 would add `swipl` to the new builds.

The new system still has a few [scaling issues](https://github.com/docker-library/meta-scripts/issues/22) we are working on, so we are now slowly moving official image builds over. Eventually we want to move them all over. As for the differences, it is mostly around not having to rely on `docker build` cache while avoiding unnecessary rebuilds; using buildkit (it's just `docker build`, but no longer the classic builder); and then being able to do some architectures' builds via GitHub actions (like `amd64`).

yosifkit:
`amd64` builds in progress; hoping for success:

https://github.com/docker-library/meta/actions/runs/7834461937
https://github.com/docker-library/meta/actions/runs/7834461854

JanWielemaker:
> `amd64` builds in progress; hoping for success:

Hi @yosifkit.    This did pass.    Thanks!     Yesterday I ran another build on an Intel machine (Intel Core Processor (Broadwell, IBRS)) that also passed fine.    

Is the problem resolved at your end, or do I need to make adjustments when doing the next release?

yosifkit:
No need to change anything; it should be fine now. We'll continue to build `swipl` on GitHib actions for the foreseeable future.

yosifkit:
> No need to change anything; it should be fine now. We'll continue to build `swipl` on GitHib actions for the foreseeable future.

Well, it seems we only lasted about 2 years using GHA for our amd64 builds and have recently switched everything back to builders in AWS and so this build failure has come back. 😢 Full logs of one of the failures are available [here](https://doi-janky.infosiftr.net/job/meta/view/builds/job/amd64/job/build/3271/consoleFull).

Builder details are Debian Trixie on a `c7i.2xlarge` EC2 instance ("compute optimized 4th Generation Intel Xeon"). If there are any other details that might help you track down the cause we are happy to provide. And if there is some config that we should change on the host, we can try that too.

---

🤔 perhaps the failure is related to https://github.com/facebook/rocksdb/pull/9978 and so would only show up on [CPUs with AVX-512 support](https://en.wikipedia.org/wiki/AVX-512#CPUs_with_AVX-512) like the Sapphire Rapids we are using now and the "Skylake-SP or Cascade Lake" we were using previously with the `m5` instance. Though that could mean the default flags might be building too specific of a binary that wouldn't work on other CPUs.

PR #16170, opened 2024-02-01
Update Scout scanner to v1.4.0

LaurentGoderre:


.external-pins/docker/scout-sbom-indexer___1
```diff
@@ -1 +1 @@
-sha256:60dae27f30cb9c7aafbd237d1f74dbf05e254f530f0580d6d79c728de83a34ee
+sha256:c9bfbbb4a90842209fd2be07f6dc7e10a762d68468570d4776e6d92ce4b07287
```
tianon, line 1: This should keep the "newline at EOF" like `echo foo > file` would provide so the bot doesn't immediately make a new branch :eyes: (https://github.com/docker-library-bot/official-images/commit/d1df5826597c0861b7f653e75daffff814a7dec1)

(See `.external-pins/update.sh` for the script the bot uses to update these :eyes:)

LaurentGoderre: Updated.

tianon (approved):

tianon:
Just for future reference: this breaks some of the paths in the SBOMs (and that's currently being investigated) :heart:

PR #16180, opened 2024-02-02
Updates to Storm 2.6.1

rzo1:
as the title says

### DIFF
--- a/library/storm
+++ b/library/storm
@@ -4,10 +4,6 @@ Maintainers: The Apache Storm Project <dev@storm.apache.org> (@asfbot),
 GitRepo: https://github.com/apache/storm-docker.git
 Architectures: amd64, arm64v8
 
-Tags: 2.6.0
-GitCommit: d1f8af26ccf800531cc773a39fac9e30efa94bca
-Directory: 2.6.0
-
 Tags: 2.6.1, 2.6, latest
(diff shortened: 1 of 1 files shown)


library/storm
```diff
@@ -4,6 +4,10 @@ Maintainers: The Apache Storm Project <dev@storm.apache.org> (@asfbot),
 GitRepo: https://github.com/apache/storm-docker.git
 Architectures: amd64, arm64v8
 
-Tags: 2.6.0, 2.6, latest
-GitCommit: 965de975c7af0ed35174b4276971c52965b0fa0b
+Tags: 2.6.0
+GitCommit: d1f8af26ccf800531cc773a39fac9e30efa94bca
 Directory: 2.6.0
+
```
tianon, line 10: Any particular reason to keep this version listed here?  Removing tags here will remove them from the "Supported" section on the Hub readme (and will prevent us from spending cycles rebuilding them on the official build servers), but the tags will still be available to users who want them.  (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)

```suggestion
```

rzo1: TIL. Yes, I will remove it.

rzo1:
Thanks @tianon

tianon:
No problem, thank you!!

tianon (approved):

PR #16183, opened 2024-02-02
Release OrientDB 3.2.27

tglman:


library/orientdb
```diff
@@ -1,17 +1,17 @@
...
-Tags: 3.2.26-tp3, 3.2-tp3
+Tags: 3.2.27-tp3, 3.2-tp3
 Architectures: amd64, arm32v7, arm64v8
 GitRepo: https://github.com/orientechnologies/orientdb-docker.git
-GitCommit: 1cf7837740ef309efc58759b0a33a3fe8c2c5278
-Directory: release/3.2.x/3.2.26-tp3
+GitCommit: 7c3a945a49494db55de300c0193da8d9dd0f51c1
+Directory: release/3.2.x/3.2.27-tp3
 
 Tags: 3.1.20, 3.1
```
tianon, line 16: Any particular reason to keep all these old versions listed here? (Are they still actively maintained?)

yosifkit (approved):

yosifkit:
Merging to get the update out; happy to continue the discussion about possibly end-of-life versions.

PR #16199, opened 2024-02-06
Update ELK

navyau09:


tianon:
> This updates ELK images to point to the original source repository so they build directly and updates the maintainers appropriately.

This was the description on the first PR that updated these to point to the original repositories -- in this case, I think you just meant you updated the versions, right?

("Updated to 7.17.18 and 8.12.1" for example)

tianon (approved):

PR #16215, opened 2024-02-08
update rocket.chat

sampaiodiego:


library/rocket.chat
```diff
@@ -1,10 +1,14 @@
...
 Maintainers: Rocket.Chat Image Team <buildmaster@rocket.chat> (@RocketChat)
 GitRepo: https://github.com/RocketChat/Docker.Official.Image.git
 
-Tags: 6.5.2, 6.5, 6, latest
-GitCommit: fb09fece013b05ce7c0d8a42b646543a75fda57a
+Tags: 6.6.0, 6.6, 6, latest
+GitCommit: 6bb91d4971eb1d1c8e7fb84ec2d646c2b9474818
+Directory: 6.6
+
+Tags: 6.5.3, 6.5
```
tianon, line 10: Any particular reason to keep these old versions listed here? (It's my understanding that they're now EOL?)

sampaiodiego: hey @tianon , I even checked the documentation before submitting the PR https://docs.rocket.chat/customer-center/support-center/premium-support-plans/support-prerequisites-and-version-durability#version-end-of-life

not sure if they bumped the supported versions a bit, but it states 6.3.x EOL is end of this month, so I left it here.

tianon: Oh interesting - I was looking at https://endoflife.date/rocket-chat but maybe that's outdated (or talking about a different version). :+1: :heart:

(Probably worth sending them a PR if it's outdated though!)

tianon (approved):

PR #16223, opened 2024-02-09
Update openjdk

yosifkit:
Changes:

- https://github.com/docker-library/openjdk/commit/f1a2ae8: Update 22 to 22 (rc1)


tianon:
https://openjdk.org/projects/jdk/22/ :eyes: (effectively, they hope this will be the GA, but it isn't set it stone yet -- in other words, we'll be removing it soon!)

tianon (approved):

PR #16226, opened 2024-02-12
MariaDB 2024 Q1 part 1 release

grooverdan:
10.10 now EOL

Added entrypoint feature to use MARIADB_AUTO_UPGRADE=1 to recreate healthcheck user where cnf file was missing.

Added notes about memory pressure, available in cgroupsv2, but hidden by container runtimes.


tianon (approved):

grooverdan:
Thanks @tianon

PR #16248, opened 2024-02-14
Update YOURLS

LeoColomb:
Automated changes from [YOURLS/images](https://github.com/YOURLS/images) via [GitHub Actions run #4336679159](https://github.com/YOURLS/images/actions/runs/7901036639).

### DIFF
--- a/library/yourls
+++ b/library/yourls
@@ -1,21 +1,21 @@
-# this file is generated via https://github.com/YOURLS/docker/blob/2ef3b5fe2539efd8dd876727376ddee2c22079b5/bin/generate-stackbrew-library.sh
+# this file is generated via https://github.com/YOURLS/images/blob/082820033f591498dbc30d4ffc0cc927197ebf26/bin/generate-stackbrew-library.sh
 
 Maintainers: YOURLS <yourls@yourls.org> (@YOURLS),
              Léo Colombaro <git@colombaro.fr> (@LeoColomb)
-GitRepo: https://github.com/YOURLS/docker.git
+GitRepo: https://github.com/YOURLS/images.git
 GitFetch: refs/heads/dist
 
 Tags: 1.9.2-apache, 1.9-apache, 1-apache, apache, 1.9.2, 1.9, 1, latest
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
-GitCommit: 2ef3b5fe2539efd8dd876727376ddee2c22079b5
+GitCommit: 082820033f591498dbc30d4ffc0cc927197ebf26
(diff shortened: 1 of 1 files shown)


yosifkit:
```dockerfile
FROM php:8.3.2-apache
```
We recommend that this stays as `php:8.3-apache` in order to get timely OS package and PHP patch-version updates. When any image in `official-images` is changed, all dependent official images are rebuilt. This is why we require that any image update be `FROM` a current/active official image.  We do not support old patch releases of `php` images, so this would quickly cause the `yourls` images to stop getting rebuild updates (when PHP `8.3.3` is released).

A few bits of feedback on the `customPostInitScripts`:
- not sure if it is desired, but the current `find` will also get all files in subdirectories as well, so you might want it limited in some way (`-mindepth 1 -maxdepth 1`?). For images we maintain, we only use scripts from the top level and let users have helper scripts or data in deeper directories.
  ```bash
  for init_script in /docker-entrypoint-init.d/* do
  ...
  ```
- the helm chart mentions just three file types so I think the loop should only process `.sh`, `.sql`, and `.php` scripts and each should likely be handled differently and then ignore anything else
  ```bash
  case "$init_script" in
    *.sh)
      if ...
      ;;
    *.php)
      php -f "$init_script"
      ...
    *) echo >&2 "ignoring $init_script" ;;
  esac
  ```
- it should do something different with non-executable `.sh` files. (perhaps `source` them [like `postgres`](https://github.com/docker-library/postgres/blob/1424abf76f421d6f7bf933d9e42bbbed866fae3a/docker-entrypoint.sh#L176), or maybe ignore them [like `nginx`](https://github.com/nginxinc/docker-nginx/blob/1f227619c1f1baa0bed8bed844ea614437ff14fb/entrypoint/docker-entrypoint.sh#L33-L34))
  ```bash
  if [ -x "$init_script" ] then "$init_script"; else ??; fi
  ```
- (for images we maintain, we don't typically add this type of functionality unless there's a temporary server involved such as MySQL or PostgreSQL, because it is trivial for users to inject their own entrypoint script or change the container command such that it does some arbitrary extra work before running the image's software, but it is ultimately up to you if you want to include/maintain this functionality)

LeoColomb:
> We recommend that this stays as `php:8.3-apache`

Understood, I'll figure out how to revert to this. 👍

> A few bits of feedback on the `customPostInitScripts`

Indeed, after a second glance, I see it raises few concerns.
@ggarson, as you submitted this addition on https://github.com/YOURLS/images/pull/263, what do you think about this?

LeoColomb:
@yosifkit Content updated based on your review. Thanks again 😊

tianon (approved):

PR #16260, opened 2024-02-15
[nats] Release v2.10.11

bruth:
Details can be found [here](https://github.com/nats-io/nats-server/releases/tag/v2.10.11)

Introduced the ppc64le arch as well.

### DIFF
--- a/library/nats
+++ b/library/nats
@@ -4,26 +4,26 @@ Maintainers: Derek Collison <derek@synadia.com> (@derekcollison),
              Neil Twigg <neil@synadia.com> (@neilalexander),
              Phil Pennock <pdp@synadia.com> (@philpennock)
 GitRepo: https://github.com/nats-io/nats-docker.git
-GitCommit: 8a2193806026b10571ca25b2bd7b8b775b035283
+GitCommit: e7153cd76185c625bfef5384c81db4c96556173a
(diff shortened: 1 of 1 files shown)


bruth:
A couple changes in our nats-docker repo introduced the use of a build arg which appears to be causing the tests in this repo to fail. See [here](https://github.com/nats-io/nats-docker/pull/140/commits/dd5212bb262783824c561de16d8b36200a4357d9#diff-e70a6853665e4e159d3cb703b51ed6ef6a7323b6fa54865de8b1cd2aa48a43be) as an example. Doing a normal `docker build` works fine. Is there some constraint for Dockerfiles in the official images repo that prevents this from working?

yosifkit:
> Is there some constraint for Dockerfiles in the official images repo that prevents this from working?

Yeah, we parse the Dockerfiles to determine build ordering and none of that supports putting variables in `FROM`/`--from` values.

bruth:
Fixed @yosifkit

albionb96:
@tianon , @yosifkit is there any reason why this is not yet released, are there any issues with this version, or is it just ongoing?

tianon (approved):

tianon:
(US-based Pacific Timezone)

bruth:
Thanks @tianon!

PR #16277, opened 2024-02-19
New Docker image for Silverpeas 6.3.4

mmoqui:
Update silverpeas Docker library to take into account Silverpeas 6.3.4

### DIFF
--- a/library/silverpeas
+++ b/library/silverpeas
@@ -6,6 +6,3 @@ Tags: 6.3.4, latest
 GitCommit: 0ab5ec380118de3c070135446eabd3d94dfa0395
 GitFetch: refs/heads/6.3.x
 
-Tags: 6.2.3-b1
-GitCommit: 9714dcc94eb558508f085835a329a44f5c3cb52e
-GitFetch: refs/heads/6.2.x
--- a/library/silverpeas
+++ b/library/silverpeas
@@ -2,8 +2,8 @@
 Maintainers: Miguel Moquillon <miguel.moquillon@silverpeas.org> (@mmoqui)
 GitRepo: https://github.com/Silverpeas/docker-silverpeas-prod.git
 
-Tags: 6.3.3, latest
-GitCommit: 58fcac8f9759d8ba09ca23bc2106a2a7c5a5217f
+Tags: 6.3.4, latest
+GitCommit: 0ab5ec380118de3c070135446eabd3d94dfa0395
 GitFetch: refs/heads/6.3.x
 
 Tags: 6.2.3-b1


tianon (approved):

tianon:
```
--2024-02-20 18:54:01--  https://silverpeas.org/files/wildfly-20.0.1.Final.zip
Resolving silverpeas.org (silverpeas.org)... 212.83.183.214
Connecting to silverpeas.org (silverpeas.org)|212.83.183.214|:443... connected.
ERROR: no certificate subject alternative name matches
	requested host name ‘silverpeas.org’.
To connect to silverpeas.org insecurely, use `--no-check-certificate'.
```

This build failure on 6.2 is odd :thinking:

whalelines:
Is this some sort of phone home mechanism?

```sh
curl -k -i --head https://www.silverpeas.org/ping
```

Is it wise to ignore TLS failures?

yosifkit (approved):

tianon:
Something to look at in the future (something we should've noticed earlier :disappointed: :see_no_evil:): does the output of `./silverpeas install` go to a volume or change based on runtime values (such as configuration)?

(If not either of those things, it should really be done in the `Dockerfile` instead.)

The ping is definitely really odd, and I'd recommend against it, but it's your choice.  If it's for tracking users, I'd recommend reading https://research.swtch.com/telemetry-opt-in (especially if you plan to ever have it do more than just track that users are running the container).  Either way, it'd be good to add some way for users to opt out (or opt in).

mmoqui:
> Something to look at in the future (something we should've noticed earlier 😞 🙈): does the output of `./silverpeas install` go to a volume or change based on runtime values (such as configuration)?
> 
> (If not either of those things, it should really be done in the `Dockerfile` instead.)

The command `./silverpeas install` in `run.sh` performs, among others things, a configuration task (in which the database setting is the main step) according to some parameters passed by the user (either by specifying them as environment variables and/or by indicating them in some configuration files  passed as a `-v` argument.  The output of the command is a log file `build-<timestamp in iso>.log` produced into the `/opt/silverpeas/log` directory. This directory is exposed as a volume. The configuration task can also update some specific configuration properties, all of them located as files in the `/opt/silverpeas/properties` directory which is also exposed as a volume.

> 
> The ping is definitely really odd, and I'd recommend against it, but it's your choice. If it's for tracking users, I'd recommend reading https://research.swtch.com/telemetry-opt-in (especially if you plan to ever have it do more than just track that users are running the container). Either way, it'd be good to add some way for users to opt out (or opt in).

The ping is just to have a glance of the geographical location of ours users. This is to know which users are actually using Silverpeas from those who are just pulling the Silverpeas image.

PR #16293, opened 2024-02-22
MariaDB urgent reissue

grooverdan:
Unfortunately the last latest release broken PHP and nodejs because of a protocol issue. The configuration item character-set-collations added in the default 11.3/11.4 image generated this.

Closes: https://github.com/MariaDB/mariadb-docker/issues/560


tianon (approved):

grooverdan:
Thanks very much @tianon for responding so very late your day.

tianon:
I just happened to still be sitting here and looking at the list; you got really lucky. :smile:

PR #16313, opened 2024-02-27
Updated Unit to 1.32.0.

oxpa:
Hello,

Pull request updates nginx-unit to the latest version (1.32.0).
Unit logs are now written to stderr to avoid mixing with application logs (if an application chooses to log to stdout).

This also updates a number of dependent libraries:
* Go: Dropped 1.20, Added 1.22
* Node: Dropped 18, Added 21
* PHP: Added 8.3
* Ruby: Added 3.3
* Python: Added 3.12 
* Rust: bumped to 1.76.0 to support wasmtime 17.


thresheek:
Looks like we have some problems in generating the library file since we moved Unit to git repo for this release.  We'll work on fixing it and then will update the PR with proper metadata.

tianon (approved):

PR #16325, opened 2024-02-29
Update spark version to 3.5.1

Yikun:
This patch add 3.5.1 version for Apache Spark https://spark.apache.org/docs/3.5.1/ [1]

[1] https://github.com/apache/spark-website/pull/502
[2] https://github.com/apache/spark-docker/pull/59

### DIFF
--- a/library/spark
+++ b/library/spark
@@ -1,7 +1,47 @@
 Maintainers: Apache Spark Developers <dev@spark.apache.org> (@ApacheSpark)
 GitRepo: https://github.com/apache/spark-docker.git
 
-Tags: 3.5.0-scala2.12-java17-python3-ubuntu, 3.5.0-java17-python3, 3.5.0-java17, python3-java17
+Tags: 3.5.1-scala2.12-java17-python3-ubuntu, 3.5.1-java17-python3, 3.5.1-java17, python3-java17
+Architectures: amd64, arm64v8
+GitCommit: 4f2d96a415c89cfe0fde89a55e9034d095224c94
+Directory: ./3.5.1/scala2.12-java17-python3-ubuntu
+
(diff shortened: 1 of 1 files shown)


Yikun:
cc @HeartSaVioR @HyukjinKwon

library/spark
```diff
@@ -1,7 +1,47 @@
...
+Architectures: amd64, arm64v8
+GitCommit: 4f2d96a415c89cfe0fde89a55e9034d095224c94
+Directory: ./3.5.1/scala2.12-java11-ubuntu
+
+Tags: 3.5.1-scala2.12-java11-python3-r-ubuntu
+Architectures: amd64, arm64v8
+GitCommit: 4f2d96a415c89cfe0fde89a55e9034d095224c94
+Directory: ./3.5.1/scala2.12-java11-python3-r-ubuntu
+
+Tags: 3.5.0-scala2.12-java17-python3-ubuntu, 3.5.0-java17-python3, 3.5.0-java17
```
yosifkit, line 44: Any particular reason to keep the `3.5.0` versions listed here?  Removing it here will remove it from the "Supported" section on the Hub readme (and will prevent us from spending cycles rebuilding it on the official build server), but the tag will still be available to users who want it.  (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)

yosifkit (approved):

library/spark
yosifkit: Merging to get `3.5.1` images building; `3.5.0` tags can be dropped in another PR if they are no longer supported.

Yikun: Thanks for reminder, I will submit another PR!

PR #16334, opened 2024-02-29
chore: upgrade chronograf maintainers

bnpfeife:


library/chronograf
```diff
@@ -1,5 +1,8 @@
 Maintainers: Jonathan A. Sternberg <jonathan@influxdata.com> (@jsternberg),
```
yosifkit, line 1: Did https://github.com/influxdata/influxdata-docker/pull/713 not work to drop `@jsternberg`? Is the `chronograf/manifest.json` not used to generate this library file?

bnpfeife: Ah my bad, I've removed him from the maintainers list. I haven't been using the manifest.json files for a while. I'll go through and remove them from the repo.

yosifkit (approved):

PR #16343, opened 2024-03-01
[Friendica] Release 2024.03-rc & fixings

nupplaphil:


tianon (dismissed):

tianon:
This looks fine, but I am a little confused about "fix IntlChar" in your title -- from the diff, this looks like it's _just_ the RC?

tianon:
I guess you intended to include https://github.com/friendica/docker/pull/243 also?

nupplaphil:
@tianon - thanks for the hint! Unfortunately we changed the fix at the wrong place (now fixed ;-) )

yosifkit (approved):

PR #16345, opened 2024-03-03
Update Clear Linux OS to 41150

clrbuilder:
Update Clear Linux OS latest+base images to release 41150

@bryteise @djklimes @bryteise


yosifkit:
The new binaries and files seem a little out of place. It looks to be related to virtual console and keyboard layout, but I'm not sure how useful those are in a container.  Are those correct for the image?

bryteise:
They aren't ideal but they go with the rest of the systemd parts that are in the image (splitting out systemd from the update mechanism that this image needs is more hassle than it is worth). The base image is largely useful to make a smaller/more targeted container image that doesn't have an updater in it.

yosifkit (approved):

PR #16349, opened 2024-03-04
chore(release): bump Kong to 3.6.1

samugi:


library/kong
```diff
@@ -1,7 +1,13 @@
...
 GitRepo: https://github.com/Kong/docker-kong.git
 
-Tags: 3.6.0-ubuntu, 3.6-ubuntu, 3.6.0, 3.6, 3, latest, ubuntu
+Tags: 3.6.1-ubuntu, 3.6-ubuntu, 3.6.1, 3.6, 3, latest, ubuntu
+GitCommit: 4dec46ee7e14ddd3a10692814728ff85adb77f25
+GitFetch: refs/tags/3.6.1
+Directory: ubuntu
+Architectures: amd64, arm64v8
+
+Tags: 3.6.0-ubuntu, 3.6.0
```
yosifkit, line 10: Any particular reason to keep this `3.6.0-ubuntu, 3.6.0` listed here? It seems like now that `3.6.1` is released, `3.6.0` is basically EOL. Removing it here will remove it from the "Supported" section on the Hub readme (and will prevent us from spending cycles rebuilding it on the official build server), but the tag will still be available to users who want it.  (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)

And I see that `3.1.*` and `3.2.*` [reached end of life](https://docs.konghq.com/gateway/latest/support-policy/#older-versions), so they should also be dropped.

samugi: thanks @yosifkit , updated.

tianon (approved):

PR #3485, opened 2017-09-22
Update Swift to 4.0

rymcol:
Updates the GitCommit for Swift 4.0 as tags 4.0, latest. Also removes the latest tag from Swift 3.

This would be my first attempt at updating an official image, feedback is welcome.

PR #16366, opened 2024-03-06
[Swift] Add support for Swift 5.10

shahmishal:


yosifkit (approved):

library/swift
```diff
@@ -6,55 +6,109 @@ Maintainers: Ted Kremenek <kremenek@apple.com> (@tkremenek),
...
 Directory: 5.9/rhel-ubi/9
 
-Tags: 5.9.2-rhel-ubi9-slim, 5.9-rhel-ubi9-slim, rhel-ubi9-slim
+Tags: 5.9.2-rhel-ubi9-slim, 5.9-rhel-ubi9-slim
 Architectures: amd64, arm64v8
 GitCommit: 51f0941d5f6d92af9df8aa89fa75841f5036c127
 Directory: 5.9/rhel-ubi/9/slim
 
-Tags: 5.9.2-windowsservercore-ltsc2022, 5.9-windowsservercore-ltsc2022, windowsservercore-ltsc2022
+Tags: 5.9.2-windowsservercore-ltsc2022, 5.9-windowsservercore-ltsc2022
```
yosifkit, line 111: Similar to https://github.com/docker-library/official-images/pull/3485#pullrequestreview-66902113, any particular reason to keep all the older `5.x` versions listed here (`5.2.x`-`5.8.x`)? Are they still supported? Removing it here will remove it from the "Supported" section on the Hub readme (and will prevent us from spending cycles rebuilding it on the official build server), but the tag will still be available to users who want it.

shahmishal: Let me get back to you on this after we have a discussion with the Swift community, specifically the Swift Core team.

PR #16369, opened 2024-03-07
Refactor CODEOWNERS, add Laurent explicitly to several things 👀

tianon:
- "give Laurent a bit more responsibility / authority on images where he's comfortable and knowledgeable 🎉"
- "make sure we check with Laurent before we update our SBOM indexer (he follows and is involved in those releases more closely than the rest of us) 👀"

@LaurentGoderre is part of our team at Docker, and he's been doing good work getting up to speed on the program and image review, especially wrangling SBOMs and the scanner, so this seems like a solid way to signify that (and work him up to more over time) :smile: :eyes:

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,10 @@
-* @tianon @yosifkit
+*  @docker-library/maintainers
+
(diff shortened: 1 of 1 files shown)


whalelines (approved):
LGTM

yosifkit (approved):

PR #16387, opened 2024-03-11
Telegraf v1.30.0

powersj:
Thank you! Includes a fix to ensure user provided groups are also passed on.

### DIFF
--- a/library/telegraf
+++ b/library/telegraf
@@ -2,7 +2,7 @@ Maintainers: Josh Powers <jpowers@influxdata.com> (@powersj),
              Sven Rebhan <srebhan@influxdata.com> (@srebhan),
              Dane Strandboge <dstrandboge@influxdata.com> (@DStrand1)
 GitRepo: https://github.com/influxdata/influxdata-docker.git
-GitCommit: fb3b40fbabc1f3f9a98b50ee9f4affd0c05f832c
+GitCommit: 958e0e5c394a241368fc81ed4d4af9286d47cfcf
 
 
 Tags: 1.28, 1.28.5
--- a/library/telegraf
+++ b/library/telegraf
@@ -2,16 +2,8 @@ Maintainers: Josh Powers <jpowers@influxdata.com> (@powersj),
              Sven Rebhan <srebhan@influxdata.com> (@srebhan),
              Dane Strandboge <dstrandboge@influxdata.com> (@DStrand1)
 GitRepo: https://github.com/influxdata/influxdata-docker.git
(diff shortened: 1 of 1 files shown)


tianon:
That `sed` seems really fragile/brittle (very easy for something to go wrong and not notice) :grimacing:

I think I've deducted correctly that https://github.com/influxdata/influxdata-docker/issues/724#issuecomment-1960860373 is the underlying issue it's solving? (`--user root:something` and wanting to keep `something` after dropping privileges?)

It's not a blocker by any means (we're fine with merging the `sed` as-is), but maybe an approach like this would be more maintainable for you?

```bash
    # honor groups supplied via 'docker run --group-add ...' but drop 'root'
    # (also removes 'telegraf' since we unconditionally add it and don't want it listed twice)
    # see https://github.com/influxdata/influxdata-docker/issues/724
    groups="telegraf"
    extra_groups="$(id -Gn)"
    for group in $extra_groups; do
        case "$group" in
            root | telegraf) ;;
            *) groups="$groups,$group" ;;
        esac
    done
    exec setpriv --reuid telegraf --regid telegraf --groups "$groups" "$@"
```

powersj:
> I think I've deducted correctly that https://github.com/influxdata/influxdata-docker/issues/724#issuecomment-1960860373 is the underlying issue it's solving? (--user root:something and wanting to keep something after dropping privileges?)

yep exactly!

> It's not a blocker by any means (we're fine with merging the sed as-is), but maybe an approach like this would be more maintainable for you?

I like this, @jdstrand, thoughts?

jdstrand:
> > It's not a blocker by any means (we're fine with merging the sed as-is), but maybe an approach like this would be more maintainable for you?
> 
> I like this, @jdstrand, thoughts?

LGTM. Cleaner, thanks!

yosifkit (approved):

wz2b:
I mentioned this is the other issue - and this is already merged - but if the intent here is to always use the TELEGRAF user, then I think this needs to be:

`extra_groups="$(id -Gn telegraf || true)"`

A more flexible approach would probably be to honor UID.

tianon:
The intent is to add the groups of the user we're running as (minus `root`) to the process running as the `telegraf` user (so that `docker run --group-add foo` with the implied root user actually works and adds `foo` to the eventual telegraf process).

tianon:
Also, this whole section of code only runs (and only even _could_ work) if we're running as root, so `docker run --user` will already "do the right thing" here :+1:

Issue #16289, opened 2024-02-21
High severity vulnerability in `libexpat` packages pulled into Alpine


sxa:
Hi.

I had a discussion on an issue with @yosifkit recently about an [openssl issue on Ubuntu](https://github.com/docker-library/official-images/issues/16225#issuecomment-1942193224).

[A user of ours](https://github.com/adoptium/containers/issues/501) has reported a high severity vulnerability with the eclipse-temurin images, and the [dockerhub site for the image](https://hub.docker.com/layers/library/eclipse-temurin/21-alpine/images/sha256-82698e23d15ada036bc176f6fb210401e0679cd0a4b1e71d05e7329982d6062c?context=explore) is showing it as a high (CVSS7.5)

[Back in 2022](https://github.com/docker-library/official-images/issues/13259) there was a rebuild done for us by @tianon to resolve another vulnerability in the same package, but as per yosifkit's comment in the first referenced issue it would be ideal if all images using libexpat which are showing this vulnerability could get updated so they get 2.6.0-r0 unless there is a reason why this should not be considered a sufficiently high severity.

Since I don't think the Alpine images necessarily have a periodic refresh, and [Alpine do not seem to want to trigger anything under their control for this](https://github.com/alpinelinux/docker-alpine/issues/377) and we have no mechanism to do so, despite the following suggestion to the raiser in that iAlpine ssue:

> _If people are getting security warnings because they use eclipse-temurin, they probably need to wait for them to release a newer image that use the new expat version you mentioned._

I suspect the person making that comment may not be aware that it is not under our control to trigger such a rebuild. Given all of the above, is this a valid justification for a trigger of a rebuild of ours, and possibly other people's, Alpine images?

tianon:
> libexpat through 2.5.0 allows a denial of service (resource consumption) because many full reparsings are required in the case of a large token for which multiple buffer fills are needed.

Any idea how widespread use/indirect invocation of `libexpat` inside Eclipse Temurin is?

sxa:
Ref Ubuntu/Debian while I haven't looked into it myself, the user who reported this to us did assert:

> The debian/ubuntu based images got rebuilt a few days ago so the problem is no longer present there

But as you say a little surprising if they haven't noted anything related to that specific CVE on they web pages.

tianon:
[...]

PR #16410, opened 2024-03-14
eclipse-temurin: trigger rebuild of alpine images

sxa:
This PR is an update to the git commits for Alpine in an attempt to force a rebuild to resolve the high severity issues that are visible in the `eclipse-temurin` images.

While my preference would be for a one-off rebuild to be performed I believe this is the only way that we can trigger such a rebuild (Noting that the Dockerfiles are unchanged, this is purely to pull in the newer expat library which is causing the vulnerability notification)

Ref: https://github.com/docker-library/official-images/issues/16289

### DIFF
--- a/library/eclipse-temurin
+++ b/library/eclipse-temurin
@@ -8,7 +8,7 @@ GitFetch: refs/heads/main
 #------------------------------v8 images---------------------------------
 Tags: 8u402-b06-jdk-alpine, 8-jdk-alpine, 8-alpine
 Architectures: amd64
-GitCommit: 5ef325612fc29e529a339ea7e35e4183cf9ae888
+GitCommit: 2dae22b74703ce78b2951bc458945fdfde5b8f50
 Directory: 8/jdk/alpine
 
 Tags: 8u402-b06-jdk-focal, 8-jdk-focal, 8-focal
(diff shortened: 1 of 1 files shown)


yosifkit:
Although there are commit changes, the build system will not rebuild unless there are build context changes (i.e., Dockerfile and scripts/files from `COPY` lines). The build system currently relies heavily on `docker build` cache, so without something to bust the cache at the appropriate layer, it will result in the same image.  We don't typically accept Dockerfile changes just to bust build cache.

Even with the discussion in https://github.com/docker-library/official-images/issues/16289, I am mostly unconvinced that this CVE warrants an image rebuild. I.E., given that it is expat that is only pulled in as a dep for `fontconfig` and `ttf-dejavu`, are there java code paths that can definitely trigger the described denial of service in expat?

We have done forced rebuilds (https://github.com/docker-library/php/pull/1451#issuecomment-1809378133) in the past, but it is currently a very manual process to find and delete all the images and their dependents across the 8 architectures (that alpine is supported on) in order to clear docker build cache, so we try to do so sparingly.

yosifkit:
Ok, we've worked through the local build cache on the build servers, so rebuilds should be happening soon.

sxa:
Thank you @yosifkit !

PR #16430, opened 2024-03-17
Update Apache Flink to 1.19.0

lincoln-lil:
This adds the new Flink 1.19.0 release:  https://flink.apache.org/2024/03/18/announcing-the-release-of-apache-flink-1.19/

and updates the tags on the 1.18 images accordingly.

### DIFF
--- a/library/flink
+++ b/library/flink
@@ -3,17 +3,32 @@
 Maintainers: The Apache Flink Project <dev@flink.apache.org> (@ApacheFlink)
 GitRepo: https://github.com/apache/flink-docker.git
 
-Tags: 1.18.1-scala_2.12-java8, 1.18-scala_2.12-java8, scala_2.12-java8, 1.18.1-java8, 1.18-java8, java8
+Tags: 1.19.0-scala_2.12-java8, 1.19-scala_2.12-java8, scala_2.12-java8, 1.19.0-java8, 1.19-java8, java8
+Architectures: amd64,arm64v8
+GitCommit: 206aa670a6f70dee91ba4474768078450a263958
+Directory: ./1.19/scala_2.12-java8-ubuntu
+
+Tags: 1.19.0-scala_2.12-java17, 1.19-scala_2.12-java17, scala_2.12-java17, 1.19.0-java17, 1.19-java17, java17
+Architectures: amd64,arm64v8
+GitCommit: 206aa670a6f70dee91ba4474768078450a263958
+Directory: ./1.19/scala_2.12-java17-ubuntu
+
+Tags: 1.19.0-scala_2.12-java11, 1.19-scala_2.12-java11, scala_2.12-java11, 1.19.0-scala_2.12, 1.19-scala_2.12, scala_2.12, 1.19.0-java11, 1.19-java11, java11, 1.19.0, 1.19, latest
+Architectures: amd64,arm64v8
+GitCommit: 206aa670a6f70dee91ba4474768078450a263958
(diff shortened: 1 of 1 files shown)


yosifkit:
```bash
        config_params+=" -D${key}=${value}"
```
Can the `key` or `value` added to `config_params` have whitespace characters in them? Using `eval` is often fraught with rough edges, so it makes me wary of code using it. I'm not certain if the `eval` is trying to allow `key/value` to reference other variables on purpose or not to know whether it is necessary or if the string could be swapped for an array to keep the items as separate string arguments:

```diff
-    local config_params=""
+    local config_params=()
...
-        config_params+=" -D${key}=${value}"
+        config_params+=("-D${key}=${value}")
...
-    if [ ! -z "${config_params}" ]; then
-        eval "${config_parser_script} ${config_dir} ${bin_dir} ${lib_dir} ${config_params}"
+    if [ "${#config_params[@]}" -gt 0 ]; then
+        "${config_parser_script}" "${config_dir}" "${bin_dir}" "${lib_dir}" "${config_params[@]}"
```

lincoln-lil:
Hi [Junrui Lee](https://github.com/JunRuiLee) could you help to confirm yosifkit's comment on these changes (refers to https://github.com/apache/flink-docker/pull/175)?

JunRuiLee:
Thank @yosifkit  for your insightful comments. I agree that using eval can be risky, especially when dealing with parameters that might include whitespace or special characters. I'll update the script as suggested.

lincoln-lil:
@yosifkit Thank you for reviewing this and thanks @JunRuiLee for the fix!
I've updated the pr include the above changes. Please take a look when you have time : )

yosifkit (approved):

tianon:
What's the purpose of the `if` statement in the `Dockerfile`?  Why would some builds have the configuration file already and some not?

tianon:
(also https://github.com/docker-library/official-images/pull/16114#issuecomment-1906744745)

JunRuiLee:
> What's the purpose of the `if` statement in the `Dockerfile`? Why would some builds have the configuration file already and some not?

Good question! Let me try to explain that for you.

The `if` statement in the `Dockerfile` ensures compatibility with both the new standard YAML configuration file, `config.yaml`, introduced in FLINK-1.19, and the legacy configuration file, `flink-conf.yaml`. To guarantee forward compatibility, Flink continues to support the legacy `flink-conf.yaml`. When the new `config.yaml` is absent, Flink automatically falls back to retrieve the legacy `flink-conf.yaml`. As a result, there are two different code paths in the Dockerfile.

lincoln-lil:
@tianon Thanks for your reviewing! 

> What's the purpose of the `if` statement in the `Dockerfile`? Why would some builds have the configuration file already and some not?
As @JunRuiLee explained above, this is a compatibility step for upgrading the standard yaml configuration file in Flink. 
After upgrading to version 2.0, the old configuration files will no longer be supported, but they can be handled compatibly in the current 1.19 version.  


Regarding the comment you left during the 1.18.1 release(https://github.com/docker-library/official-images/pull/16114#issuecomment-1906744745), this might be a case of forgetting to follow up 😅. 
I have created a Jira https://issues.apache.org/jira/browse/FLINK-34746 to ensure that the next release 
switches to the new CDN address, do you think that's okay? (Because considering that Flink 1.19.0 has been
officially announced on 18/03, users are already asking about the ready time of the Docker image)

PR #16433, opened 2024-03-18
Update Couchbase Server CE+EE to 7.6.0

udkyo:
Adding new Couchbase Server release.

### DIFF
--- a/library/couchbase
+++ b/library/couchbase
@@ -1,30 +1,40 @@
 Maintainers: Couchbase Docker Team <docker@couchbase.com> (@cb-robot)
 GitRepo: https://github.com/couchbase/docker
 
-Tags: 7.2.4, enterprise-7.2.4, enterprise, latest
-GitCommit: 24bae6aab6520dc97c362b56150db9658d041873
+Tags: 7.6.0, enterprise-7.6.0, enterprise, latest
+GitCommit: 203e62f36ce055be75955e073d0000249bf2c040
+Directory: enterprise/couchbase-server/7.6.0
+Architectures: amd64, arm64v8
+
+Tags: community-7.6.0, community
+GitCommit: 7148bc241bcdcf8961c34f8adf76ebfb2860ed83
+Directory: community/couchbase-server/7.6.0
+Architectures: amd64, arm64v8
+
+Tags: 7.2.4, enterprise-7.2.4
+GitCommit: 8205e18d139813ac3f72f6dfef9ee9ffbc382e0f
 Directory: enterprise/couchbase-server/7.2.4
 Architectures: amd64, arm64v8
 
-Tags: community-7.2.4, community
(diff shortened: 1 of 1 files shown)


yosifkit:
In Docker Official Images, we try hard to avoid multi-stage builds in almost every instance. Even though we maintain the gcc image, I'm not sure I'd recommend using it to build binaries compatible with a specific libc of a distribution; it'd be better to just use the gcc package from the distribution. So, we would rather that it be just a regular non-multistage build (i.e., `apt install`, build, and `apt purge` all in one `RUN`). It will help make it clearer for users viewing docker history where things came from. See also the FAQ section on [multi-stage builds](https://github.com/docker-library/faq/tree/76862506fde541f964df742db3b5abe549d9beb6#multi-stage-builds). You also should pin the `git clone` to get a specific commit id so it has better reproducibility.

I am mildly concerned about using a custom fork of `runit`. Would it confuse users when they use other images containing the same tools? 🤔

ceejatec:
To your last point, IMHO it's a non-goal for a container image to provide tools for general use. This container image is intended to host Couchbase Server, and it contains the necessary tools for that. Users should not generally be running their own programs inside the container at all, let alone depending on the specific behaviour of the 'init' program in use.

As for the use of the gcc image, I think I understand your concern. I believe I can modify it to be a single-stage build, although it'll be difficult to ensure that it doesn't end up with significant additional "cruft" in that runit build layer since we'll need to remove all of gcc and anything it brought in. It would be far easier and safer to make it a multi-stage build and just ensure we use the same FROM line for both stages, rather than using the separate "gcc" image. What's the reasoning behind avoid multi-stage builds in general?

All that said, this is the Dockerfile we've internally tested with Server 7.6.0, which is now released, and it's the Dockerfile that was used to create our "unofficial" couchbase/server:7.6.0 image. Making either of those changes will take at least a few days to implement and validate, and we're already nearly 48 hours past our 7.6.0 release date. Would it be acceptable to commit this version for 7.6.0, with the understanding that we'll correct the issues for 7.6.1?

I also agree that we should check out a specific SHA of our runit fork; I'll make that change in the follow-up Dockerfile as well.

whalelines:
> What's the reasoning behind avoid multi-stage builds in general?

@ceejatec that is addressed in the FAQ @yosifkit pointed you to:

>> See also the FAQ section on [multi-stage builds](https://github.com/docker-library/faq/tree/76862506fde541f964df742db3b5abe549d9beb6#multi-stage-builds).

ceejatec:
@whalelines @yosifkit Example #2 of the "likely acceptable" use cases under the FAQ entry seems to exactly match our scenario.

ceejatec:
@yosifkit @whalelines I've pushed an update which includes the changes you suggested; please re-review.

yosifkit (approved):

PR #16437, opened 2024-03-18
[security fix] varnish: update to 7.5.0, 7.4.3 and 6.0.13

gquintard:
https://varnish-cache.org/security/VSV00014.html#vsv00014

PR #16446, opened 2024-03-19
varnish: update tags

gquintard:
sorry, in https://github.com/docker-library/official-images/pull/16437 I forgot to update the tags...


yosifkit:
I think some architecture-specific images might have finished building already; did you want to re-add the old tags to fix the image contents?

yosifkit (approved):

PR #16447, opened 2024-03-19
Added a new "otel" variant for nginx mainline.

thresheek:
This adds an opentelemetry module variant developed by F5/NGINX: https://nginx.org/en/docs/ngx_otel_module.html

I've decided to make a separate variant based on the main image instead of extending it because the module build-depends on a fairly large chunk of C++ code from multiple projects, which takes around 10 minutes to compile and link on an 8-core amd64 machine.  This is why it's currently limited to amd64 and arm64v8, which nginx.org provides builds for. Users can build them on less popular architectures as the instructions are still provided in the dockerfiles.

Also, it's currently only available for the "mainline" branch, with "stable" to follow in the future.


yosifkit (approved):

tianon:
@thresheek I hope you know how delightful you are to work with :smile: :heart:

PR #16475, opened 2024-03-25
Update tomcat

yosifkit:
Changes:

- https://github.com/docker-library/tomcat/commit/1ef6e6c: Update 8.5 to 8.5.100
- https://github.com/docker-library/tomcat/commit/76e168e: Update 10.1 to 10.1.20


tianon:
Note for users: looking at https://tomcat.apache.org/tomcat-85-eol.html, this is _very_ likely to be the final 8.5 release of Tomcat.

See also https://lists.apache.org/thread/bf8gx7bckyj32qjfwf34flmjp38r9h4q, especially:

> This will likely be the final release of Apache Tomcat 8.5.

tianon (approved):

PR #16479, opened 2024-03-26
Update elk 7.17.19

navyau09:


tianon:
> This updates ELK images to point to the original source repository so they build directly and updates the maintainers appropriately.

This is not accurate :sweat_smile: (it appears to just be a copy of my comment from https://github.com/docker-library/official-images/pull/15808)

In the future, feel free to leave the description blank if you don't have anything important to add or point out about the new releases (pointing to the release notes would also be very appropriate).

tianon:
On the `Dockerfile`s, I would note that `apt-get clean` is a no-op in the `ubuntu` based images, and that `apt-get clean literally-anything-here` is the same as `apt-get clean` (any other arguments are simply ignored).

tianon (approved):

mark-vieira:
> On the `Dockerfile`s, I would note that `apt-get clean` is a no-op in the `ubuntu` based images, and that `apt-get clean literally-anything-here` is the same as `apt-get clean` (any other arguments are simply ignored).

Thanks. We use a templated Dockerfile which supports multiple package managers and lack a condition for this specific instance. I presume this is benign though, since we're calling it in the same layer as other commands which are appropriate so there's no real penalty for doing so?

tianon:
Yeah, it just doesn't do anything (successfully!) :smile: :+1:

PR #16490, opened 2024-03-27
eclipse-temurin: add jdk22 release

gdams:


yosifkit:
Adding `Builder: buildkit` seems a little odd, is there a specific feature that you want to use from the buildkit builder or just migrating early? In the coming months we hope to automatically move all images (except windows) over to using the buildkit builder, but it is still a work in progress.

gdams:
> Adding `Builder: buildkit` seems a little odd, is there a specific feature that you want to use from the buildkit builder or just migrating early? In the coming months we hope to automatically move all images (except windows) over to using the buildkit builder, but it is still a work in progress.

I was using it to test RISCV but happy to remove it, also noting that it seems to be faster than the old way of building

yosifkit (approved):

PR #16519, opened 2024-04-01
Update docker

yosifkit:
Changes:

- https://github.com/docker-library/docker/commit/9b7730d: Update 26 to compose 2.26.0


yosifkit:
Bad update (downgrade)

PR #16524, opened 2024-04-02
Updated mageia images

juanluisbaptiste:
Hi!

Long overdue, but here are the update for the mageia images. I'm also adding the latest version 9.


library/mageia
```diff
@@ -1,24 +1,24 @@
...
 
-Tags: 8,latest
-GitCommit: 067e6df19c568e101a42dc71b77f6a2de5992c70
+Tags: 9,latest
+GitCommit: 2365d6c0c7c87fb7e120348bac4c3f1bd1e21af7
 GitFetch: refs/heads/dist
 Architectures: amd64, arm32v7, arm64v8
 amd64-Directory: dist/8/x86_64
 arm32v7-Directory: dist/8/armv7hl
 arm64v8-Directory: dist/8/aarch64
```
yosifkit, line 10: This isn't quite correct. This is still pointing at the `8` directories on the `9` image. I think it just needs to swap to `dist/9/...` on all the `*-Directory` values.

```suggestion
amd64-Directory: dist/9/x86_64
arm32v7-Directory: dist/9/armv7hl
arm64v8-Directory: dist/9/aarch64
```

juanluisbaptiste: Ooops! fixed now :smile:

yosifkit:
No worries on the diff-pr comment failure. It's just too long for a comment. I'll generate it tomorrow into a gist.

yosifkit:
Diff for 79007e681e60c5e717a393c44646b53a1b9fa0d8: https://gist.github.com/docker-library-bot/3ded1a6b4a6edb2fb560dc9a7d784e93

yosifkit (approved):

yosifkit:
> While we don't actively recommend using labels in Docker Official Images, if an image maintainer wants to have labels, that is fine, but Image label names should adhere to the image spec: https://github.com/opencontainers/image-spec/blob/v1.1.0/annotations.md.
>
> \- https://github.com/docker-library/official-images/issues/3540#issuecomment-530925319

juanluisbaptiste:
> > While we don't actively recommend using labels in Docker Official Images, if an image maintainer wants to have labels, that is fine, but Image label names should adhere to the image spec: https://github.com/opencontainers/image-spec/blob/v1.1.0/annotations.md.
> > - [#3540 (comment)](https://github.com/docker-library/official-images/issues/3540#issuecomment-530925319)

Sure, I will remove them.

PR #16531, opened 2024-04-03
emqx 5.6.0

id:


tianon:
```diff
+    apt-get clean; \
```

This doesn't actually do anything in the base images :sweat_smile:

https://github.com/debuerreotype/debuerreotype/blob/c5abe5d7e2c6db0a153e6068ce6a8f2ba2c29415/scripts/debuerreotype-minimizing-config#L87-L109

(it's a ~harmless no-op though, so not a blocker if you'd like us to merge as-is)

id:
Got it, thanks. Let's merge as is, I'll fix it in the next version.

yosifkit (approved):

PR #16546, opened 2024-04-05
Update bonita, add 2024.1

rbioteau:


yosifkit:
```Dockerfile
FROM eclipse-temurin:17.0.10_7-jre-alpine
```
Unless you want to update the image on every `eclipse-temurin:17.*` minor update, we recommend using a more generic like `eclipse-temurin:17-jre-alpine` so that we can automatically rebuild the `bonita` image (for example, if there is a `17.0.10_8` release).

rbioteau:
> ```dockerfile
> FROM eclipse-temurin:17.0.10_7-jre-alpine
> ```
> 
> Unless you want to update the image on every `eclipse-temurin:17.*` minor update, we recommend using a more generic like `eclipse-temurin:17-jre-alpine` so that we can automatically rebuild the `bonita` image (for example, if there is a `17.0.10_8` release).

Thanks for the suggestion, we were on the reproducible side of things but is sounds reasonable enough for us after all.

LaurentGoderre:
@rbioteau you can still pin to a specific sha to stay with a specific combination

LaurentGoderre:
The image gets a SHA but it doesn't check the integrity of the package against it. It seems like it didn't check in previous versions either.

tianon:
Instead of copying the contents of Temurin's entrypoint, you should invoke `/__cacert_entrypoint.sh` at the appropriate point; something like this would work:

```dockerfile
ENTRYPOINT ["/__cacert_entrypoint.sh", "/opt/files/startup.sh"]
```

Alternatively, you could make sure all your uses of `exec ...` become `exec /__cacert_entrypoint.sh ...` but I think this change is the cleanest.

tianon:
```diff
+  host=$(echo $(hostname -i) | cut -f1 -d ' ' | xargs)
```

Also, this is kind of weird -- `echo $(hostname -i)` instead of just `hostname -i` seems to be trying to make it space-separated, and the prior pipe to `xargs` would then do the same, but now with the `cut` this just keeps getting more and more confusing.

Do you have an example use case for what this code is trying to solve?

(We used to use `hostname -i` in Cassandra, but ended up swapping to parsing `ip address` because of limitations in that approach: https://github.com/docker-library/cassandra/pull/146, https://github.com/docker-library/cassandra/blob/3b22e7fc1b586c619beffaa0081238a445a526db/docker-entrypoint.sh#L17-L31)

rbioteau:
@LaurentGoderre 

> The image gets a SHA but it doesn't check the integrity of the package against it. It seems like it didn't check in previous versions either.

Are you talking about the Bonita archive ? The checksum check is performed [here](https://github.com/bonitasoft/bonita-distrib/blob/dcac85a2670e9adfa99c057c3468b9923dc31301/docker/Dockerfile#L38C32-L38C73)

LaurentGoderre:
@rbioteau my bad, I haven't seen this way of validating before

PR #16550, opened 2024-04-06
Add lts-and-current images to allow Gradle to use Java 22

keeganwitt:
As noted in the [compatibility guide](https://docs.gradle.org/current/userguide/compatibility.html#java), Gradle supports compiling with Java 22 using toolchains, but doesn't yet support executing Gradle. This means Gradle must be launched with a supported version of Java (such as the last LTS release) and then use toolchains to configure compilation using a second JDK on the image. I've tried to document this [here](https://github.com/keeganwitt/docker-gradle?tab=readme-ov-file#lts-and-current-images).

### DIFF
--- a/library/gradle
+++ b/library/gradle
@@ -6,63 +6,78 @@ GitRepo: https://github.com/keeganwitt/docker-gradle.git
 
 Tags: 8.7.0-jdk8, 8.7-jdk8, 8-jdk8, jdk8, 8.7.0-jdk8-jammy, 8.7-jdk8-jammy, 8-jdk8-jammy, jdk8-jammy
 Architectures: amd64, arm32v7, arm64v8, ppc64le
-GitCommit: 9a1868d61f289a711881d5af0299308968b0fc65
+GitCommit: 5ca02d087b9989b0d54bc4a62a797c86c1b77d38
 Directory: jdk8
 
 Tags: 8.7.0-jdk8-focal, 8.7-jdk8-focal, 8-jdk8-focal, jdk8-focal
 Architectures: amd64, arm32v7, arm64v8, ppc64le
-GitCommit: 9a1868d61f289a711881d5af0299308968b0fc65
+GitCommit: 5ca02d087b9989b0d54bc4a62a797c86c1b77d38
 Directory: jdk8-focal
 
 Tags: 8.7.0-jdk11, 8.7-jdk11, 8-jdk11, jdk11, 8.7.0-jdk11-jammy, 8.7-jdk11-jammy, 8-jdk11-jammy, jdk11-jammy
 Architectures: amd64, arm32v7, arm64v8, ppc64le, s390x
-GitCommit: 9a1868d61f289a711881d5af0299308968b0fc65
+GitCommit: 5ca02d087b9989b0d54bc4a62a797c86c1b77d38
 Directory: jdk11
 
 Tags: 8.7.0-jdk11-focal, 8.7-jdk11-focal, 8-jdk11-focal, jdk11-focal
 Architectures: amd64, arm32v7, arm64v8, ppc64le, s390x
-GitCommit: 9a1868d61f289a711881d5af0299308968b0fc65
+GitCommit: 5ca02d087b9989b0d54bc4a62a797c86c1b77d38
 Directory: jdk11-focal
 
 Tags: 8.7.0-jdk11-alpine, 8.7-jdk11-alpine, 8-jdk11-alpine, jdk11-alpine
 Architectures: amd64
-GitCommit: 9a1868d61f289a711881d5af0299308968b0fc65
+GitCommit: 5ca02d087b9989b0d54bc4a62a797c86c1b77d38
 Directory: jdk11-alpine
 
 Tags: 8.7.0-jdk17, 8.7-jdk17, 8-jdk17, jdk17, 8.7.0-jdk17-jammy, 8.7-jdk17-jammy, 8-jdk17-jammy, jdk17-jammy
 Architectures: amd64, arm32v7, arm64v8, ppc64le, s390x
-GitCommit: 9a1868d61f289a711881d5af0299308968b0fc65
+GitCommit: 5ca02d087b9989b0d54bc4a62a797c86c1b77d38
 Directory: jdk17
 
 Tags: 8.7.0-jdk17-focal, 8.7-jdk17-focal, 8-jdk17-focal, jdk17-focal, 8.7.0-jdk-focal, 8.7-jdk-focal, 8-jdk-focal, jdk-focal, 8.7.0-focal, 8.7-focal, 8-focal, focal
 Architectures: amd64, arm32v7, arm64v8, ppc64le, s390x
-GitCommit: 9a1868d61f289a711881d5af0299308968b0fc65
+GitCommit: 5ca02d087b9989b0d54bc4a62a797c86c1b77d38
 Directory: jdk17-focal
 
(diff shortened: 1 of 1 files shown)


tianon:
I'm concerned about adding a new "lts-and-current" variant/alias and the support burden that will be for you over time for what's _hopefully_ a temporary problem. :sweat_smile: :grimacing:

I think I'm understanding this correctly, but just to confirm, the problem this is solving is that users want to use Gradle to compile their applications against JDK 22, but Gradle itself will not run on 22, correct?

If my understanding of the problem is correct, wouldn't it be simpler to have the `jdk22` images just quietly embed 21 (JRE-only, even?) for now, so Gradle itself can be run on 21 but 22 can be used for compiling?

(Apologies if I've misunderstood or have oversimplified the complexity here :smile: :heart:)

keeganwitt:
> I'm concerned about adding a new "lts-and-current" variant/alias and the support burden that will be for you over time for what's _hopefully_ a temporary problem. 😅 😬

It's a recurring "temporary" problem. We had users asking for Java 20, 21 (which was LTS), and now 22 support. Gradle always lags in supporting the newest Java, often by months. And for non-LTS releases, they're only supported 6 months, so that means the window of time in which you have to test new Java versions with Gradle might be only a few months before it's no longer supported. I share the concern about adding more maintenance burden (I waited when the Java 21 question came up and it took far longer than I expected for support to land in Gradle).

> I think I'm understanding this correctly, but just to confirm, the problem this is solving is that users want to use Gradle to compile their applications against JDK 22, but Gradle itself will not run on 22, correct?

Yep, exactly.

> If my understanding of the problem is correct, wouldn't it be simpler to have the `jdk22` images just quietly embed 21 (JRE-only, even?) for now, so Gradle itself can be run on 21 but 22 can be used for compiling

That would be another valid solution for this. The reason I chose a new name was I wanted to make it clear this is different from any regular image, because I set global Gradle properties to make it easier to use toolchains for this use case, which I don't do for the other images. So whenever Gradle finally supports 22 fully (assuming they do so before September when 22 is EOL), I would release a normally named jdk22 tag which would only contain Java 22 and won't contain the global Gradle properties.

I suppose we could have the jdk22 image temporarily set the properties and have the other JDK installed, but I thought it was confusing and potentially breaking to fundamentally alter how an image operates.

I definitely welcome more thoughts on this. It's kinda a messy problem.

tianon:
Yeah, totally fair, that makes sense.  I'm not sure what the best answer is -- there's not a cleaner way to convince Gradle itself to use an installation of JRE 21 (maybe in a non-standard path?) and _not_ use/expose that for users without changing the global configuration like you've had to? :disappointed:

For the recurring case, is "current-and-lts" always a sufficient workaround?  In other words, might there eventually be a case where the latest LTS _and_ the latest "current" are both versions Gradle can't handle?

keeganwitt:
> there's not a cleaner way to convince Gradle itself to use an installation of JRE 21 (maybe in a non-standard path?) and not use/expose that for users without changing the global configuration like you've had to? 😞

If there is, I haven't found it yet. I'm fairly certain there isn't. The one place I'll double check is if the Gradle bin script has something in it I can use to customize which Java gets used.

> For the recurring case, is "current-and-lts" always a sufficient workaround? In other words, might there eventually be a case where the latest LTS and the latest "current" are both versions Gradle can't handle?

This actually will probably happen with Java 25. It will be the next LTS and when it comes out, the latest Gadle will not be able to be launched with 25, but would be able to with Java 21 and 24, but 24 won't be supported at that point, so it'd have to be 21 still. This is a case where the latest current and latest LTS are the same release. Hopefully that made sense. When that happens I was thinking the image would be 21 and 25 (since it will be 21 and 22, 23, 24 before that). Once Gradle supports 25, then it could be just 25 (or maybe we leave it as 21 and 25 if that's less confusing).

tianon:
Gah yeah, that's a messy situation :sob:

Looking through https://github.com/gradle/gradle/blob/3ceb6ffc8c33313642b02fd9ff048c0f59ad4a10/gradlew I don't see anything particularly helpful for this case -- I wonder if they'd be open to something like `${GRADLE_JAVA_HOME:-$JAVA_HOME}` or `: ${GRADLE_JAVA_HOME:=$JAVA_HOME}` and then replace all uses of `JAVA_HOME` :thinking:

keeganwitt:
The file you referenced is actually part of the wrapper, which usually isn't used with this image (it downloads Gradle and launches it). The launcher script I believe is https://github.com/gradle/gradle/blob/6d38851488eff56a6ab4b5a38f9ece3b4d421a5f/platforms/jvm/plugins-application/src/main/resources/org/gradle/api/internal/plugins/unixStartScript.txt. But I don't see anything we can manipulate there either, just `JAVA_HOME`.

tianon:
Yeah, it's got very similar logic. :disappointed:

keeganwitt:
https://docs.gradle.org/current/userguide/build_environment.html says that `org.gradle.java.home` is just for configuring the path to the JDK used for building, and not the client VM. `JAVA_HOME` seems to be the only option to configure the client VM. But even if we had a `GRADLE_JAVA_HOME`, we'd still have to configure `org.gradle.java.home` globally, since by default Gradle will use the client JDK for the daemon also.

keeganwitt:
I think the question of whether there's other ways to tell Gradle which JDK to use is a separate question from the maintenance burden of an lts-and-current image and the potential confusion it could cause. That seems kinda unavoidable given this support lag that Gradle has. Given that, should we consider moving forward with these images?

donalmurtagh:
> Given that, should we consider moving forward with these images?

I would appreciate it if someone could review/release these images. Given that Java has a 6-month release cadence, if we wait much longer, v22 will no longer be the latest.

yosifkit:
In `gradle:jdk-21-and-22-jammy` and `-alpine` the Dockerfiles set `JAVA_LTS_HOME`, but that path is non-existent in the base images, and I don't see a symlink created. Was it supposed to just point to `/opt/java/openjdk` or `$JAVA_HOME`?

```dockerfile
ENV JAVA_LTS_HOME /opt/java/openjdk21
```

keeganwitt:
> JAVA_HOME

Ah, good catch!

yosifkit (approved):

keeganwitt:
> Not a bocker but would it be possible to use more descriptive variable name and make them `ENV` to help readability. Something like
> 
> ```dockerfile
> ENV JDK_VERSION_LTS=21.0.2
> ENV JDK_VERSION_CURRENT=22.0.0
> ```

You mean to use in paths like `/opt/java/openjdk21`? Or you just wanted these new variables?

keeganwitt:
> Not a bocker but would it be possible to use more descriptive variable name and make them `ENV` to help readability. Something like
> 
> ```dockerfile
> ENV JDK_VERSION_LTS=21.0.2
> ENV JDK_VERSION_CURRENT=22.0.0
> ```

I can see why it might be useful to have this metadata, but the versions (except for the Graal image) aren't directly controlled by me and when the upstream Adoptium image gets a security patch for the JVM, we'd be shipping a version with a wrong version in the env var.

yosifkit:
If you wanted to condense it a little, you could combine the `FROM` and the `COPY`, but it is fine either way.
```diff
-FROM eclipse-temurin:22-jdk-alpine AS java_current
-
 FROM eclipse-temurin:21-jdk-alpine

-COPY --from=java_current /opt/java/openjdk /opt/java/openjdk22
+COPY --from=eclipse-temurin:22-jdk-alpine /opt/java/openjdk /opt/java/openjdk22
```

PR #16556, opened 2024-04-08
Update pypy

yosifkit:
Changes:

- https://github.com/docker-library/pypy/commit/d8bcd6c: Merge pull request https://github.com/docker-library/pypy/pull/83 from infosiftr/dpkg-query-search
- https://github.com/docker-library/pypy/commit/8b4ced2: Fix `dpkg-query --search` to be more specific

### DIFF
--- a/library/pypy
+++ b/library/pypy
@@ -6,23 +6,23 @@ GitRepo: https://github.com/docker-library/pypy.git
 
 Tags: 3.10-7.3.15-bookworm, 3.10-7.3-bookworm, 3.10-7-bookworm, 3.10-bookworm, 3-7.3.15-bookworm, 3-7.3-bookworm, 3-7-bookworm, 3-bookworm, bookworm
 Architectures: amd64, arm64v8, i386
-GitCommit: 6c5c41354e29aef40de9a1af9fe02c9ea39e0e77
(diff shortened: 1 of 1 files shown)


whalelines:
What's this story with all these awk script updates?

```diff
-		| awk '/=>/ { so = $(NF-1); if (index(so, "/usr/local/") == 1) { next }; gsub("^/(usr/)?", "", so); print so }' \
+		| awk '/=>/ { so = $(NF-1); if (index(so, "/usr/local/") == 1) { next }; gsub("^/(usr/)?", "", so); printf "*%s\n", so }' \
```

FWIW, `print "*" so` would have saved you a few characters.

tianon:
https://github.com/docker-library/python/pull/858 was the OG, we just didn't cycle through every usage of the pattern until now (and then prompted by https://github.com/docker-library/haproxy/pull/226 to finally do so)

tianon (approved):

tianon:
(I also realized while reviewing these today that doing `print "/" so; print "/usr/" so` would technically be even better / more correct, but decided to wait to cycle all those when/if we see an instance of an actual issue due to this still being more generic with the wildcard prefix)

PR #16567, opened 2024-04-08
Update rakudo-star to 2024.03

AntonOks:
Adding a bookworm tag for consistency and clarity and switching the FROM alpine Docker image to "alpine:latest"...

### DIFF
--- a/library/rakudo-star
+++ b/library/rakudo-star
@@ -4,12 +4,12 @@ Maintainers: Moritz Lenz <moritz.lenz@gmail.com> (@moritz),
              Anton Oks <anton.oks@gmx.de> (@AntonOks)
 GitRepo: https://github.com/Raku/docker
 
-Tags: latest, 2024.02
+Tags: latest, 2024.03, bookworm
 Architectures: amd64, arm64v8
-GitCommit: 1357b3acb550569e7ecf2da22fc32c32319fb872
-Directory: 2024.02/bookworm
+GitCommit: 64657bbc83c2536e3d2dcd8f075dde4e206b1421
+Directory: 2024.03/bookworm
 
-Tags: alpine, 2024.02-alpine
+Tags: alpine, 2024.03-alpine
 Architectures: amd64, arm64v8
-GitCommit: 1357b3acb550569e7ecf2da22fc32c32319fb872
-Directory: 2024.02/alpine3.19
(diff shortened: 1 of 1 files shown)


tianon:
Switching to the implicit `alpine:latest` is what I like to call a "square wheel" -- it's technically a rolling tag, but when that square wheel rolls, you might experience some unexpected breakage as a result, so we recommend being explicit (and bumping manually when the wheel rolls and you've done at least a surface-level verification of compatibility).

AntonOks:
> Switching to the implicit `alpine:latest` is what I like to call a "square wheel" -- it's technically a rolling tag, but when that square wheel rolls, you might experience some unexpected breakage as a result, so we recommend being explicit (and bumping manually when the wheel rolls and you've done at least a surface-level verification of compatibility).

@tianon

Thanks for your hint / opinion. We discussed pro's and con's in https://github.com/Raku/docker/pull/60 and part of that is my [thoughts on this case here](https://github.com/Raku/docker/pull/60#discussion_r1545792753).

From my understanding, the "rakudo-star" container was supported, if at all, on a "personal and best effort level". I doubt there were "verification" or the like efforts done...

tianon:
Ah nice, thank you for the references (and for already having thought about / discussed it) :heart:

tianon (approved):

AntonOks:
> Switching to the implicit `alpine:latest` is what I like to call a "square wheel" -- it's technically a rolling tag, but when that square wheel rolls, you might experience some unexpected breakage as a result, so we recommend being explicit (and bumping manually when the wheel rolls and you've done at least a surface-level verification of compatibility).

Finally, it's about automating merge requests to this repo here... based on GitHub workflows in our "rako/docker" repo... If you have some nice-and-easy examples from other images here, which I / we could just copy-paste to get our rakudo-star "offical-images" up2date, including the "FROM" images of our Dockerfiles, I would be fine either way ;)

And finally, from my understanding and I may be very wrong here... our "bookwork" also points to a "somehow latest stable" and not an "explicit" version, right:
- https://github.com/Raku/docker/blob/master/2024.03/bookworm/Dockerfile

tianon:
`FROM buildpack-deps:bookworm-scm` is explicitly Debian Bookworm (so when Trixie is released, it will _not_ automatically update :smile:)

For automating image update PRs, you'll want to review https://github.com/docker-library/faq#can-i-use-a-bot-to-make-my-image-update-prs :+1:

PR #16572, opened 2024-04-09
Add JDK 22 clojure variants

cap10morgan:

PR #16579, opened 2024-04-10
Update clojure

Quantisan:

### DIFF
--- a/library/clojure
+++ b/library/clojure
@@ -2,7 +2,7 @@ Maintainers: Paul Lam <paul@quantisan.com> (@Quantisan),
              Wes Morgan <wes@wesmorgan.me> (@cap10morgan)
 Architectures: arm64v8, amd64
 GitRepo: https://github.com/Quantisan/docker-clojure.git
-GitCommit: 31b7d3f1003eeafd155bc504e78f8a515655ea48
+GitCommit: 513d5b7ce920767aa872ef8c3a411757b53fd448
 
 
 Tags: latest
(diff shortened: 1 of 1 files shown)


cap10morgan:
Apologies, I had misunderstood the situation in #16572. The missing upstream images were older distro versions that will never get new JDK versions. This PR should only build on valid upstream images.

alexander-yakushev:
Apparently, `eclipse-temurin:22-jdk-jammy` doesn't have `curl` (while `eclipse-temurin:21-jdk-jammy` had it). Nice.

cap10morgan:
> Apparently, `eclipse-temurin:22-jdk-jammy` doesn't have `curl` (while `eclipse-temurin:21-jdk-jammy` had it). See https://github.com/adoptium/containers/issues/255.

I'm working on a fix on our end.

alexander-yakushev:
Here's my solution, you may consider taking it if you aren't too far ahead with yours yet https://github.com/Quantisan/docker-clojure/pull/221

tianon:
I would suggest also adding `-f`/`--fail` to all these `curl` invocations, but given they're all immediately followed by a checksum validation, I don't think it's a blocking concern (just something to keep in mind for the next update). :+1:

tianon:
(While you're at it, you might want to add `-S`/`--show-error` as well, which updates `-s`/`--silent`'s behavior to actually print error messages :sob:)

tianon (approved):

cap10morgan:
Great suggestions, thanks @tianon!

PR #16583, opened 2024-04-10
Almalinux auto-update - 20240410 17:47:25

almalinuxautobot:
This is an auto-generated commit. Any concern or issues, please contact or email AlmaLinux OS Foundation cloud-infra@almalinux.org (@AlmaLinux)

### DIFF
--- a/library/almalinux
+++ b/library/almalinux
@@ -1,64 +1,39 @@
-# This file is generated using https://github.com/almalinux/docker-images/blob/b27b2d450d0db9bd0495c92a1b79c4cfa29d35bd/gen_docker_official_library
+# This file was generated on https://github.com/AlmaLinux/container-images/actions/runs/8635583348
 Maintainers: The AlmaLinux OS Foundation <cloud-infra@almalinux.org> (@AlmaLinux)
-GitRepo: https://github.com/AlmaLinux/docker-images.git
+GitRepo: https://github.com/AlmaLinux/container-images.git
 
-
-Tags: 8, 8.9, 8.9-20231124
-GitFetch: refs/heads/al8-20231124-amd64
-GitCommit: b5c3f2dea7120eee009287009173081dfc079e86
-amd64-File: Dockerfile-x86_64-default
-arm64v8-GitFetch: refs/heads/al8-20231124-arm64v8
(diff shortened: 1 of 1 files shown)


andrewlukoshko:
This PR is created by our new GitHub Actions pipeline from https://github.com/AlmaLinux/container-images
We switched to Fedora-like compact library definition file and maintain 8 and 9 branches with rootfs tarballs and Dockerfiles for AlmaLinux 8 and 9 respectively:
https://github.com/AlmaLinux/container-images/tree/8
https://github.com/AlmaLinux/container-images/tree/9

yosifkit:
🤔❓ why the move from `https://github.com/AlmaLinux/docker-images` to `https://github.com/AlmaLinux/container-images` and not a GitHub rename to keep issues, PRs, and history (plus a URL redirect)?

With the `tar.xz` files in many old branches, did the repo just get too big that GitHub didn't like it?  I think @tianon has worked around this issue for `debian` rootfs files by only keeping the most recent build per arch+release (plus it speeds up cloning and other git operations). On the plus side, old commits are still directly fetchable from GitHub, they just aren't part of the repository tree.

andrewlukoshko:
@yosifkit The reason to change repo is completely new pipeline to build and push images redesigned from scratch. We kept Dockerfiles so resulting images are the same with the same packages/files but changed everything else.
Previous pipeline was complicated and hard to maintain, and also relies on private Jenkins, some of its unmaintained plugins and Groovy scripts.
The new one is very easy and relies on GitHub Actions only. This also makes the whole pipeline public which is important for Open Source project.
All old issues, pull requests, and history are irrelevant to the new pipeline as well so it's not a problem.
We're already using this new repo and pipeline to build UBI-like images for Docker user repo https://hub.docker.com/u/almalinux and other registries and would like to use it as well to push PRs to Docker Library. This will allow us to archive our old https://github.com/AlmaLinux/docker-images repo completely.

Sorry that this change makes this particular PR harder to review but the following ones will be easy again.

yosifkit (approved):

yosifkit:
One quick note is to make sure that the next almalinux build doesn't keep the old `tar.xz` files in the git history. See [this comment](https://github.com/docker-library/official-images/pull/179#issuecomment-56233346) for context.

PR #16595, opened 2024-04-12
Releasing Neo4j 5.19.0

glindroth:


LaurentGoderre:
What is the purpose of this line:

```
sed -i 's/Package Type:.*/Package Type: docker ubi8/' $NEO4J_HOME/packaging_info; \
```

jennyowen:
@LaurentGoderre 
Inside Neo4j there's a text file called `packaging_info` which has diagnostic information about which version of neo4j is installed and by what installation method. It's useful for understanding and replicating bugs and such. 
Anyway, I noticed that for our docker images it says that neo4j was installed from tar, and doesn't mention that it runs in Docker, so I thought it would be a good idea to change the packaging info file to say docker + base image.

 Basically it's a small change that can be undone if deemed necessary. It just fixes some minor diagnostic data.

yosifkit (approved):

PR #16597, opened 2024-04-12
Update mongo-express

LaurentGoderre:
Cleans the yarn cache preventing dev dependencies in the image.

https://github.com/mongo-express/mongo-express-docker/commit/075234363b3c008c7c2e97de8a3639e23a75cfcd


tianon (approved):

tianon:
FYI the `node` image no longer supports Alpine 3.18 _or_ 3.19 so we're due for another bump of Alpine versions here :eyes:

PR #16620, opened 2024-04-16
TomEE Added 10.0.0-M1, 9.1.3, dropped 8.x

scriptmonkey:


library/tomee
```diff
@@ -1,193 +1,164 @@
...
...emurin-microprofile, 9.1.3-Temurin-microprofile, 9.1.3-jre11-microprofile, 9.1.3-microprofile, 9.1.3-jre11-Temurin-ubuntu, 9.1.3-Temurin-ubuntu, 9.1.3-jre11-ubuntu, 9.1.3-ubuntu, 9.1.3-jre11-Temurin, 9.1.3-Temurin, 9.1.3-jre11, 9.1.3, 9.0-jre11-Temurin-ubuntu-microprofile, 9.0-Temurin-ubuntu-microprofile, 9.0-jre11-ubuntu-microprofile, 9.0-ubuntu-microprofile, 9.0-jre11-Temurin-microprofile, 9.0-Temurin-microprofile, 9.0-jre11-microprofile, 9.0-microprofile, 9.0-jre11-Temurin-ubuntu, 9.0-Temurin-ubuntu, 9.0-jre11-ubuntu, 9.0-ubuntu, 9.0-jre11-Temurin, 9.0-Temurin, 9.0-jre11, 9.0, 9-jre11-Temurin-ubuntu-microprofile, 9-Temurin-ubuntu-microprofile, 9-jre11-ubuntu-microprofile, 9-ubuntu-microprofile, 9-jre11-Temurin-microprofile, 9-Temurin-microprofile, 9-jre11-microprofile, 9-microprofile, 9-jre11-Temurin-ubuntu, 9-Temurin-ubuntu, 9-jre11-ubuntu, 9-ubuntu, 9-jre11-Temurin, 9-Temurin, 9-jre11, 9, jre11-Temurin-ubuntu-microprofile, Temurin-ubuntu-microprofile, jre11-ubuntu-microprofile, ubuntu-microprofile, jre11-Temurin-microprofile, Temurin-microprofile, jre11-microprofile, microprofile, jre11-Temurin-ubuntu, Temurin-ubuntu, jre11-ubuntu, ubuntu, jre11-Temurin, Temurin, jre11, latest
```
tianon, line 6: This (and following lines) have a lot of `9.0` in them that I think should be `9.1`? (it's hard to see because there's just _so many_ tags, but I think that's what we're seeing in the sorted tag list in our diff :+1:)

scriptmonkey: @tianon excellent catch.   One day I'll get tags right the 1st time. lol.

Incoming commit.

tianon: Might be worth scripting, if you're doing these updates by hand :sweat_smile:

(I know I'd sure be nervous if I were maintaining this matrix by hand! :joy:)

scriptmonkey: I have a script that gets me 90% of they way.  I just  neglected to change 9.0 to 9.1.  Sometimes it's the simple things.

yosifkit (approved):

PR #16623, opened 2024-04-17
Update the API Firewall to v0.7.2

afr1ka:


tianon (approved):

tianon:
Once 0.7.1 is built/pushed, you'll probably want to send a follow-up to remove it (maybe tomorrow to be sure?) :+1: :heart:

afr1ka:
Hello @tianon,
We are currently working on another release. It will be available in 2 weeks. I think it would be better to remove 0.7.1 along with the next release. So as not to bother you twice. 
Thanks.

PR #16639, opened 2024-04-22
Fedora 2024-04-22

cverna:
Fedora 40 is now the latest


cverna:
> `layer.tar` as a filename for gzip-compressed content is a little odd, but should be mostly harmless 😄

Yeah, we change the tooling to build the images for Fedora 40+.  I ll change this so that it matches what we had for older releases :smile:

PR #16646, opened 2024-04-23
[security] Update ruby

tianon:
Changes:

- https://github.com/docker-library/ruby/commit/a2b957f: Merge pull request https://github.com/docker-library/ruby/pull/446 from infosiftr/security
- https://github.com/docker-library/ruby/commit/53646d3: [security] Update to 3.3.1, 3.2.4, 3.1.5, 3.0.7

> After this release, Ruby 3.0 reaches EOL. In other words, this is expected to be the last release of Ruby 3.0 series. We will not release Ruby 3.0.8 even if a security vulnerability is found (but could release if a severe regression is found). We recommend all Ruby 3.0 users to start migration to Ruby 3.3, 3.2, or 3.1 immediately.

https://www.ruby-lang.org/en/news/2024/04/23/ruby-3-0-7-released/

### DIFF
--- a/library/ruby
+++ b/library/ruby
@@ -4,117 +4,117 @@ Maintainers: Tianon Gravi <admwiggin@gmail.com> (@tianon),
              Joseph Ferguson <yosifkit@gmail.com> (@yosifkit)
 GitRepo: https://github.com/docker-library/ruby.git
 
-Tags: 3.3.0-bookworm, 3.3-bookworm, 3-bookworm, bookworm, 3.3.0, 3.3, 3, latest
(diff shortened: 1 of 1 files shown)


LaurentGoderre (approved):

LaurentGoderre:
Minor issue is that any upstream image using the `RUBY_MAJOR` env that isn't there anymore would break. The name of the variable is a bit odd though.

tianon:
Only on 3.1 and 3.0 where that change hadn't applied yet.  We could make a pretty good argument for skipping it on these two releases (especially 3.0), but I think we should get this in first and then worry about that since we're not removing anything for the EOL yet so a follow up is pretty easy.

PR #16650, opened 2024-04-23
Update clojure

Quantisan:


cap10morgan:
We're removing our boot variants as it is no longer being updated, sees very little usage, and we want to consolidate the two remaining build tool variants into images that install both of them.

PR #16652, opened 2024-04-23
Updated nginx

thresheek:
Updated nginx stable to 1.26.0 and moved to Alpine 3.19 / Debian Bookworm.  This also adds otel module variant for stable as was previously introduced for mainline.

Mainline is also updated to hopefully fix njs 0.8.4+quickjs build on arm32v5.

With that, some simplifications and minor fixes to Dockerfiles were merged.

### DIFF
--- a/library/nginx
+++ b/library/nginx
@@ -1,64 +1,74 @@
-# this file is generated via https://github.com/nginxinc/docker-nginx/blob/9cb278860bdcea48abc0bc770a29ead3fc9a1fe6/generate-stackbrew-library.sh
+# this file is generated via https://github.com/nginxinc/docker-nginx/blob/435f9c2a06cf2d5fd65767d2f0696b9c988b814b/generate-stackbrew-library.sh
 
 Maintainers: NGINX Docker Maintainers <docker-maint@nginx.com> (@nginxinc)
 GitRepo: https://github.com/nginxinc/docker-nginx.git
 
 Tags: 1.25.5, mainline, 1, 1.25, latest, 1.25.5-bookworm, mainline-bookworm, 1-bookworm, 1.25-bookworm, bookworm
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
-GitCommit: 29d5001c15e7a2c7b34402c35254cc55199f8cc8
+GitCommit: 66c0f94b4c764b1a531528e8c242ad94497011f6
 Directory: mainline/debian
 
 Tags: 1.25.5-perl, mainline-perl, 1-perl, 1.25-perl, perl, 1.25.5-bookworm-perl, mainline-bookworm-perl, 1-bookworm-perl, 1.25-bookworm-perl, bookworm-perl
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
-GitCommit: 29d5001c15e7a2c7b34402c35254cc55199f8cc8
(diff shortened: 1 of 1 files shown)


tianon:
Everything looks really nice except this guy:

```diff
-                && curl -f -O https://hg.nginx.org/pkg-oss/archive/${NGINX_VERSION}-${PKG_RELEASE}.tar.gz \
-                && PKGOSSCHECKSUM=\"74000f32ab250be492a8ae4d408cd63a4c422f4f0af84689973a2844fceeb8a3e7e12b04d7c6dac0f993d7102d920a5f60e6f49be23ce4093f48a8eb1ae36ce5 *${NGINX_VERSION}-${PKG_RELEASE}.tar.gz\" \
-                && if [ \"\$(openssl sha512 -r ${NGINX_VERSION}-${PKG_RELEASE}.tar.gz)\" = \"\$PKGOSSCHECKSUM\" ]; then \
+                && curl -f -O https://hg.nginx.org/pkg-oss/archive/93ac6e194ad0.tar.gz \
+                && PKGOSSCHECKSUM=\"d56d10fbc6a1774e0a000b4322c5f847f8dfdcc3035b21cfd2a4a417ecce46939f39ff39ab865689b60cf6486c3da132aa5a88fa56edaad13d90715affe2daf0 *93ac6e194ad0.tar.gz\" \
+                && if [ \"\$(openssl sha512 -r 93ac6e194ad0.tar.gz)\" = \"\$PKGOSSCHECKSUM\" ]; then \
```

That's a little weird, isn't it?  If this is expected, maybe you could add a comment that points to where this random commit comes from so it's easier to trace / validate?

thresheek:
Yeah... we don't really have a good way to tag separate module versions on the packaging repo for nginx...  Typically we get everything ready before the release happens so the modules will just share the tags used by an actual nginx release, but this time we had to apply an additional patch after a release, so...

This changeset ID corresponds to https://hg.nginx.org/pkg-oss/rev/93ac6e194ad0 which (hopefully) fixes the issue as reported by @yosifkit when njs FTBFS on arm32v5: https://github.com/nginxinc/docker-nginx/issues/881

We also sometimes do a similar change when we update njs versions not in the nginx release cycle:
 - https://github.com/nginxinc/docker-nginx/commit/ce8322287b7def616fd077ea21c9c045c50ac461#diff-e14594f6bb9b29ac0146c04e78ae33e8e92f2f00709fefab97359dea31b94d11R47
 - https://github.com/nginxinc/docker-nginx/commit/7f1ef355dea083761951da16ab02ea2c37addbdd#diff-e14594f6bb9b29ac0146c04e78ae33e8e92f2f00709fefab97359dea31b94d11L48
 - https://github.com/nginxinc/docker-nginx/commit/3e71631e0e757f13265e0bd1d1e30e937ad49347#diff-e14594f6bb9b29ac0146c04e78ae33e8e92f2f00709fefab97359dea31b94d11R47

Most of the time previously I used actual numbers are mercurial guarantees them to be unique, but lately resorted to use short sha id of changesets;  that's mostly to get used to sha ids, since we're moving our nginx-related projects and infra over to github from mercurial..

Hope this clarifies things!

tianon:
Ah yeah, so this is temporary?  That makes sense.  I guess anyone familiar with Mercurial could walk back from that to the URL for browsing the repo anyhow and figure out which branch it came from. :+1:

thresheek:
Yep, we'll revert back to the usual `${NGINX_VERSION}-${PKG_RELEASE}` when 1.27.0 is released (which is probably next month).

tianon (approved):

PR #16653, opened 2024-04-24
Update Debian (esp. for CVE-2024-2961)

tianon:
See https://github.com/debuerreotype/docker-debian-artifacts/issues/216 (especially Buster users!!)


tianon:
Diff contains the expected versions for bookworm, bullseye, and sid/unstable. 👍

yosifkit (approved):

tianon:
Just to reiterate again explicitly: this does _not_ include Testing/Trixie (which is not officially "security supported"), and does _not_ include Buster/oldoldstable.

> Since Buster is not fixed yet, it will not be included in this rebuild -- I don't think this is severe enough to either wait for it or to rebuild again in a day or two if it gets an update, so users who know they're on Buster should be looking at mitigations 👀 🙈

> (Also, even Buster LTS ends on June 30th, 2024, so uh, also be looking to upgrade your systems 😬)

PR #16661, opened 2024-04-24
Add Couchbase Server EE 7.2.5

ceejatec:


tianon:
reviewer's note: runit install block is identical to the one in https://github.com/docker-library/official-images/pull/16433 :+1: :heart:

(FWIW, `apt-get clean` is a no-op in both the `debian` and `ubuntu` base images because they already do that automatically on every `apt-get install`, but it's harmless to include)

tianon (approved):

PR #16676, opened 2024-04-27
feat: Add s390x to rust images

Muscraft:
This adds the `s390x` architecture to some Debian versions for Rust. 

See: [rust-lang/docker-rust#198](https://github.com/rust-lang/docker-rust/pull/198)


tianon (approved):
Ooh, exciting!  Let's make some mainframes Rusty! :joy: :heart:

PR #16698, opened 2024-04-30
[8.13.2] Update ELK images

navyau09:
Update elk images

### DIFF
--- a/library/elasticsearch
+++ b/library/elasticsearch
@@ -3,10 +3,10 @@ GitRepo: https://github.com/elastic/dockerfiles.git
 Directory: elasticsearch
 Builder: buildkit
 
-Tags: 8.13.0
+Tags: 8.13.1
 Architectures: amd64, arm64v8
 GitFetch: refs/heads/8.13
-GitCommit: 4269a02cf8a56844db30e5e74b23087840ee6fcb
+GitCommit: 4fdbd17dca428a5c64601f380656c124b116fe6f
 
 Tags: 7.17.20
 Architectures: amd64, arm64v8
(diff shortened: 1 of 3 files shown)


LaurentGoderre:
@navyau09  there is still the issue that using the `arch` commands gives the kernel architecture and not the container. It should be switched to using `dpkg --print-architecture` or `apk --print-arch`. As mentioned here: https://github.com/docker-library/official-images/pull/16520#issuecomment-2043738623

navyau09:
@LaurentGoderre Could you please verify again, if this looks good?

LaurentGoderre (approved):

LaurentGoderre:
Can we close #16697 in favor of this?

tianon:
This is good for now, but please do revisit https://github.com/elastic/logstash/pull/16063#discussion_r1577000627 (adding at the very least `set -e` to all semi-colon-using `RUN` lines, if not the full `set -eux` I suggest there).

tianon:
```diff
-RUN \
- curl -Lo - https://artifacts.elastic.co/downloads/logstash/logstash-8.13.0-linux-$(arch).tar.gz | \
+RUN curl -Lo - https://artifacts.elastic.co/downloads/logstash/logstash-8.13.3-linux-$(arch).tar.gz | \
```

Wait a minute, this is 8.13.3 but the tags are 8.13.2 - I think that's a mistake?

(We should also apply the same "avoiding `arch` in favor of querying the image userspace environment" logic to these lines, but they're pre-existing usage so I feel less strongly about enforcing that for now.)

LaurentGoderre:
Good catch @tianon . It seems only logstash got the bump.

PR #16699, opened 2024-04-30
Update julia

yosifkit:
Changes:

- https://github.com/docker-library/julia/commit/92a88f4: Update 1.10 to 1.10.3


tianon (approved):
CI failures are unfortunately not unexpected :see_no_evil:

PR #16700, opened 2024-04-30
[ros] jazzy 0.10.0-4 -> 0.11.0-1

mikaelarguedas:
Bump package versions for ROS Jazzy.

Initial image build failed due to version bump between image release and image build : https://doi-janky.infosiftr.net/job/multiarch/job/amd64/job/ros/

Connected to https://github.com/osrf/docker_images/pull/742


tianon (approved):
Nice catch, thanks for being proactive about it! :rocket:

PR #16718, opened 2024-05-03
Update Monica to v4.1.2

asbiin:
Update Monica to [v4.1.2](https://github.com/monicahq/monica/releases/tag/v4.1.2).


parkr:
[v4.1.2](https://github.com/monicahq/monica/releases) was just released. Should we jump straight to that?

tianon:
Actual environment variables set on the container will override the `.env` values, right? (they're just used as defaults?)

As to 4.1.2, that depends on whether you care about whether there's a 4.1.1 tag available - I don't think it's _terribly_ important (and thus you should feel free to go straight to 4.1.2 if you'd like), but you might feel otherwise. :+1:

asbiin:
Thank you @parkr I'll update this version right now.

asbiin:
> Actual environment variables set on the container will override the `.env` values, right? (they're just used as defaults?)

Exactly. The `.env` values are default one that make the docker run with the best values, but any environment variable will have the priority over those values.

tianon (approved):

PR #16719, opened 2024-05-03
Update ALT Linux - 20240503

FNDenisovna:


whalelines:
Looks like a lot of files are moving from `/` to `/usr`. Is that intentional?

Obirvalger:
Yes, we did usrmerge.

tianon (approved):
Hopefully the dropped architectures are also intentional/expected (arm32v7 and ppc64le) -- if not, easy enough to fix in a follow-up. :+1:

PR #16734, opened 2024-05-07
Update UBI pins

yosifkit:


tianon (approved):
Unix timestamp based version numbers, that's a new one! :smile:

(https://catalog.redhat.com/software/containers/ubi9/ubi-minimal/615bd9b4075b022acc111bf5?architecture=amd64&image=6633b5136d81a5634a616c41; "9.4-949.1714662671")

PR #16743, opened 2024-05-08
Update mageia

juanluisbaptiste:
Update mageia 9 and cauldron, and remove mageia 8 (EOL).


yosifkit:
It looks like the Dockerfiles refer to a different `tar.xz` filename than is in the commit: https://github.com/juanluisbaptiste/docker-brew-mageia/commit/29b615c11a0a5851f07f05a7c369e7e2bd9b8b85.

```console
Step 2/3 : ADD rootfs.tar.xz /
ADD failed: file not found in build context or excluded by .dockerignore: stat rootfs.tar.xz: file does not exist
```

juanluisbaptiste:
Yes, I just noticed it too, a change on the build script had this unintended effect that went unnoticed.

juanluisbaptiste:
@yosifkit fixed!

tianon (approved):

Issue #527, opened 2015-03-03
Promotion of Fedora base image built in our build system "as is"


vpavlin:
Hi,

Problem:
We build Fedora base images in our Koji build system [1]. They are not only rootfs but complete (i.e. load-able) Docker images [2]. To get them to Docker Hub registry as official images, we need to extract layer.tar, upload it to github and then use Dockerfile with Automated builds [3]. This means we lose all metadata added during build - mainly the image ID at the moment, could be f.e. labels in the future if the PR gets merged.

It also requires non-trivial amount of work (where most of it could be probably automated, TBH) to get from "hey, there is new image in Koji" to "hey, there is new image in Docker Hub", which is mostly done by @lsm5 at the moment.

Proposed solution:
One solution we thought about at DockerCon Europe could be that we would push our image under fedora/* repository and send a link to Koji build to you, guys. You could then review that those 2 match and promote the pushed image between official images - would this work? What are the challenges you and us would have to deal with?

Is there any other solution you could think of?

Thanks,
Vašek

[1] http://koji.fedoraproject.org/koji/ 
[2] http://koji.fedoraproject.org/koji/taskinfo?taskID=9038227
[3]  https://github.com/fedora-cloud/docker-brew-fedora

tiborvass:
Hmmm, @tianon ?

tianon:
I'm definitely interested in improving this, especially with the future possibility of having signed artifacts stay signed all the way through.

One major problem I see with accepting these kinds of tarballs directly is that review gets much harder since we're then reviewing a discrete tarball instead of a source repository (which is more manageable when it's just the rootfs, since problems there are usually obvious).  Problems like what we discussed on the changes in #497 would have a much higher potential for slipping through since they'd be baked inside said tarball, and then they'd be causing the issues we discussed there out in the wild instead of being caught earlier (which is the whole point of the review process).

vpavlin:
@tianon I don't think this is an issue. We would definitely provide all the sources - ks file is already available in Koji and we could make metadate also available if that's your requirement - i.e. the review process could stay, just would happen after build.

tianon:
It's more the logistical problem of reviewing/verifying raw binary
artifacts than much of anything else.  Even if the source is available,
we'd need a way to easily verify that the artifacts actually match the
source (ie right now, we only review the source because we then build the
artifacts from that source, so we have the implicit guarantee that the two
match).

As the process stands right now, we can do 95+% of our review of most PRs
[...]

PR #16754, opened 2024-05-10
Update rocky linux containers for 9.4 and 8.10

NeilHanlon:


NeilHanlon:
This will need to be kicked, probably -- my branches are still uploading

NeilHanlon:
marking draft; need to rebuild Base image for regression

NeilHanlon:
@gionn it's ready now! :)

tianon:
Are `/image`, `/run`, and `/dev` all included and non-empty intentionally?  (The first seems like something related to how you're building these, the second is just to verify, the third often causes issues with certain runtimes that cope poorly with non-empty `/dev` inside images.)

NeilHanlon:
besides /image, the others are owned by the `filesystem` package on Enterprise Linux, so I think they are expected here, though I do agree they probably should be empty. I will give that a look today and see.

We did change how we are building images for 9.x in this release, so I know there are some deltas that we're still working to clean up. /images, for example, comes from the image build process, but probably can be removed.

I also noticed we're leaving a machine-id in, which we probably shouldn't.

tianon:
Friendly ping; what's the status here? :bow: :heart:

jamshid:
Sorry, don't mean to complain as I don't know of any security issues in 8.9. But it seems weird that the rockylinux:8 image is still 8.9 when 8.10 was released over 4 months ago?

anderbubble:
You can find the images released by the RESF at https://hub.docker.com/r/rockylinux/rockylinux. They are up-to-date.

jamshid:
> You can find the images released by the RESF at https://hub.docker.com/r/rockylinux/rockylinux. They are up-to-date.

Ah thanks I was pulling `rockylinux:8`. Didn't realize new images are only published under the `rockylinux/` path `rockylinux/rockylinux:8`.

```
% docker run -ti --pull always rockylinux:8 cat /etc/os-release
8: Pulling from library/rockylinux
Digest: sha256:9794037624aaa6212aeada1d28861ef5e0a935adaf93e4ef79837119f2a2d04c
Status: Image is up to date for rockylinux:8
NAME="Rocky Linux"
VERSION="8.9 (Green Obsidian)"
```

```
% docker run -ti --pull always rockylinux/rockylinux:8 cat /etc/os-release
8: Pulling from rockylinux/rockylinux
584b9ab5a0f2: Pull complete 
Digest: sha256:e8a49c5403b687db05d4d67333fa45808fbe74f36e683cec7abb1f7d0f2338c6
Status: Downloaded newer image for rockylinux/rockylinux:8
NAME="Rocky Linux"
VERSION="8.10 (Green Obsidian)"
```

yosifkit:
Apologies, I was waiting for a follow up to https://github.com/docker-library/official-images/pull/16754#issuecomment-2147455473. Once the conversations here are resolved we can merge this PR, and the updated image will be published.

jamshid:
@yosifkit is it _intended_ that image `rockylinux:8` is (eventually) the same as `rockylinux/rockylinux:8` ? 
I'm wondering if I need to change my image name references in build scripts and Dockerfiles or if I can leave them as `rockylinux:8`.

NeilHanlon:
Thanks all for your patience, have been tied up with other things, and, as the updated images are in the rocky namespace, has not been prioritized.

@jamshid - @yosifkit is just the docker-library maintainer. I am the maintainer of the rocky container images.

In general, the intention is that the rocky images are eventually consistent with the library ones, but the official source will always be the rockylinux namespaces on docker hub or quay.io. (and in the future, on our own registry).

The docker library images will continue to be updated, but at more of a best effort speed. Not all the images we create are eligible for docker library (e.g., UBI/toolkit images), so you'll always need to use the rockylinux namespace for those.

While the content of the base and minimal images between the library and rockylinux namespaces are identical, the manifests _are_ different, due to how the images are composed in this repository.

tianon:
> The docker library images will continue to be updated, but at more of a best effort speed.

This is honestly somewhat concerning -- should we deprecate https://hub.docker.com/_/rockylinux and point existing users to https://hub.docker.com/r/rockylinux/rockylinux instead?

(To be clear, this isn't my ideal case -- I'd much rather see it updated/maintained here. :heart:)

> While the content of the base and minimal images between the library and rockylinux namespaces are identical, the manifests _are_ different, due to how the images are composed in this repository.

To that end, you might be interested in https://github.com/docker-library/official-images/issues/527#issuecomment-2089150649 (which describes how to switch to our `oci-import` builder that would allow you to maintain identical digests between images elsewhere and images here).

tbugfinder:
Indeed for image consumers this is confusing and inconsistent. Either remove rockylinux:<tag> or sync it with rockylinux/rockylinux.

NeilHanlon:
> >The docker library images will continue to be updated, but at more of a best effort speed.

> This is honestly somewhat concerning -- should we deprecate https://hub.docker.com/_/rockylinux and point existing users to https://hub.docker.com/r/rockylinux/rockylinux instead?

> (To be clear, this isn't my ideal case -- I'd much rather see it updated/maintained here. ❤️)

Agreed with you, though that does become slightly at odds with comments like [this](https://github.com/docker-library/official-images/pull/16754#issuecomment-2438759669), because I can't maintain parity between the Docker Library images and the ones in my namespace, given we have variants which are not currently permitted in the Official Library collection. Still, this doesn't mean we can't promote the subset of images to the Library collection, having them more consistent will help.

> >While the content of the base and minimal images between the library and rockylinux namespaces are identical, the manifests are different, due to how the images are composed in this repository.

> To that end, you might be interested in https://github.com/docker-library/official-images/issues/527#issuecomment-2089150649 (which describes how to switch to our oci-import builder that would allow you to maintain identical digests between images elsewhere and images here).

Indeed, that is interesting and perhaps that's a model that would be easier to handle than the current one in a future with better automation around this on my end.

NeilHanlon:
> Indeed for image consumers this is confusing and inconsistent. Either remove rockylinux: or sync it with rockylinux/rockylinux.

If all of rockylinux/rockylinux (UBI images, e.g.) cannot be synced, would you rather see a subset synced to library/rockylinux?

It's also worth noting that the rocky linux team intends to introduce its own registry endpoint in the near future as a centralized point to sync images to other registries, so there's probably some mix of things which can happen here, but I want to give a consistent experience.

matt-undo:
This seems stalled. Is it possible to resuscitate it? Also, we have Rocky 9.5 images too now. Will those be added?

PR #16764, opened 2024-05-13
New SapMachine images and tags

RealCLanger:
We want to add support for all active Ubuntu versions and hence also need to add some tags.

### DIFF
--- a/library/sapmachine
+++ b/library/sapmachine
@@ -1,82 +1,212 @@
 Maintainers: Christoph Langer <sapmachine@sap.com> (@realclanger), Christian Halstrick <sapmachine@sap.com> (@chalstrick)
 GitRepo: https://github.com/SAP/SapMachine-infrastructure.git
 
-Tags: 11-jre-headless-ubuntu, 11-jre-headless-ubuntu-jammy, 11-jre-headless-ubuntu-22.04, 11.0.23-jre-headless-ubuntu, 11.0.23-jre-headless-ubuntu-jammy, 11.0.23-jre-headless-ubuntu-22.04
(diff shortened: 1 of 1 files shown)


yosifkit:
With Ubuntu `23.10` going end of life in July, does it make sense to add it now?

tianon:
Agreed, 23.10 doesn't seem like a good base given the option of _both_ surrounding LTS releases (unless there's something compelling about 23.10 that we're missing?)

It also seems a little strange to add new 20.04 variants so far into its life (and only a year away from the EOL), but less objectionable if you feel strongly about supporting it.

RealCLanger:
I just added all Ubuntu releases that are in support currently and the plan is to drop these once they are out of standard maintenance. As for 23.10, we had created dockerfiles but forgot to add them here. Since it is superseded by 24.04 now, I guess we could altogether skip it. Will update the PR.

yosifkit:
It seems that `jre` & `jre-headless` is not here for `11`, `17`, and `21` on Ubuntu 20.04; is that correct and on purpose?

Everything else looks correct, so I think we can merge. If these are accidentally missing, they can be added in another PR.

yosifkit (approved):

tianon:
> I just added all Ubuntu releases that are in support currently and the plan is to drop these once they are out of standard maintenance.

In general, I would suggest skipping all non-LTS releases of Ubuntu unless there is a really compelling use case for doing otherwise (for example, perhaps the latest LTS does not have a version of OpenSSL or similar that is _required_ by SapMachine, and thus a newer release is required), _especially_ because their support windows are dramatically shorter.

RealCLanger:
> > I just added all Ubuntu releases that are in support currently and the plan is to drop these once they are out of standard maintenance.
> 
> In general, I would suggest skipping all non-LTS releases of Ubuntu unless there is a really compelling use case for doing otherwise (for example, perhaps the latest LTS does not have a version of OpenSSL or similar that is _required_ by SapMachine, and thus a newer release is required), _especially_ because their support windows are dramatically shorter.

Let's see. Non LTS versions usually have one year of support. A reason that we might want to have them is that they could have fewer vulnerability problems.

tianon:
Ubuntu very actively supports their LTS releases, and Canonical culturally treats CVEs as "all hands on-deck" events, so I would be very skeptical of a CVE argument for non-LTS Ubuntu releases (in fact, I might even try to argue the opposite -- because it's not LTS, it will likely be a lower priority task queue for getting CVE fixes out promptly).

PR #16797, opened 2024-05-21
[nats] Release v2.10.16

bruth:
Details can be found [here](https://github.com/nats-io/nats-server/releases/tag/v2.10.16)

### DIFF
--- a/library/nats
+++ b/library/nats
@@ -4,26 +4,26 @@ Maintainers: Derek Collison <derek@synadia.com> (@derekcollison),
              Neil Twigg <neil@synadia.com> (@neilalexander),
              Phil Pennock <pdp@synadia.com> (@philpennock)
 GitRepo: https://github.com/nats-io/nats-docker.git
-GitCommit: b39950ece3064f20a2a5e25806d093793e455769
+GitCommit: 8d800ec8480ff944f4ee16619ef84082bdc14cf5
 GitFetch: refs/heads/main
 
-Tags: 2.10.14-alpine3.19, 2.10-alpine3.19, 2-alpine3.19, alpine3.19, 2.10.14-alpine, 2.10-alpine, 2-alpine, alpine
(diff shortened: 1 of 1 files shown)


bruth:
@yosifkit reviewers used to automatically get assigned. Has the workflow changed for anyone on the Docker team to review and merge the PRs? If so, understood and I will not ping you :)

tianon:
(I wish the teams themselves could be public, _especially_ when their existence is already public due to being part of the appropriate file: https://github.com/docker-library/official-images/blob/6a521d5304d3fbb1fe4f7e4f08c8aef76d5e3478/.github/CODEOWNERS#L1 :upside_down_face:)

bruth:
@tianon Ah missed that, thanks. I see `a team` was assigned.

tianon:
I think it's probably fine to leave it, but I'd just note the warning that `setcap` might not round-trip the graph drivers successfully (so I'd consider the `setcap` addition to the `Dockerfile` mostly best-effort).

I'm guessing that's probably to be able to bind to "privileged" ports, which is already enabled by default for containers in recent Docker versions (since in a private network namespace, the concept of "privileged ports" doesn't really hold much meaning), so there's other mitigations in place already. :+1:

tianon (approved):

tianon:
You might also enjoy https://github.com/moby/moby/issues/8460 and https://github.com/moby/moby/pull/41030 (which is the PR that finally fixed this for the engine :partying_face:).

bruth:
@tianon It turns out that adding this setcap change broke a few people when pulling the image for more restrictive environments. We reverted the change in the Dockerfile. The NATS binary did not change of course, so we are safe to open a new PR with an update to the GitCommit to the new nats-docker version to re-build/push the image for 2.10.16?

yosifkit:
Yeah, that's fine.

PR #16805, opened 2024-05-22
[nats] Release v2.10.16 - fix image

bruth:
This is a follow-up to #16797 to remove a setcap call in the Dockerfile which caused issues for users in more constrained environments.

The NATS binary remains the same.


yosifkit (approved):

bruth:
Thanks @yosifkit!

PR #16821, opened 2024-05-23
MariaDB 2024 Q2 part 1 release

grooverdan:
The new 11.4 release is a little late going though QA so it will occur later. In the mean time we keep the 11.3 as the latest tag so it isn't a downgrade.


yosifkit (approved):

grooverdan:
Thanks @yosifkit

PR #16834, opened 2024-05-25
Update drupal

tianon:
Changes:

- https://github.com/docker-library/drupal/commit/bbcaf7d: Move "rc" alias to 11 pre-releases (instead of 10.3)
- https://github.com/docker-library/drupal/commit/bfbda3f: Merge pull request https://github.com/docker-library/drupal/pull/261 from infosiftr/alpine3.20
- https://github.com/docker-library/drupal/commit/21e8f7b: Add 10.3 and 11.0 pre-releases
- https://github.com/docker-library/drupal/commit/75145d9: Update to alpine 3.20

### DIFF
--- a/library/drupal
+++ b/library/drupal
@@ -1,8 +1,128 @@
-# this file is generated via https://github.com/docker-library/drupal/blob/afe9df8e4e007f6f30a2f1e22909929104d55ca6/generate-stackbrew-library.sh
+# this file is generated via https://github.com/docker-library/drupal/blob/bbcaf7d512f79bec6f48e17671366bbc20f87d78/generate-stackbrew-library.sh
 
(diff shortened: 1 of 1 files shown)


LaurentGoderre:
Is there a reason why the rc alias goes to the version 10 rc vs the version 11 rc since there are two different rc here?

tianon:
Ah good catch -- because that alias is still maintained manually, and had been updated previously/pre-emptively to point to 10.3-rc but then we got 11 and didn't update it.  Fix incoming.

tianon:
Fixed in https://github.com/docker-library/drupal/commit/bbcaf7d512f79bec6f48e17671366bbc20f87d78, commit here incoming.

yosifkit (approved):

PR #16835, opened 2024-05-25
Update wordpress

tianon:


yorkshireboy6:
where do i find the ref for setting up a contain on aws? eg "wordpress 5.1.2"

LaurentGoderre:
@tianon this is only to add the new arch?

tianon:
> @tianon this is only to add the new arch?

Yep, this is "the parent image now has a new architecture, so that trickles down"

yosifkit (approved):

PR #16843, opened 2024-05-26
Update photon image

gerrit-photon:
Update photon image


zyyw:
@tianon do we have a timeline when this PR will be merged?

yosifkit (approved):

Issue #16618, opened 2024-04-16
Odoo release number in tag


LaurentGoderre:
I am wondering if we should add the odoo release in a tag. For example:

```
17.0-20240416
16.0-20240416
15.0-20240416
```

LaurentGoderre:
@d-fence

d-fence:
Well, why not. I think that it could help some users to pin a particular release.

PR #16848, opened 2024-05-27
Odoo: Create tag for release numbers

LaurentGoderre:
/cc @d-fence

### DIFF
--- a/library/odoo
+++ b/library/odoo
@@ -2,15 +2,15 @@ Maintainers: Christophe Monniez <moc@odoo.com> (@d-fence)
 GitRepo: https://github.com/odoo/docker
 GitCommit: ea74918b211215176bda85854a771be1c5c97871
 
-Tags: 17.0, 17, latest
+Tags: 17.0-20240522, 17.0, 17, latest
 Architectures: amd64, arm64v8, ppc64le
 Directory: 17.0
 
-Tags: 16.0, 16
+Tags: 16.0-20240522, 16.0, 16
 Architectures: amd64, arm64v8, ppc64le
 Directory: 16.0
 
-Tags: 15.0, 15
+Tags: 15.0-20240522, 15.0, 15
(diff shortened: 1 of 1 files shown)


LaurentGoderre:
Resolves #16618

LaurentGoderre:
@whalelines how so?

LaurentGoderre:
That I am not sure. I just augmented what was there

library/odoo
```diff
@@ -2,15 +2,15 @@ Maintainers: Christophe Monniez <moc@odoo.com> (@d-fence)
 GitRepo: https://github.com/odoo/docker
 GitCommit: ea74918b211215176bda85854a771be1c5c97871
 
-Tags: 17.0, 17, latest
+Tags: 17.0, 17, 17.0-20240522, 17-20240522, latest
 Architectures: amd64, arm64v8, ppc64le
 Directory: 17.0
 
-Tags: 16.0, 16
+Tags: 16.0, 16, 16.0-20240522, 16-20240522
```
yosifkit, line 9: We don't rely on it often, but ordering the tags from most specific version to least specific version should minimize the diff here and in future PRs. We use the last tag in the list (the least specific) to help the diff minimize the "rename from -> rename to" output.
```suggestion
Tags:  16.0-20240522, 16-20240522, 16.0, 16
```

d-fence:
hmm 🤔 will we have to accumulate the date tags in the file or just replace it ?

LaurentGoderre:
@d-fence replace it.

d-fence:
I'm ok with that then. Thanks for the PR

yosifkit:
Yep, the old tags will stay on Docker Hub regardless of whether or not it is in the `library/odoo` file, so a new date tag each time is fine and expected with this change.

yosifkit (approved):

PR #16866, opened 2024-05-29
caddy: v2.8.0 release

francislavoie:
https://github.com/caddyserver/caddy/releases/tag/v2.8.0

### DIFF
--- a/library/caddy
+++ b/library/caddy
@@ -1,98 +1,52 @@
 # this file is generated with gomplate:
 # template: https://github.com/caddyserver/caddy-docker/blob/e05f66898af9a0a694af637db2724f91d9d82ed7/stackbrew.tmpl
-# config context: https://github.com/caddyserver/caddy-docker/blob/b0a0cdbc0537af7b8ee85b480262414d2179455f/stackbrew-config.yaml
+# config context: https://github.com/caddyserver/caddy-docker/blob/8e45629b3f219609ebd9ae2dfb70df0e73765b5f/stackbrew-config.yaml
 Maintainers: Dave Henderson (@hairyhenderson),
              Francis Lavoie (@francislavoie)
 
-Tags: 2.8.0-rc.1-alpine, 2.8-alpine
-SharedTags: 2.8.0-rc.1, 2.8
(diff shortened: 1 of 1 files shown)


tianon:
Sure!  Currently, the only "meaningful" versions to support are really only the two you've got (LTSC2022 and 1809) -- we don't officially support anything beyond that here (yet).  It's important to support both if you want users to be able to use the images effectively because they essentially have to match the image OS version to their host OS version (or at least "level of compatibility").  In newer versions of Windows, Microsoft _is_ working to relax this (see https://learn.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/version-compatibility for example), but it's going to take a bit before that work hits Windows Server versions in a meaningful way, and even then there'll always be a transition period. :smile:

(1809 is effectively Windows Server 2019)

tianon (approved):

francislavoie:
Okay, thanks, good to know. Dependabot has been annoying us about bumping it up but I wasn't comfortable with doing it without knowing the implications. Glad we didn't then!

tianon:
Oh interesting -- I guess I'll go check your recent PRs to see for myself, but I'm curious what it's suggesting you bump _to_ :sweat_smile:

tianon:
Oh man, https://github.com/caddyserver/caddy-docker/pull/344 is so hilariously wrong I can't believe it's recommending it with a straight face -- Windows is popular enough that I think they need to rush to implement some special cases for those recommendations! :grimacing:

francislavoie:
We might just have it configured incorrectly, I dunno. But yeah :grimacing:

tianon:
https://github.com/caddyserver/caddy-docker/blob/31e47e815d156068674d986aa64fa03998df952a/.github/dependabot.yml looks pretty vanilla to me!  Those other two PRs are _also_ probably not what you want (you've chosen to "pin" to `golang:1.22` on purpose -- switching to something [more](https://github.com/caddyserver/caddy-docker/pull/346) or [less](https://github.com/caddyserver/caddy-docker/pull/345) specific, it can't seem to make up its mind, is definitely also unhelpful :joy:)

PR #16872, opened 2024-05-30
MariaDB-11.4.2 LTS/ 11.5.1 RC release + UBI

grooverdan:
UBI Base images added for 10.6 /10.11 / 11.4, the LTS releases
+ the new RC release 11.5.1.

### DIFF
--- a/library/mariadb
+++ b/library/mariadb
@@ -1,19 +1,29 @@
-# this file is generated via https://github.com/MariaDB/mariadb-docker/blob/b4ff4700163e71e066a8f1d9df01137e2100eca1/generate-stackbrew-library.sh
+# this file is generated via https://github.com/MariaDB/mariadb-docker/blob/bd767859579606f8af92013309e1a9aa626d24da/generate-stackbrew-library.sh
 
 Maintainers: Daniel Black <daniel@mariadb.org> (@grooverdan),
              Daniel Bartholomew <dbart@mariadb.com> (@dbart),
              Faustin Lammler <faustin@mariadb.org> (@fauust)
 GitRepo: https://github.com/MariaDB/mariadb-docker.git
 
-Tags: 11.4.1-rc-jammy, 11.4-rc-jammy, 11.4.1-rc, 11.4-rc
-Architectures: amd64, arm64v8, ppc64le, s390x
-GitCommit: e0677724cbe9f932a568c8f85bca6c14d460bbb4
-Directory: 11.4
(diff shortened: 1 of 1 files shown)


tianon:
```diff
+	curl --fail https://pagure.io/fedora-web/websites/raw/master/f/sites/getfedora.org/static/keys/FF8AD1344597106ECE813B918A3872BF3228467C.txt --output /tmp/epelkey.txt ; \
+	gpg --import-options show-only --import  --with-colons /tmp/epelkey.txt  | grep '^pub:-:4096:1:8A3872BF3228467C:1631033579:::-:' ; \
+	rpmkeys --import /tmp/epelkey.txt ; \
```

This is better, for sure, but not quite as far as I meant -- this validates that the key you expect is included, but not that it is exclusive (so in theory extra keys could be packed into that file and would be happily imported and used for verification).

What I would suggest is probably something more like this:

```dockerfile
	curl --fail https://pagure.io/fedora-web/websites/raw/master/f/sites/getfedora.org/static/keys/FF8AD1344597106ECE813B918A3872BF3228467C.txt --output /tmp/epelkey.txt ; \
	GNUPGHOME="$(mktemp -d)"; export GNUPGHOME; \
	gpg --batch --import /tmp/epelkey.txt; \
	gpg --batch --export FF8AD1344597106ECE813B918A3872BF3228467C | rpmkeys --import; \
	rm -rf "$GNUPGHOME" /tmp/epelkey.txt; unset GNUPGHOME; \
```

(or re-using that same `GNUPGHOME` for the later MariaDB server dance as well, again with full fingerprint in the `export` line)

```diff
+LABEL name="MariaDB Server" \
+	vendor="MariaDB Community" \
+	version="10.11.8" \
+	release="Refer to Annotations org.opencontainers.image.{revision,source}" \
+	summary="MariaDB Database" \
+	description="MariaDB Database for relational SQL"
```

:grimacing: do they really not support a fallback to better/more appropriate/correct label names??  (that sounds like an OpenShift certification bug they should probably fix :sob:)

grooverdan:
Thanks. Done. `rpmkey --import` doesn't take stdin and resisted using a `<( gpg --export  ...)` syntax as an arg.

I cringed on the Openstack format "standard" too. I'll find a place to bug report it.

tianon:
Very nice, thank you :heart: :heart: :heart:

tianon (approved):

PR #16873, opened 2024-05-30
caddy: v2.8.1 release

francislavoie:
https://github.com/caddyserver/caddy/releases/tag/v2.8.1


LaurentGoderre (approved):

francislavoie:
Oh wow - just noticed you're from Ottawa @LaurentGoderre :grinning: both @hairyhenderson and I are as well!

tianon:
Perhaps y'all can get together over a pint of maple syrup :eyes: :maple_leaf: :heart:

tianon (approved):

francislavoie:
@tianon it seems like the `amd64` and `arm64` builds are lagging behind by quite a bit. Is there anything we can do to increase priority on those or something? Every release, we keep getting complaints from people because they tend to take many hours to get their turn in the build queue.

whalelines:
They all appear to be available now. We are investigating ways to make the build system more scalable, but the reality is that some updates trigger thousands of builds and it just takes time to work through them all.

PR #16878, opened 2024-05-30
Update ruby

yosifkit:
Changes:

- https://github.com/docker-library/ruby/commit/c5a2801: Update 3.3 to 3.3.2


tianon (approved):

longkt90:
could you please help to bump to 3.3.3 as well? here is the link https://github.com/docker-library/official-images/compare/docker-library:master...docker-library-bot:ruby?expand=1 

Tried but it limits to collaborators, we need 3.3.3 to fix a segment fault error

Issue #16829, opened 2024-05-23
"operation not permitted", a `libseccomp` story


yosifkit:
This issue is to show example errors that happen when `dockerd`, and/or `libseccomp`\* are not up-to-date when using images based on a newer Linux distribution release.

----

TL;DR: 
`libseccomp` lets you configure allowed syscalls for a process. Docker sets a default seccomp profile for all containers such that only certain syscalls are allowed and everything else is blocked (so, newer syscalls that are not yet known to `libseccomp` or `docker` are blocked\*\*).

- verify that it is `libseccomp` by running the image with `--security-opt seccomp=unconfined`
  - use this only as a test; it is less secure to keep running them "unconfined"
  - learn more: https://docs.docker.com/engine/security/seccomp/
- one fix:
  - update `libseccomp` and `docker` on the host running the containers
- one workaround:
  - switch to an older image:

     Many Docker Official Images have tags that reference the Linux distribution and some DOI support two concurrently. Check their Docker Hub description to see which tags are currently maintained.

     for example, if `16` started failing when it moved to Debian Bookworm, then `16-bullseye` is a valid workaround until the host can be updated. And similar for `16-alpine` and `16-alpine3.19`.
    ![image](https://github.com/docker-library/official-images/assets/896423/d7152094-ddb5-46f8-8157-8c97a0d49a01)

----

One example is that, on some platforms, `tar -xzf file.tar.gz` uses `fchmodat2` in Alpine 3.20 and fails. This syscall isn't in Docker's seccomp profile until Docker 26.0.0, 25.0.3, 24.0.10, and 23.0.10 (see https://github.com/moby/moby/pull/47341 and linked backports).

example:
```console
+ wget -O memcached.tar.gz [https://memcached.org/files/memcached-1.6.27.tar.gz](https://memcached.org/files/memcached-1.6.27.tar.gz%1B[0m)
Connecting to memcached.org (107.170.231.145:443)
saving to 'memcached.tar.gz'
memcached.tar.gz     100% |********************************| 1161k  0:00:00 ETA
'memcached.tar.gz' saved
+ echo 'baf2e7494e1f62d275ff29a99f270abbdb923f75  memcached.tar.gz'
+ sha1sum -c -
memcached.tar.gz: OK
+ mkdir -p /usr/src/memcached
+ tar -xzf memcached.tar.gz -C /usr/src/memcached '--strip-components=1'
[...]

PR #16888, opened 2024-05-31
Release Chronograf 1.10.5

powersj:
Thank you!

### DIFF
--- a/library/chronograf
+++ b/library/chronograf
@@ -3,7 +3,7 @@ Maintainers: Brandon Pfeifer <bpfeifer@influxdata.com> (@bnpfeife),
              Sven Rebhan <srebhan@influxdata.com> (@srebhan),
              Dane Strandboge <dstrandboge@influxdata.com> (@DStrand1)
 GitRepo: https://github.com/influxdata/influxdata-docker
-GitCommit: 1a58dee982ca7f110b9d7478f55c6ee2837ec3bb
+GitCommit: 0e63266d966f2b8917440aa4a146ce03cbfbf1de
 
 Tags: 1.7, 1.7.17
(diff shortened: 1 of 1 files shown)


LaurentGoderre (dismissed):

tianon:
:grimacing: I do not recommend using `FROM alpine:3` -- while convenient, it's what I'd call a "square wheel" and when it rolls, you might get unexpected breakage, so we prefer to see explicit testing/verification/bumping. :bow:

yosifkit:
To expand upon @tianon's comment: Being from the minor version is safer (e.g. `FROM alpine:3.20`) but specifying the specific patch (`3.20.0`) would just create extra work (more PRs to keep up-to-date). A version like `3.20` would allow the image to be automatically rebuilt for "stable" updates (e.g. when `3.20.1` is released). But a plain `3` would mean a rebuild at the `3.21.0` release which is more likely to have breaking changes, aka the "square wheel"; we've seen everything from major library changes that causes the build to fail (like openssl to libressl and back to openssl) to the apk-provided libraries/binaries using newer syscalls causing the container to fail to run for many users because of `libseccomp` denials (https://github.com/docker-library/official-images/issues/16829).

powersj:
I will always defer to your recommendations. Let me update our images to go back to specific minor releases and push an update once I get a review internally.

> But a plain 3 would mean a rebuild at the 3.21.0 release

Right, this was the behavior we thought we wanted. We get requests each time a new Alpine minor release comes out and I had assumed the risk of auto-moving was low for our products and the Alpine images.

LaurentGoderre (approved):

yosifkit (approved):

PR #16914, opened 2024-06-04
Update wordpress

tianon:
Changes:

- https://github.com/docker-library/wordpress/commit/4d3f10a: Update beta to 6.6-beta1

### DIFF
--- a/library/wordpress
+++ b/library/wordpress
@@ -64,47 +64,47 @@ Architectures: amd64, arm32v6, arm32v7, arm64v8, i386, ppc64le, riscv64, s390x
 GitCommit: 509adb58cbc7463a03e317931df65868ec8a3e92
 Directory: cli/php8.3/alpine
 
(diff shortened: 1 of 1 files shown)


yosifkit:
🤔The `beta` images are moving to `6.6` https://make.wordpress.org/core/6-6/#release-schedule. `6.5.4` release is scheduled for tomorrow: https://make.wordpress.org/core/2024/05/24/wordpress-6-5-4-an-upcoming-maintenance-release/#content; any reason to wait for `6.5.4`?

yosifkit (approved):

tianon:
Nope -- they maintain a _lot_ of older versions very aggressively, but we don't maintain builds of any of them (https://github.com/docker-library/wordpress/issues/701#issuecomment-1121688268), so unless they put it out before the GA of 6.6, I don't think we should worry about it too much.

PR #16925, opened 2024-06-06
Official Convertigo 8.3.0 release!

nicolas-albert:
Thx!

### DIFF
--- a/library/convertigo
+++ b/library/convertigo
@@ -1,12 +1,9 @@
 Maintainers: Nicolas Albert <nicolasa@convertigo.com> (@nicolas-albert), Olivier Picciotto <olivier.picciotto@convertigo.com> (@opicciotto)
 GitRepo: https://github.com/convertigo/convertigo
-
-Tags: 8.3.0, 8.3, latest
 GitCommit: b2b3387e1d01034c010817e14de30f2ce60febdc
-Architectures: amd64, arm64v8
-Directory: docker/default
 
 Tags: 8.3.0, 8.3, latest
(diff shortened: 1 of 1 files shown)


yosifkit:
```diff
+arm32v7-Directory: docker/arm32v7
+arm32v7-GitCommit: 19e10c9fe9c26f40b7f9b355a5c4e6e44511bcee
+arm32v7-GitFetch: refs/heads/hotfix
```

> Images must have a single `Dockerfile` per entry in the library file that can be used for multiple architectures. This means that each supported architecture will have the same `FROM` line.
>
> \- https://github.com/docker-library/official-images/blob/768a4bd403a91b6ebcf00404bab8a4f2b1d36e1e/README.md#multiple-architectures

While the `library/` files can specify a different Dockerfile & context per architecture, that functionality is reserved for base images where that is necessary.

It is unfortunate that the newer JDK LTS ([`eclipse-temurin:21-jdk-jammy`](https://github.com/docker-library/official-images/blob/768a4bd403a91b6ebcf00404bab8a4f2b1d36e1e/library/eclipse-temurin#L369-L371)) does not currently have an `arm32v7` image while the old LTS ([`eclipse-temurin:17-jdk-jammy`](https://github.com/docker-library/official-images/blob/768a4bd403a91b6ebcf00404bab8a4f2b1d36e1e/library/eclipse-temurin#L256-L258)) does.

nicolas-albert:
Sorry to haven't follow this rule.
I can drop the arm32v7 until it supports jdk21 (if it happens).
I'll not move all arch to jdk17.

yosifkit (approved):

PR #16926, opened 2024-06-06
[Swift] Add Swift 5.10.1 with additional platforms

shahmishal:
* Adding 5.10.1 with Ubuntu 24.04, Ubuntu 23.10, Debian 12 and Fedora 39
* Removing 5.4.3, 5.3.3 and 5.2.5

### DIFF
--- a/library/swift
+++ b/library/swift
@@ -6,57 +6,77 @@ Maintainers: Ted Kremenek <kremenek@apple.com> (@tkremenek),
 GitRepo: https://github.com/apple/swift-docker.git
 GitFetch: refs/heads/main
 
-Tags: 5.10.0, 5.10, 5.10.0-jammy, 5.10-jammy, jammy, latest
+Tags: 5.10.1, 5.10, 5.10.1-jammy, 5.10-jammy, jammy, latest
 Architectures: amd64, arm64v8
-GitCommit: ea035798755cce4ec41e0c6dbdd320904cef0421
+GitCommit: 53c4d44e4b556d7bab6ff94c80e5bd0444cdc17d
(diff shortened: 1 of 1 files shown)


shahmishal:
@yosifkit If possible, we would like to get this merged soon.

yosifkit:
In general, I would suggest skipping all non-LTS releases of Ubuntu unless there is a really compelling use case for doing otherwise (for example, perhaps the latest LTS does not have a version of OpenSSL or similar that is required by Swift, and thus a newer release is required), especially because their support windows are dramatically shorter.

tianon:
Also, it seems worth giving a heads up that while CentOS 7 is not _technically_ officially EOL yet, the Docker image of it has not been updated in so long (November 16, 2020) that it might as well be EOL for years now.

shahmishal:
Most likely this will be the last release of CentOS 7, as EOL is end of the month.

shahmishal:
>In general, I would suggest skipping all non-LTS releases of Ubuntu unless there is a really compelling use case for doing otherwise (for example, perhaps the latest LTS does not have a version of OpenSSL or similar that is required by Swift, and thus a newer release is required), especially because their support windows are dramatically shorter.

Would you prefer I drop Ubuntu 23.10 from the list?

yosifkit:
> Would you prefer I drop Ubuntu 23.10 from the list?

That is up to you and what your users might expect. It is end of life in July (https://ubuntu.com/about/release-cycle), so it is a short window.

shahmishal:
Let's keep it for now, and we can remove it in the future.

yosifkit (approved):

PR #16941, opened 2024-06-07
Removed EOL server version on EOL PHP versions

LaurentGoderre:
https://www.zend.com/products/zend-server#paragraph-15027

### DIFF
--- a/library/php-zendserver
+++ b/library/php-zendserver
@@ -3,12 +3,6 @@ Maintainers: Heigo Sinilind <hsinilind@perforce.com> (@hsinilind),
 GitRepo: https://github.com/zendtech/php-zendserver-docker.git
 GitCommit: bcd65b82acb4f3b5b67f8c657c52f08bc52d8789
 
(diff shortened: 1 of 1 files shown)


LaurentGoderre:
@rainviigipuu  is this right as well?

yosifkit:
Friendly ping @rainviigipuu or @hsinilind. As the `php-zendserver` image maintainers, does this PR (removing EOL version of PHP zendserver) and https://github.com/docker-library/official-images/pull/17016 seem ok to you?

Removing it here will remove it from the "Supported" section on the Hub readme, but the tag will still be available to users who want to pull it from Docker Hub.

tianon:
Also, it's been three years since this image has had any meaningful updates (https://github.com/docker-library/official-images/pull/10276) -- I appreciate what Laurent is doing here in trying to help out, but ultimately we need to either fix that situation (since I imagine upstream has had _some_ kind of changes in the last three years) or put a deprecation notice on it to clearly communicate the status to users.

PR #16950, opened 2024-06-10
Telegraf v1.31.0

powersj:
New minor release - thank you!

### DIFF
--- a/library/telegraf
+++ b/library/telegraf
@@ -2,29 +2,26 @@ Maintainers: Josh Powers <jpowers@influxdata.com> (@powersj),
              Sven Rebhan <srebhan@influxdata.com> (@srebhan),
              Dane Strandboge <dstrandboge@influxdata.com> (@DStrand1)
 GitRepo: https://github.com/influxdata/influxdata-docker.git
-GitCommit: 13c39ad36b360ff6c48284d1fd296239f4fc6e4e
+GitCommit: c0beea141b6a373b9d3feb2b1b7c99f20542b707
 
-
-Tags: 1.28, 1.28.5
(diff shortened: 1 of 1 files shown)


library/telegraf
```diff
@@ -2,29 +2,26 @@ Maintainers: Josh Powers <jpowers@influxdata.com> (@powersj),
...
-Tags: 1.28-alpine, 1.28.5-alpine
-Architectures: amd64, arm64v8
-Directory: telegraf/1.28/alpine
-
 Tags: 1.29, 1.29.5
 Architectures: amd64, arm32v7, arm64v8
 Directory: telegraf/1.29
 
 Tags: 1.29-alpine, 1.29.5-alpine
-Architectures: amd64, arm64v8
```
tianon, line 21: Just to be sure, this is intentionally dropped, right? (this will remove arm64v8 from 1.29)

yosifkit, line 21: Is the removal of `arm64v8` here and on `1.30-alpine` intentional?

powersj: definitely not intentional, pushed a change to re-add

powersj: definitely not intentional, pushed a change to re-add

tianon (approved):

denschub:
I see that the tag has been pushed 5 hours ago to https://hub.docker.com/layers/library/telegraf/1.31/images/sha256-eb37f58646a901dc7727cf448cae36daaefaba79de33b5058dab79aa4c04aefb?context=explore, but I see no layers in there, and I also can't pull the image - I just get a "no matching manifest for linux/amd64 in the manifest list entries".

Is there something wrong, or is this just some heavily delayed cache?

PR #16952, opened 2024-06-10
Update ALT Linux - 20240610

FNDenisovna:
Add new branch p11

### DIFF
--- a/library/alt
+++ b/library/alt
@@ -1,7 +1,7 @@
 Maintainers: ALT Linux Team Cloud <alt-cloud@altlinux.org> (@alt-cloud),
              Alexey Shabalin <shaba@altlinux.org> (@shaba),
              Gleb Fotengauer-Malinovskiy <glebfm@altlinux.org> (@glebfm),
-             Mikhail Gordeev <obirvalger@altlunux.org> (@obirvalger)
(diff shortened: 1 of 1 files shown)


library/alt
```diff
@@ -1,7 +1,7 @@
 Maintainers: ALT Linux Team Cloud <alt-cloud@altlinux.org> (@alt-cloud),
              Alexey Shabalin <shaba@altlinux.org> (@shaba),
              Gleb Fotengauer-Malinovskiy <glebfm@altlinux.org> (@glebfm),
-             Mikhail Gordeev <obirvalger@altlunux.org> (@obirvalger)
+             Nadezhda Fedorova <fedor@altlunux.org> (@fedor)
```
tianon, line 4: Welcome!  I think you meant @FNDenisovna here (this is a GitHub handle), right?

FNDenisovna: Thank you. Sorry to that, i'm just newbee) Fixed.

tianon (approved):
Nice, looks good -- thank you, and welcome again! :heart:

(I'm guessing it's intentional that "latest" stays at p10 for now because I don't see anything about p11 being GA, but if that's not right, a follow-up to move "latest" is fine too :+1:)

PR #16962, opened 2024-06-11
fix(kong): rm 2.8 arm64 variants

curiositycasualty:
This PR removes the 2.8 arm64 variants. This is a Product decisions being made so that our Community Edition (OSS) product matches the platforms supported by our Enterprise Edition (without exceeding them).

### DIFF
--- a/library/kong
+++ b/library/kong
@@ -29,10 +29,10 @@ Tags: 2.8.4-alpine, 2.8.4, 2.8
 GitCommit: 1c31704cdc9bbd2c0a20e5479eb307140339582b
 GitFetch: refs/tags/2.8.4
 Directory: alpine
-Architectures: amd64, arm64v8
+Architectures: amd64
 
 Tags: 2.8.4-ubuntu, 2.8-ubuntu
 GitCommit: 1c31704cdc9bbd2c0a20e5479eb307140339582b
 GitFetch: refs/tags/2.8.4
 Directory: ubuntu
-Architectures: amd64, arm64v8
+Architectures: amd64


yosifkit (approved):

yosifkit:
Going to merge since the build failures are not related to removing architectures.

I would recommend that `3.4.2`, `3.5.0`, and `3.6.1` should be looked at to see if the build failures can be fixed since they look like they would prevent us from rebuilding on a base image update (`ubuntu:jammy`).

curiositycasualty:
@yosifkit Thanks. Yeah, we just migrated our package storage. So I'll have at look at these failures.

jaxtonw:
@curiositycasualty @yosifkit why did we remove the arm64 images, and not at least retain the prior images built with arm64? Removing these images--that were working just fine--is catastrophic to my teams build. If we're not supporting arm64 moving forward, that's okay, but removing the existing images has caused lots of turmoil.

tianon:
The content-addressable nature of the registry means these are not _removed_ removed -- just no longer part of the supported tag.  See https://github.com/docker-library/repo-info/commit/3c990addae7966904aec01db57b7c102718d6116 for a record of the removal that includes the digest of the old images (`kong@sha256:c4077ace27127ee0574ff87fb6ea1f052717b7e6abacca36e2481a8eb969b908` for the `kong:2.8-ubuntu` manifest list which still contained `arm64` and `kong@sha256:beae4144f5beab7a9eecb2633b5311436d8fe2060a817ae5c327619084a06a82` for the `kong:2.8-alpine` manifest list which did).

PR #16970, opened 2024-06-12
Add official image for Apache Kafka

KrishVora01:
Hi there, this aims to introduce a Docker Official Image for [Apache Kafka](https://kafka.apache.org/).

Apache Kafka is an open-source distributed event streaming platform used by thousands of companies for high-performance data pipelines, streaming analytics, data integration, and mission-critical applications.
To know more about Apache Kafka, please read the [documentation](https://kafka.apache.org/documentation/).

# Checklist for Review

### DIFF
--- a/library/kafka
+++ b/library/kafka
@@ -1,0 +1,8 @@
+# This file is generated via https://github.com/apache/kafka/blob/2432a1866e7bd270f1073cea7c9594e111aed78b/docker/generate_kafka_pr_template.py
+Maintainers: The Apache Kafka Project <dev@kafka.apache.org> (@ApacheKafka)
+GitRepo: https://github.com/apache/kafka.git
+
+Tags: 3.7.0, latest
+Architectures: amd64,arm64v8
+GitCommit: b7dcae44ffb29f854385cef959519a3d0baad55e
+Directory: ./docker/docker_official_images/3.7.0/jvm 


whalelines (changes requested):
Thank you for your interest in the Docker Official Images program and for your submission to create a `kafka` DOI!

Here is an initial review of this PR:

1.  `USER root` is unnecessary

2.  What is the purpose of executing the logic in `jsa_launch` at build time? Why could the .jsa files not be generated ahead of time and distributed with kafka? If they need machine information, why would they not be done at container startup?

3.  `apk update` is unnecessary with `--no-cache`

4.  `apk upgrade` is not allowed as it makes the resulting image dependent on when the build occurs and breaks the caching/determinism of rebuild calculations based on updated parent images.

5.  Specific, individual GPG keys should be referenced by ID. The current implementation provides no protection from an attacker that controls the means of distribution, i.e., if they replace the source they can replace the KEYS file.

6.  All invocations of `gpg` should include `--batch`

7.  The kafka tarball should be verified prior to being unpacked

8.  The DOI build system sets appropriate annotations, so we recommend not setting labels, which have been deprecated. See https://github.com/opencontainers/image-spec/blob/main/annotations.md .

9.  Is `apk cache clean` necessary since you have previously specified `--no-cache`?

10.  It is odd to see so many Dockerization-specific shell scripts. Is running in Docker so different that these are not part of the upstream distribution?

11. In general, application files should not be owned by the user the application is running as. Is there are reason the files under `/etc/kafka/docker` and `/etc/kafka/docker/launch` are owned by/writeable by the `appuser`?

12.  Per archive.apache.org service's notice, it should only be used as a last resort fallback for when a release has been superseded, and the standard apache.org CDN should be preferred for actively supported releases instead

13.  In Docker Official Images, we try hard to avoid multi-stage builds in almost every instance. Since this does a multi-stage build to not clean up after generating the `jsa` files, we would rather that it be just a regular non-multistage build. Perhaps the JSA stage can be eliminated (see 2) or, if it is necessary for it to run at build time, to can be moved to the main stage with proper cleanup. It will help make it clearer for users viewing docker history where things came from. See also the FAQ section on [multi-stage builds](https://github.com/docker-library/faq/tree/76862506fde541f964df742db3b5abe549d9beb6#multi-stage-builds).

Thanks again!

VedarthConfluent:
@whalelines Thanks for providing valuable feedback!
For few of the comments here are my thoughts:-

> In Docker Official Images, we try hard to avoid multi-stage builds in almost every instance. Since this does a multi-stage build to not clean up after generating the jsa files, we would rather that it be just a regular non-multistage build. Perhaps the JSA stage can be eliminated (see 2) or, if it is necessary for it to run at build time, to can be moved to the main stage with proper cleanup. It will help make it clearer for users viewing docker history where things came from. See also the FAQ section on [multi-stage builds](https://github.com/docker-library/faq/tree/76862506fde541f964df742db3b5abe549d9beb6#multi-stage-builds).

Multi stage build allows us to just build the required artifact and use it in the final image. While I agree it is possible to move that logic in the final image and do the cleanup, but it will require more changes and testing. Multi stage build just seems like a cleaner solution. In the FAQ section it was mentioned that multi stage builds are allowed. If this is not too much of an issue would it be possible to keep the multi stage build? Otherwise we will go with doing everything in the single stage and cleaning it up, as you recommended.

> It is odd to see so many Dockerization-specific shell scripts. Is running in Docker so different that these are not part of the upstream distribution?

These scripts make the usage of docker image more convenient. We have moved some logic in the upstream itself, but these scripts are essential. Some logic which especially facilitates bringing up the java process up itself, cannot be moved into upstream. Maybe in future this can be reduced by moving more logic into upstream. But for now, we want to keep it this way.

> In general, application files should not be owned by the user the application is running as. Is there are reason the files under /etc/kafka/docker and /etc/kafka/docker/launch are owned by/writeable by the appuser?

For `/etc/kafka/docker/launch` we don't need the write access, but at runtime we generate and modify files inside `/etc/kafka/docker`, hence write access is required to this folder. Maybe this can be improved in the future. But this is the current behaviour.

> Per archive.apache.org service's notice, it should only be used as a last resort fallback for when a release has been superseded, and the standard apache.org CDN should be preferred for actively supported releases instead

This is something that we confirmed from upstream's maintainers and they recommended us to use archive.apache.org


Except for these points, we will be incorporating rest of the suggestions in the Dockerfile.

KrishVora01:
Hi @whalelines ! Thank you for the suggestions. 
We have raised a PR in Apache Kafka's repository with the changes you have suggested (the remaining comments have been addressed by @VedarthConfluent above). The new Dockerfile that we are proposing is [here](https://github.com/apache/kafka/pull/16768/files#diff-3893b64afdbd7b689ebe1d60f4b89ee9e76a8dab4101f5696eef79822dc2e0ea) . 
Could you please have a look at it, and give us your review on whether the new changes follow the best practices? 
Thank you so much!

tianon:
Based on the status of my review at https://github.com/apache/kafka/pull/16768#pullrequestreview-2675480788, I'm updating the status of this PR to "draft" for now. When it's ready for re-review, please remove the draft status and leave a comment (GitHub unfortunately does not notify maintainers for draft state changes).

PR #16977, opened 2024-06-12
Update haskell

develop7:
Add GHC 9.10

Fixes haskell/docker-haskell#131


tianon:
Debian Buster's LTS team maintenance ends June 30th, 2024 :grimacing:

So old versions will need to be updated or removed very very soon, but new versions should definitely be built on something newer :sweat_smile:

develop7:
@tianon thank you for the heads-up, upgrading from buster is certainly on the list

yosifkit (approved):

Issue #16302, opened 2024-02-24
More RISC-V builds - how can I help?


IngwiePhoenix:
Hello there!

I have a stable VisionFive2 on a 6.6.0 kernel and I want to use it for something good - and since most common image tags, such as `alpine:3.19` and the likes - do not exist, I would like to know how I can help to improve that situation. Compute is costly, thus build servers cost a lot; I know that... so, I would love to help, if possible.

1. What is the process of joining, partaking or sharing RISC-V images to be added to the standart library?
2. Where do I find the raw Dockerfiles that I can then use to build the images - or is there some CI/CD script that can run in, let's say Jenkins, that I could utilize by just spinning up a Jenkins instance?

Thank you and kind regards,
Ingwie

tianon:
As you've noted `alpine:3.19` and friends don't include `riscv64` which is a bit complicated, and not something we can fix in the Docker Official Images directly (and without fixing that, most images won't have support since they're understandably based on stable releases of their upstream distributions).

I know in Debian there's been some great progress towards this (`riscv64` is an official archive architecture now and _might_ be included in Debian Trixie if they can get through the transition work they're undergoing in time).  I think the `#debian-riscv` IRC channel on OFTC is the most active place where that gets discussed, although https://lists.debian.org/debian-riscv/ is also used for linking riscv-related bugs to.

I believe https://gitlab.alpinelinux.org/alpine/aports/-/issues/13269 is still the best place to track the efforts within the Alpine community.

(I'm going to go ahead and close this since there's not really anything we'll do about this here, but feel free to continue discussing.)

gdams:
@tianon is there any hope of an ubuntu base image for RISC-V at some point?

whalelines:
I would imagine that would be in Canonical's plans since they are supporting RISC-V now, but you would have to check with Canonical for a definitive answer.

https://github.com/docker-library/official-images/blob/master/library/ubuntu

tianon:
Yes, definitely -- they had it previously, and they do plan to bring it back, but I think they've had some challenges in their infrastructure/build process that made that complicated.  They definitely still are planning to do so.

gdams:
> Yes, definitely -- they had it previously, and they do plan to bring it back, but I think they've had some challenges in their infrastructure/build process that made that complicated. They definitely still are planning to do so.

[...]

PR #16978, opened 2024-06-12
Update Debian

tianon:
Reproducer's note: this is a little bit of a complicated one due to https://bugs.debian.org/1072675, so it's been generated by pointing directly to the new snapshot-mlm-01.debian.org server for now (although hijacked at the DNS level, with cache, so all the image-internal references, etc are still appropriately snapshot.debian.org).  See especially https://bugs.debian.org/1072675#15 (eventually, all those snapshots will appear on the main snapshot.debian.org service webpage). :heart:

### DIFF
--- a/library/debian
+++ b/library/debian
@@ -7,34 +7,34 @@ GitRepo: https://github.com/debuerreotype/docker-debian-artifacts.git
 GitCommit: ea6ede8f53deaae64dd492001410b689d127e810
 # https://github.com/debuerreotype/docker-debian-artifacts/tree/dist-amd64
 amd64-GitFetch: refs/heads/dist-amd64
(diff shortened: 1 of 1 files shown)


yosifkit (approved):

yosifkit:
For Debian, you'd need to pull `debian:unstable` (aka `debian:sid`) as those are the only stages of Debian releases that have `riscv64` support. Canonical has not yet (re)published `riscv64` artifacts for their Ubuntu container images. See also these comments: https://github.com/docker-library/official-images/issues/16302#issuecomment-1965091680 and https://github.com/docker-library/official-images/issues/16302#issuecomment-1995002636

PR #16979, opened 2024-06-13
Update Joomla!

Llewellynvdm:
Changes:

- joomla-docker/docker-joomla@dc41da6: Update all images with new entrypoint.
- joomla-docker/docker-joomla@4d6ee3a: Refactor the whole entrypoint template file to a better code standard.
- joomla-docker/docker-joomla@9b1db2a: Fix implementation of JOOMLA_SMTP_HOST.
- joomla-docker/docker-joomla@2df41cc: Add install extensions via JOOMLA_EXTENSIONS_URLS and JOOMLA_EXTENSIONS_PATHS.

### DIFF
--- a/library/joomla
+++ b/library/joomla
@@ -6,75 +6,75 @@ GitRepo: https://github.com/joomla-docker/docker-joomla.git
 
 Tags: 5.1.1-php8.1-apache, 5.1-php8.1-apache, 5-php8.1-apache, php8.1-apache
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
-GitCommit: d3713dc96c3bb1fd08be4239dab19a6ad78e867d
+GitCommit: dc41da6f1db9d744478ae04ebed834af60cf5328
 Directory: 5.1/php8.1/apache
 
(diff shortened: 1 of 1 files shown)


Llewellynvdm:
The changes are to our `entrypoint.sh` script and since its used in all our images, the change appears to by large, but its actually just one file that was refactored.

Let me know what you think I should do... that actual change is the ability to [install extensions](https://github.com/joomla-docker/docker-joomla/commit/2df41cc) to Joomla during deployment. The [refactoring](https://github.com/joomla-docker/docker-joomla/commit/4d6ee3a) of the `entrypoint.sh` script was purely to improve the readability of everything.

yosifkit:
>```bash
>if [ -n "$MYSQL_PORT_3306_TCP" ]; then
>...
>: "${JOOMLA_DB_PASSWORD:=$MYSQL_ENV_MYSQL_ROOT_PASSWORD}"
>```

Container links' use is [heavily discouraged](https://docs.docker.com/network/links/). The shared and port environment variables from legacy container links shouldn't be used in new scripts. It looks like these uses have always been in the `joomla` image. I would suggest phasing out their use. In the [`wordpress` images](https://github.com/docker-library/wordpress/blob/0ece07b7af742f5c6a1858bce22da1c7f5233c41/wp-config-docker.php#L59), we just made it default to connect to the `mysql` DNS name, since that is the result that we'd get from checking for `MYSQL_PORT_3306_TCP` anyway.

yosifkit (approved):

PR #16998, opened 2024-06-17
New Docker image for Silverpeas 6.4

mmoqui:
Update silverpeas Docker library to take into account Silverpeas 6.4

### DIFF
--- a/library/silverpeas
+++ b/library/silverpeas
@@ -3,9 +3,9 @@ Maintainers: Miguel Moquillon <miguel.moquillon@silverpeas.org> (@mmoqui)
 GitRepo: https://github.com/Silverpeas/docker-silverpeas-prod.git
 
 Tags: 6.4, latest
-GitCommit: e4e766a0a20d9fd3d138a693f70fed56954a8847
+GitCommit: c3054e40add18fd254afdd07a2b206be6a4d3530
 GitFetch: refs/heads/6.4.x
 
 Tags: 6.3.5
-GitCommit: 73ae9580e6b2aa09b10b722fdff13cfcc61e9d34
+GitCommit: 126477050a7890b45fd57eebe471bfdcff20643d
 GitFetch: refs/heads/6.3.x
--- a/library/silverpeas
+++ b/library/silverpeas
@@ -2,7 +2,10 @@
 Maintainers: Miguel Moquillon <miguel.moquillon@silverpeas.org> (@mmoqui)
 GitRepo: https://github.com/Silverpeas/docker-silverpeas-prod.git
 
-Tags: 6.3.5, latest
-GitCommit: 126477050a7890b45fd57eebe471bfdcff20643d
-GitFetch: refs/heads/6.3.x
+Tags: 6.4, latest
+GitCommit: e4e766a0a20d9fd3d138a693f70fed56954a8847
+GitFetch: refs/heads/6.4.x
 
+Tags: 6.3.5
+GitCommit: 73ae9580e6b2aa09b10b722fdff13cfcc61e9d34
+GitFetch: refs/heads/6.3.x


whalelines:
```diff
+  if [ -f ${SILVERPEAS_HOME}/configuration/config.properties ]; then
+    echo "The configuration file ${SILVERPEAS_HOME}/configuration/config.properties already exists. Does nothing"
+    return
+  fi
```

Best to quote the argument to `-f` (and no need to use curly braces here)

```sh
if [ -f "$SILVERPEAS_HOME/configuration/config.properties" ]; then
```

```diff
+# Migrate the JCR from Apache Jackrabbit 2 to Apache Jackrabbit Oak
+# For doing we have to find out where the JCR home directory is located.
+migrate_jcr() {
+  # figure out the data home directory (by default it is into the Silverpeas home directory)
+  data_home=`grep "SILVERPEAS_DATA_HOME=" ${SILVERPEAS_HOME}/configuration/config.properties  | cut -d '=' -f 2 | xargs`
```

Best to quote here too

```sh
data_home=`grep "SILVERPEAS_DATA_HOME=" "$SILVERPEAS_HOME/configuration/config.properties"  | cut -d '=' -f 2 | xargs`
```

```diff
+  # figure out now the JCR home directory (by default it is located into the data home directory)
+  jcr_home=`grep "JCR_HOME[ ]*=" ${SILVERPEAS_HOME}/configuration/config.properties  | cut -d '=' -f 2 | xargs`
```

This regular expression looks more like a glob pattern. What are you trying to match? Also, quote/brace as above.

```diff
+  if [ "Z${jcr_home}" = "Z" ]; then
+    jcr_home="${data_home}/jcr"
+  else    
+    jcr_home=`sed -e "s/SILVERPEAS_DATA_HOME/data_home/g" <<< "${jcr_home}"`
```

Do you want `SILVERPEAS_DATA_HOME` replaced with the literal string `data_home` or the value of the `data_home` variable?

```diff
+    jcr_home=`eval echo -e "${jcr_home}"`
```

Perhaps the value replacement happens here (`SILVERPEAS_DATA_HOME` is always preceded by `$` in `$jcr_home`)?

```diff
+  jcr_dir=`dirname ${jcr_home}`
+  if [ -d "${jcr_dir}/jackrabbit" ] && [ ! -d "${jcr_dir}/jcr/segmentstore" ]; then
+    echo "Migrate the JCR from Apache Jackrabbit 2 to Apache Jackrabbit Oak..."
+    /opt/oak-migration/oak-migrate.sh "${jcr_dir}/jackrabbit" "${jcr_dir}/jcr"
+    test $? -eq 0 || exit 1
+  fi
```

Typically it is better to test the exit value of a command directly

```sh
if ! /opt/oak-migration/oak-migrate.sh "$jcr_dir/jackrabbit" "$jcr_dir/jcr"; then
  exit 1
fi
```

tianon:
```dockerfile
COPY src/oak-migrate.zip /opt/
```

This is definitely a big eyebrow-raiser - can you elaborate on it?  Why commit the `.zip` to Git?  If it's a canonical upstream artifact, shouldn't it be downloaded from somewhere instead or maybe even be part of the silverpeas distribution itself?  (If it isn't, I'm struggling to see a strong reason for it to be stored in Git as a zip file. :sweat_smile:)

```diff
+  && ./silverpeas assemble || eval "cat ../log/build-* && exit 1" \
```

I don't think you need `eval` here, and probably meant to use `{ ...; }` or `( ... )` instead?  Also mixing `&&` and `||` in a long chain like this is likely to cause problems -- I'd suggest adding `set -eux` (at the very least `set -e`) to the `RUN` block and converting all the `&&` to `;` instead, but alternatively, you should be able to get clever with `{}` or `()` to make sure the precedence works correctly.

mmoqui:
> Best to quote the argument to -f (and no need to use curly braces here)

Ok for quoting vars referring a path (in case there are some specific characters like spaces; it shouldn't be happen here). 

> This regular expression looks more like a glob pattern. What are you trying to match? Also, quote/brace as above.

The goal is to get the value of the `JCR_HOME` var in the `config.properties` configuration file. This var refers the path of the JCR home directory for Silverpeas (JCR stands for Java Content Repository, a kind of a document-oriented database structured into a tree).

> Do you want SILVERPEAS_DATA_HOME replaced with the literal string data_home or the value of the data_home variable?

The value of the `data_home` variable. The `JCR_HOME` value can be made up of another variable, `SILVERPEAS_DATA_HOME`. The value of latter was fetched previously in the script (in the first lines of the `migrate_jcr` function). The goal is to expand (id est _replace_) this var by the value of `data_home`.

> Perhaps the value replacement happens here (SILVERPEAS_DATA_HOME is always preceded by $ in $jcr_home)?

Yes, the variable expansion is done here by the `eval` statement. (Yes, `SILVERPEAS_DATA_HOME` is always prefixed by either `$` or is wrapped by `${}`)

> This is definitely a big eyebrow-raiser - can you elaborate on it? Why commit the .zip to Git? I

Yes, you're right.

> I don't think you need eval here, and probably meant to use { ...; } or ( ... ) instead?  [...]

Ok

whalelines:
> > Best to quote the argument to -f (and no need to use curly braces here)
> 
> Ok for quoting vars referring a path (in case there are some specific characters like spaces; it shouldn't be happen here).

Best to approach this defensively, over time _shouldn't_ can soften to _didn't think it would_.

> > >     jcr_home=`grep "JCR_HOME[ ]*=" ${SILVERPEAS_HOME}/configuration/config.properties  | cut -d '=' -f 2 | xargs`
> > 
> > This regular expression looks more like a glob pattern. What are you trying to match? Also, quote/brace as above.
> 
> The goal is to get the value of the JCR_HOME var in the config.properties configuration file.

I looked at the `/opt/silverpeas/configuration/sample_config.properties` in the current `silverpeas:latest` image. It had this line:

```
#JCR_HOME = ${SILVERPEAS_DATA_HOME}/jcr
```

So it seems the regular expression is using `[ ]*` to capture zero or more space characters. The square brackets are unnecessary. Also, do you want to replace the value regardless of whether the line is commented out? Perhaps this command is more in line with your intention?

```sh
jcr_home=$(grep "^JCR_HOME *=" ${SILVERPEAS_HOME}/configuration/config.properties  | cut -d '=' -f 2 | xargs`
```

Also, it looks like you are using `xargs` to strip whitespace from the value since `cut` will leave the spaces since it it splitting on the equal sign. This may not work the way you expect:

```sh
$ echo 'foo = bar ' | cut -d '=' -f 2 | xargs -I % echo X%X
Xbar X
```

(the above test was run in `silverpeas:latest`)

mmoqui:
>  Also, do you want to replace the value regardless of whether the line is commented out? Perhaps this command is more in line with your intention?

the `config_sample.properties` file is a sample configuration file in which all the supported properties are defined with their default value. When such a property isn't declared or redefined in the `config.properties` file, then the configuration task uses the hard-coded default value of the property. The regexp here is to get the value of a possible declared `JCR_HOME` property in `config.properties`. If not defined, then the default value is used.

> This may not work the way you expect: [...]
The white spaces between the first character of the value and the = sign are stripped. Any tailing spaces are taken into account because those characters is considered to be part of the value (as it is in Java).

yosifkit:
```dockerfile
RUN \
...\
  && wget -nc https://www.silverpeas.org/files/oak-migrate.zip \
  && mkdir -p /opt/oak-migration \
  && unzip oak-migrate.zip -d /opt/oak-migration/ \
```

Any files downloaded should be verified as much as possible (https://github.com/docker-library/official-images/blob/48dadc42dc8aab072c754bb7811f0a63c16499e5/README.md#image-build). Using `https` and embedding a checksum in the Dockerfile can be used when no signature is published. This will provide a natural build cache bust when the remote artifact changes (and prevent an accidental fetch of a newer artifact when the image is rebuilt for a base image update).

tianon (approved):
Those `eval echo -e "${data_home}"` style variable assignments still make me nervous, but in practice they're probably going to be mostly OK unless someone's config file has some really unexpected content/structure.

(The `trap` also only applies during the runtime of the startup script, which is correct but worth noting.)

PR #17013, opened 2024-06-19
Update to Flink 1.19.1

hlteoh37:


library/flink
```diff
@@ -3,19 +3,19 @@
 Maintainers: The Apache Flink Project <dev@flink.apache.org> (@ApacheFlink)
 GitRepo: https://github.com/apache/flink-docker.git
 
-Tags: 1.19.0-scala_2.12-java8, 1.19-scala_2.12-java8, scala_2.12-java8, 1.19.0-java8, 1.19-java8, java8
+Tags: 1.19.1-scala_2.12-java8, 1.19-scala_2.12-java8, scala_2.12-java8, 1.19.1-java8, 1.19-java8, java8
 Architectures: amd64,arm64v8
-GitCommit: 20017e8f0e81d54fe74c0f9f6a3a988ea609be8f
+GitCommit: b464e80c06bd302a6a76dae6b3c8fc2382ec5b57
```
yosifkit, line 8: It looks like https://github.com/apache/flink-docker/pull/196 was merged as a different commit id; so, the `GitCommit` for these three `1.19.1-*` blocks should be https://github.com/apache/flink-docker/commit/f77b347d0a534da0482e692d80f559f47041829e instead of https://github.com/apache/flink-docker/commit/b464e80c06bd302a6a76dae6b3c8fc2382ec5b57.

hlteoh37: The approval workflow seems to be failing because the commits for Flink 1.17 + 1.18 are different than latest. I've updated the PR to reflect this.

hlteoh37:
@yosifkit You're right 😅 I've created a new PR here! Sorry for the hassle....!

https://github.com/docker-library/official-images/pull/17039

PR #17016, opened 2024-06-20
Update php-zend-server to supported Ubuntu

LaurentGoderre:
Bring in changes from https://github.com/zendtech/php-zendserver-docker/pull/30

### DIFF
--- a/library/php-zendserver
+++ b/library/php-zendserver
@@ -1,7 +1,7 @@
 Maintainers: Heigo Sinilind <hsinilind@perforce.com> (@hsinilind),
              Rain Viigipuu <rviigipuu@perforce.com> (@rainviigipuu)
 GitRepo: https://github.com/zendtech/php-zendserver-docker.git
-GitCommit: bcd65b82acb4f3b5b67f8c657c52f08bc52d8789
+GitCommit: e6be4f91651f96b9497718a21586787669e82d3f
 
 Tags: 8.5, 8.5-php5.6, 5.6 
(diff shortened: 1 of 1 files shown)


library/php-zendserver
```diff
@@ -1,9 +1,10 @@
 Maintainers: Heigo Sinilind <hsinilind@perforce.com> (@hsinilind),
              Rain Viigipuu <rviigipuu@perforce.com> (@rainviigipuu)
 GitRepo: https://github.com/zendtech/php-zendserver-docker.git
-GitCommit: bcd65b82acb4f3b5b67f8c657c52f08bc52d8789
+GitCommit: e6be4f91651f96b9497718a21586787669e82d3f
 
-Tags: 8.5, 8.5-php5.6, 5.6 
+Tags: 8.5, 8.5-php5.6, 5.6
+GitCommit: bcd65b82acb4f3b5b67f8c657c52f08bc52d8789
 Directory: 8.5/5.6
```
yosifkit, line 8: `8.5` doesn't change between these [commits](https://github.com/zendtech/php-zendserver-docker/compare/bcd65b82acb4f3b5b67f8c657c52f08bc52d8789...e6be4f91651f96b9497718a21586787669e82d3f), so I don't think it needs to be kept back. Changing it will just add complexity to (what looks like) a manually maintained library file.

LaurentGoderre: Ok, was being over cautious

PR #17022, opened 2024-06-21
nginx maintenance updates

thresheek:
No tag updates, just a couple of fixes:

 - compatibility with Alpine's apk-tools 2.14.2+ when signing self-built packages, fixing the codepath where binary packages are built from source.

 - new gpg signing keys for debian-based images, fixing an expired key issue for downstreams.

I initially thought a gpg issue is best fixed by dropping the keys from the resulting image altogether; but then I thought that our next step would be to introduce another set of keys for the future releases;  and that will undoubtedly cause even more confusion to downstreams.   So let's keep the keys as they are and update them properly.

### DIFF
--- a/library/nginx
+++ b/library/nginx
@@ -5,7 +5,7 @@ GitRepo: https://github.com/nginxinc/docker-nginx.git
 
 Tags: 1.27.0, mainline, 1, 1.27, latest, 1.27.0-bookworm, mainline-bookworm, 1-bookworm, 1.27-bookworm, bookworm
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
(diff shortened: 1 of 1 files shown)


thresheek:
FWIW, here's a commit that introduces a couple new GPG keys: https://github.com/nginxinc/docker-nginx/commit/a6f7d140744f8b15ff4314b8718b3f022efc7f43 (and where to fetch them to check if they're legit)

tianon (approved):
Normally a flag like `--allow-untrusted` would be an eyebrow-raiser, but in this case it makes sense -- you need to tell `apk index` to trust the `.apk` files you _just built_ so it will generate an `APKINDEX` file which you then immediately sign. :smile:

tianon:
(also `(aptly named "signing key 2" and "signing key 3")` gave me a good giggle :smile: :heart:)

PR #17029, opened 2024-06-21
Update drupal

tianon:
Changes:

- https://github.com/docker-library/drupal/commit/e4b6db2: Update 10.1
- https://github.com/docker-library/drupal/commit/971657c: Update 10.3-rc
- https://github.com/docker-library/drupal/commit/d0dadcd: Update 10.3 to 10.3.0, composer 2


yosifkit:
Ah, this needs a `generate-stackbrew-library.sh` update to move the `latest` and `10` tags to the `10.3` release now that it is released.

PR #17037, opened 2024-06-21
NeuroDebian: add Ubuntu noble (nd24.02) and Debian trixie (nd130), make nd120 (current Debian stable) to be default

yarikoptic:


tianon (approved):
:heart:

PR #17060, opened 2024-06-26
Add newly supported architectures

keeganwitt:


ArvinB:
Looking forward to seeing this get merged!
Thank you @keeganwitt for making this happen.

tianon (approved):

ArvinB:
@keeganwitt How long does it take before the images are built and pushed up to dockerhub?

keeganwitt:
> @keeganwitt How long does it take before the images are built and pushed up to dockerhub?

I'm not sure. It looks like the [job](https://doi-janky.infosiftr.net/job/multiarch/job/s390x/job/gradle/) finished, but I [don't see it yet](https://hub.docker.com/_/gradle/tags?page=&page_size=&ordering=&name=8.8.0-jdk22-jammy) in Docker Hub. I'm sure it will be some time today, but I'm not sure how soon.

PR #17074, opened 2024-06-27
Added Redis 7.4 RC2

dmaier-redislabs:


library/redis
```diff
@@ -2,14 +2,25 @@
 
 Maintainers: Tianon Gravi <admwiggin@gmail.com> (@tianon),
              Joseph Ferguson <yosifkit@gmail.com> (@yosifkit)
-GitRepo: https://github.com/docker-library/redis.git
+             David Maier <david.maier@redis.com> (@dmaier-redislabs)
```
tianon, line 5: These are comma-separated, but realistically I think @yosifkit and I should probably be removed now -- we can't really be considered "maintainers" of a repository we don't have commit bits on, so even for the older content we're not really the maintainers anymore. :sweat_smile:

library/redis
```diff
@@ -2,14 +2,25 @@
...
+GitCommit: c828d8c6d982f0d45e1e8f564aae50b42f6d1ec0
+Directory: 7.4-rc/alpine
+
+Tags: 7.4-rc1, 7.4-rc1-bookworm
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
 GitCommit: 0d682fed252b85f39d2033294eab217be02f95a1
 Directory: 7.4-rc/debian
 
-Tags: 7.4-rc1-alpine, 7.4-rc-alpine, 7.4-rc1-alpine3.20, 7.4-rc-alpine3.20
+Tags: 7.4-rc1-alpine, 7.4-rc1-alpine3.20
```
tianon, line 23: These `rc1` versions should be removed with `rc2`; removing tags here will remove them from the "Supported" section on the Hub readme (and will prevent us from spending cycles rebuilding them on the official build servers), but the tags will still be available to users who want them.  (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)

library/redis
yosifkit: If you adjust the content on line 59-63 of [generate-stackbrew-library.sh](https://github.com/redis/docker-library-redis/blob/master/generate-stackbrew-library.sh#L59-L63), then the output of that script should be contents of this `library/redis` file.

In other words:
```console
$ # from redis checkout
$ ./generate-stackbrew-library.sh > ../path-to-official-images-checkout/library/redis
```

library/redis
dmaier-redislabs: OK, this should be fixed now. Part of the issue on my side yesterday night was that the `generate-stackbrew-library` script didn't execute without errors in my development environment. So I just went ahead and generated the entries manually. I fixed the issue now and the new output should be what you expect from the script.

tianon (approved):
Nicely done :heart:

PR #17083, opened 2024-07-01
Scientific Linux reached end of life

sl-team:


jcpunk:
Starting July 1 2024 Scientific Linux no longer recieves updates.

jcpunk:
Sure, do you have an example I should copy?

whalelines:
Typically the deprecation notice is prefixed to the description in the `README-short.txt` file.

https://github.com/docker-library/docs/blob/master/centos/README-short.txt
https://github.com/docker-library/docs/blob/master/express-gateway/README-short.txt

More extensive guidance can be placed at the top of the `content.md` file, e.g., https://hub.docker.com/_/openjdk (the java directory in the docs has since been removed).

jcpunk:
Is https://github.com/docker-library/docs/pull/2459 what you're looking for?

whalelines:
Yes, thanks!

tianon:
Ok, https://github.com/docker-library/docs/pull/2459 has now successfully propagated to both https://hub.docker.com/_/sl _and_ https://hub.docker.com/r/amd64/sl so we should be good to go. :bow: :heart:

yosifkit (approved):

tianon (approved):

PR #17091, opened 2024-07-01
Update Debian (esp. for 12.6 + 11.10)

tianon:


tianon:
Oh, this is notably _also_ the EOL of Debian Buster!  Let me grab a list of affected maintainers.

tianon:
FYI affected maintainers for the Debian Buster EOL/removal:

- `couchdb`: @wohali @janl @nickva
- `elixir`: @c0b @tsloughter (@getong)
- `erlang`: @c0b (@getong)
- `haskell`: @tarleb @develop7
- `haxe`: @andyli
- `mono`: @directhex @akoeplinger
- `neurodebian`: @yarikoptic
- `node`: @nodejs (@starefossen @hmalphettes @jlmitch5 + @mikeal @LaurentGoderre @SimenB @PeterDaveHello -- it would be great if we could update `library/node` with some humans listed so that pings like this are less manual)
- `perl`: @PeterMartini @zakame
- `rust`: @sfackler @Muscraft

yosifkit (approved):

PR #17104, opened 2024-07-02
[security] Update golang

tianon:
Changes:

- https://github.com/docker-library/golang/commit/539882f: Update 1.22 to 1.22.5
- https://github.com/docker-library/golang/commit/4e6e435: Update 1.21 to 1.21.12

### DIFF
--- a/library/golang
+++ b/library/golang
@@ -59,108 +59,108 @@ Directory: 1.23-rc/windows/nanoserver-1809
 Builder: classic
 Constraints: nanoserver-1809, windowsservercore-1809
 
-Tags: 1.22.4-bookworm, 1.22-bookworm, 1-bookworm, bookworm
-SharedTags: 1.22.4, 1.22, 1, latest
+Tags: 1.22.5-bookworm, 1.22-bookworm, 1-bookworm, bookworm
+SharedTags: 1.22.5, 1.22, 1, latest
 Architectures: amd64, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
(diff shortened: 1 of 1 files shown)


yosifkit (approved):

ldemailly:
thanks! I wonder if this could be automated based on trusting go.dev/dl/ ?

tianon:
It is; that's how https://github.com/docker-library/golang/commit/539882f and https://github.com/docker-library/golang/commit/4e6e435 were created.  This PR (and the review of it) are the only manual steps involved (and opening the PR itself is only manual because we often have scraping issues with various upstreams where the automated updates are flaky and thus "flap" between versions occasionally and we don't want to open PRs/create noise for those frivolous commits).

See also https://github.com/docker-library/faq#an-images-source-changed-in-git-now-what

ldemailly:
oh, awesome, thanks for having answered with the details!
so I see (I think all green/done on https://doi-janky.infosiftr.net/job/multiarch/view/images/view/golang/

yet

https://hub.docker.com/v2/repositories/library/golang/tags/1.22.5 is still 404 - where else can see the progress?

ldemailly:
this is probably the wrong spot to ask (where is the right one) but any ETA/issue with

https://hub.docker.com/layers/library/golang/1.22.5/images/sha256-c8715e797d0705204fdeeaf791773c80d5a036c021779984b4c817d84f9b19cb?context=explore

to get the rest of the architectures? (arm etc missing)

ldemailly:
filled https://github.com/docker-library/golang/issues/525 in case it'd help visibility

PR #17105, opened 2024-07-02
Aerospike 7.1.0.2

pvinh-spike:

### DIFF
--- a/library/aerospike
+++ b/library/aerospike
@@ -3,14 +3,14 @@ Maintainers: Lucien Volmar <lucien@aerospike.com> (@volmarl),
              Phuc Vinh <pvinh@aerospike.com> (@pvinh-spike),
              Kevin Porter <kporter@aerospike.com> (@kportertx)
 
-Tags: ee-7.1.0.2, ee-7.1.0.2_1
+Tags: ee-7.1.0.2, ee-7.1.0.2_2
 Architectures: amd64, arm64v8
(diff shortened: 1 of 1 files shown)


tianon:
Hmm, there doesn't appear to be any change / version bump in the `Dockerfile` changes here?

pvinh-spike:
Images rebuilt due to Ubuntu 22.04 base image updated.

tianon:
I'm confused -- that's already how _all_ images in this program work, automatically.  In other words, this PR won't actually rebuild anything (it will only retag things, because no source has changed), and _all_ existing tags are already either rebuilt/repushed or in queue to be rebuilt automatically due to the base image update.

You can see this via the `org.opencontainers.image.created` annotations on https://oci.dag.dev/?image=aerospike:ce-7.1.0.2.

pvinh-spike:
Since the immutable tag 7.1.0.2_2 not recorded, I'm closing the PR.  Thanks

pvinh-spike:
N/A

tianon:
Just to be 100% explicitly clear: _none_ of our tags are "immutable" -- we do not support that concept, and do not plan to.
(See https://github.com/docker-library/official-images#what-are-official-images, especially that fourth bullet point.)

If users want or need an "immutable" reference to an image, that's the purpose of the content-addressable image digest ("pull by digest").

pvinh-spike:
Understood.  Thanks for explaining.

PR #17121, opened 2024-07-04
Removed Swift variant of EOL Centos

LaurentGoderre:


yosifkit:
Since CentOS 7 is end of life, we'd like to drop the CentOS variants of the `swift` images. @tkremenek @shahmishal @tomerd @hamin @swizzlr is this change ok with you? Would you rather make the change in your own PR?

yosifkit (approved):

yosifkit:
Going ahead with removing the end-of-life centos-based swift images. We're slowly migrating images to our newer build system and don't want to repush these EOL images.

PR #17094, opened 2024-07-01
Remove all tags of CentOS (now fully EOL)

tianon:
https://www.redhat.com/en/topics/linux/centos-linux-eol

PR #17126, opened 2024-07-05
Remove EOL CentOS versions of IBM Semeru Runtimes

LaurentGoderre:
This is in response to https://github.com/docker-library/official-images/pull/17094#issuecomment-2201439800, in order to drop CentOS-based IBM Semeru Runtimes images since the CentOS 7 image has been dropped from Docker Official Images.


yosifkit:
Since CentOS 7 is end of life, we'd like to drop the CentOS variants of the `ibm-semeru-runtimes` images. @jayasg12  is this change ok with you? Would you rather make the change in your own PR?

jayasg12:
Hi @yosifkit I have raised an PR https://github.com/docker-library/official-images/pull/17192 by removing CentOS entry , please review and merge, thanks !!

PR #16801, opened 2024-05-22
alpine: add 3.20.0 as latest

ncopa:
Also remove the EOL 3.16

PR #17132, opened 2024-07-06
haxe: remove alpine3.16, add alpine3.20

andyli:
https://github.com/docker-library/official-images/pull/16801#issuecomment-2211490133


tianon (approved):
Nice, thank you!

PR #17136, opened 2024-07-08
Update Bonita base alpine image version to 3.20

rbioteau:
* Remove EOL Bonita 2022.1 version
* Update all Bonita version base image to alpine 3.20
* Remove a maintainer
* Fix deprecated `ENV` assignment format

Related to https://github.com/docker-library/official-images/pull/16801#issuecomment-2211490133


tianon (approved):
All the unnecessary conversions of `ENV foo bar` to `ENV foo=bar` make this diff noisier than it needs to be, but they should be harmless.

PR #17151, opened 2024-07-10
Add a beta registry pre-release

milosgajdos:
See: https://github.com/distribution/distribution/releases/tag/v3.0.0-beta.1

### DIFF
--- a/library/registry
+++ b/library/registry
@@ -7,10 +7,6 @@ Tags: 3.0.0-beta.1
 Architectures: amd64, arm64v8, arm32v6, arm32v7, ppc64le, s390x
 GitCommit: a943e89c3efe06134cd9a4b439203c5341082cbc
 
-Tags: 3.0.0-alpha.1
-Architectures: amd64, arm64v8, arm32v6, arm32v7, ppc64le, s390x
-GitCommit: 6356e8dbe1969f7340ca25a9874faa26e6e6f71d
-
 Tags: 2.8.3, 2.8, 2, latest
 Architectures: amd64, arm64v8, arm32v6, arm32v7, ppc64le, s390x
 GitCommit: 39dd72feaab7066334829d6945c54bc51a0aee98
--- a/library/registry
+++ b/library/registry
@@ -1,7 +1,11 @@
 Maintainers: Milos Gajdos (@milosgajdos),
              Sebastiaan van Stijn (@thajeztah)
-GitRepo: https://github.com/docker/distribution-library-image.git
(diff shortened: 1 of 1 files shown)


thaJeztah (commented):
LGTM

library/registry
```diff
@@ -1,8 +1,12 @@
...
              Sebastiaan van Stijn (@thajeztah)
-GitRepo: https://github.com/docker/distribution-library-image.git
+GitRepo: https://github.com/distribution/distribution-library-image.git
 GitFetch: refs/heads/master
 
+Tags: 3.0.0-beta.1
+Architectures: amd64, arm64v8, arm32v6, arm32v7, ppc64le, s390x
+GitCommit: a943e89c3efe06134cd9a4b439203c5341082cbc
+
 Tags: 3.0.0-alpha.1
```
tianon, line 10: The release of `beta.1` means that `alpha.1` is now effectively unsupported, right? (ie, there won't ever be an `alpha.2` now that `beta.1` is out, right?)

So this section should be removed/replaced. :+1:

tianon:
Yeah, unfortunately we do need at least GitHub handles (email address is optional) -- we use it most often for notifying maintainers of things (such as the ping on this PR, but also pings like https://github.com/docker-library/official-images/pull/16801#issuecomment-2211490133 for base image EOL notifications).  I really wish GitHub had some way to be able to specify that a "team" should be public so we could let folks list a team here instead, but it won't let us @-mention teams from a different organization (even if we're members of it!)

library/registry
milosgajdos: Ok I binned that: https://github.com/docker-library/official-images/pull/17151/commits/698473001c2ae3277d810726e0be480261ad3667

milosgajdos:
> I really wish GitHub had some way to be able to specify that a "team" should be public so we could let folks list a team here instead, but it won't let us @-mention teams from a different organization (even if we're members of it!)

Yeah this would be the most welcome...alas 😮‍💨

tianon (approved):

library/registry
thaJeztah: Good catch; thank you! 🙏

thaJeztah:
> but it won't let us @-mention teams from a different organization (even if we're members of it!)

would it be an option to create teams (with no additional permissions) in this org? So just for pinging / possibly review approval? I guess that could involve extra work to maintain though

tianon:
Yeah, that would then be _us_ maintaining the list, which I think would be even more error prone than y'all doing so :sweat:

PR #17156, opened 2024-07-11
Add tags for skipped Jetty versions 10.0.21, 11.0.21 and 12.0.10

lachlan-roberts:
Add tags for Jetty versions `10.0.21`, `11.0.21`, `12.0.10` which were skipped in the previous PR where we upgraded straight to `10.0.22`, `11.0.22` & `12.0.11`.

see https://github.com/docker-library/official-images/pull/17138

### DIFF
--- a/library/jetty
+++ b/library/jetty
@@ -119,6 +119,46 @@ Architectures: amd64, arm64v8
 Directory: eclipse-temurin/12.0/jdk17
 GitCommit: 65b114c0ee20b162c7df650bb1b95a0949bea68f
 
+Tags: 12.0.10-jre21-alpine, 12.0.10-jre21-alpine-eclipse-temurin
+Architectures: amd64
+Directory: eclipse-temurin/12.0.10/jre21-alpine
+GitCommit: ca69d374a1c300d36d75973007e15ca0fd77f2e2
+
+Tags: 12.0.10-jre21, 12.0.10-jre21-eclipse-temurin
+Architectures: amd64, arm64v8
+Directory: eclipse-temurin/12.0.10/jre21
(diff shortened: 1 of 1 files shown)


tianon:
Just to be clear, this is intended as a temporary backfill, not a long-term addition, right? :eyes: :heart:

joakime:
> Just to be clear, this is intended as a temporary backfill, not a long-term addition, right? 👀 ❤️

Sorry,
It was a missed release (due to holidays and vacations).
It should stick around, and (hopefully) shouldn't conflict with other existing tags of the same id.

We have identified this gap in process, and we don't anticipate a similar behavior in the future.

lachlan-roberts:
@tianon Yes it is just a temporary addition because we missed those releases here.

In future PRs here this file will return to its original format, I just included these extra sections to add tags for the skipped versions.

@joakime I believe when this is merged the `10.0.21`, `11.0.21`, `12.0.10` tags will stick around without having to be included in this file. Then I will delete the folders such as https://github.com/jetty/jetty.docker/tree/aaf4578f1e60c1907d8df7c943f6e2d8bc5f574c/eclipse-temurin/10.0.21 in the `eclipse/jetty.docker` repository. Then after new releases of Jetty this file will just include tags for the latest versions of `9.4.x`, `10.0.x`, `11.0.x` and `12.0.x`.

yosifkit (approved):

tianon:
Indeed, sorry for not getting back to this quicker to explain, but you're exactly right -- removing tags here will remove them from the "Supported" section on the Hub readme (and will prevent us from spending cycles rebuilding them on the official build servers), but the tags will still be available to users who want them.  (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.) :heart:

PR #17192, opened 2024-07-17
IBM Semeru updated to drop CentOS images

jayasg12:
This PR is for dropping CentOS-based IBM Semeru Runtimes images since the CentOS 7 image has been deprecated and reached End of Life.

### DIFF
--- a/library/ibm-semeru-runtimes
+++ b/library/ibm-semeru-runtimes
@@ -7,54 +7,54 @@ GitFetch: refs/heads/ibm
 #-----------------------------openj9 v8 images---------------------------------
 Tags: open-8u412-b08-jdk-focal, open-8-jdk-focal
 Architectures: amd64, ppc64le, s390x, arm64v8
-GitCommit: a3c82d720a98bbdd4cb70f1db22eb27060af2164
+GitCommit: 85c863fc29210104181c7b8ef9d24f3e52fcc87e
 Directory: 8/jdk/ubuntu/focal
 File: Dockerfile.open.releases.full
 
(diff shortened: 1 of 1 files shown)


LaurentGoderre:
@jayasg12 this also bumps the version of tomcat. Is that ok?

LaurentGoderre (approved):

yosifkit (approved):

jayasg12:
@yosifkit Is there a way to add the CentOS deprecation message in this page as we have dropped the tag from the list :  https://hub.docker.com/_/ibm-semeru-runtimes

tianon:
I guess that depends on where you're hoping to add it -- if you just want it in the description somewhere, somewhere in `content.md` would be an appropriate place.

pitinga:
And about CentOS Stream 9? Could not be used for a CentOS image?

Now will be only Semeru with Ubuntu 20 or Ubuntu 22?

yosifkit:
> And about CentOS Stream 9? Could not be used for a CentOS image?

>
> \- https://github.com/docker-library/official-images/blob/955a9134b650f4e5c1052e377b47a1b2014d3521/README.md#repeatability

The CentOS images are no longer provided to Docker Official Images (https://github.com/docker-library/official-images/pull/11831, https://github.com/docker-library/docs/pull/2205, https://github.com/docker-library/official-images/pull/17094, and https://github.com/docker-library/docs/pull/2461), so no other Official Image can be `FROM` them. For whatever reasons, updates stopped being provided to the Docker Official Images and no one from CentOS upstream has come back to provide `stream*` images.

PR #17224, opened 2024-07-22
eclipse-temurin: bump JDK17 and 21

gdams:
Other versions will be available shortly.

Also removes the EOL Centos 7 images

Note the bump from Alpine 3.19 to 3.20


yosifkit (approved):

yosifkit:
Just a warning for users of the plain tags that are moving from Ubuntu Jammy to Noble; the `ubuntu` user and group (id 1000) now exists in the base image and might conflict if you are also creating a user or group at id 1000:  https://github.com/docker-library/official-images/pull/17231#issuecomment-2245772484

PR #17229, opened 2024-07-22
Update gcc

tianon:
Changes:

- https://github.com/docker-library/gcc/commit/3583d3c: Merge pull request https://github.com/docker-library/gcc/pull/108 from infosiftr/eol-11
- https://github.com/docker-library/gcc/commit/b5055bc: Remove EOL 11


tianon:
The comment updates mean this will cause rebuilds of the other versions too, but on the flip side, if we _don't_ merge this, we'll rebuild 11 for every base image update, so IMO we should pay the short term pain instead of the longer term.

yosifkit (approved):

PR #17231, opened 2024-07-23
Apache Storm 2.6.3

rzo1:
As the title says

### DIFF
--- a/library/storm
+++ b/library/storm
@@ -4,10 +4,10 @@ Maintainers: The Apache Storm Project <dev@storm.apache.org> (@asfbot),
 GitRepo: https://github.com/apache/storm-docker.git
 Architectures: amd64, arm64v8
 
-Tags: 2.6.2, 2.6, latest
-GitCommit: d8c821ffb599be733400c98cc10fe4c05966b552
-Directory: 2.6.2
+Tags: 2.6.3, 2.6, latest
+GitCommit: e7117105eb207e0ffad071d429e7c3de8c0864ce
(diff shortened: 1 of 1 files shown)


LaurentGoderre (approved):

cjdcordeiro:
> @woky @cjdcordeiro is adding a new UID 1000 `ubuntu` user to the Ubuntu images _really_ intentional, or an oversight of some kind?
> 
> ```diff
> +# Remove the default user 'ubuntu' and its group
> +RUN userdel -r ubuntu && \
> +    groupdel ubuntu || true \
> ```
> 
> (This is definitely already causing some downstream breakage as folks upgrade or have their base images upgraded underneath them, as seen here, so _really_ might be worth reconsidering. 😬)

Hi,

yes it was intentional and specifically to have a consistent `ubuntu` uid across different Ubuntu formats (VMs, containers, ...) that developers can easily use for nonroot application containers.

The change has been introduced in 23.04.

tianon:
:grimacing: alright, thanks for confirming!

tianon (approved):

yosifkit:
@rzo1, it looks like `storm:2.6.3` (`latest`) now also needs this same user fix as `storm:2.6.3-jre17` since `eclipse-temurin:11-jre` is also updated to Ubuntu Noble from Jammy (https://github.com/docker-library/official-images/pull/17235). Alternatively swapping to `eclipse-temurin:11-jre-jammy` to stay on Ubuntu Jammy would work as well.

PR #17235, opened 2024-07-23
eclipse-temurin: update JDK11 and 22

gdams:
JDK8 will be available soon


yosifkit:
Just a warning for users of the plain tags that are moving from Ubuntu Jammy to Noble; the `ubuntu` user and group (id 1000) now exists in the base image and might conflict if you are also creating a user or group at id 1000:  https://github.com/docker-library/official-images/pull/17231#issuecomment-2245772484

yosifkit (approved):

PR #17249, opened 2024-07-24
eclipse-temurin: add JDK8

gdams:
This closes out our July release cycle

Also contains the rootless CA certs changes made in https://github.com/adoptium/containers/pull/572


yosifkit (approved):

yosifkit:
Just a warning for users of the plain `8` tags that are moving from Ubuntu Jammy to Noble; the `ubuntu` user and group (id 1000) now exists in the base image and might conflict if you are also creating a user or group at id 1000:  https://github.com/docker-library/official-images/pull/17231#issuecomment-2245772484

PR #17254, opened 2024-07-25
Update library/liquibase to 4.29.0

liquibot:

### DIFF
--- a/library/liquibase
+++ b/library/liquibase
@@ -2,12 +2,12 @@ Maintainers: Jake Newton <docker@liquibase.com> (@jnewton03)
 Architectures: arm64v8, amd64
 GitRepo: https://github.com/liquibase/docker.git
 
-Tags: 4.28, 4.28.0, 4.28.0.1, latest
+Tags: 4.29,4.29.0, latest
 GitFetch: refs/heads/main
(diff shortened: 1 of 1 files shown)


library/liquibase
```diff
@@ -2,12 +2,12 @@ Maintainers: Jake Newton <docker@liquibase.com> (@jnewton03)
 Architectures: arm64v8, amd64
 GitRepo: https://github.com/liquibase/docker.git
 
-Tags: 4.28, 4.28.0, 4.28.0.1, latest
+Tags: , , latest
```
yosifkit, line 5: Looks like this is an automation failure somewhere in the chain to making the PR and seems to be reflected in the PR title "Update library/liquibase to".

yosifkit: cc maintainer: @jnewton03

tianon: (see also https://github.com/docker-library/faq#can-i-use-a-bot-to-make-my-image-update-prs, especially point 2. which we should update to note CI failures :joy:)

tianon (approved):

tianon:
(FYI, it does not notify us when things move from draft to "ready for review", so a comment is appropriate after those transitions :+1:)

PR #17281, opened 2024-07-31
Release Plone 5.2.14

avoinea:


tianon (approved):

tianon:
It's been quite some time now since a Plone image update -- notably, the only remaining tag is `FROM python:3.8-slim-bullseye` which is even no longer supported.  Should we look to add a deprecation notice to the image, or perhaps there's an update coming very soon?

See also:
- https://github.com/docker-library/official-images/pull/17687 (`python:3.8` EOL / removal)
- https://github.com/docker-library/official-images/pull/17382
- https://github.com/plone/plone.docker/issues/161

tianon:
I've made a simplified alternative to https://github.com/docker-library/docs/pull/2479 in https://github.com/docker-library/docs/pull/2583. :heart:

PR #17283, opened 2024-07-31
Add the signer image to the bashbrew environment

LaurentGoderre:


tianon (changes requested):
Sorry, but I would prefer to keep this pinning explicitly closer to the place we're using it, especially while the shape of the tool/process continues to evolve and settle. :see_no_evil:

(Otherwise we won't notice breakage unless/until it gets used, because our pins here are a notable blind spot in our automated testing.)

PR #17287, opened 2024-07-31
Update library/liquibase to 4.29.1

liquibot:

### DIFF
--- a/library/liquibase
+++ b/library/liquibase
@@ -2,12 +2,12 @@ Maintainers: Jake Newton <docker@liquibase.com> (@jnewton03)
 Architectures: arm64v8, amd64
 GitRepo: https://github.com/liquibase/docker.git
 
-Tags: 4.29,4.29.0, latest
+Tags: 4.29, 4.29.1, latest
 GitFetch: refs/heads/main
-GitCommit: 6903fb3ab80cde8a9a763fd230bfa66449fe5ca5
+GitCommit: 07e271e1a0d33b7670c5abcaf940e83bc0d17418
 File: Dockerfile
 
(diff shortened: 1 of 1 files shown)


library/liquibase
```diff
@@ -2,12 +2,12 @@ Maintainers: Jake Newton <docker@liquibase.com> (@jnewton03)
...
 GitFetch: refs/heads/main
-GitCommit: 6903fb3ab80cde8a9a763fd230bfa66449fe5ca5
+GitCommit: cd836a9f9e1f89fde5458096c38b7b41b8f6ea7b
 File: Dockerfile
 
-Tags:  4.29-alpine, 4.29.0-alpine, alpine
+Tags: 4.29-alpine, 4.29.1-alpine, alpine
 GitFetch: refs/heads/main
-GitCommit: 6903fb3ab80cde8a9a763fd230bfa66449fe5ca5
+GitCommit: cd836a9f9e1f89fde5458096c38b7b41b8f6ea7b
```
tianon, line 12: These should both be https://github.com/liquibase/docker/commit/07e271e1a0d33b7670c5abcaf940e83bc0d17418 instead, right? (there's no actual `Dockerfile` changes between these two commits otherwise)

jandroav: Yes @tianon  , you are right. We need to add a small fix after https://github.com/liquibase/docker/blob/07e271e1a0d33b7670c5abcaf940e83bc0d17418/.github/workflows/create-release.yml#L91 to use that new commit.

I manually fixed it for this PR (rebased).

Thanks

LaurentGoderre (approved):

yosifkit (approved):

PR #17310, opened 2024-08-06

adivinho:


whalelines:
What sets `PERCONA_TELEMETRY_DISABLE` to `0` such that the test in this added conditional would work?

```diff
-# PERCONA_TELEMETRY_DISABLE is handled at the very beginning of call-home.sh
-/call-home.sh -f "PRODUCT_FAMILY_PS" -v "${PS_TELEMETRY_VERSION}" -d "DOCKER" ${CALL_HOME_OPTIONAL_PARAMS} &> /dev/null || :
-
-exec "$@"
+if [[ ${PERCONA_TELEMETRY_DISABLE} -ne "0" ]]; then
+  exec "$@" --percona_telemetry_disable=1
+else
+  /usr/bin/telemetry-agent-supervisor.sh &
+  exec "$@"
+fi
```

Why is it necessary for the telemetry script to `sleep infinity` rather than just exit?

adivinho:
> What sets PERCONA_TELEMETRY_DISABLE to 0 such that the test in this added conditional would work?
The data collection may be disabled via setting an environment variable PERCONA_TELEMETRY_DISABLE=1 during Linux package installation or Docker container startup.

For example, on a Docker container (this action disables the continuous telemetry part as well):
docker run -e PERCONA_TELEMETRY_DISABLE=1 percona/percona-server

adivinho:
> Why is it necessary for the telemetry script to sleep infinity rather than just exit?

https://github.com/percona/telemetry-agent?tab=readme-ov-file#continuous-telemetry

whalelines:
To clarify my questions a bit more:

The `telemetry-agent-supervisor.sh` calls two commands, redirecting the stdout and stderr to files. Since these commands are not started in the background, presumably they are either short-lived processes or fork and exit on their own. With that being the case, why is it necessary for the `telemetry-agent-supervisor.sh` script to remain running?

The conditional for the value of the `PERCONA_TELEMETRY_DISABLE` variable explicitly checks if the value is "0". I see nothing that sets the value explicitly to "0". As such, when will that conditional ever be true, short of the user setting it explicitly to zero? The documentation indicates you should set it to one to disable it, but as I read it, it seems by default that conditional would not evaluate to true, so perhaps the docs should say that telemetry is disabled by default and you should set PERCONA_TELEMETRY_DISABLE to zero to enable it.

surbhat1595:
> The telemetry-agent-supervisor.sh calls two commands, redirecting the stdout and stderr to files. Since these commands are not started in the background, presumably they are either short-lived processes or fork and exit on their own. With that being the case, why is it necessary for the telemetry-agent-supervisor.sh script to remain running?

The second command, i.e. /usr/bin/percona-telemetry-agent  runs in the background and has to be always running and will exit only in case the container is killed. In case of any issues, it will try re-starting the percona-telemetry-agent process 3 times and then exit. Since percona-telemetry-agent process has to always be running and to avoid any zombie processes being created when the supervisor script exits, we have the supervisor script always running. 

> The conditional for the value of the PERCONA_TELEMETRY_DISABLE variable explicitly checks if the value is "0". I see nothing that sets the value explicitly to "0". As such, when will that conditional ever be true, short of the user setting it explicitly to zero? The documentation indicates you should set it to one to disable it, but as I read it, it seems by default that conditional would not evaluate to true, so perhaps the docs should say that telemetry is disabled by default and you should set PERCONA_TELEMETRY_DISABLE to zero to enable it.

No, the percona-telemetry-agent process runs in both conditions, when PERCONA_TELEMETRY_DISABLE is set to "0" and even when PERCONA_TELEMETRY_DISABLE is unintialized. In the condition `if [[ ${PERCONA_TELEMETRY_DISABLE} -ne "0" ]];` -ne is an arithmetic operator, in an arithmetic context the empty string evaluates to zero since an unset variable expands to the empty string. All these conditions have been tested before publishing the docker images.

surbhat1595:
> +  /usr/bin/telemetry-agent-supervisor.sh &
> 
> Running unsupervised background processes in the container is not generally allowed in Official Images since that then requires a third process to supervise them: handling and sending signals, reaping zombies, and deciding what to do if one or more of them exit (or don't exit when asked to). Unsupervised background processes are processes that don't have a responsible parent process. Even if their PID relationship puts them in a parent/child relationship (like because of an exec), the parent does not take responsibility for the child (tracking/restarting, forwarding signals, reaping, etc). Most of this is instead better handled at a higher level in functionality provided by dockerd or in an orchestration of multiple containers like docker-compose and kubernetes (though neither seem a great fit for a "sidecar" telemetry process).
> See also https://github.com/docker-library/official-images/blob/e070cf74d7f816287583930d2eeb9df1a93ad06e/README.md#init.

The reason we added a `sleep infinity ` command in the telemetry-agent-supervisor.sh is because the process might turn into a zombie process if the script exits. We agree that it can be handled better using supervisord etc, but it seemed to be too heavy considering our requirement. Nevertheless, we shall check if there are ways to improve this

> +    chown  mysql /usr/bin/telemetry-agent-supervisor.sh; \
> 
> Files from a previous layer (like COPY) should not be modified in a later layer as it duplicates content and increases the final image size. Permissions/ownership should be done via values tracked in git (like executability) and/or --chown and --chmod flags on COPY ([docs.docker](https://docs.docker.com/reference/dockerfile/#copy---chown---chmod)).

Sure, Thank you, we shall try replacing chmod/chown commands with the above mentioned ways.

whalelines:
> > The telemetry-agent-supervisor.sh calls two commands, redirecting the stdout and stderr to files. Since these commands are not started in the background, presumably they are either short-lived processes or fork and exit on their own. With that being the case, why is it necessary for the telemetry-agent-supervisor.sh script to remain running?
> 
> The second command, i.e. /usr/bin/percona-telemetry-agent runs in the background and has to be always running and will exit only in case the container is killed. In case of any issues, it will try re-starting the percona-telemetry-agent process 3 times. Since percona-telemetry-agent process has to always be running and to avoid any zombie processes being created when the supervisor script exits, we have the supervisor script always running.

I am confused by your response. Does `/usr/bin/percona-telemetry-agent` fork itself and run in the background or does it stay in the foreground and the `telemetry-agent-supervisor.sh` monitors it? In other words, are those three retries done because of transient startup errors or to recover from any issue the agent may have during the lifetime of the container?

Here is the script in question:

```sh
#!/bin/bash

# phase-0 telemetry
/call-home.sh -f "PRODUCT_FAMILY_PS" -v "${PS_TELEMETRY_VERSION}" -d "DOCKER" ${CALL_HOME_OPTIONAL_PARAMS} &> /dev/null || :

# phase-1 telemetry
for i in {1..3}; do
    /usr/bin/percona-telemetry-agent >> /var/log/percona/telemetry-agent.log 2>> /var/log/percona/telemetry-agent-error.log
    if [ $? -eq 0 ]; then
      break
    fi
    sleep 5
done
sleep infinity
```

tianon:
Either way, and to try and make this more explicitly clear, a generic process supervisor (either in shell or an explicit supervisor tool like supervisord) is unfortunately not going to be acceptable here.

tianon:
_All_ images specified in `library/percona` are now failing to build -- if this is not rectified soon, we will be disabling them. :heart:

EvgeniyPatlan:
> Either way, and to try and make this more explicitly clear, a generic process supervisor (either in shell or an explicit supervisor tool like supervisord) is unfortunately not going to be acceptable here.

@tianon and what be the suggested way?

tianon:
If both processes are necessary and must stay resident/running, a separate container for each or a proper manager/worker model as seen in Apache, HAProxy, NGINX, etc.

surbhat1595:
@tianon , We followed the official documentation from Docker [here](https://docs.docker.com/engine/containers/multi-service_container/) which suggests using either a wrapper script or supervisord for managing multiple processes in a container. Our implementation uses a wrapper script ( telemetry-agent-supervisor.sh)  as per the guidance, structured like this:

> percona_bash_script
>    ├── mysqld
>    └── telemetry_agent

Alternatively, we also explored supervisord which is also mentioned in the above doc,

> supervisord
>    ├── mysqld
>    └── telemetry_agent

Also, the docker images you mentioned as a reference, i.e. Apache, HAProxy, NGINX, etc. run one process in entrypoint which forks sub-processes. Since you mentioned that supervisord or a wrapper script is not acceptable, could you kindly clarify what exactly needs to be changed in our approach or any alternative method you would recommend to run multiple processes in a container? 
Thank you.

yosifkit:
> Also, the docker images you mentioned as a reference, i.e. Apache, HAProxy, NGINX, etc. run one process in entrypoint which forks sub-processes. Since you mentioned that supervisord or a wrapper script is not acceptable, could you kindly clarify what exactly needs to be changed in our approach or any alternative method you would recommend to run multiple processes in a container?

Without making the telemetry process a child managed by `mysqld` (similar to NGINX and its worker processes), there isn't really an alternative within Docker Official Images. I guess a "hacky" supervisor configuration could be hidden from our review by making it and the telemetry script part of a `percona-server-*` packaging. However, that should also make it more directly supported/tested within Percona (and bugs would not be considered just a quirk of containers) and then used by all users of Percona Server whether or not they use the Docker images.

Running multiple processes side by side within a single container is complex and there are many edge cases that we can't always remember and review in PRs to official images. To ensure that all images meet the high quality users expect from Docker Official Images, we restrict them to a single process (aside from those that spawn and manage their own worker processes).

---

Just for clarification as to why multi-process containers are so complex. Here are just a few edges and considerations that are needed for a multi-process container; this list is neither exhaustive nor an acceptance checklist:
- minimum of 3 processes
  - supervisor process: one process responsible for managing the other processes, signal handling/forwarding, and exiting
  - `main` processes: the child processes that if any of them exits should cause the container to exit
  - `sidecar` processes: the children that if they exit, should just be restarted by the supervisor (or ignored?)
    - it could be zero `sidecar` processes and more than one `main` process
- when should the supervisor exit (causing the whole container to stop and cause all other processes to be forcibly killed)
- what exit code should the supervisor use?
- supervisor reaping (the supervisor is expected to handle pid 1 responsibilities, like reaping zombies, as well as notice when the `main` or `sidecar` processes exit)
- forwarding signals
  - which signals are sent to which children?
  - which signals should cause the specific child to exit?
  - is there a timeout on signals that expect the child to exit (and when should that cause the supervisor to exit)
    - how does that interact with user-configured Docker or Kubernetes stop timeouts?
- what to do if a `sidecar` process exits unexpectedly (immediately or after a long time)
  - does the `sidecar` need a startup grace period (like `initialDelaySeconds` in Kubernetes liveness probes)
  - is the `sidecar` required to be ready before starting the `main` process? Or the other way around?
    - what is "ready"
    - how does a user reflect the `sidecar` processes in a health check (or Kubernetes liveness/readiness probes)?
  - is the response different if the `sidecar` has failed 10 times, or 100 times?
- docker users expect most logs will be available on stdout and stderr
  - does each process (supervisor, `main`, and `sidecar`) need a prefix on their output to clarify where it came from
  - changing logging configuration now can be done in multiple, possibly conflicting, places (e.g., `my.cnf`, supervisor config, and supervisor child-specific config)

PR #17332, opened 2024-08-09
Odoo: update 15.0-17.0 to release 20240809

d-fence:
Hello,

here are the latest updates of Odoo supported versions.

Thanks

### DIFF
--- a/library/odoo
+++ b/library/odoo
@@ -1,16 +1,16 @@
 Maintainers: Christophe Monniez <moc@odoo.com> (@d-fence)
 GitRepo: https://github.com/odoo/docker
-GitCommit: 0a794bf1c202f1d7ca409ce30c159350dc541100
+GitCommit: 801d34e4b41dbf16f81650b9e0c6fbbd302d5fb6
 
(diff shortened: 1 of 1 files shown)


library/odoo
```diff
@@ -1,6 +1,6 @@
 Maintainers: Christophe Monniez <moc@odoo.com> (@d-fence)
 GitRepo: https://github.com/odoo/docker
-GitCommit: 0a794bf1c202f1d7ca409ce30c159350dc541100
+GitCommit: b4cccf3ccae004aebfd334aa2639499496151598
 
 Tags: 17.0-20240730, 17.0, 17, latest
```
tianon, line 5: You probably want to update all these date references too (to `20240809`) :eyes:

(The way we mitigate this oversight for repos we maintain is by writing a script that generates this `library/XXX` content directly from our repo metadata, in case that's an idea that would be interesting/helpful for you :eyes:)

d-fence: Nice catch, thanks.

tianon (approved):

PR #17335, opened 2024-08-09
Remove dangling children of Debian Buster

tianon:
Debian Buster is EOL, and as such these images need to be either upated or removed (they are not receiving meaningful rebuilds anymore).

(see https://github.com/docker-library/official-images/pull/17091#issuecomment-2201099734)

I'm happy to accept these back if they are updated with a suitably supported base image.


tianon:
Also FYI @getong (`erlang`, `elixir`)

whalelines (approved):
LGTM

yosifkit (approved):

yarikoptic:
Is there some effort to archive such older/EOLed images?
For the use cases of "reproducible science" it might be desired/critical to be able to access such old bases.

tianon:
Oh sorry, I should've included my standard blurb!

yarikoptic:
oh, thanks!  I will to remember!

PR #17340, opened 2024-08-10
Update Semeru release versions

jayasg12:


tianon:
You probably wanted to update the version numbers in the `Tags:` too, right? (FWIW, we usually write a script that generates the appropriate `library/XXX` contents for images we maintain, especially so we're less likely to miss tag updates :see_no_evil:)

jayasg12:
@tianon Thank you for pointing out missing tag update. I have updated the tags and pushed the changes. Please review.

tianon (approved):

PR #17352, opened 2024-08-13
Fedora Update 20240813

cverna:


library/fedora
```diff
@@ -4,7 +4,7 @@ GitRepo: https://github.com/fedora-cloud/docker-brew-fedora.git
 Tags: 39
 Architectures: amd64, arm64v8, ppc64le, s390x
 GitFetch: refs/heads/39
-GitCommit: 8e39969bbb6fb9482cede31c19e969c48252939b
+GitCommit: e493147
```
tianon, line 7:
```suggestion
GitCommit: e493147cb7383bda672f4295c09e435d83047efa
```

library/fedora
```diff
@@ -13,7 +13,7 @@ ppc64le-Directory: ppc64le/
 Tags: 40, latest
 Architectures: amd64, arm64v8, ppc64le, s390x
 GitFetch: refs/heads/40
-GitCommit: 7f9dbbf47c41b95d34c7722927bb93816956cfe1
+GitCommit: 978adc7
```
tianon, line 16:
```suggestion
GitCommit: 978adc7dc216f09fca95de7174734748710fe1df
```

library/fedora
```diff
@@ -22,7 +22,7 @@ ppc64le-Directory: ppc64le/
 Tags: 41, rawhide
 Architectures: amd64, arm64v8, ppc64le, s390x
 GitFetch: refs/heads/41
-GitCommit: 68b26415c2ff3605aed2f7c737b74048b23d7fdf
+GitCommit: c527fe0
```
tianon, line 25:
```suggestion
GitCommit: c527fe0bcd2f1d8fbbbe61467ad50546567a655b
```

tianon (changes requested):
Can you please continue using the full 64-character commit hashes? :innocent:

cverna:
> Can you please continue using the full 40-character commit hashes? 😇

Sure, just curious any special reason for it? It's faster for me to create the PR with the short commit hashes 😄

tianon:
> Sure, just curious any special reason for it? It's faster for me to create the PR with the short commit hashes 😄

Oh, yeah -- they're a checksum of the actual commit contents, so they're the way we can verify the contents we fetch.  Generating a collision of a _short_ commit ID is going to be pretty trivial (and won't be likely to be detected/blocked by GitHub) whereas collisions of the full SHA1 commit ID are going to be _much_ more expensive to generate _and_ will be detected/blocked by GitHub. :eyes:
(see also https://github.blog/news-insights/company-news/sha-1-collision-detection-on-github-com/)

If you're copying these from the command-line (or CI), you can use something like `git rev-parse HEAD` (or `git log -1 --format=format:%H` which is the one I usually end up reaching for because I forget about `rev-parse` :joy:).

If you're copying them from the GitHub web UI, you can press <kbd>y</kbd> (the only keyboard shortcut on GitHub I actually use/like!) and it'll put you at a permalink that includes the full commit hash in the URL _or_ you can go to the "commits" / "History" view and there's a direct copy button for the full hash. :+1:

PR #17368, opened 2024-08-15
Updated nginx to latest versions

thresheek:
mainline gets updated to 1.27.1
stable gets updated to 1.26.2

while at it, Alpine is bumped for both mainline and stable to 3.20

also implemented a way to specify a different dynamic module packaging release version (1 vs 2) to workaround currently published packaging sources;  might likely come in handy in the future when we'd need to bump nginx "base" package independently of all other modules.

### DIFF
--- a/library/nginx
+++ b/library/nginx
@@ -3,72 +3,72 @@
 Maintainers: NGINX Docker Maintainers <docker-maint@nginx.com> (@nginxinc)
 GitRepo: https://github.com/nginxinc/docker-nginx.git
 
-Tags: 1.27.0, mainline, 1, 1.27, latest, 1.27.0-bookworm, mainline-bookworm, 1-bookworm, 1.27-bookworm, bookworm
+Tags: 1.27.1, mainline, 1, 1.27, latest, 1.27.1-bookworm, mainline-bookworm, 1-bookworm, 1.27-bookworm, bookworm
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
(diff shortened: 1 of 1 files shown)


library/nginx
```diff
@@ -3,72 +3,72 @@
...
 Directory: mainline/debian-perl
 
-Tags: 1.27.0-otel, mainline-otel, 1-otel, 1.27-otel, otel, 1.27.0-bookworm-otel, mainline-bookworm-otel, 1-bookworm-otel, 1.27-bookworm-otel, bookworm-otel
+Tags: 1.27.1-otel, mainline-otel, 1-otel, 1.27-otel, otel, 1.27.1-bookworm-otel, mainline-bookworm-otel, 1-bookworm-otel, 1.27-bookworm-otel, bookworm-otel
 Architectures: amd64, arm64v8
-GitCommit: 3180cdbec313dc4a9f6dd1109ae66adaf98f11fb
+GitCommit: e78cf70ce7b73a0c9ea734c9cf8aaaa283c1cc5a
 Directory: mainline/debian-otel
 
-Tags: 1.27.0-alpine, mainline-alpine, 1-alpine, 1.27-alpine, alpine, 1.27.0-alpine3.19, mainline-alpine3.19, 1-alpine3.19, 1.27-alpine3.19, alpine3.19
```
jnoordsij, line 21: This (and also stable) should be `alpine3.20` now; see also https://github.com/nginxinc/docker-nginx/pull/917

thresheek: Oh indeed, thanks for catching that!

yosifkit (approved):

PR #17388, opened 2024-08-19
maven: 3.9.9

carlossg:


tianon (approved):
Noisier diff than it needed to be thanks to https://github.com/carlossg/docker-maven/pull/487 (https://github.com/moby/buildkit/issues/5130), but harmless otherwise

PR #17389, opened 2024-08-19
Update buildpack-deps

tianon:
Changes:

- https://github.com/docker-library/buildpack-deps/commit/5d682bd: Add back focal+risc64 bits removed in https://github.com/docker-library/buildpack-deps/pull/163

More `riscv64` on Ubuntu variants :eyes:


tianon:
(I'm doing a quick build test to make sure we don't need to bring back the focal bits I removed in https://github.com/docker-library/buildpack-deps/pull/163/files)

yosifkit (approved):

PR #17395, opened 2024-08-20
[oraclelinux] Updating 8 and 9 for ELSA-2024-5534 ELSA-2024-5529 ELSA-2024-5530 ELSA-2024-5524

amsteinb-orcl:
This update incorporates fixes for CVE-2024-6345, CVE-2024-2398, CVE-2024-1737, CVE-2024-1975

https://linux.oracle.com/errata/ELSA-2024-5530.html
https://linux.oracle.com/errata/ELSA-2024-5524.html
https://linux.oracle.com/errata/ELSA-2024-5534.html
https://linux.oracle.com/errata/ELSA-2024-5529.html

### DIFF
--- a/library/oraclelinux
+++ b/library/oraclelinux
@@ -7,7 +7,7 @@ amd64-GitFetch: refs/heads/dist-amd64
 amd64-GitCommit: 53082608c83ad0df59197d2ac14f1c1403764bf6
 # https://github.com/oracle/container-images/tree/dist-arm64v8
 arm64v8-GitFetch: refs/heads/dist-arm64v8
-arm64v8-GitCommit: 31a1708d05cf315d7200eae275ab891c152a522c
+arm64v8-GitCommit: 584963be844e344c6b385c93ce2e3399c65207b4
 
 Tags: 9
 Architectures: amd64, arm64v8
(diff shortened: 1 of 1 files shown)


amsteinb-orcl:
Republished with amd64v8 commit hash due to missing OL9 full image caused by rebuild.

tianon (approved):
I love the `-etc/X11/xorg.conf.d/` :smile:

(distro maintenance is always a really interesting set of problems and tradeoffs, and I'm sure there's probably a good story underneath this one being included before _and_ it being removed :smile:)

amsteinb-orcl:
@tianon I am missing something. Where do you see the etc/X11/xorg.conf.d reference?

tianon:
That's in the diff -- it was apparently an empty directory in the previous `oraclelinux:9` image that was removed in the updated one.  I'm guessing it was probably owned by/provided by some core package in the base OS and the reasons for that are no longer applicable (or the "ownership" of that directory has simply shifted to another package which isn't in the minimal base set).  Not spicy at all, but definitely the kind of thing I find (personally) interesting and there's almost always a story behind changes like that. :smile:

PR #17398, opened 2024-08-20
Update phpmyadmin to the latest commit

williamdes:
- Some changes for rootless
- Support new ENVs for SSL


yosifkit:
`RUN chown ...` is going to duplicate the file added by `COPY`, so it should use `COPY --chown` instead (https://docs.docker.com/reference/dockerfile/#copy---chown---chmod).  It will require a `Builder: buildkit` in the `library/phpmyadmin` file (like [rabbitmq](https://github.com/docker-library/official-images/blob/3f6a23256b56df28f09ac035d6cdc0ace235de58/library/rabbitmq#L6)) so that we use the correct docker builder (since the classic builder doesn't understand it).

`buildkit` will eventually be the default for Docker Official Images, but it is still an active migration.

tianon (approved):

PR #17403, opened 2024-08-21
eclipse-temurin: add support for riscv64

gdams:
Also has some minor reformatting changes to the entrypoint.sh

For now we're only adding Ubuntu Noble support, we could add previous Ubuntu versions if there was enough interest.

I'm not sure why the diff comment is failing

### DIFF
--- a/library/eclipse-temurin
+++ b/library/eclipse-temurin
@@ -9,34 +9,34 @@ Builder: buildkit
 #------------------------------v8 images---------------------------------
 Tags: 8u422-b05-jdk-alpine, 8-jdk-alpine, 8-alpine
 Architectures: amd64
-GitCommit: c031e924f7de45f02a902e2cb88958ab0f343f59
+GitCommit: 07677395574f5d3462c3b6fdf5f6c4a0a350b683
 Directory: 8/jdk/alpine
 
 Tags: 8u422-b05-jdk-focal, 8-jdk-focal, 8-focal
 Architectures: amd64, arm32v7, arm64v8, ppc64le
-GitCommit: c031e924f7de45f02a902e2cb88958ab0f343f59
+GitCommit: 07677395574f5d3462c3b6fdf5f6c4a0a350b683
 Directory: 8/jdk/ubuntu/focal
 
 Tags: 8u422-b05-jdk-jammy, 8-jdk-jammy, 8-jammy
 Architectures: amd64, arm32v7, arm64v8, ppc64le
-GitCommit: c031e924f7de45f02a902e2cb88958ab0f343f59
+GitCommit: 07677395574f5d3462c3b6fdf5f6c4a0a350b683
 Directory: 8/jdk/ubuntu/jammy
 
 Tags: 8u422-b05-jdk-noble, 8-jdk-noble, 8-noble
 SharedTags: 8u422-b05-jdk, 8-jdk, 8
 Architectures: amd64, arm32v7, arm64v8, ppc64le
-GitCommit: c031e924f7de45f02a902e2cb88958ab0f343f59
+GitCommit: 07677395574f5d3462c3b6fdf5f6c4a0a350b683
 Directory: 8/jdk/ubuntu/noble
 
 Tags: 8u422-b05-jdk-ubi9-minimal, 8-jdk-ubi9-minimal, 8-ubi9-minimal
 Architectures: amd64, arm64v8, ppc64le
-GitCommit: c031e924f7de45f02a902e2cb88958ab0f343f59
+GitCommit: 07677395574f5d3462c3b6fdf5f6c4a0a350b683
 Directory: 8/jdk/ubi/ubi9-minimal
 
 Tags: 8u422-b05-jdk-windowsservercore-ltsc2022, 8-jdk-windowsservercore-ltsc2022, 8-windowsservercore-ltsc2022
 SharedTags: 8u422-b05-jdk-windowsservercore, 8-jdk-windowsservercore, 8-windowsservercore, 8u422-b05-jdk, 8-jdk, 8
 Architectures: windows-amd64
-GitCommit: c031e924f7de45f02a902e2cb88958ab0f343f59
+GitCommit: 07677395574f5d3462c3b6fdf5f6c4a0a350b683
 Directory: 8/jdk/windows/windowsservercore-ltsc2022
 Builder: classic
 Constraints: windowsservercore-ltsc2022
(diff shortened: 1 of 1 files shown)


gdams:
I've put the diff in a gist as I think it's too long for a comment https://gist.github.com/gdams/de1eec781fc16585276553c7685ff6c8

yosifkit:
Summary of the changes that I see in the diff. Hopefully I didn't miss something.

1. `riscv64` added to architectures with Dockerfile changes to support it :+1:
1. `sh` -> `bash`: hopefully this'll re-fix (and prevent issues for) those using dotted environment variables
1. `entrypoint.sh` templating (removal of unused blocks per variant): This seems good. 
1. boilerplate Apache 2.0 license headers: No problems here. :+1:
1. `ENV` space -> `=`: buildkit linting-induced churn adding to the noise (https://github.com/moby/buildkit/issues/5130).
1. executable bit removed from git for `entrypoint.sh`: This with the next are a bit confusing. What issue is solved by changing from one to the other? Has something gone wrong with tracking executability via git that the `--chmod` via `COPY` was the better solution?
1. executable bit added to `entrypoint.sh` via `COPY` (`--chmod`)

So, just a question on the executability bit. :arrow_up:

gdams:
Thanks for the summary @yosifkit. That's correct, at one point we broke an update because the bits changed. We figured that now that build kit supports the chmod flag in the COPY command it made the most sense to explicitly define the permissions there. Are we doing this correctly?

yosifkit:
> Are we doing this correctly?

It looks accurate, just a bit unusual is all.

Newest change is just https://github.com/adoptium/containers/pull/636. 👍 Everything seems fine to me.

yosifkit (approved):

gdams:
@yosifkit looks like CI is failing: https://doi-janky.infosiftr.net/job/multiarch/job/riscv64/job/eclipse-temurin/14/console

[0.024s][error][os] Syscall: RISCV_FLUSH_ICACHE not available; error='Operation not permitted' (errno=EPERM)

This is the error caused by https://github.com/moby/moby/issues/44319 which we've managed to work around locally by starting the container with 

--security-opt seccomp=unconfined

gdams:
CC @luhenry who was looking at this recently for Adoptium

yosifkit:
🤔 That's super strange as we are running new enough versions of both Docker and Buildkit to have the updated seccomp profile. We'll keep looking on our side to see what we can figure out.

https://github.com/moby/moby/blob/v23.0.11/profiles/seccomp/default.json#L575-L585
https://github.com/moby/buildkit/blob/v0.13.2/vendor/github.com/docker/docker/profiles/seccomp/default.json#L572-L582

tianon:
I tested with Docker v26 from Debian unstable and it _also_ fails with the same error (and libseccomp2 is extremely up-to-date :joy:), so something really suspicious is happening here, because as far as I can tell, the profile looks right. :thinking:

gdams:
For what it's worth, it builds okay on my Mac using buildkit specifying riscv64 as the platform

gdams:
> For what it's worth, it builds okay on my Mac using buildkit specifying riscv64 as the platform

@yosifkit @tianon is there a way we could build the image on an intel host rather than a riscv64 one?

yosifkit:
> is there a way we could build the image on an intel host rather than a riscv64 one?

Unfortunately, we cannot; none of our build hosts are setup with qemu user mode emulation and our build orchestration cannot handle scheduling a "cross" build like this.

sxa:
The simplest bypass for this for now would be to avoid running `java` i.e. exclude the `-Xshare:dump` which re-populates the shared class cache in the docker environment. We should be safe without it although it may have an impact on startup performance. I'll be honest I wasn't aware we were doing the cache regeneration until now ... but I'll note that it likely means the shared class cache in the docker image isn't the same as the ones in our tarballs ...

The 'verification' step after that in the Dockerfiles also try tun run `java -version` which I'd expect to suffer a similar fate if we switch off the cache generation, so maybe we need to bypass that on riscv64 for now too if we can't run in another environment. I'm not sure if any further tests are invoked after the dockerfile is run that might trip up on this too.

We do need to be clear that end-users will hit this too if they run the containers without the `seccomp` (or `--privileged`) options.

tianon:
> We do need to be clear that end-users will hit this too if they run the containers without the `seccomp` (or `--privileged`) options.

Yeah, this is the part I'm most worried about -- as far as I can tell, this is "fixed" in all the seccomp profiles and fully supported in libseccomp, so I'm not even sure _why_ it's failing or how the profile would need to change to make it succeed.

Generally, we shouldn't have users running with `--privileged`, and `--security-opt seccomp=unconfined` is a slightly better workaround, although still not ideal.

gdams:
I've done a bit of digging here and discovered that the issue seems to be with the architecture detection in the seccomp profile.

running with this will fail:

```json
		{
			"names": [
				"riscv_flush_icache"
			],
			"action": "SCMP_ACT_ALLOW",
			"includes": {
				"arches": [
					"riscv64"
				]
			}
		},
```

but if you remove the includes it will work:

```json
		{
			"names": [
				"riscv_flush_icache"
			],
			"action": "SCMP_ACT_ALLOW"
		},
```

I'm not a seccomp expert so I might have to delegate to @tianon or @yosifkit but this does seem to be a good starting point. I did try playing around with the arches values such as `riscv` but to no avail. I've checked that the arches value maps to `uname -m` which it does so I'm not sure where the seccomp actually gets it's architecture value from

yosifkit:
Ah, that gives me a clue as to where the problem might be. Let me go check `moby/moby`.

gdams:
@yosifkit could it be this file that needs to be modified?

https://github.com/moby/moby/blob/5c96ed41097e09cc34d99e8a4ba11e4b300b6d52/profiles/seccomp/seccomp_linux.go#L28-L62

yosifkit:
I was looking at the same spot; it doesn't have a mapping for `riscv64` and I think that's the problem. Can you file a bug in moby/moby? (If you want to make a PR too, I wouldn't say no. ❤️)

gdams:
@yosifkit see:

https://github.com/moby/moby/issues/48454 and https://github.com/moby/moby/pull/48455

gdams:
@yosifkit will I need to raise the same PR in the buildkit repo?

tianon:
I believe buildkit uses Moby's profile, so it should pick it up once it's in + vendoring updated :+1:

gdams:
@tianon do we need to wait for the next releases of moby now or can we unblock the CI?

tianon:
I can backport the patch to the builds of moby we use, but it's significantly easier (and already done, as you noticed :joy:) for me to backport the patch to our _buildkit_ builds, and the timing is perfect because I _think_ Temurin was in the next set of images moving to our newer build system that would use those buildkit builds: https://github.com/docker-library/meta-scripts/pull/78 :eyes:

PR #17407, opened 2024-08-21
Update openjdk

yosifkit:
Changes:

- https://github.com/docker-library/openjdk/commit/9561b4f: Update 23

According to the schedule this should be the "Final Release Candidate" for 23: https://openjdk.org/projects/jdk/23/. That means this will likely be the final 23 available in the `openjdk` tags since we only continue to provide them for pre-releases.

Reminder to use one of the other Official Images for GA releases: https://github.com/docker-library/docs/blob/f80da10a27ea507d6e916d33267c3217b5905d8a/openjdk/README.md#deprecation-notice


tianon (approved):
it's the final countdown!

PR #17421, opened 2024-08-23
Update tomcat

yosifkit:
`riscv64` 🎉


tianon (approved):
It's a risc, but we'll take it

PR #17446, opened 2024-08-28
Test that setuptools and wheel are installed (or not)

tianon:
In Python 3.12+, setuptools and wheel should be removed (but should be present in older versions).

Technically this should be tested on Python 2 as well, but I opted not to because the interfaces for doing so are different and that only exists in a single variant of PyPy (that frankly I don't care too much about verifying the "correctness" of as much).

https://github.com/docker-library/python/pull/954

### DIFF
--- a/test/tests/python-imports/container.py
+++ b/test/tests/python-imports/container.py
@@ -1,8 +1,8 @@
-import platform
+import platform, sys
 
 isWindows = platform.system() == 'Windows'
 isNotPypy = platform.python_implementation() != 'PyPy'
-isCaveman = platform.python_version_tuple()[0] == '2'
+isCaveman = sys.version_info[0] == 2
 
 if not isWindows:
     import curses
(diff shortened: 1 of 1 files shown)


tianon:
https://github.com/docker-library/python/pull/954#issuecomment-2316165609

In addition to that verification, I've _also_ verified that this works fine with current `pypy:2-slim` and `pypy:3-slim`.

tianon:
To illustrate what "failure" looks like:
```console
$ test/tests/python-imports/run.sh python:3.12-alpine3.20
Traceback (most recent call last):
  File "/tmp/test-dir/python-imports/./container.py", line 32, in <module>
    assert(hasSetuptools == shouldHaveSetuptoolsAndWheel)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

whalelines (approved):
I think I have followed the double negatives properly

yosifkit (approved):

PR #17459, opened 2024-08-30
Update `python-hy` test version numbers/constraints

tianon:
I've updated the logic a little bit so that this is easier to update correctly in the future (in theory -- I'm sure Python 4 is going to show up shortly *just* to spite me).


tianon:
```console
$ DOCKER_HOST=ssh://my-windows-test-box test/tests/python-hy/run.sh python:3.12-windowsservercore
4
42
```

yosifkit (approved):

PR #17461, opened 2024-09-01
Update Clear Linux OS to 42280

clrbuilder:
Update Clear Linux OS latest+base images to release 42280

@bryteise @djklimes @bryteise


bryteise:
I think we are going to wait for the next image update after looking that issue over.

tianon (approved):

PR #17462, opened 2024-09-01
emqx 5.8.0

id:


tianon (approved):
All the older versions here are still actively supported as well, right?
(If not, just a reminder that we should remove them from the file :heart:)

id:
Yes. 5.1.x family EOL is December this year.

PR #17468, opened 2024-09-03
MariaDB - bug fix - healthcheck.sh

grooverdan:
healthcheck.sh had some code paths where an incorrect healthy status could have been returned.

Also correct a noisy install where the messages didn't relate to the container release.


tianon (approved):

grooverdan:
Thank you.

PR #17469, opened 2024-09-03
Update XWiki LTS version to 15.10.12

mflorea:


tianon (approved):
I think `~` is a slightly dangerous choice of a `sed` separator (given the special meaning the shell will often apply to it), but that's not actually new here. :smile:

PR #17470, opened 2024-09-03
Fedora Update 2024-09-03

cverna:


tianon:
Does that libssl update include an early release of the fix for today's CVE-2024-6119 (https://openssl-library.org/news/secadv/20240903.txt)?  If not, did you want to respin this to include the fix for that across all the Fedora images, or would you rather wait for the next image update?

cverna:
Closing in Favor of https://github.com/docker-library/official-images/pull/17564 😄

PR #17471, opened 2024-09-03
[Friendica] update 2024.09-dev

ne20002:
update 2024.09-dev 
 - revert from bookworm to bullseye due to problems with dependency


tianon:
Hopefully the problem is short-term :crossed_fingers:

(bullseye _just_ transitioned from being supported by the Debian Security Team to being supported instead by the Debian LTS Team)

tianon (approved):

PR #17475, opened 2024-09-03
Update ruby

tianon:
Changes:

- https://github.com/docker-library/ruby/commit/04175a1: Update 3.3 to 3.3.5

### DIFF
--- a/library/ruby
+++ b/library/ruby
@@ -34,34 +34,34 @@ Architectures: amd64, arm32v6, arm32v7, arm64v8, i386, ppc64le, s390x
 GitCommit: f69cac9888c7909b402a1bc21a87331429318aae
 Directory: 3.4-rc/alpine3.19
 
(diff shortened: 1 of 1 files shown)


zavan:
A note that now Ruby follows a [bi-monthly release schedule](https://www.ruby-lang.org/en/news/2024/09/03/3-3-5-released/) for patch versions:


> ### Release Schedule
> As previously [announced](https://www.ruby-lang.org/en/news/2024/07/09/ruby-3-3-4-released/), we intend to release the latest stable Ruby version (currently Ruby 3.3) every 2 months following a .1 release.
> 
> We expect to release Ruby 3.3.6 on November 5th and Ruby 3.3.7 on January 7th. If any significant changes arise that impact a large number of users, we may release a new version earlier than scheduled.

zavan:
More details on the above: https://rubyonrails.org/2024/10/15/new-maintenance-policy-and-eol-annouments

PR #17549, opened 2024-09-12
Redis CE 8.0 Milestone 1 release for Debian

adamiBs:
Related links:

1. https://github.com/redis/docker-library-redis/commit/8f1a23f9fd78e9d227bcc59ed43f571614151a63
2. https://github.com/redis/redis/releases/tag/8.0-m01

### DIFF
--- a/library/redis
+++ b/library/redis
@@ -1,8 +1,12 @@
-# This file was generated via https://github.com/redis/docker-library-redis/blob/d3809987c491cc61fa4f3c05ef0f4044e7fb6000/generate-stackbrew-library-mac.sh
-
-Maintainers: David Maier <david.maier@redis.com> (@dmaier-redislabs),
+Maintainers: Adam Ben Shmuel <adam.ben-shmuel@redis.com> (@adamiBs),
              Yossi Gottlieb <yossi@redis.com> (@yossigo)
 GitRepo: https://github.com/redis/docker-library-redis.git
+
+Tags: 8.0-M01, 8.0-M01-bookworm
+Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
+GitCommit: af8fe134a94d9d3ac4c696a5d8fd0096e7df6794
+GitFetch: refs/heads/release/8.0
+Directory: debian
 
 Tags: 7.4.0, 7.4, 7, latest, 7.4.0-bookworm, 7.4-bookworm, 7-bookworm, bookworm
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x


tianon:
Because our build system needs to handle a complex dependency tree between images, we unfortunately cannot support variables inside dependency statements (`FROM` lines, in our case).  Additionally, `ADD` with a remote file has a lot of poor behaviors such that we cannot support it (again, especially not with a variable, but in general).

I also don't know that we're falling within the constraints of https://github.com/docker-library/faq#multi-stage-builds

(there are also some other failures in the CI, such as the branching / metadata that tells our system _what_ to build and where to get it)

adamiBs:
I've removed the variables (`ARG`) and the retrieval of the remote source code tarball using `ADD`.
Please let me know if this setup complies with the requirements from the official build system's setup.
Thanks
@tianon

yosifkit:
I haven't looked at the full content of the Dockerfile, but here are a couple changes to fix the build failures:

From the "Diff Comment" test (and also affecting the build test):
```console
cp: cannot stat 'tar/redis_8.0-M01-bookworm/debian/docker-entrypoint.sh': No such file or directory
```
Change the `COPY` from https://github.com/redis/docker-library-redis/blob/6c299a6986fa7aabaf9683002dc6c528a6969dc3/debian/Dockerfile#L163C1-L164C1 to `COPY docker-entrypoint.sh /usr/local/bin/` so that it is relative to the `Directory` in the relevant section of the `library/redis` file.

From the "Naughty" test:
```console
   - commit 6c299a6986fa7aabaf9683002dc6c528a6969dc3:
     - is not in the specified ref GitFetch: refs/heads/master
```
This one will need a `GitFetch: refs/heads/release/8.0` added the `8.0` section of the `library/redis` file to tell our tools which branch the commit comes from.

adamiBs:
@yosifkit I see that the diff CI passes now. Thanks for the explanations and directing me towards a setup that would work in Docker's CI system!

yosifkit:
The entrypoint changes are causing the image to fail on the `override-cmd` test. This test helps ensure our ["Consistency"](https://github.com/docker-library/official-images/blob/050756878846f85a0cb039a629c4245de74550b1/README.md#consistency) requirement, which basically means that `docker run -it --rm redis bash` works for `bash` or any binary in the image. Because of this, I'd recommend wrapping your new functionality to only run if `redis-server` is the command being run. I'd also recommend using the argv array instead of relying on `eval` (like the `set --` line earlier in the script) and checking that the `.so` is likely to work (like existence, size, or that it isn't a directory).  So, roughly something like this?

```sh
if [ "$1" = 'redis-server' ]; then
	modules_dir='/usr/local/lib/redis/modules'

	for module in "$modules_dir"/*.so; do
		# TODO warn/error if the `.so` doesn't have a size, `-s`?
		# or is a directory, `-d`?
		# or isn't readable by the running user, `-r`? (since we'll be the `redis` user at this point)
		if [ ! -d "$module" ] && [ -s "$module" ]; then
			set -- "$@" --loadmodule "$module"
		fi
	done
fi

exec "$@"
```

adamiBs:
@tianon / @yosifkit Could you please take a look so that I could know if we should modify anything?
Thanks 🙏

tianon:
I still agree with @yosifkit that you ought to switch from building up a string and then `eval`ing it to using `set` to add things to `"$@"` directly (https://github.com/docker-library/official-images/pull/17549#issuecomment-2350726819), but I don't think that's a blocker.

tianon (approved):
It looks like 7.0 is EOL (unless I'm reading the dates wrong); if so, it should be removed here.

To be clear / as a refresher, removing tags here will remove them from the "Supported" section on the Hub readme (and will prevent us from spending cycles rebuilding them on the official build servers), but the tags will still be available to users who want them.  (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)

adamiBs:
Thank you for approving @tianon 

I will look into both of the comments that you left and will address them accordingly in a future PR.

PR #17553, opened 2024-09-13
Update library/liquibase to 4.29.2

liquibot:

### DIFF
--- a/library/liquibase
+++ b/library/liquibase
@@ -4,10 +4,10 @@ GitRepo: https://github.com/liquibase/docker.git
 
 Tags: 4.29, 4.29.2, latest
 GitFetch: refs/heads/main
-GitCommit: 079c4169361e913b9477fc1fd93692974d4f10c5
+GitCommit: 13d063767623e282539b232a3a9ed19f4b3d7bbd
 File: Dockerfile
 
 Tags: 4.29-alpine, 4.29.2-alpine, alpine
 GitFetch: refs/heads/main
-GitCommit: 079c4169361e913b9477fc1fd93692974d4f10c5
+GitCommit: 13d063767623e282539b232a3a9ed19f4b3d7bbd
 File: Dockerfile.alpine


jandroav:
This PR fixes CVE-2024-34158

tianon:
> This PR fixes [CVE-2024-34158](https://github.com/advisories/GHSA-j7vj-rw65-4v26)

I'm confused - does liquibase parse Go code?

jnewton03:
hi @tianon, no but a dependency (lpm) does.  Here is the PR to update LPM:
https://github.com/liquibase/liquibase-package-manager/pull/435

And the PR that updated LPM in our Image:
https://github.com/liquibase/docker/pull/334

tianon:
I'm still missing something -- LPM doesn't appear to parse Go code either, so wasn't and isn't vulnerable to CVE-2024-34158. :thinking:

I ran [`govulncheck`](https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck) against v0.2.7's source (which required me to delete one or the other of `cmd/lpm/darwin.go` or `cmd/lpm/windows.go` because otherwise the code as-is is redeclaring `main` which isn't ideal and there are Go native ways to solve whatever problem that redefinition exists for, like build tags), and it confirmed that none of the code there was vulnerable.

I also ran it with `-mode=binary` on the release binaries from https://github.com/liquibase/liquibase-package-manager/releases/tag/v0.2.7 and got the same result ("Your code is affected by 0 vulnerabilities.").

The update looks fine, but I want to be very clear that CVE-2024-34158 is not something this was/is vulnerable to.

jnewton03:
hey @tianon, all great questions.  Anytime there are issues in our Image highlighted by Scout or Trivy we can't pretty much guarantee customers are going to reach out and ask us to fix it even if it's not vulnerable.  So we've found it's easier to just go ahead and patch it, run E2E, and if it's green we'll apply the fix and re-ship the Image.  Here's where we saw the golang alert:
https://github.com/liquibase/docker/actions/runs/10825809551/job/30035496105

yosifkit (approved):

PR #17561, opened 2024-09-16
Create devcontainer.json

fdrm12:


yosifkit:
With no explanation, this is indistinguishable from spam. We have no need for `devcontainer.json`.

PR #17571, opened 2024-09-17
[Swift] Release Swift 6.0 Docker images

shahmishal:


tianon (approved):
The number of supported versions is getting long again (5.5, 5.6, 5.7, 5.8, 5.9, 5.10, and now 6,.0) -- are any of those versions end-of-life? :eyes:

PR #17572, opened 2024-09-17
Update ALT Linux - 20240917

FNDenisovna:


tianon:
```diff
+LABEL org.opencontainers.image.licenses="GPLv3"
```

This seems kind of strange -- surely not _all_ of the software contained within the `alt` image is GPLv3, right?  (and there's no "main" software here, like in an image for PHP, MySQL, etc)

FNDenisovna:
Hi, @tianon .
According to our license we distribute a software kit under the terms of gplv3, so we want to mark this point with container info. By software kit we mean both free and non-free components licensed by copyright owners, and we hold copyright to the kit as a composite work.

PR #17587, opened 2024-09-19
Update Redis Versions List & Entrypoint Implementation for 8.0

adamiBs:
- Removal of Redis 7.0.
- Replace usage of `eval` with `exec` in the entrypoint script to avoid the risk of arbitrary code execution.

Relates to:
- https://github.com/docker-library/official-images/pull/17549#pullrequestreview-2310314318
- https://github.com/docker-library/official-images/pull/17549#issuecomment-2356427812


tianon (approved):
(I think you might've left in the `command='exec "$@"'` line above the `for` loop, but it should be harmless :+1:)

PR #17591, opened 2024-09-19
Add support for Semeru 23 container images and update tomcat version.

ramachandrabolisetti:
This update contains:

- Semeru 23 container images added.
- Tomcat version update.

### DIFF
--- a/library/ibm-semeru-runtimes
+++ b/library/ibm-semeru-runtimes
@@ -7,54 +7,54 @@ GitFetch: refs/heads/ibm
 #-----------------------------openj9 v8 images---------------------------------
 Tags: open-8u422-b05-jdk-focal, open-8-jdk-focal
 Architectures: amd64, ppc64le, s390x, arm64v8
-GitCommit: 72028e92c6d8f289986ae96707bb440bb6de0861
+GitCommit: 1212e4fe213cb5b4c65eb260ccbbc40a7eadfb5d
 Directory: 8/jdk/ubuntu/focal
 File: Dockerfile.open.releases.full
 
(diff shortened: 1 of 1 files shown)


tianon:
22 is now EOL as well, right? (and thus should be removed?) :eyes:

ramachandrabolisetti:
I will discuss on the EOS and I will remove 22 tag once the timeline is confirmed. Thanks !!

tianon:
```
+FROM ubuntu:23.04
```

The images tagged "jammy" aren't from jammy :sweat_smile:

(FWIW, these _could_ be `FROM ubuntu:jammy` so it's easier to keep them in sync correctly)

tianon (approved):

yosifkit:
`ibm-semeru-runtimes` images are failing to rebuild after base image updates (Ubuntu rebuilds) because the "old" tomcat is no longer available through the CDN. Can you provide some updated Dockerfiles? Perhaps they could fallback to `https://archive.apache.org/dist/` when it is no longer available on the primary Apache CDN? It is up to you if you want to add the archive fallback. (We have a block that does this for our [`cassandra` images](https://github.com/docker-library/cassandra/blob/1e3d5732f34ceb9e77870d0be9501515f917cc60/5.0/Dockerfile#L107-L118)).
```console
+ curl -LfsSo /tmp/tomcat/tomcat.tar.gz https://dlcdn.apache.org/tomcat/tomcat-9/v9.0.95/bin/apache-tomcat-9.0.95.tar.gz
curl: (22) The requested URL returned error: 404 
```

PR #17593, opened 2024-09-19
Add Corretto23 images

chadrako:
Add Corretto23 official images now that Corretto23 is available for general access


yosifkit:
22 is now EOL as well, right? (and thus should be removed?) :eyes:

yosifkit (approved):

Rudometov:
Corretto 22 is supported until October, though I wouldn't expect updates there

Issue #16404, opened 2024-03-13
Add LoongArch architecture support


zhaixiaojuan:
The LoongArch architecture (LoongArch) is an Instruction Set Architecture (ISA) that has a RISC style.

Documentations:
ISA:
https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN.html
ABI:
https://loongson.github.io/LoongArch-Documentation/LoongArch-ELF-ABI-EN.html
More docs can be found at:
https://loongson.github.io/LoongArch-Documentation/README-EN.html

The alpine community is currently adding support for the LoongArch architecture, and the community has already merged some patches. According to the current progress, the alpine upstream community should soon support the LoongArch architecture.
After the alpine upstream community supports the LoongArch architecture, can the Docker official image consider adding a LoongArch architecture image? If you need machine or other help, we will fully support you.

Key PR merged by alpine community:
https://gitlab.alpinelinux.org/alpine/tsc/-/issues/72
https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/61611
https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/61612

whalelines:
Thank you for the request to add support for LoongArch to DOI, @zhaixiaojuan! We really appreciate your interest in DOI.

Unfortunately at this time we have no plans to add support for LoongArch. We will reevaluate this decision as the situation evolves.

Thanks again and take care!

zhaixiaojuan:
@tianon @whalelines 
Thank you very much for your reply. Currently, go, gcc, kernel, llvm and other upstreams already support the loongarch architecture, and alpine and debian systems are also integrating patches for the loongarch architecture.

I very much hope that the official image can also support the loongarch architecture. But at the moment I don't know exactly what needs to be done to achieve this goal. I hope to get some specific evaluation indicators. I will prepare in advance based on these indicators and strive to achieve this goal as soon as possible. Can you help us?

Thank you very much and look forward to your reply and guidance.

whalelines:
You have noted some metrics we track but there are other internal metrics as well as support and compliance considerations. As such, we cannot communicate a checklist of prerequisites. Ultimately, supporting additional architectures for DOI requires a significant investment from Docker and while we appreciate your interest and enthusiasm, at this time we cannot justify taking on an additional architecture. As mentioned previously, we will continue to monitor the situation.

zhaixiaojuan:
Thank you very much for your reply

[...]

PR #17607, opened 2024-09-23
alpine: edge snapshot 20240923, add loong64

ncopa:
Introduce loongarch64

### DIFF
--- a/library/alpine
+++ b/library/alpine
@@ -1,13 +1,14 @@
 Maintainers: Natanael Copa <ncopa@alpinelinux.org> (@ncopa)
 GitRepo: https://github.com/alpinelinux/docker-alpine.git
 
-Tags: 20240807, edge
-Architectures: arm64v8, arm32v6, arm32v7, ppc64le, riscv64, s390x, i386, amd64
+Tags: 20240923, edge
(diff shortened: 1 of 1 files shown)


tianon:
As noted in https://github.com/docker-library/official-images/issues/16404#issuecomment-2369792476, this unfortunately isn't something we can/will support currently.  We're happy to accept the image refresh, but the Loongson architecture will need to be dropped from it.  Thanks!

xen0n:
Hi, I'm an individual LoongArch porter who has recently been notified of the `loong64` support situation as mentioned above. In order to move forward and unblock things such as [`linux-loongarch64-musl` support for this .NET SDK unofficial distribution](https://github.com/loongson-community/dotnet-unofficial-build/issues/3) I've been working on lately, would it be okay to distribute the `linux/loong64` Alpine images under another Docker Hub account or another container registry altogether, and preferably linking to the alternate location from various doc pages?

ncopa:
> would it be okay to distribute the `linux/loong64` Alpine images under another Docker Hub account or another container registry altogether, and preferably linking to the alternate location from various doc pages?

This is exactly the thing I want avoid.

PR #17608, opened 2024-09-23
Update Debian

tianon:


tianon:
Gah this still includes more `bullseye` architectures than it should :disappointed:

tianon:
(converting to draft while I decide whether I want to do more manual effort to shrink that list or just continue to let it ride :disappointed:)

tianon:
Nice, looks good now

PR #17612, opened 2024-09-24
Gradle 8.10.2 and Java 23

keeganwitt:


keeganwitt:
While replacing Java 22 with 23, the new noble eclipse-temurin image introduced a breaking change, which required additional changes on my side. Any feedback on that is welcome: https://github.com/keeganwitt/docker-gradle/pull/295#issuecomment-2370004190

tianon:
> While replacing Java 22 with 23, the new noble eclipse-temurin image introduced a breaking change, which required additional changes on my side. Any feedback on that is welcome: [keeganwitt/docker-gradle#295 (comment)](https://github.com/keeganwitt/docker-gradle/pull/295#issuecomment-2370004190)

Yeah, it's kind of an irritating situation all around, and there isn't really a _good_ answer, unfortunately :disappointed:

Another possible option is to use `useradd --non-unique` to just keep using UID 1000 and letting _both_ users have it, but that has downsides too.  If you're satisfied with the option you've chosen though, it's a fine one (and users who were relying on UID 1000 can take the breaking opportunity to more accurately hard-code their assumption by switching to numeric `--user 1000:1000`/`user: '1000:1000'` instead of relying on the image :eyes:).

tianon (dismissed):
(approving so we're ready to merge if you're happy with the solution you've chosen, but I want to make sure you've got time to consider "two UID 1000 users" before we merge in case you didn't already think about that one and whether you like it any worse/better)

keeganwitt:
What are the downsides of two users with the same uid? They can share files in a way the user might not expect?

keeganwitt:
If I change the UID, should I choose a higher number to make it less likely this happens again (like 3000)? Or maybe it should be a random number?

tianon:
Sharing the same UID shouldn't really have any obvious ill effect -- the kernel / filesystems only care about UID/GID, the names are a userspace abstraction.  Most programs will parse the file and just choose one or the other based on the way they've implemented their parsing (for a reverse lookup).  There might be some programs that emit warnings or errors, but IMO that'd be a bug in the program.

Normally our advice is to choose the service's main default port number when in doubt, but there's not really a default port number for something like Gradle, so it's a hard call either way.  I don't think it's very likely this happens again, but it's not impossible, so choosing a higher number wouldn't be a terrible plan.  It would also make the change more obvious if it happens to break any users, I think.  On the other hand, the main benefit of using 1000:1000 is that it's most likely to match the user's host system, and Gradle is a tool that I think is probably likely to be used in that way (so it's easier for users who don't understand how all this works to get it "right").

keeganwitt:
If I wanted to share the same uid and gid, how would I do that? `groupadd --system --gid 1000 gradle` fails with `groupadd: GID '1000' already exists`. and `useradd --system --gid 1000 --uid 1000 --shell /bin/bash --create-home gradle` fails with `useradd: UID 1000 is not unique`.

tianon:
`useradd --non-unique` :eyes: (and `groupadd --non-unique`)

keeganwitt:
> `useradd --non-unique` 👀 (and `groupadd --non-unique`)

Thank you!

I finally found the upstream issue complaining about this breakage (but I can't find any issue where they discussed this change before it was made): https://bugs.launchpad.net/cloud-images/+bug/2005129. In it, they mention two other options. One is to delete the user (`userdel -r ubuntu` or `touch /var/mail/ubuntu && chown ubuntu /var/mail/ubuntu && userdel -r ubuntu`), the other is to reassign the user (`usermod -l gradle ubuntu`). I wonder if I should consider the reassign option. That kinda seems cleaner.

tianon (approved):

PR #17622, opened 2024-09-26
Update spark 4.0.0-preview2

yaooqinn:
https://github.com/apache/spark-docker/tree/master/4.0.0-preview2

### DIFF
--- a/library/spark
+++ b/library/spark
@@ -1,25 +1,45 @@
 Maintainers: Apache Spark Developers <dev@spark.apache.org> (@ApacheSpark)
 GitRepo: https://github.com/apache/spark-docker.git
 
-Tags: 4.0.0-preview1-scala2.13-java17-python3-ubuntu, 4.0.0-preview1-python3, 4.0.0-preview1
+Tags: 4.0.0-preview2-scala2.13-java21-python3-ubuntu, 4.0.0-preview2-java21-python3, 4.0.0-preview2-java21
 Architectures: amd64, arm64v8
-GitCommit: b9f1f8e8ebed1959c2be3864a114b52f67519092
-Directory: ./4.0.0-preview1/scala2.13-java17-python3-ubuntu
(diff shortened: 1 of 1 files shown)


tianon:
```diff
+    ln -s "$(basename $(ls /opt/spark/examples/jars/spark-examples_*.jar))" /opt/spark/examples/jars/spark-examples.jar; \
```

This chain is definitely eyebrow-raising :sweat_smile:

Can you explain a bit what it's trying to do?  Does `/opt/spark/examples/jars/spark-examples_*.jar` match multiple things, or just one?  Perhaps the `--relative` flag to `ln` is what you're looking for here so you can avoid both subshells?

yaooqinn:
This is a user experience improvement for referencing it for quickstart w/ spark examples.

FYI, https://github.com/apache/spark-docker/pull/67

tianon:
Sorry, let me be more clear -- `ls` in any scripting context is almost always a "code smell" because it is primarily a user interface, not intended for scripting use.

In this case, I think what you meant by this:

```diff
+    ln -s "$(basename $(ls /opt/spark/examples/jars/spark-examples_*.jar))" /opt/spark/examples/jars/spark-examples.jar; \
```

was really much more simply stated as:

```diff
+    ln -sr /opt/spark/examples/jars/spark-examples_*.jar /opt/spark/examples/jars/spark-examples.jar; \
```

(with what should be the exact same end result, but with less fiddly subshell layers in between where things could unintentionally go wrong)

tianon:
Another alternative that's still valid, but only a single subshell (still not ideal, but more correct than the current that's shelling out to `ls` as well):

```diff
+    ln -s "$(basename /opt/spark/examples/jars/spark-examples_*.jar)" /opt/spark/examples/jars/spark-examples.jar; \
```

(The _ideal_ solution would be to ship this symlink as part of the original upstream dist, but that's perhaps a separate topic.)

tianon (approved):

PR #17629, opened 2024-09-26
Update SBOM indexer to 1.14.1

tianon:
This fixes the issue with empty SBOMs in the new GHA images / Ubuntu 22.04 update.

See also:

- https://github.com/docker-library/meta-scripts/pull/82 (which we can revert after this goes in)
- https://github.com/anchore/syft/pull/3269


whalelines (approved):
LGTM

https://oci.dag.dev/?image=docker%2Fscout-sbom-indexer%3A1-doi

PR #17634, opened 2024-09-27
maven: JDK 23

carlossg:


patpatpat123:
Thank you Carlos for this

tianon:
Not a blocker, and I plan to merge in a moment in spite of it (because it'll be a trivial follow-up if it's an issue), but this discrepancy in the list of packages installed between these different Alpine variants is expected/intentional, right?

```diff
-FROM eclipse-temurin:8-jdk-alpine
+FROM amazoncorretto:17-alpine
 
-RUN apk add --no-cache bash procps curl tar openssh-client
+RUN apk add --no-cache bash openssh-client
```

tianon (approved):

carlossg:
> Not a blocker, and I plan to merge in a moment in spite of it (because it'll be a trivial follow-up if it's an issue), but this discrepancy in the list of packages installed between these different Alpine variants is expected/intentional, right?

thanks for checking, yes it is fine, different distros have different packages for backwards compatibility

tianon:
FYI, JDK 23 is now EOL, so these should all be removed again :sweat_smile:

PR #17635, opened 2024-09-27
Haskell: upgrade to stack 3.1.1

jhrcek:


tianon (approved):

tianon:
FWIW, it's probably also worth looking at Bookworm now that Bullseye is in LTS (still two years in LTS and coverage for all the architectures y'all care about here, but something to be aware of / looking at :eyes:)

PR #17636, opened 2024-09-27
Release SonarQube 10.7 and 9.9.7

jCOTINEAU:
hello everyone 

This PR contains the releases of SonarQube 10.7.0 and 9.9.7.

Thanks a lot in advance

### DIFF
--- a/library/sonarqube
+++ b/library/sonarqube
@@ -3,44 +3,44 @@ Maintainers: Carmine Vassallo <carmine.vassallo@sonarsource.com> (@carminevassal
              Davi Koscianski Vidal <davi.koscianski-vidal@sonarsource.com> (@davividal)
 GitRepo: https://github.com/SonarSource/docker-sonarqube.git
 Architectures: amd64, arm64v8
-GitCommit: 29b65682b4f8dbb38bd303b35b3ceaa3a39a0e40
+GitCommit: 50701464f6c6e1ec93b0692fc9525f970a2b5d93
 
-Tags: 9.9.6-community, 9.9-community, 9-community, lts, lts-community
+Tags: 9.9.7-community, 9.9-community, 9-community, lts, lts-community
 Architectures: amd64, arm64v8
 Directory: 9/community
 
-Tags: 9.9.6-developer, 9.9-developer, 9-developer, lts-developer
+Tags: 9.9.7-developer, 9.9-developer, 9-developer, lts-developer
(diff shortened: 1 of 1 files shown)


tianon:
```diff
+    apt-get clean;
```

This is unnecessary, and is a no-op in your base images: :eyes:

https://oci.dag.dev/layers/eclipse-temurin@sha256:c2872fbbc15dc6cd39f606eb912279bc41b5e791314f371a3a9c796977e702ea/etc/apt/apt.conf.d/docker-clean

Linking to the same config file in `debian:bookworm-slim` because it has comments that help explain the purpose/intent/effect better:

https://oci.dag.dev/layers/debian@sha256:7095ea629c4563714b9655137db2eacd456eb3eea0eb8a2b0a4a6b0b187220a9/etc/apt/apt.conf.d/docker-clean

```diff
+VOLUME ["${SQ_DATA_DIR}", "${SQ_EXTENSIONS_DIR}", "${SQ_LOGS_DIR}", "${SQ_TEMP_DIR}"]
```

Just to make sure, these are all "runtime state" directories, right?  See https://github.com/docker-library/official-images/pull/2437#issuecomment-266578827 for a good explanation of how we recommend using `VOLUME`s to avoid some of their rougher edges :eyes:
(and of course, users can always mount anywhere -- no `VOLUME` necessary for that :+1:)

jCOTINEAU:
Hello @tianon, thanks a lot for the prompt feedback.

For the volumes, I think they are now all "runtime state." The docker image is meant to be used locally with docker to try out Sonarqube, and in this case, all the documentation / docker-compose examples are using those volumes by default.

For production, our supported target deployment platform is kubernetes/Openshift through our helm chart, and there also, all volumes are being mounted and used by the chart directly. (even the extension one, which is populated by an initContainer on demand).

And we decided to make those volumes explicit in the Dockerfile because it looks like its an OpenShift best practice  from [this guide](https://docs.openshift.com/container-platform/4.16/openshift_images/create-images.html) as the volume definition will be highlighted to users.

Hope this answers the point, let me know if some more clarifications are needed.

For the apt clean I'll take a deeper look but I cannot find the comment you are mentioning in the link, would it prevent from merging and releasing this PR ?

Thanks a lot again for the kind feedbacks, 

Cheers

tianon (approved):

tianon:
Thanks for updating / confirming!

PR #17637, opened 2024-09-27
Update cassandra

tianon:
Changes:

- https://github.com/docker-library/cassandra/commit/ad1d8a9: Update 4.1 to 4.1.7
- https://github.com/docker-library/cassandra/commit/ef383f1: Update 4.0 to 4.0.14


tianon:
I've retried the CI for 5.0.0 several times and it's not wanting to work -- it's _just_ the smoke test failing, and it's not even one of the versions updated here, so I'm going to merge without it.

PR #17640, opened 2024-09-28
Update a bunch of bullseye-based `Architectures:`

tianon:
See https://wiki.debian.org/LTS and https://github.com/docker-library/official-images/pull/17608


tianon:
Obviously there will be a few rounds of this before we're 100% updated, but this gives us a start.

tianon:
Ahahaha, amazing, because I've included so many here, they're already cascading, but their generate scripts aren't currently designed to handle that -- normally we try to have an escape hatch for things that scrape the library files directly from the source-of-truth such that setting `BASHBREW_LIBRARY` in the environment can override it, which would let me do multiple passes on this, but that's not something we support in the copy-pasta block of code we use to calculate architectures for these, so I'll need to noodle on this problem more next week.

tianon:
Ok, cascading updates should be included here now :crossed_fingers:

tianon:
image

"238 successful checks" :hot_face:

yosifkit (approved):

PR #17644, opened 2024-09-30
Release Neo4j 5.24.1

ericsporre:
Hello! 

Adding a new colleague as maintainer, sorry about some confusion in a previous PR caused by this.

Thanks!

### DIFF
--- a/library/neo4j
+++ b/library/neo4j
@@ -1,27 +1,27 @@
 Maintainers: Jenny Owen <jenny.owen@neo4j.com> (@jennyowen),
-             Gustav Lindroth <gustav.lindroth@neo4j.com> (@glindroth),
-             Eric Sporre <eric.sporre@neo4j.com> (@ericsporre)
+             Eric Sporre <eric.sporre@neo4j.com> (@ericsporre),
+             Tobias Ericsson <tobias.ericsson@neo4j.com> (@neo-tobias)
 GitRepo: https://github.com/neo4j/docker-neo4j-publish.git
 
-Tags: 5.23.0-community-bullseye, 5.23-community-bullseye, 5-community-bullseye, 5.23.0-community, 5.23-community, 5-community, 5.23.0-bullseye, 5.23-bullseye, 5-bullseye, 5.23.0, 5.23, 5, community-bullseye, community, bullseye, latest
+Tags: 5.24.1-community-bullseye, 5.24-community-bullseye, 5-community-bullseye, 5.24.1-community, 5.24-community, 5-community, 5.24.1-bullseye, 5.24-bullseye, 5-bullseye, 5.24.1, 5.24, 5, community-bullseye, community, bullseye, latest
 Architectures: amd64, arm64v8
-GitCommit: 8e55c5a5629327d318a63703f4f4f361c97c6293
-Directory: 5.23.0/bullseye/community
+GitCommit: 585f74b23c369847e4445a3b9f6a30a3c5d80697
+Directory: 5.24.1/bullseye/community
 
-Tags: 5.23.0-enterprise-bullseye, 5.23-enterprise-bullseye, 5-enterprise-bullseye, 5.23.0-enterprise, 5.23-enterprise, 5-enterprise, enterprise-bullseye, enterprise
+Tags: 5.24.1-enterprise-bullseye, 5.24-enterprise-bullseye, 5-enterprise-bullseye, 5.24.1-enterprise, 5.24-enterprise, 5-enterprise, enterprise-bullseye, enterprise
 Architectures: amd64, arm64v8
-GitCommit: 8e55c5a5629327d318a63703f4f4f361c97c6293
-Directory: 5.23.0/bullseye/enterprise
+GitCommit: 585f74b23c369847e4445a3b9f6a30a3c5d80697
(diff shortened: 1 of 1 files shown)


tianon:
Sorry for the delay!

Looking over the diff, I'm very concerned about the OpenSSL-related changes -- I see those are related to FIPS, and limited to UBI9, which seems odd.  Is there no easier way to get/enable FIPS mode for OpenSSL in UBI9 than compiling it from source yourself?  Do you have a plan for updating/maintaining that over time to keep up with things like new releases, especially with relation to security updates to OpenSSL?

To be explicit/clear, I think this would be _lot_ better if we could find some cleaner/more supported way to enable FIPS mode in the OpenSSL that's presumably already shipped in the image, or at least the one maintained by the upstream distribution. :see_no_evil:

jennyowen:
Hi @tianon,

I totally understand the skepticism about compiling and installing OpenSSL from source in the Dockerfile. It's not great at all.

The reason for the change is that we wanted to add a feature so that anyone who needs to run Neo4j in a FIPS compatible way can do so, but that requires a FIPS certified version of OpenSSL to be installed in the image. Currently that would be OpenSSL versions 3.0.0, 3.0.8 or 3.0.9 [citation](https://openssl-library.org/source).

For some reason Redhat only have 3.0.7 in their repositories (not even 3.0.0), so I can't install one of the certified versions from there sadly. I also couldn't find any official docker images that provide specific OpenSSL version builds, so there was nowhere I could just do a neat little `COPY --from` from 😢.
Installing from source seems to be the OpenSSL recommended way if trustworthy binaries aren't available.

That's how I ended up at the current version of our code. I 10000% agree that having some upstream provider of trustworthy OpenSSL binaries would be waaaay preferable though. 

Regarding maintenance, I don't think keeping the OpenSSL installation up to date will be too much of an issue, since OpenSSL themselves don't seem to FIPS certify versions very often (3.0.9 was originally certified on 23/08/2022 [citation](https://csrc.nist.gov/projects/cryptographic-module-validation-program/certificate/4282)). The latest stable version they've published on their website is 3.3.2, but the latest on rhel is a super old 3.0.7, and debian 11 uses 1.1.1f which is end of life. 
Perhaps there are other factors at play (I confess I'm no security expert), but the certification cadence doesn't seem too burdensome?

That being said, I totally understand you need to maintain the quality of docker official images. If you think the code is unpublishable as an official image, then I'll revert the changes and update this PR to be a normal version bump.

yosifkit:
```diff
-    gpg --batch --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys \
-        595E85A6B1B4779EA4DAAEC70B588DFF0527A9B7 \
-        B42F6819007F00F88E364FD4036A9C25BF357DD4; \
+    gpg --batch --keyserver hkps://keys.openpgp.org --recv-keys \
+        6380DC428747F6C393FEACA59A84159D7001A4E5 \
+        A21FAB74B0088AA361152586B8EF1A6BA9DA2D5C; \
```
Just because this looked a little odd to me in wondering why `tini` doesn't fail to `gpg --verify` with its key being removed: `595E85A6B1B4779EA4DAAEC70B588DFF0527A9B7` is a subkey of `6380DC428747F6C393FEACA59A84159D7001A4E5` (https://keyserver.ubuntu.com/pks/lookup?search=595E85A6B1B4779EA4DAAEC70B588DFF0527A9B7&fingerprint=on&op=index)

---

It looks like the RedHat provided openssl-fips library is just pending certification/review: https://access.redhat.com/articles/compliance_activities_and_gov_standards. Maybe that's enough to just use that and not build the upstream OpenSSL in the image?

jennyowen:
I would be ok with using the redhat openssl-fips library in general, but the FIPS certification is *required* to guarantee compliance. Having pending certification isn't enough sadly. 
If it was pending with an expectation of being done in a few weeks then I'd be happy to wait before submitting these changes. However [according to NIST](https://csrc.nist.gov/projects/cryptographic-module-validation-program/modules-in-process/modules-in-process-list) it's been pending certification since December 2023, so I've no confidence in it being done any time soon.

Incidentally, we have to do a new bug fix release either today or tomorrow, so I'll revert these openssl changes when we update this PR. I'll have to resubmit them afterwards though, so your feedback is still very much appreciated!

ericsporre:
Hello again! Updated the PR with the new neo4j release along with the revert of the openssl changes. Thanks!

yosifkit:
```bash
    eval "$base_variable_name=$secret_value"
```
I think this might lead to some accidental evaluation of the secret (since the variable is expanded before the `eval` evaluates the string). I think either escaping the `$` or switching to export should prevent it:
```bash
    eval "$base_variable_name=\$secret_value"
# or 
    export "$base_variable_name"="$secret_value"
```
The `export` is how we do it in the [`postgres` images](https://github.com/docker-library/postgres/blob/a37e929682e8de45a3304a5bf9d63210c2e0a680/docker-entrypoint.sh#L23).

yosifkit (approved):

prasanna-ranga:
Hi @jennyowen , so If I want to use neo4j in FIPS mode, do I have to built the openssl in fips mode? 
I see you have reverted the changes ? any reason for removing fips again for 5.24.0?

PR #17382, opened 2024-08-16
Remove several (persistently) failing builds

tianon:
Each of these has been failing for some period of time with no fix/mitigation forthcoming.  We're happy to see them re-enabled if they are fixed/updated.

- `eclipse-mosquitto`: https://github.com/eclipse/mosquitto/issues/3055
- `node`: ~https://github.com/docker-library/official-images/pull/16714#issuecomment-2096455085, https://github.com/nodejs/docker-node/issues/2107~ https://github.com/docker-library/official-images/pull/17394
- `plone`: https://github.com/plone/plone.docker/issues/161, https://github.com/docker-library/official-images/pull/12915#issuecomment-1426176158
- `rethinkdb`: ~https://github.com/rethinkdb/rethinkdb-dockerfiles/issues/67 (https://github.com/rethinkdb/rethinkdb/issues/7153)~ builds fixed
- `satosa`: https://github.com/IdentityPython/satosa-docker/issues/11#issuecomment-2270115644

PR #17662, opened 2024-10-02
eclipse-mosquitto 2.0.19

ralight:


library/eclipse-mosquitto
```diff
@@ -1,12 +1,9 @@
 Maintainers: Roger Light <roger@atchoo.org> (@ralight)
 GitRepo: https://github.com/eclipse/mosquitto.git
-GitCommit: f4400fa422eacac8417efbc45dd1284526dce8d4
+GitCommit: e96ddfc6e58f7087b9fa57a717d7f56d2c00f9fe
 Architectures: amd64, arm32v6, arm64v8, i386, ppc64le
```
yosifkit, line 4: We disabled your `s390x` builds in https://github.com/docker-library/official-images/pull/17382, but I think the images that are left should work for it now (https://github.com/eclipse/mosquitto/issues/3055). A quick local (emulated) build test works; do you want to add `s390x` back?
```suggestion
Architectures: amd64, arm32v6, arm64v8, i386, ppc64le, s390x
```

ralight: Yes please! I didn't want to complicate the update, but if you're happy it's building then that's ok. I've added the change to the PR.

yosifkit (approved):

PR #17582, opened 2024-09-18
eclipse-temurin: add JDK23 images

gdams:
Skipping nanoserver for now as it's broken

PR #17666, opened 2024-10-02
Update Jetty version to 12.0.14

lachlan-roberts:
Update Jetty Version
`12.0.13` -> `12.0.14`

### DIFF
--- a/library/jetty
+++ b/library/jetty
@@ -99,15 +99,15 @@ Architectures: amd64, arm64v8
 Directory: eclipse-temurin/12.0/jre17
 GitCommit: f0c6a3e167722283788502f39ee521ffdac911a0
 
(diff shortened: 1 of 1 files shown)


yosifkit:
The failing "naughty" test is about `eclipse-temurin:22-*` reaching end of life. Their JDK 22 images were replaced by JDK 23 images in https://github.com/docker-library/official-images/pull/17582. So, the `jetty:12.0.14-jdk22` and `jetty:12.0.14-jdk22-alpine` images should be dropped or replaced with JDK 23 based images

lachlan-roberts:
@yosifkit thanks, I have now updated those images to use JDK23

tianon (approved):

PR #17679, opened 2024-10-06
[Friendica] update 2024.09-dev

ne20002:
Updated dependencies

- apcu 5.1.23 -> 5.1.24
- redis 6.0.2 -> 6.1.0


ne20002:
I'm not sure why the two tests failed because of failure of downloading files. It worked with the last version (before the last push fixing the architectures) and works for me locally.
Maybe someone can just trigger a new build? It may have been a temporarily problem on pecl server.

tianon:
Yep, looks like a retry did the trick :+1:

tianon (approved):

PR #17709, opened 2024-10-10
mediawiki: fix ldd parsing logic

christian-heusel:
As it turns out the liblua5 shared object is needed for runtime aswell
and having it missing produces a warning i.e. when running "php -v". Fix
this by installing it before the build time dependencies that are to be
removed later on.

Fixes https://phabricator.wikimedia.org/T376447

Signed-off-by: Christian Heusel <christian@heusel.eu>

### DIFF
--- a/library/mediawiki
+++ b/library/mediawiki
@@ -3,7 +3,7 @@ Maintainers: Kunal Mehta <legoktm@debian.org> (@legoktm),
              Christian Heusel <christian@heusel.eu> (@christian-heusel)
 GitRepo: https://github.com/wikimedia/mediawiki-docker.git
 GitFetch: refs/heads/main
-GitCommit: 0b41a035fabebeda3d6b4ccc39e2c5fc3ce1cc19
+GitCommit: 4708aabf22334d933a45f714e28a78722b36427a
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, ppc64le
 
 # current stable version


tianon:
Hmm, that's odd -- the whole point of https://github.com/wikimedia/mediawiki-docker/blob/0b41a035fabebeda3d6b4ccc39e2c5fc3ce1cc19/1.42/apache/Dockerfile#L47-L53 is to catch deps like this.

I wonder if it just needs similar fixes to what we did in https://github.com/docker-library/wordpress/commit/7c79eeb7a24484fa9a5aca3161947c821b6b1e0e + https://github.com/docker-library/wordpress/commit/2007d5d69bfa13fd2af075a786bbf544e9f6cbe8 (maybe easier to read in https://github.com/docker-library/wordpress/blob/d6690c71a24ad7a434d1de15f382c41279932fc3/latest/php8.2/apache/Dockerfile#L74-L80)?

christian-heusel:
@tianon thanks for investigating! 😊 I will have a look and incorporate your patch into our repository 😅

christian-heusel:
@tianon could you re-review? 😊 I have reverted the wrong fixed I initially thought was right and implemented your suggestions 😃

tianon (approved):

tianon:
Glad we got it fixed, and hopefully in a way that makes sure you don't see other cases of it in the future (if more libraries need to be added). :+1:

Keepstacking22:
I dnt get this what is it I have to do it jus keeps takin me through loop holes

PR #17734, opened 2024-10-16
[Friendica] update 2024.09-rc

ne20002:
- removed version 2024.03
- added version 2024.09-rc


ne20002:
If possible, it would be helpfull for me to get a short advice regarding the questions in https://github.com/friendica/docker/issues/265 from the docker-library maintainers.

tianon (approved):

PR #17743, opened 2024-10-16
Updated TomEE 10.0.0-M2 to -M3

scriptmonkey:


tianon (approved):

scriptmonkey:
@tianon, thank you for the quick turn around!!

PR #17752, opened 2024-10-18
Add Python 3.14-rc

jnoordsij:
Not sure if/how this slipped under the radar, but the 3.14-rc variants were not yet added here (last PR was #17687).

*Note:* I've deliberately only included addition of the new 3.14-rc images and skipped over the changes to other variants, as they seem not to be substantial enough to warrant a rebuild. Feel free to replace this with a full update however if that's to be preferred.

### DIFF
--- a/library/python
+++ b/library/python
@@ -5,6 +5,53 @@ Maintainers: Tianon Gravi <admwiggin@gmail.com> (@tianon),
 GitRepo: https://github.com/docker-library/python.git
 Builder: buildkit
 
+Tags: 3.14.0a1-bookworm, 3.14-rc-bookworm
+SharedTags: 3.14.0a1, 3.14-rc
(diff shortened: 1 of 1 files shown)


yosifkit:
Yeah, I was slightly waiting on https://github.com/docker-library/python/pull/979 / https://github.com/docker-library/python/pull/980, but it's a bit more complicated than just dropping the `PROFILE_TASK` lines 😢. So, I think adding 3.14-rc without that is fine.

jnoordsij:
I see, all fine to me! I'll just leave this open for anyone interested to track when the new 3.14-rc image will be available.

tianon:
Closing in favor of https://github.com/docker-library/official-images/pull/17757, which now includes https://github.com/docker-library/python/pull/980 :bow:

As always, thank you for your diligence, @jnoordsij :heart:

PR #17773, opened 2024-10-22
Update to Flink 2.0-preview1

reswqa:
The preview release of flink 2.0.

NB. jdk1.8 is no longer supported in flink 2.0


tianon:
2.0-preview1 is a pre-release, right?  So is it intentional that aliases like `latest` have moved to it already?  Is it recommended that users upgrade to it already, or just try it out?

reswqa:
Thanks @tianon, you are right! Tags like `latest`/`java11` shouldn't bind to 2.0 for preview release. I have updated this PR then.

yosifkit (approved):

PR #17779, opened 2024-10-23
Update photon image

gerrit-photon:
Update photon image


michellew-vmware:
Hi,

Could you please hold this PR and we will have a new update later today. thank you.

PR #17807, opened 2024-10-29
Redis CE 8.0 Milestone 2 release for Debian and Alpine

maxb-io:
Added support for Alpine

### DIFF
--- a/library/redis
+++ b/library/redis
@@ -2,9 +2,15 @@ Maintainers: Adam Ben Shmuel <adam.ben-shmuel@redis.com> (@adamiBs),
              Yossi Gottlieb <yossi@redis.com> (@yossigo)
 GitRepo: https://github.com/redis/docker-library-redis.git
 
-Tags: 8.0-M01, 8.0-M01-bookworm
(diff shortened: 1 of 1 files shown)


adamiBs (commented):
LGTM @maxb-io

yosifkit:
While I don't agree with using `break-system-packages` and it is unlikely for a user to install python in the `redis` image, `PIP_BREAK_SYSTEM_PACKAGES=1` should not be an `ENV` and should instead be limited to just where it is needed. So, it can probably just be an `export` in the `RUN` line.

Associated PEP discussion: https://discuss.python.org/t/pep-668-marking-python-base-environments-as-externally-managed/10302

maxb-io:
@yosifkit Thanks for the feedback! I've updated the Dockerfile to remove PIP_BREAK_SYSTEM_PACKAGES=1 from the ENV and added as an inline variable before the relevant command in the specific RUN line instead, so it's limited to where it's needed. Let me know if there's anything else you'd like adjusted.

PR #17835, opened 2024-11-04
Update Tomcat version for Semeru containers.

jayasg12:
Updated tomcat version to 9.0.96


tianon (approved):
I'd love to see this updated with an explicit fallback as described in https://github.com/docker-library/official-images/pull/17591#issuecomment-2442922629 so we don't have this build issue in the future, but this is fine (we'll just need to stay on top of Tomcat version bumps).

PR #17851, opened 2024-11-06
Updates for the release of WebSphere and Open Liberty 24.0.0.11

gkertasef:


tianon:
I just noticed that your GitHub handle in these files is probably not right :sweat_smile: (the @ gseferi should probably be @gkertasef, right?)

tianon (approved):

PR #17889, opened 2024-11-12
Update ELK images for 8.15.4

navyau09:
Update ELK images for 8.15.4


yosifkit (dismissed):
To avoid a possible merge conflict and reduce the number of image update PRs, multiple version updates can be combined when they happen at the same time (like https://github.com/docker-library/official-images/pull/17888)

LGTM

tianon (approved):

PR #17892, opened 2024-11-12
Update to BuildKit 0.16

tianon:
- https://github.com/moby/buildkit/releases/tag/v0.14.0
  - https://github.com/moby/buildkit/releases/tag/dockerfile%2F1.8.0
- https://github.com/moby/buildkit/releases/tag/v0.15.0
  - https://github.com/moby/buildkit/releases/tag/dockerfile%2F1.9.0
- https://github.com/moby/buildkit/releases/tag/v0.16.0
  - https://github.com/moby/buildkit/releases/tag/dockerfile%2F1.10.0

I have been using this version for my own personal (https://hub.docker.com/u/tianon) image builds for a while now with a build system that's basically exactly the same as DOI's, and haven't noticed any major issues. 🤞

(That being said, the delta is a *lot*, so we should still be cautious IMO.  At the *very* least, we should wait for the current Debian rebuild queue to quiet down before considering this.)


tianon:
(That diff would be a lot smaller if we updated to the rebuild of 0.13 for the Debian base image updates first :sweat_smile:  Not sure it's worth it though)

yosifkit (approved):
Yolo, `0.16` seems fine.

PR #17917, opened 2024-11-16
Update tianon/buildkit:0.13

tianon:
This is the Debian rebuild (but see also https://github.com/docker-library/official-images/pull/17892).


tianon:
(Oh, and also https://github.com/tianon/dockerfiles/pull/846 because I'm tired)

yosifkit:
Just went with `0.16` instead (https://github.com/docker-library/official-images/pull/17892).

PR #17926, opened 2024-11-18
Update images after the recent releases

Felixoid:
Here are some changes besides the update:

- `focal-lts` is renamed to `lts-focal`, like other places.

And I have a question regarding update frequency. What would be `too many updates` tempo?


Felixoid:
Thank you, it addresses the question

PR #17960, opened 2024-11-22
[ros] Add ROSDISTRO_PKGS_SYNC_DATE env var

mikaelarguedas:
This allows user to know on what official package sync the image is based on. This allow them to install any other package of the ecosystem  at the same sync date (from http://snapshots.ros.org/noetic/).

This also permits to bust the docker cache when new version of packages are available which has been a long lasting issue https://github.com/osrf/docker_images/issues/112

### DIFF
--- a/library/ros
+++ b/library/ros
@@ -9,7 +9,7 @@ GitRepo: https://github.com/osrf/docker_images.git
 
 Tags: noetic-ros-core, noetic-ros-core-focal
 Architectures: amd64, arm32v7, arm64v8
-GitCommit: 27cc0b68263bbbb10bb58dd814efc0a6b0a01ec7
+GitCommit: cf1c3c02d0657c05b8b19ac98d58895ff79c989c
 Directory: ros/noetic/ubuntu/focal/ros-core
 
 Tags: noetic-ros-base, noetic-ros-base-focal, noetic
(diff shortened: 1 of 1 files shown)


tianon:
https://github.com/osrf/docker_images/issues/112#issuecomment-361326462 and all the following discussion is definitely still accurate :see_no_evil:

ruffsl:
Late to the party, holidays...

> instead of just a simple timestamp so that the cache busts are contextually meaningful instead of simply time-based.

I'd have to argue that this isn't just a time stamp, it is the atomic release that distro maintainers choose to cut that includes a sync all core packages - of differing individual semantic versions - that have been tested and validated to work with each other in aggregate as a snapshot. End user have been requesting such an contextual ENV that would denote the sync version for quite a while. It may be more apt to compare it to something like Ubuntu 24.04.1 vs 24.04.2 etc, only that the version is expressed as the date it is released, rather than a semantic decimals to give better context and grounding for the version number.

tianon:
Sure, that's fair, but the value of the variable is unused during build.  When we perform an automated rebuild of the image in a month or two, it might pull in newer packages than those that were tested, verified, and validated, and now that "semantic version" is no longer accurate.  If the value of the variable were _used_ in some way to actually pull in the expected set/versions of packages (a metapackage, a repository mirror/snapshot, etc), it would be less objectionable because it would actually function in the way you've described.

mikaelarguedas:
Reverting that change for now as this has been preventing docker image updates for a while.

An alternative to both provide the sync date to user, having reproducible builds and allowing end user to update/upgrade packages would be to:
- Set the sync date
- Install deb packages from the corresponding snapshot from snapshots.ros.org
- Point the apt source back to packages.ros.org for users to use

Looks a bit twisted meaning that each dockerfile will have to do this apt source gymnastics..

PR #17976, opened 2024-11-26
Update ELK Images for 8.15.5

vincentDeuschle:


tianon (approved):
Update looks fine -- revisiting https://github.com/docker-library/official-images/pull/17158#issuecomment-2223474718 (and the comment link chain) would still be a good idea though. :+1:

PR #17982, opened 2024-11-27
update rocket.chat

sampaiodiego:


sampaiodiego:
need help on how to proceed with the following error https://github.com/docker-library/official-images/actions/runs/12054469860/job/33612565432?pr=17982#step:4:11

is this something we can't do? or I just need to specify an architecture? if we can't do that, I can do what they do on [their Dockerfile](https://github.com/denoland/deno_docker/blob/main/bin.dockerfile#L14).. 

for context, this is the new code that triggered the error:

```
COPY --from=denoland/deno:bin-1.37.1 /deno /usr/local/bin/deno
```

tianon:
Yep, unfortunately that's not something we can allow :sweat_smile:; https://github.com/docker-library/official-images#repeatability

The code in their `Dockerfile` is _also_ not acceptable :see_no_evil:  Instead of relying on `TARGETARCH`, it should be doing userspace architecture detection (`dpkg --print-architecture` / `apk --print-arch`), and ought to include _some_ kind of validation of the downloaded binary (checksum, signature, etc).

---

However, given it's for an otherwise optional plugin, is it really something you want installed by default for _all_ users?  See https://github.com/docker-library/docs/blob/388c9515404a569d8eea75ae9a3c2e6b3b7a3dee/wordpress/README.md#adding-additional-libraries--extensions for an example of how we handle this in images like WordPress (that have a similar problem, where it's unreasonable for us to include _everything_ users might need for every plugin they might install :sweat_smile:).

Ultimately it's up to you, but I want to make sure you're aware it's a tradeoff (it's ~easy for users to add more things to an image they consume, but it's somewhere between hard and impossible for them to remove things due to the immutable nature of container images). :heart:

sampaiodiego:
thx @tianon for the very detailed explanation.. for some reason I thought that rule would apply only to `FROM` command but not for a `COPY` 🤦  but it makes sense..

for this case I'll need to come up with a solution to install Deno, since it is a mandatory dependency for something crucial to Rocket.Chat.. I'll probably open a PR on our repo and ask for help there if that's ok.. tia

sampaiodiego:
closing in favor of #17984 created by the new automation

PR #17986, opened 2024-11-27
Update photon image

gerrit-photon:
Update photon image


michellew-vmware:
Please not publish the PR. We will have a new update soon.

PR #17997, opened 2024-12-02
Add Fluentd v1.18.0

kenhys:
Replace Fluentd v1.17.x with new stable v1.18.0 release.

ref. https://github.com/fluent/fluentd/security/policy

### DIFF
--- a/library/fluentd
+++ b/library/fluentd
@@ -17,13 +17,13 @@ GitCommit: ddb5019b433339f9eef991ca373445deb6e4d2b1
 Directory: v1.16/debian
 
 # Alpine images
-Tags: v1.17.1-1.0, v1.17-1, latest
+Tags: v1.18.0-1.0, v1.18-1, latest
 Architectures: amd64, arm32v6, arm64v8, i386, ppc64le, s390x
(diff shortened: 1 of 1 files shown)


yosifkit (approved):

yosifkit:
```diff
+ # Don't purge $GEM_DIR/gems/GEM/lib/GEM/ext because it might contain runtime .so (e.g json)
+ && find $GEM_DIR -maxdepth 3 -type d -name test -or -name ext -or -name spec -or -name benchmark | xargs -r rm -rfv \
```
😕🤔 this `find` doesn't look like it matches the comment and might delete more than expected?

kenhys:
This is expected change. ref. https://github.com/fluent/fluentd-docker-image/pull/401

Before: without `-maxdepth 3`

It removes $GEM_DIR/gems/GEM/lib/GEM/ext. (e.g. /usr/local/bundle/gems/json-2.7.4/lib/json/ext)
But it should not do it because it contains runtime .so files.

After: with `-maxdepth 3`

It keeps $GEM_DIR/gems/GEM/lib/GEM/ext. (e.g. /usr/local/bundle/gems/json-2.7.4/lib/json/ext)
It is expected behavior.

yosifkit (approved):

PR #18004, opened 2024-12-02
Update Debian

tianon:
Notably, this finally includes https://github.com/debuerreotype/docker-debian-artifacts/pull/186, and thus uses the `oci-import` builder. :eyes:

(See also https://github.com/docker-library/bashbrew/issues/51, https://github.com/docker-library/bashbrew/pull/61, https://github.com/docker-library/official-images/pull/13769, https://github.com/debuerreotype/debuerreotype/pull/149 :eyes:)

### DIFF
--- a/library/debian
+++ b/library/debian
@@ -4,137 +4,169 @@
 Maintainers: Tianon Gravi <tianon@debian.org> (@tianon),
              Paul Tagliamonte <paultag@debian.org> (@paultag)
 GitRepo: https://github.com/debuerreotype/docker-debian-artifacts.git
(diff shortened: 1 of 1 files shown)


tianon:
FWIW, I made a few minor adjustments after reviewing that diff (https://github.com/debuerreotype/debuerreotype/compare/debuerreotype:5a69d0d...debuerreotype:3fa2fda), but they're only in `index.json` which our build tooling throws away anyhow, so I didn't regenerate for those and plan to just let them be part of the next update instead.

yosifkit (approved):

PR #18008, opened 2024-12-03
Update ELK Images for 7.17.26

vincentDeuschle:


yosifkit (approved):

tianon:
https://github.com/docker-library/official-images/pull/17158#issuecomment-2223474718

> https://github.com/docker-library/official-images/pull/16747#issuecomment-2102952435 👀 (which would help make sure it fails sooner, such that if it fails on our official build infrastructure that might use aggressive build cache that the cache is "busted" in the appropriate place and the rebuild has a chance at success)
> 
> To elaborate, it's failing on `RUN tar -zxf /tmp/elasticsearch.tar.gz --strip-components=1` (in the Elasticsearch case), meaning the `RUN curl` line won't be retried on a rebuild attempt on our official infrastructure, so it will just fail indefinitely if we were to have a similar issue with the download there.

> > Friendly reminder: https://github.com/docker-library/official-images/pull/16698#issuecomment-2096609598 (`set -e`/`set -ux` and avoiding `arch`; `curl -fS` would be good too, but likely fail in other ways if `curl` were to fail)

> > > This is good for now, but please do revisit https://github.com/elastic/logstash/pull/16063#discussion_r1577000627 (adding at the very least `set -e` to all semi-colon-using `RUN` lines, if not the full `set -eux` I suggest there).

> > > (We should also apply the same "avoiding `arch` in favor of querying the image userspace environment" logic to these lines...)

PR #18009, opened 2024-12-03
Updates for the release of WebSphere and Open Liberty 24.0.0.12

wraschke:


yosifkit:
It looks like there is a bug somewhere in updating because all of the `.sh` scripts lost their execute bit in `git` for `24.0.0.12`:

```diff
old mode 100755
new mode 100644
```

wraschke:
Thank you, @yosifkit . We've fixed the problem and will just make a new PR.

PR #18062, opened 2024-12-10
Add Laurent as a codeowner to more library files

yosifkit:
Let's add @LaurentGoderre to a few more images as a follow up to https://github.com/docker-library/official-images/pull/16369.

Also, sort the list.


tianon (approved):
:+1:

PR #18080, opened 2024-12-12
Fixed reference to pruned crane image

LaurentGoderre:


yosifkit (approved):
LGTM.  What do you think, @tianon?

tianon:
\*shakes fist\* @jonjohnsonjr

tianon (approved):

PR #18092, opened 2024-12-13
[Friendica] update to latest memcached version

ne20002:
Fixes #https://github.com/friendica/docker/issues/269


ne20002:
Is the failed test caused by some connection hickup?

yosifkit (approved):

yosifkit:
Yeah, it was just flaky gpg that we see quite often.

Thank you!

PR #18114, opened 2024-12-16
Update ELK Images

jmlrt:
Update ELK Images

### DIFF
--- a/library/elasticsearch
+++ b/library/elasticsearch
@@ -8,12 +8,12 @@ Architectures: amd64, arm64v8
 GitFetch: refs/heads/8.16
 GitCommit: 905b04aed7c279576c093de38e3dfa9d11c45d04
 
-Tags: 8.15.5
-Architectures: amd64, arm64v8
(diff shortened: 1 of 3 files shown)


tianon:
Will 8.15 or 8.16 have any further patch updates? 

If not, they should be removed; removing tags here will remove them from the "Supported" section on the Hub readme (and will prevent us from spending cycles rebuilding them on the official build servers), but the tags will still be available to users who want them.  (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)

jmlrt:
> Will 8.15 or 8.16 have any further patch updates?

Hey @tianon, 8.16 will still get new patch but not 8.15 as we are transitioning to a new release policy for supporting the 2 last minors of the current major version (in addition to the last minor of the previous major). I'll update the PR manually, then, and we'll work on implementing this in our automation 👍🏻

tianon (approved):
Thanks! 👍

(`# For wolfi we pick it from the blessed wolfi package registry.` 👀)

PR #18124, opened 2024-12-17
Add amazonlinux releases

sktomer:
Updated Packages for Amazon Linux 2023:
- python3-libs-3.9.20-1.amzn2023.0.2
- amazon-linux-repo-cdn-2023.6.20241212-0.amzn2023
- system-release-2023.6.20241212-0.amzn2023
- python3-3.9.20-1.amzn2023.0.2

(update other releases to not rely on EOL alpine:3.17)

### DIFF
--- a/library/amazonlinux
+++ b/library/amazonlinux
@@ -12,21 +12,21 @@ Maintainers: Amazon Linux <amazon-linux@amazon.com> (@amazonlinux),
 GitRepo: https://github.com/amazonlinux/container-images.git
 GitCommit: cc7a1876866f4056fa73a789a5b758358151c189
 
-Tags: 2023, latest, 2023.6.20241121.0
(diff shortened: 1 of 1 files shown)


LaurentGoderre:
I'm not sure that using the `latest` for alpine is a good idea. It would be better to pin it to 3.21 instead.

sktomer:
I can make that change if you insist. We discussed that it would be low risk and the pre-checks would inform us if anything is broken.

tianon:
Do you mean you have proactive CI that would notice when `:latest` changes and notify you of any breakage?  (as long as you've got that, I think `:latest` is probably mostly OK here)

sktomer:
yes please merge this. also please close this as a duplicate: https://github.com/docker-library/official-images/pull/18139 (the author, my colleague is out of office today).

tianon (approved):

PR #18140, opened 2024-12-19
Updated Unit to 1.34.0

thresheek:
This also bumps language variants to:
 - php 8.4
 - python 3.13

Note that this new release also incorporates Rust-based code into the main binary, which means build times are now longer and the resulting images have a bigger size.

### DIFF
--- a/library/unit
+++ b/library/unit
@@ -1,123 +1,123 @@
-# this file is generated via https://github.com/nginx/unit/blob/624debcf17ea7faab01fa841bd4dcd9f308cf306/pkg/docker/Makefile
+# this file is generated via https://github.com/nginx/unit/blob/d8acad350a52a20918c46c09cb0a0f5479400923/pkg/docker/Makefile
 
 Maintainers: Unit Docker Maintainers <docker-maint@nginx.com> (@nginx)
 GitRepo: https://github.com/nginx/unit.git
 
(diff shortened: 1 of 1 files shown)


LaurentGoderre:
I opened a PR to optimize this and ensure rust is not left behind in the image: https://github.com/nginx/unit/pull/1522

tianon:
I don't love the duplication (between variants or between this and Rust), but I think it's probably OK.  I do wonder how many of these are the same compilation arguments, and thus could possibly do something like `COPY --from=unit:X.Y.Z-other-variant` instead of building again.  I left some comments on https://github.com/nginx/unit/pull/1522 also.

tianon:
(Not leaving behind Rust in the final images _is_ probably important though :eyes:)

LaurentGoderre:
@tianon I tried the copy approach but each variant seems to be compiled with different libraries depending on the variant.

thresheek:
First of all, thanks for the suggested improvements!  I did not expect you to go as far as submitting a PR upstream :-)

I'm pretty sure unitd can be copied between the variants, since they are all supposed to be the same binaries.  The modules for different languages are supposed to be the only difference between the images.

LaurentGoderre:
Ah! O will amend my PR

yosifkit (approved):

PR #18143, opened 2024-12-20
Add Windows Server 2025 to allowed external pins

tianon:
Depends on:
- https://github.com/docker-library/meta-scripts/pull/102
- https://github.com/docker-library/bashbrew/pull/112

See also:
- https://github.com/docker-library/bashbrew/pull/111
- https://github.com/actions/runner-images/issues/11228
  - https://github.com/actions/runner-images/issues/10806#issuecomment-2555585078


tianon:
Arguably, we should also be removing ltsc2019 / 1809 :thinking:

https://learn.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/base-image-lifecycle (end of mainstream support was "01/09/2024")

tianon:
> Arguably, we should also be removing ltsc2019 / 1809 🤔
> 
> https://learn.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/base-image-lifecycle (end of mainstream support was "01/09/2024")

The "extended support" _is_ legit though, so I think we should maybe pick a project like GHA to mirror here and keep support for 2019 roughly as long as they do (they dropped 2016 - https://github.com/actions/runner-images/pull/5994, which is also still in extended support mode).

tianon:
Rebased, updated pins to latest, updated to bashbrew 0.1.13 (for https://github.com/docker-library/bashbrew/pull/112)

tianon:
> The "extended support" _is_ legit though, so I think we should maybe pick a project like GHA to mirror here and keep support for 2019 roughly as long as they do (they dropped 2016 - [actions/runner-images#5994](https://github.com/actions/runner-images/pull/5994), which is also still in extended support mode).

Oh, well there it is: https://github.blog/changelog/2025-04-15-upcoming-breaking-changes-and-releases-for-github-actions/#windows-server-2019-is-closing-down

> We’re beginning the process of closing down the Windows server 2019 hosted runner image, following our [N-1 OS support policy](https://github.com/actions/runner-images?tab=readme-ov-file#software-and-image-support). This image will be fully retired by June 30, 2025. We recommend updating workflows use `windows-2022` or `windows-2025`.

(FYI @gdams)

PR #18170, opened 2024-12-28
fix 1.42.4 update script, add 1.43.0, remove EOL 1.41

christian-heusel:
Link to our PR https://github.com/wikimedia/mediawiki-docker/pull/152

### DIFF
--- a/library/mediawiki
+++ b/library/mediawiki
@@ -3,38 +3,38 @@ Maintainers: Kunal Mehta <legoktm@debian.org> (@legoktm),
              Christian Heusel <christian@heusel.eu> (@christian-heusel)
 GitRepo: https://github.com/wikimedia/mediawiki-docker.git
 GitFetch: refs/heads/main
-GitCommit: 597ec9bc8565eefd398bc6831130fff2443b6c8a
+GitCommit: 1be3b2bf0efd0f09e4cf2e706b2ecd92f3680fab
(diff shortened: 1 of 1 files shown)


tianon:
Apologies for the delay (holiday PTO :see_no_evil:).

```diff
+	gpg --fetch-keys "https://www.mediawiki.org/keys/keys.txt"; \
```

Unfortunately, this is not acceptable -- the specific keys need to be verified in some way (checksum or full fingerprint).  See https://github.com/docker-library/tomcat/pull/327 for an approach that might be reasonable in your case (essentially: download the keys, import them, export specific keys via full fingerprint, wipe, import exported keys, verify).

Another reasonable alternative (depending on what problem this change is trying to solve) would be to swap to the https://keys.openpgp.org/ keyserver, which is very reliable, but has the added requirement that key holders verify their email addresses (for GDPR compliance).

christian-heusel:
@tianon thanks for your feedback! The issue was that the key was not available on the regular servers. I have now uploaded it to `keyserver.ubuntu.com` and reverted to the previous explicit key list.

yosifkit (approved):

bheiskell:
Hey folks, I don't know if this expected, but I went to pull the new 1.43 image and the tags look empty:

image

Compared to the 1.42 images which also appear to have been updated:

image

PR #18180, opened 2025-01-02

adivinho:


tianon:
```diff
+    curl -O https://dl.fedoraproject.org/pub/epel/RPM-GPG-KEY-EPEL-9; \
+    rpm --import RPM-GPG-KEY-EPEL-9; \
```

This key needs to be verified/exported via full fingerprint or checksum in some way like the keys above (in the same `Dockerfile`).

adivinho:
> This key needs to be verified/exported via full fingerprint or checksum in some way like the keys above (in the same `Dockerfile`).

Verification has been added.

tianon (approved):

PR #18183, opened 2025-01-04
erlang/otp 27.2, 26.2.5.6, 25.3.2.16

getong:


LaurentGoderre (dismissed):

yosifkit:
Given that Alpine 3.21 has GCC 14 by default, these builds will run into https://github.com/erlang/otp/issues/9211, won't they?

getong:
It looks this bug still not fix, https://github.com/erlang/otp/pull/9234
I roll back to alpine 3.20

getong:
done, all change to use alpine 3.20

yosifkit (approved):

PR #18213, opened 2025-01-09
Aerospike 8.0.0.0-rc2

pvinh-spike:

### DIFF
--- a/library/aerospike
+++ b/library/aerospike
@@ -3,14 +3,14 @@ Maintainers: Lucien Volmar <lucien@aerospike.com> (@volmarl),
              Phuc Vinh <pvinh@aerospike.com> (@pvinh-spike),
              Kevin Porter <kporter@aerospike.com> (@kportertx)
 
-Tags: ee-7.2.0.6, ee-7.2.0.6_1
+Tags: ee-8.0.0.0-rc2, ee-8.0.0.0-rc2_1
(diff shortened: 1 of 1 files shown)


library/aerospike
```diff
@@ -3,14 +3,14 @@ Maintainers: Lucien Volmar <lucien@aerospike.com> (@volmarl),
              Phuc Vinh <pvinh@aerospike.com> (@pvinh-spike),
              Kevin Porter <kporter@aerospike.com> (@kportertx)
 
-Tags: ee-7.2.0.6, ee-7.2.0.6_1
+Tags: ee-8.0.0.0-rc2, ee-8.0.0.0-rc2_1
```
yosifkit, line 6: Given that `8.0.*` is a Developer Preview Release, did you mean to remove the `7.2` images? Or did you want to keep both until `8.0` is no longer a preview build?

I am not sure if `7.2.x` would have a new version in parallel to possibly other `8.0.x` pre-releases, but it might confuse users if the [Docker Hub page](https://hub.docker.com/_/aerospike) no longer has `7.2` in the "Supported tags" list and only lists the `8.0` preview.

pvinh-spike: Thanks for the input.  I'm closing the pr and wait for the GA.

pvinh-spike:
We'll wait for the 8.0 GA.  Thanks

PR #18229, opened 2025-01-14
Update CrateDB to version 5.9.6

romseygeek:


tianon (approved):

mfussenegger:
It looks like the `linux/arm64/v8` tag is missing (https://hub.docker.com/_/crate/tags) despite the library file still listing `Architectures: amd64, arm64v8`. Do we need to change anything about that, or is this some kind of hiccup?

tianon:
Sorry, we had a hiccup in our system -- it was corrected yesterday and builds should be (long since) up now. :+1:

PR #18246, opened 2025-01-15
Revert "Add amazonlinux releases"

sktomer:
This reverts commit 20b02e5599a8e5b14d3c6c55cae1c2f09c12ae9c.

The new releases are configured with repositories referencing an rsync security fix.

However the patch has a regression reported upstream and hence as a precaution, we want to rollback to previous known good images.


sktomer:
Will open a new PR to restart tests.

PR #18247, opened 2025-01-15
Rollback amazonlinux releases. This reverts commit 20b02e5.

sktomer:
This reverts commit 20b02e5.

The new releases are configured with repositories pulling in an rsync security fix. However the patch has a regression reported upstream and hence as a precaution, we want to rollback to previous known good images.


yosifkit (approved):

sktomer:
@yosifkit maybe you need to click merge button as well ? :-)

PR #18251, opened 2025-01-16
Update ruby

tianon:
Changes:

- https://github.com/docker-library/ruby/commit/59bb4e0: Update 3.3 to 3.3.7


tianon:
As a reminder, and as seen in the diff above, this includes https://github.com/docker-library/ruby/pull/497#issuecomment-2590877455 (less packages in `slim` and `alpine` image variants).

LaurentGoderre (approved):

PR #18320, opened 2025-01-26
Apache Storm 2.8.0

rzo1:
Updates to Apache Storm 2.8.0 with two notable changes

- (1) We removed the version based directory in the image to allow for a better UX for end users
- (2) Storm 2.8.0 only supports Java 17+ now, so the base image will also be Java 17 based now.

### DIFF
--- a/library/storm
+++ b/library/storm
@@ -5,9 +5,9 @@ GitRepo: https://github.com/apache/storm-docker.git
 Architectures: amd64, arm64v8
 
 Tags: 2.8.0, 2.8, latest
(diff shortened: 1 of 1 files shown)


yosifkit:
Changes seem fine, but it seems to be missing the version bump on [line 36](https://github.com/apache/storm-docker/blob/eba09e577f8a43ac96496272f035c0ce662a2bac/2.8.0/Dockerfile#L36)

rzo1:
Really good catch! Will fix that

yosifkit (approved):

PR #18327, opened 2025-01-28
Upgrade stable version to 17.0.0

manuelleduc:


library/xwiki
```diff
@@ -2,36 +2,36 @@ Maintainers: XWiki Core Dev Team <committers@xwiki.org> (@xwiki)
...
 Architectures: amd64, arm64v8
-Directory: 16/mariadb-tomcat
-GitCommit: beed20b75ae4c952be757cec1502151d2d2a7939
+Directory: 17/mariadb-tomcat
+GitCommit: 381926879a24fb09d3b7611b883c78eb4576d3a2
 
-# old LTS
-Tags: 15, 15.10, 15.10.16, 15-mysql-tomcat, 15.10-mysql-tomcat, 15.10.16-mysql-tomcat
+# LTS
+Tags: 16, 16.10, 16.10.2, 16-mysql-tomcat, 16.10-mysql-tomcat, 16.10.2-mysql-tomcat, mysql-tomcat, lts-mysql-tomcat, lts-mysql, lts
```
yosifkit, line 21: Duplicate tag here with `mysql-tomcat` that is also up in `17*` (line 5), so probably can be dropped from here.  This'll also need a rebase since I just merged https://github.com/docker-library/official-images/pull/18325.

manuelleduc: Addressed in e07eb642015f70bc8d7160cb117b098ec0bb1870

yosifkit (approved):

Issue #17896, opened 2024-11-13
Platform now added in manifest


pbo-linaro:
From an x64 machine,

```
$ docker run -it --rm --pull always docker.io/arm64v8/debian
Error: choosing an image from manifest list docker://arm64v8/debian:latest: no image found in image index for architecture amd64, variant "", OS linux
```

Same when trying to run an amd64 image from an arm64 host.
This used to work until very recently (~ weeks), and I suspect the image is now built with an explicit `--platform` parameter on your side.

Even though it's possible to explicitly add a --platform on our side, it used to be very convenient to not have to do it for cross compilation scenarios and multi stage builds involving multiple architectures. It's definitely a niche usage though.

Was this changed recently?
What was the rationale?

yosifkit:
Yes, this happened with https://github.com/docker-library/meta-scripts/pull/93. We finally have switched all our builds over to `buildkit`, so they now include attached provenance (the item with `unknown` os and architecture in the image index).  (Similar issue in https://github.com/docker-library/python/issues/933)

```json
$ docker run -i --rm gcr.io/go-containerregistry/crane manifest arm64v8/debian | jq .
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.index.v1+json",
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:e2554fa0cd89eec5865aef4bf094c8d646059d37c42b44a6655839ea46ef3f90",
      "size": 1041,
      "platform": {
        "os": "linux",
        "architecture": "arm64",
        "variant": "v8"
      },
      "annotations": {
        "com.docker.official-images.bashbrew.arch": "arm64v8",
        "org.opencontainers.image.base.name": "scratch",
        "org.opencontainers.image.created": "2024-11-12T00:55:54Z",
        "org.opencontainers.image.revision": "dde41269a09feb7d5b046133bdd54c918ecbf1ca",
        "org.opencontainers.image.source": "https://github.com/debuerreotype/docker-debian-artifacts.git#dde41269a09feb7d5b046133bdd54c918ecbf1ca:bookworm",
        "org.opencontainers.image.url": "https://hub.docker.com/_/debian",
        "org.opencontainers.image.version": "bookworm"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
[...]

PR #18375, opened 2025-02-04
REL-3521 Release SonarQube Community Build 25.2

davividal:
Hello team!

We just released a new SonarQube Community Build: 25.2!

Please let me know if I need to adjust anything here. :)

### DIFF
--- a/library/sonarqube
+++ b/library/sonarqube
@@ -4,6 +4,7 @@ Maintainers: Carmine Vassallo <carmine.vassallo@sonarsource.com> (@carminevassal
 GitRepo: https://github.com/SonarSource/docker-sonarqube.git
 Architectures: amd64, arm64v8
 GitCommit: 1edbd4a3f756a8b6eba3f039be62fd9c4862e9b9
+Builder: buildkit
 
 Tags: 2025.1.0-developer, 2025.1-developer, 2025-lta-developer, developer
 Architectures: amd64, arm64v8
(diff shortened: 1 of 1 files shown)


tianon:
```diff
+COPY --chown=sonarqube:root --chmod=550 run.sh sonar.sh ${SONARQUBE_HOME}/docker/
```

This is weird -- isn't `sonarqube` the user the container runs as?  Why should these scripts be owned by that user?  (That makes them potentially something that could be modified at runtime, which will then be run if the container/host ever restarts.)

jCOTINEAU:
> ```diff
> +COPY --chown=sonarqube:root --chmod=550 run.sh sonar.sh ${SONARQUBE_HOME}/docker/
> ```
> 
> This is weird -- isn't `sonarqube` the user the container runs as? Why should these scripts be owned by that user? (That makes them potentially something that could be modified at runtime, which will then be run if the container/host ever restarts.)

Hello @tianon, thanks for the kind feedback.

We reworked a bit the permission system so it work for both openshift (any uid and gid=0) as well as other user's use case (uid=1000 and gid!=0) and applied this to everything.

Nonetheless here with what you describe are'nt we safe with `chmod=550` as neither SonarQube user as owner or part of root group will be able to write to this file ?

What would be your best practice over there ? 

Thanks again for the nice discussions

davividal:
Hey @tianon. Thanks again for the feedback.

I assume that the images can rely on Buildkit [since Nov 2024](https://github.com/docker-library/official-images/issues/17896#issuecomment-2477155783), isn't that correct?

If Buildkit is supported, what am I missing here?

yosifkit (approved):

PR #18383, opened 2025-02-05
Releasing neo4j 5.26.2

neo-tobias:


jennyowen:
We just so happened to have a new version branch of neo4j release today. I hope you don't mind it being in the same PR as this?

yosifkit (approved):

yosifkit:
> We just so happened to have a new version branch of neo4j release today. I hope you don't mind it being in the same PR as this?

That's perfect. If they are drastically different, then it can increase our review time, but when the new version is mostly the same as previous versions, one PR is recommended and makes less churn for all of us (and removes the possibility of merge conflicts on the `library/neo4j` file).

PR #18386, opened 2025-02-05
Update CrateDB 5.9.9

BaurzhanSakhariev:
Point to a commit with updated base image

### DIFF
--- a/library/crate
+++ b/library/crate
@@ -6,7 +6,7 @@ GitRepo: https://github.com/crate/docker-crate.git
 
 Tags: 5.9.9, 5.9, latest
 Architectures: amd64, arm64v8
-GitCommit: 4311253b84f943e5662ad014ce264d0ffe3b96da
+GitCommit: c83963c3aefaf30e0a17a3162e8692b723d48781
 
 Tags: 5.8.6, 5.8
 Architectures: amd64, arm64v8


mfussenegger:
> It would be preferable to update the OS on a version change (on 5.9.10) for example

Could do - but just to give some context: CrateDB 5.9.8 included a OpenDAL update to fix a bug, but we missed that the OpenDAL update came with a higher glibc version requirement affecing the ARM build. Because of that the 5.9.8 and 5.9.9 docker images are broken for ARM while the tarball is fine as long as they're running on a system with a recent enough glibc.

yosifkit (approved):

yosifkit:
> > It would be preferable to update the OS on a version change (on 5.9.10) for example
> 
> Could do - but just to give some context: CrateDB 5.9.8 included a OpenDAL update to fix a bug, but we missed that the OpenDAL update came with a higher glibc version requirement affecing the ARM build. Because of that the 5.9.8 and 5.9.9 docker images are broken for ARM while the tarball is fine as long as they're running on a system with a recent enough glibc.

Yeah, it isn't necessarily great to change OS versions outside of a bump or tag change, but it is sometimes necessary. Considering that it is a fix and that users are less likely to be building their "application" inside a database container, this is fine.

LaurentGoderre:
@mfussenegger thank you for the context and explanation! Makes sense

PR #18387, opened 2025-02-05
Update clickhouse to the latest state

robot-ch-test-poll4:
This is an automatic PR to update `clickhouse` according to the latest releases.

Maintainers: Misha f. Shiryaev <felixoid@clickhouse.com> (@Felixoid),
             Max Kainov <max.kainov@clickhouse.com> (@mkaynov),
             Alexander Sapin <alesapin@clickhouse.com> (@alesapin)


Felixoid:
The question regarding the robot. Should we use a single particular robot to make PRs?

Usually, we have a fleet of robots and choose the one with the most remaining requests. But we can use @robot-clickhouse

In this case, should it be added to the maintainers in the LDF or not?

yosifkit:
> The question regarding the robot. Should we use a single particular robot to make PRs?
>
> Usually, we have a fleet of robots and choose the one with the most remaining requests. But we can use @robot-clickhouse

Any "clickhouse" robot user should be fine; ones without a merged PR here will just have to wait for an "approve and run" from us for GitHub actions.

> In this case, should it be added to the maintainers in the LDF or not?

Just regular users there is recommended. The maintainers list is mostly if we need to email you directly for security/sensitive issues and for PR notification. For example, when your bot opens a PR, you get notified by the callouts in the diff message so you know changes are being made for `clickhouse` and can tell us if they should or should not be merged yet. We usually assume that a bot PR is good unless a maintainer comments, or we see something off in the diff (which you then would get our comments asking for clarification or changes).

yosifkit:
From the test failures:
> 	'override-cmd' [3/3]...failed

This is testing the ["Consistency"](https://github.com/docker-library/official-images/blob/429dd22ac3022ce45cbd0d22116d93957a2af0ca/README.md#consistency) requirement of Docker Official Images. The short description being that `docker run -it clickhouse:version [binary]` works for any binary in the image without extra output printed or any additional work from the user (i.e., not having to use `--entrypoint`).

Felixoid:
Hmm, I am not sure what it doesn't like..

```
> docker run -it clickhouse:25.1.3.23.ubuntu bash
/entrypoint.sh: neither CLICKHOUSE_USER nor CLICKHOUSE_PASSWORD is set, disabling network access for user 'default'
root@8512ebe6adc7:/var/lib/clickhouse#
```

Is the log line from the `/entrypoint.sh` an issue here?

Felixoid:
I've took a look at postgres entrypoint, it looks like executing functions is the best approach here

Felixoid:
Well, I'd rather to go into [this](https://github.com/ClickHouse/ClickHouse/pull/75643/files#diff-65b30e60f953b0f91424fca7c7b9f257fe4cc1e6f69c0cf65bf931015912370c) direction

Felixoid:
Can you, please, review the [entrypoint.sh](https://github.com/ClickHouse/docker-library/blob/improved-entrypoint/server/25.1.3.23/entrypoint.sh)?

It's in progress of the review in our PR as well, so now I need to sync everything https://github.com/ClickHouse/ClickHouse/pull/75643

tianon:
Making functions for everything to remove the log line seems like overkill, but if that's the direction you want, it's unobjectionable (and as you noted, there's prior art).

tianon:
Logic in the script looks OK too :+1:

tianon:
> The question regarding the robot. Should we use a single particular robot to make PRs?

The only other consideration I'd add over what's already been stated is making sure there's only one bot PR open at a time, but if several bots update that PR, that's totally fine too (and looks like it's already handled via your branch naming choice :heart:). :+1:

Felixoid:
The PR is updated from https://github.com/ClickHouse/ClickHouse/actions/runs/13435831977/job/37537768853, for the reverence.

It has updated the existing branch, as expected.

tianon (approved):

PR #18403, opened 2025-02-07
Update python-hy test to Hy 1.0.0

tianon:


tianon:
Quoting https://github.com/docker-library/pypy/pull/86#issuecomment-2641540524 here, because if @yosifkit agrees with me, we should close this PR instead:

> I'm adding an asterisk there because the more I look at it, the more I no longer think this test actually makes sense for us to keep.  Originally it was helpful as a deeper test of Python, but I don't think it's caught us any Python bugs, and just puts us at the forefront of Hy integration problems like this one.
> 
> "But Tianon," you might say, "isn't that a good thing?"  Yes!  Yes, it is.  HOWEVER, we literally have the https://hub.docker.com/_/hylang image which would run into these _same_ integration issues, but in a more appropriate place (such that they're only affecting updates to the Hy image, not PyPy or Python).
> 
> So in short, I think we should actually perhaps remove that test instead, as much as I love it and find it extremely cute, since we'll keep it anyways on all the variants of the Hy image (but without that hairy `if` statement for filtering supported versions, because it'll only even _run_ for supported versions, pre-filtered).

yosifkit:
I agree; it makes sense to not block `python` and `pypy` on `hy` integration issues.

PR #18408, opened 2025-02-07
New Docker image for Silverpeas 6.3.6

mmoqui:
Update silverpeas Docker library to take into account Silverpeas 6.3.6


mmoqui:
This PR is an error (triggered by the release of version 6.3.6  that is also an error. This version is equivalent to 6.3.5 and it doesn't carry nothing more)

PR #18418, opened 2025-02-10
Remove `python-hy` test

tianon:
Originally this was helpful as a deeper test of Python, but I don't think it's caught us any Python bugs, and just puts us at the forefront of Hy integration problems.

"But Tianon," you might say, "isn't that a good thing?" Yes! Yes, it is. HOWEVER, we literally have the https://hub.docker.com/_/hylang image which would run into these same integration issues, but in a more appropriate place (such that they're only affecting updates to the Hy image, not PyPy or Python).

So in short, we should remove that test instead, as much as I love it and find it extremely cute, since we'll keep it anyways on all the variants of the Hy image (but without that hairy if statement for filtering supported versions, because it'll only even run for supported versions, pre-filtered).

See also:
- https://github.com/docker-library/pypy/pull/86#issuecomment-2641540524
- https://github.com/docker-library/official-images/pull/18403


tianon:
Success: https://github.com/docker-library/official-images/actions/runs/13247211569?pr=18418

yosifkit (approved):

PR #18428, opened 2025-02-11
Update ELK images for 8.17.2

StamatisKourk:


yosifkit:
Unless there is a larger change for us to review in one version and not the other (this and #18427), please use one PR for simple version bumps to help minimize notifications and the possibility of merge conflicts.

yosifkit (approved):

PR #18435, opened 2025-02-11
Update golang

tianon:
Changes:

- https://github.com/docker-library/golang/commit/aa7af9a: Merge pull request https://github.com/docker-library/golang/pull/549 from infosiftr/go-1.24
- https://github.com/docker-library/golang/commit/04edff0: Bump to go 1.24.0 release

PR #18438, opened 2025-02-12
Update python

tianon:
Changes:

- https://github.com/docker-library/python/commit/5d789378: Update 3.14-rc to 3.14.0a5


yosifkit (approved):

yosifkit:

PR #18442, opened 2025-02-12
Update docker

yosifkit:
Changes:

- https://github.com/docker-library/docker/commit/e5fa4d1: Update 28-rc to compose 2.33.0
- https://github.com/docker-library/docker/commit/2738fa1: Update 27 to compose 2.33.0


yosifkit:

LaurentGoderre (approved):

PR #18443, opened 2025-02-12
Update Windows pins

tianon:


yosifkit:

So, we need to not add the 2025 updates until the GitHub actions runner images are ready. Can you split this PR to exclude the 2025 updates?

tianon:
Yeah... :sob: 

https://github.com/docker-library/official-images/pull/18447

(this is not a great situation to be in, but nothing we can do about it right now :disappointed:)

tianon:
https://github.com/yosifkit/fuzzy-octo-engine/actions/runs/13571858257 looks like we're good to go? :eyes:

yosifkit:
> https://github.com/yosifkit/fuzzy-octo-engine/actions/runs/13571858257 looks like we're good to go? 👀

Oh, yeah, I think the windows-2025 runner migration is close enough to fully migrated. There are 0-2 failures out of 60 runs across each of the actions pipelines in the last 6 hours and that's not much worse that what we usually get with different external failures across our builds.

yosifkit (approved):

PR #18445, opened 2025-02-12
Update python

yosifkit:
Changes:

- https://github.com/docker-library/python/commit/4ce3b2ae: Update 3.14-rc

https://www.python.org/downloads/release/python-3140a5/:
> #### Note
> This release initially contained source tarballs that didn't match the final v3.14.0a5 Git tag. Around 2025-02-12 0815 UTC, the source tarballs were replaced with the correct versions. This happened because this release took three attempts due to a bug that failed the Windows PGO builds, and the release manager accidentally uploaded the source tarballs from the second attempt.

### DIFF
--- a/library/python
+++ b/library/python
@@ -8,32 +8,32 @@ Builder: buildkit
 Tags: 3.14.0a5-bookworm, 3.14-rc-bookworm
 SharedTags: 3.14.0a5, 3.14-rc
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, ppc64le, s390x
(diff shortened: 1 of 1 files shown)


LaurentGoderre (approved):

yosifkit:
Windows ltsc2025 failures are currently expected (https://github.com/actions/runner-images/issues/11582). It should be fine for our builds since we pin to specific Windows images and can delay moving the 2025 pin until the GitHub runners are ready. This will mean that the Windows 2025 images will not get their "Patch Tuesday" rebuilds until the runner images are updated.

PR #18452, opened 2025-02-13
Update to Apache Flink 1.19.2 and 1.20.1

afedulov:
- Add new release `1.19.2`
- Add new release `1.20.1`
- Remove EOL `1.18` release

### DIFF
--- a/library/flink
+++ b/library/flink
@@ -3,12 +3,12 @@
 Maintainers: The Apache Flink Project <dev@flink.apache.org> (@ApacheFlink)
 GitRepo: https://github.com/apache/flink-docker.git
 
-Tags: 2.0-preview1-scala_2.12-java17, 2.0-scala_2.12-java17, scala_2.12-java17, 2.0-preview1-java17, 2.0-java17, java17
+Tags: 2.0-preview1-scala_2.12-java17, 2.0-scala_2.12-java17, 2.0-preview1-java17, 2.0-java17
 Architectures: amd64,arm64v8
 GitCommit: 3a8ba480ff327674b0a090c6ce417f0437576e92
 Directory: ./2.0/scala_2.12-java17-ubuntu
 
-Tags: 2.0-preview1-scala_2.12-java11, 2.0-scala_2.12-java11, scala_2.12-java11, 2.0-preview1-scala_2.12, 2.0-scala_2.12, scala_2.12, 2.0-preview1-java11, 2.0-java11, java11, 2.0-preview1, 2.0, latest
+Tags: 2.0-preview1-scala_2.12-java11, 2.0-scala_2.12-java11, 2.0-preview1-scala_2.12, 2.0-scala_2.12, 2.0-preview1-java11, 2.0-java11, 2.0-preview1, 2.0
 Architectures: amd64,arm64v8
 GitCommit: 3a8ba480ff327674b0a090c6ce417f0437576e92
 Directory: ./2.0/scala_2.12-java11-ubuntu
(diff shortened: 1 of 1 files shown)


tianon:
```diff
diff --git a/flink_1.19-java17/docker-entrypoint.sh b/flink_1.19-java17/docker-entrypoint.sh
index e081109..e884375 100755
--- a/flink_1.19-java17/docker-entrypoint.sh
+++ b/flink_1.19-java17/docker-entrypoint.sh
@@ -65,19 +65,19 @@ set_config_options() {
     local bin_dir="$FLINK_HOME/bin"
     local lib_dir="$FLINK_HOME/lib"
 
-    local config_params=()
+    local config_params=""
 
     while [ $# -gt 0 ]; do
         local key="$1"
         local value="$2"
 
-        config_params+=("-D${key}=${value}")
+        config_params+=" -D${key}=${value}"
 
         shift 2
     done
 
-    if [ "${#config_params[@]}" -gt 0 ]; then
-        "${config_parser_script}" "${config_dir}" "${bin_dir}" "${lib_dir}" "${config_params[@]}"
+    if [ ! -z "${config_params}" ]; then
+        eval "${config_parser_script} ${config_dir} ${bin_dir} ${lib_dir} ${config_params}"
     fi
 }
 
```

This change seems like it was unintentional (there's no explicit reference to it in either [the commit](https://github.com/apache/flink-docker/commit/d0258b835f42cabbdfb4d6c04320fe79ae9d1f0e) or [the PR](https://github.com/apache/flink-docker/pull/212) that introduced it). :thinking:

Can you elaborate on the intent / problem it's solving?  Some of the potentially unintentional side-effects of it are things in the same realm as SQL injection -- config params containing `$` that now parse variable references where they shouldn't, directories or values with spaces in their paths causing unintentional behavior, etc.

afedulov:
@tianon it seems that whoever did the last 1.19 release (1.19.1), did not update the base `docker-entrypoint.sh` on top of which we run `./add-version.sh -r 1.19 -f 1.19.2` [[link]](https://github.com/apache/flink-docker/blob/dev-1.19/add-version.sh), according to the release workflow [instructions](https://github.com/apache/flink-docker/tree/master?tab=readme-ov-file#release-workflow). 
https://github.com/apache/flink-docker/blob/dev-1.19/docker-entrypoint.sh#L66-L74
https://github.com/apache/flink-docker/blob/dev-1.20/docker-entrypoint.sh#L66-L74

Basically the `dev-1.19/docker-entrypoint.sh` is out of sync from what was actually generated/published for the last 1.19.x release. Seems like the 1.20 branch was used instead.

afedulov:
fixing on our side https://github.com/apache/flink-docker/pull/215

afedulov:
@tianon fixed in flink-docker: https://github.com/apache/flink-docker/commit/e2765a6d67449ccd22bc75050449b199a3a91302
PTAL

yosifkit:
```yaml
Tags: scala_2.12-java11, scala_2.12, java11, latest
Tags: scala_2.12-java17, java17
```
You are certain that you want to move these generic tags from `1.20` releases to the `2.0-preview` images?

afedulov:
@yosifkit thanks for catching this! It is an unfortunate effect of our release scripts not having the notion of a "preview" release (2.0 is the first major release since 9 years). I moved the tags back into a stable version https://github.com/docker-library/official-images/pull/18452/commits/cbf007c3274d95255abadeab2afc2d949a85b5bb, PTAL

afedulov:
@tianon could you please take look if all concerns are addressed now? We are blocked on this step for announcing the Flink releases.

yosifkit (approved):
LGTM.

These are a couple removed tags that you might still want. I didn't want to hold this up further since they can be added back in another PR.
```diff
-flink:java8
-flink:scala_2.12-java8
```

PR #18461, opened 2025-02-14
MariaDB 2025 Q1 Part 2 release

grooverdan:
11.8 as new RC in a LTS  version
11.7 now stable

11.6 now eol.

### DIFF
--- a/library/mariadb
+++ b/library/mariadb
@@ -6,25 +6,25 @@ Maintainers: Daniel Black <daniel@mariadb.org> (@grooverdan),
 GitRepo: https://github.com/MariaDB/mariadb-docker.git
 Builder: buildkit
 
-Tags: 11.7.1-ubi9-rc, 11.7-ubi9-rc, 11.7.1-ubi-rc, 11.7-ubi-rc
+Tags: 11.8.1-ubi9-rc, 11.8-ubi9-rc, 11.8.1-ubi-rc, 11.8-ubi-rc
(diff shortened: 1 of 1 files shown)


yosifkit (approved):

yosifkit:
> 33.80 + microdnf install -y MariaDB-backup-11.7.1 MariaDB-server-11.7.1
> 33.92 error: No package matches 'MariaDB-backup-11.7.1'

And here I was just about to file an issue about `mariadb:11.7.1-ubi-rc` failing to rebuild on an updated UBI base image, so this update is perfectly timed. 🎉

(Or rather `10.7.2` is the cause of the old package being unavailable 😆)

rfay:
I note that the 11.8 images do not have arm64 variants, and there is no 11.8 master tag. I assume all of those will be coming?

yosifkit:
Oh, you mean on Docker Hub? They are all in queue to build, but since we also merged an Alpine image update (#18459), there are quite a few images being rebuilt from those updated base images. They will show up eventually; you can see the current layer of the queue in the trigger jobs: like https://doi-janky.infosiftr.net/job/meta/view/triggers/job/arm64v8/job/trigger/ for `arm64v8`.

grooverdan:
> microdnf install -y MariaDB-backup-11.7.1

looking into it.

> and there is no 11.8 master tag.

11.8 will happen on GA in a few months. Its rc only at the moment.

PR #18482, opened 2025-02-18
Update postgres

tianon:
Changes:

- https://github.com/docker-library/postgres/commit/a537d60: Update 17 to bookworm 17.3-3.pgdg120+1, bullseye 17.3-3.pgdg110+1

### DIFF
--- a/library/postgres
+++ b/library/postgres
@@ -6,12 +6,12 @@ GitRepo: https://github.com/docker-library/postgres.git
 
 Tags: 17.3, 17, latest, 17.3-bookworm, 17-bookworm, bookworm
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
-GitCommit: 22dad776d9f858f5fb1940ac165be76aa8521e49
+GitCommit: a537d6002b1a4bb92eb88e1e894332a76b1d2e6b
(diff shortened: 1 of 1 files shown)


tianon:
I think this is just https://salsa.debian.org/postgresql/postgresql/-/commit/88678696ae7648abffdb5c2f91bc7ccd89f1b77c, but there's likely a 17.4 release that's imminent so it's probably worth waiting at least a little bit to see if that surfaces.

tianon:
Oh, looking at https://git.postgresql.org/gitweb/?p=postgresql.git;a=summary, we're probably getting releases of all 13-17 majors.

tianon:
Ah, from https://git.postgresql.org/gitweb/?p=postgresql.git;a=commitdiff;h=848e2c6abc941267a4b2eb00a57c34ef3165ffb3;hp=3977bd2985de8c5e4b8b00f67dfec28dcfb343c9 it appears to be (in part) to fix the same regression, so it wouldn't be a bad idea to get this update out that backports the fix for 17.3 while we wait for the wheels to turn on the other major updates.

tianon:
Yeah, we won't see 17.4 until Thursday: https://bugs.debian.org/1096243#25

yosifkit (approved):

PR #18514, opened 2025-02-24
Update Debian

tianon:
Notably, this drops `e2fsprogs` (and deps) from Trixie+ :partying_face:

(Thanks for processing it, @paultag :heart:)

### DIFF
--- a/library/debian
+++ b/library/debian
@@ -7,34 +7,34 @@ GitRepo: https://github.com/debuerreotype/docker-debian-artifacts.git
 GitCommit: 7935fc7dd049cb343df42037c152f570069d274f
 # https://github.com/debuerreotype/docker-debian-artifacts/tree/dist-amd64
 amd64-GitFetch: refs/heads/dist-amd64
-amd64-GitCommit: 32138bf51ca52d8a4a8dcf24953d464fbedbf4bb
+amd64-GitCommit: c75e1cbd17082ba26de19b4a979aab7e7b4dccd9
 # https://github.com/debuerreotype/docker-debian-artifacts/tree/dist-arm32v5
 arm32v5-GitFetch: refs/heads/dist-arm32v5
-arm32v5-GitCommit: 94fa58d43704694fc21cb5e03c575075110ca125
+arm32v5-GitCommit: 561f5827807fd32a244f7bc8125fa9198645018d
 # https://github.com/debuerreotype/docker-debian-artifacts/tree/dist-arm32v7
 arm32v7-GitFetch: refs/heads/dist-arm32v7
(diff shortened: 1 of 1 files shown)


yosifkit (approved):

yosifkit:
> is this normal because trixie is not stable yet?

Yes, packages can be removed from the required (minbase) set while it is still `testing` and thus not yet `stable`. `e2fsprogs` not being installed in the image will be the standard for all Debian releases starting from Trixie, so I'd recommend installing it if you need it.
```dockerfile
FROM debian:trixie
RUN set -e; apt-get update; apt-get install -y e2fsprogs; rm -rf /var/lib/apt/lists/*
```

tianon:
Yep, this is a known and expected hazard of using testing before release (even during freeze, things like this are less common, but not impossible). :heart:

paultag:
@kroese if you'll notice there's no `apt install e2fsprogs` in any dockerfiles or bootstrapping script. The docker images reflect the most minimal set of what you can get in Debian, called `minbase`.

FWIW, with this change in particular, I am responsible for the action which did this (processing the override in the Debian archive), but I didn't request it, the maintainers of e2fsprogs requested this. This was not done on anything to do with Docker anywhere.

To state it again clearly -- there's no Docker build toolchain now or ever which was ever responsible for the installation of e2fsprogs in Debian images directly. This was only ever done because of the package priority. I understand operating systems can be complex to understand and from your point of view on the outside it may feel a bit opaque but that is a pretty significant point here.

The maintainers of e2fsprogs in Debian lowered the priority of their package. If you feel like the Debian maintainer(s) of e2fsprogs should continue to ensure that package is shipped on every Debian system in perpetuity even when no ext filesystem is being used to avoid breaking the few dockerfiles out there that needed it and replied on this implicit behavior, feel free to open an issue on the BTS and ask them to raise the priority of the package again. I don't know for sure, but I'm not convinced the e2fsprogs maintainers will be convinced by this. I don't think I am, either.

The related Debian issue requesting the correction of the priority was http://bugs.debian.org/1096180

Hope that helps!

paultag:
@kroese No worries at all! Just trying to help provide some background as to how this all came to be and be sure I don't leave you with a bad expectation of what could happen in the future.

Relatedly, since I didn't say it -- I am very sorry this broke your Dockerfile, I know it's a stressful/annoying thing to randomly see a build fail due to something that didn't change. I hope you didn't spend too much time tracking it down! When we did the override change we ping'd the debian installer/debian boot team to make sure it didn't break the installer, but it's tough to know who else is relying on the implicit package set that's marked "Required" without an explicit apt-get.

Hope you have a great day :)

PR #18551, opened 2025-03-01
Remove unsupported Java versions and add UBI images

keeganwitt:
When I rebased the 6 and 7 branches off master, I missed that we added Java versions that are not officially supported by those older Gradle versions. https://docs.gradle.org/current/userguide/compatibility.html says that Java 16 and newer require Gradle 7 and Java 20 and newer require Gradle 8.

### DIFF
--- a/library/gradle
+++ b/library/gradle
@@ -4,530 +4,344 @@ GitRepo: https://github.com/keeganwitt/docker-gradle.git
 
 # Gradle 8.x
 
-Tags: 8.12.1-jdk21, 8.12-jdk21, 8-jdk21, jdk21, 8.12.1-jdk21-noble, 8.12-jdk21-noble, 8-jdk21-noble, jdk21-noble, latest, 8.12.1-jdk, 8.12-jdk, 8-jdk, jdk, 8.12.1, 8.12, 8, 8.12.1-jdk-noble, 8.12-jdk-noble, 8-jdk-noble, jdk-noble, 8.12.1-noble, 8.12-noble, 8-noble, noble
+Tags: 8.13-jdk21, 8-jdk21, jdk21, 8.13-jdk21-noble, 8-jdk21-noble, jdk21-noble, latest, 8.13-jdk, 8-jdk, jdk, 8.13, 8, 8.13-jdk-noble, 8-jdk-noble, jdk-noble, 8.13-noble, 8-noble, noble
 Architectures: amd64, arm64v8, ppc64le, riscv64, s390x
 GitFetch: refs/heads/master
-GitCommit: 40b84264cea73e19cb7e916446e1bf025f0f89d2
+GitCommit: 774ac009567d0640091b8856705ad6a78b4468ae
 Directory: jdk21-noble
 
-Tags: 8.12.1-jdk21-jammy, 8.12-jdk21-jammy, 8-jdk21-jammy, jdk21-jammy, 8.12.1-jdk-jammy, 8.12-jdk-jammy, 8-jdk-jammy, jdk-jammy, 8.12.1-jammy, 8.12-jammy, 8-jammy, jammy
+Tags: 8.13-jdk21-jammy, 8-jdk21-jammy, jdk21-jammy, 8.13-jdk-jammy, 8-jdk-jammy, jdk-jammy, 8.13-jammy, 8-jammy, jammy
 Architectures: amd64, arm64v8, ppc64le, s390x
 GitFetch: refs/heads/master
-GitCommit: 40b84264cea73e19cb7e916446e1bf025f0f89d2
+GitCommit: 774ac009567d0640091b8856705ad6a78b4468ae
 Directory: jdk21-jammy
 
-Tags: 8.12.1-jdk21-alpine, 8.12-jdk21-alpine, 8-jdk21-alpine, jdk21-alpine, 8.12.1-jdk-alpine, 8.12-jdk-alpine, 8-jdk-alpine, jdk-alpine, 8.12.1-alpine, 8.12-alpine, 8-alpine, alpine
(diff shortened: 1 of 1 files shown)


keeganwitt:
@tianon Side note, this required a change to the script to skip non-existent directories: https://github.com/keeganwitt/docker-gradle/commit/4ec63841d8e35d80bb05959abc44e5e31c6f7ae5.

tianon:
Oh interesting, although I didn't actually check out each branch, so `-d` in the local won't be testing the branch you're targeting. :see_no_evil:

I would guess you probably wanted that following line to change to `if ! dockerfile="$(git show "$commit:$dir/Dockerfile")"; then continue; fi` instead so that it skips if it can't find the `Dockerfile`, but even that doesn't make much sense because that `directories` variable comes from `git ls-tree` which _is_ doing a directory listing of the correct branch. :thinking:

Did you merge the directory deletions back to those other branches too?

tianon:
It's also a bit strange that this is removing `Tags: 8.12.1-jdk-lts-and-current-graal, 8.12-jdk-lts-and-current-graal, 8-jdk-lts-and-current-graal, jdk-lts-and-current-graal, 8.12.1-jdk-lts-and-current-graal-noble, 8.12-jdk-lts-and-current-graal-noble, 8-jdk-lts-and-current-graal-noble, jdk-lts-and-current-graal-noble, 8.12.1-jdk-21-and-23-graal, 8.12-jdk-21-and-23-graal, 8-jdk-21-and-23-graal, jdk-21-and-23-graal, 8.12.1-jdk-21-and-23-graal-noble, 8.12-jdk-21-and-23-graal-noble, 8-jdk-21-and-23-graal-noble, jdk-21-and-23-graal-noble` -- that's not intentional, is it?  That seems like it's possibly/probably a side effect of that less-than-ideal `-d`?

keeganwitt:
> Oh interesting, although I didn't actually check out each branch, so `-d` in the local won't be testing the branch you're targeting. 🙈
> 
> I would guess you probably wanted that following line to change to `if ! dockerfile="$(git show "$commit:$dir/Dockerfile")"; then continue; fi` instead so that it skips if it can't find the `Dockerfile`, but even that doesn't make much sense because that `directories` variable comes from `git ls-tree` which _is_ doing a directory listing of the correct branch. 🤔
> 
> Did you merge the directory deletions back to those other branches too?

Yes, I made the script change in all branches and deleted the directories from the branches that shouldn't have had them.

keeganwitt:
> It's also a bit strange that this is removing `Tags: 8.12.1-jdk-lts-and-current-graal, 8.12-jdk-lts-and-current-graal, 8-jdk-lts-and-current-graal, jdk-lts-and-current-graal, 8.12.1-jdk-lts-and-current-graal-noble, 8.12-jdk-lts-and-current-graal-noble, 8-jdk-lts-and-current-graal-noble, jdk-lts-and-current-graal-noble, 8.12.1-jdk-21-and-23-graal, 8.12-jdk-21-and-23-graal, 8-jdk-21-and-23-graal, jdk-21-and-23-graal, 8.12.1-jdk-21-and-23-graal-noble, 8.12-jdk-21-and-23-graal-noble, 8-jdk-21-and-23-graal-noble, jdk-21-and-23-graal-noble` -- that's not intentional, is it? That seems like it's possibly/probably a side effect of that less-than-ideal `-d`?

Hmm. Yea, it must be. But I'm not sure why.

tianon:
I ran the script and compared the output to running the script with https://github.com/keeganwitt/docker-gradle/commit/4ec63841d8e35d80bb05959abc44e5e31c6f7ae5 reverted, and I got the exact same output in both cases. :sweat_smile:

Mine also still includes `Tags: 8.13-jdk-lts-and-current-graal, 8-jdk-lts-and-current-graal, jdk-lts-and-current-graal, 8.13-jdk-lts-and-current-graal-noble, 8-jdk-lts-and-current-graal-noble, jdk-lts-and-current-graal-noble, 8.13-jdk-21-and-23-graal, 8-jdk-21-and-23-graal, jdk-21-and-23-graal, 8.13-jdk-21-and-23-graal-noble, 8-jdk-21-and-23-graal-noble, jdk-21-and-23-graal-noble`, so something is going weird here. :thinking:

keeganwitt:
> I ran the script and compared the output to running the script with [keeganwitt/docker-gradle@4ec6384](https://github.com/keeganwitt/docker-gradle/commit/4ec63841d8e35d80bb05959abc44e5e31c6f7ae5) reverted, and I got the exact same output in both cases. 😅
> 
> Mine also still includes `Tags: 8.13-jdk-lts-and-current-graal, 8-jdk-lts-and-current-graal, jdk-lts-and-current-graal, 8.13-jdk-lts-and-current-graal-noble, 8-jdk-lts-and-current-graal-noble, jdk-lts-and-current-graal-noble, 8.13-jdk-21-and-23-graal, 8-jdk-21-and-23-graal, jdk-21-and-23-graal, 8.13-jdk-21-and-23-graal-noble, 8-jdk-21-and-23-graal-noble, jdk-21-and-23-graal-noble`, so something is going weird here. 🤔

It looks like the directory variable contains trailing whitespace and that's why it's being seen as not existing.

yosifkit:
Unsure if file permissions will be an issue with users switching to or starting on the UBI based images, but something to look out for or maybe mention in documentation.

yosifkit (approved):

tianon:
> It looks like the directory variable contains trailing whitespace and that's why it's being seen as not existing.

Oh strange, are you on Windows? (and thus getting a `\r` in there somewhere?)

keeganwitt:
> > It looks like the directory variable contains trailing whitespace and that's why it's being seen as not existing.
> 
> Oh strange, are you on Windows? (and thus getting a `\r` in there somewhere?)

Yes, I was running with Cygwin. Maybe that was the issue.

tianon:
Might need to add `| tr -d '\r'` in there somewhere to eat carriage returns in the appropriate place. :eyes:

PR #9379, opened 2021-01-06
Reword the Official Images description

wglambert:
Add short bullet points for the core ideas of the program, and a few sentences as a summary

PR #18552, opened 2025-03-01
Update Zend Server versions to latest releases

rainviigipuu:


tianon:
You might've noticed there's no actual changes in the diff, and the CI didn't pick up any image builds.  If you check the library file, you'll note the entry references have all been replaced with a comment pointing to https://github.com/docker-library/docs/pull/2475, which points in particular to https://github.com/docker-library/official-images/pull/16941#issuecomment-2195225594 -- if we want to resurrect this image into the actively supported set, we'll need some assurance that it will be actively maintained and that you plan to be responsive to issues with it / GitHub pings about it.

rainviigipuu:
Hello,

I updated the library/php-zenderver file with the only Zend Server version that is currently supported. This version will have updates before the official end of support: https://www.zend.com/products/zend-server. Until then we would like to have updated package available via Docker official images. We are going to update the images once the new version becomes available.

I have question however what should we do once the EOS support time arrives? The same way that it is done now, that we just update the library/php-zendserver by removing all the tags and adding a link to information that it is now end of support?

Is there some timeline how long the tags can be available after end of support?

rainviigipuu:
And I do have another question related to the images that are available here: https://hub.docker.com/_/php-zendserver/tags. While I hope that at least 2021 version will have no vulnerabilities once it gets updated, then for example some old images (like 2018 or 9.1 etc) are not going to receive any updates as they are not supported anymore. So the question(s) are:

Is there a way to get rid of those old images there?
Is that common practice that projects have that kind of old images available indefinitely?
Is there a way to update those old images to the latest version before end of support without adding tag-s for them? I understand the ones with tag-s are considered actively supported?

Just trying to understand what is the practice here so we can do some informed decisions what to do (or what we can do) with them. 

Best regards,
Rain

LaurentGoderre:
Could you please rebase this so there is only your commit in the PR?

tianon:
Am I understanding correctly that there is a period of time during which Zend Server has _no_ "supported" versions?  How does that work?

Regarding old versions, we can remove them, but we typically do not unless a maintainer explicitly requests it because they are often of historical interest to users.

Am I understanding that you are agreeing to maintaining this image more actively in the future than we've seen in the past?

Also, relevant to https://github.com/docker-library/official-images#what-are-official-images, I should verify before we agree to resurrection here that Zend Server is Open Source (pursuant to [the OSI's definition](https://opensource.org/osd) of such)?

rainviigipuu:
Zend Server is reaching EOM at the end of 2026 and EOL at the end of 2027: https://portal.perforce.com/s/article/Zend-Support-Lifecycle

Yes, I am agreeing to maintain it more actively.

I have to confirm the open source part and  I will get back to you on that one. However, Zend Server has never been fully open source. Have there been any changes during the past 3-5 years regarding of official images being open source and that makes it not possible the have Zend Server published in official images?

tianon:
> Have there been any changes during the past 3-5 years regarding of official images being open source and that makes it not possible the have Zend Server published in official images?

It was always somewhat of a focus for us (although a bit of a loose one, as you experienced :sweat_smile:), but in https://github.com/docker-library/official-images/pull/9379 we clarified it more strongly, and in https://github.com/docker-library/official-images/pull/11473 added it explicitly to the image review checklist to be more strongly enforced.

So, I think the best place to go from here is probably to update the image deprecation text on Docker Hub to provide a pointer to where you'll (presumably) maintain the images in the future so that users can easily find the new location.  I'm also happy to connect you to the product team for Docker's Trusted Content programs (DOI, DVP, DSOS) for further discussion.

PR #18584, opened 2025-03-07
TomEE: build temurin jre21-alpine based images for arm64v8 as well

jungm:
On older JRE versions Temurin did not support arm64v8 on alpine, however this changed with JRE 21. See: https://adoptium.net/supported-platforms/

IMO there is no reason to build these images for amd64 only


scriptmonkey:
Agreed!

PR #18586, opened 2025-03-07
Update Tomcat version for Semeru Releases

jayasg12:
- ~Handling various repositories for base image~
- Tomcat version update

### DIFF
--- a/library/ibm-semeru-runtimes
+++ b/library/ibm-semeru-runtimes
@@ -7,194 +7,194 @@ GitFetch: refs/heads/ibm
 #-----------------------------openj9 v8 images---------------------------------
 Tags: open-8u442-b06-jdk-focal, open-8-jdk-focal
 Architectures: amd64, ppc64le, s390x, arm64v8
-GitCommit: 84ca03b712e3f51142fc458271b62e10bb9efa76
+GitCommit: 4b7a7c4704cf1d6b2241aa09f8dacef518cf6e74
 Directory: 8/jdk/ubuntu/focal
 File: Dockerfile.open.releases.full
 
 Tags: open-8u442-b06-jdk-jammy, open-8-jdk-jammy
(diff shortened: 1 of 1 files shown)


tianon:
Unfortunately, our build system cannot reasonably support variadic `FROM` values because it needs to calculate the full image dependency tree.

AdamBrousseau:
@tianon is there a mechanism to have the base image configurable at build time that is acceptable to DockerHub official builds? We're trying to avoid having multiple sets of Dockerfiles while allowing a use case where other users want to pull base images from other registry sites.

tianon:
The best/simplest answer is `docker buildx build --build-context`: https://docs.docker.com/reference/cli/docker/buildx/build/#build-context (then users can even swap in totally different base images, although that will likely break the build in adorable ways)

Another way to handle this would be templates, but it sounds like that doesn't solve the use case you're targeting in a very obvious way (unless the users in question are OK with regenerating from the templates before building).  It _is_ kind of odd to have users explicitly building the images themselves though, not pulling them from somewhere else and building `FROM` them. :thinking:

library/ibm-semeru-runtimes
```diff
@@ -7,194 +7,194 @@ GitFetch: refs/heads/ibm
 #-----------------------------openj9 v8 images---------------------------------
 Tags: open-8u442-b06-jdk-focal, open-8-jdk-focal
 Architectures: amd64, ppc64le, s390x, arm64v8
-GitCommit: a278ee463c7dcda550133dccd661058a152df006
+GitCommit: 84ca03b712e3f51142fc458271b62e10bb9efa76
```
yosifkit, line 10: Can you update these `GitCommit` values to include https://github.com/ibmruntimes/semeru-containers/pull/144?

yosifkit (approved):

PR #18608, opened 2025-03-11
Official Convertigo 8.3.3 release!

nicolas-albert:
Thx!


nicolas-albert:
I did the PR before publishing the release public on GH!
Sorry :'(

tianon:
Instead of `sudo`, I would suggest using `setpriv` (which is more similar
in operation to `gosu` and less of a potential security risk than having
`sudo` in the image, and probably already installed in the image).

nicolas-albert:
I used `setpriv` to run tomcat and it's now ok for me.

tianon (approved):

PR #18629, opened 2025-03-13
Improve `toc.sh` to support more of GitHub's edge cases

tianon:
Most notably, this allows emoji in headers to generate the correct anchors.

See https://gist.github.com/tianon/75e267d9137b1c2978031b66b3a98987 for a bunch of test cases I worked through this with (and found bugs in GitHub's implementation while reverse engineering it, which is fun, and quirks of the library I was copying that aren't correct for the GitHub implementation).

I've also verified that this causes no changes to the TOC here or over in the docs repo (as expected).

(I dedicate this to / blame @thompson-shaun, with love :heart:)


tianon:
@thompson-shaun wanted a way to exclude all the "H1" on the page (for example, a `FOO.md` that starts with a `# Foo` to be cute), and because I implemented correct support for "repetition" (the same header twice) that gets a little complicated since you have to parse everything (since it might affect suffixes) but only emit H2+, so here's the smallest change I've found that accomplishes that well:

```diff
diff --git a/toc.sh b/toc.sh
index ed049b611..d4b3e323d 100755
--- a/toc.sh
+++ b/toc.sh
@@ -57,7 +57,9 @@ jq --raw-input --null-input --raw-output '
 					)
 					// error("repetition level too deep on #\($anchor) (\($line)) at line \(input_line_number)")
 				) as $finalAnchor
-				| .toc += "\("\t" * ($level-1) // "")\(.levels[$level]).\t[\($cap.heading)](#\($finalAnchor))\n"
+				| if $level > 1 then
+					.toc += "\("\t" * ($level-2) // "")\(.levels[$level]).\t[\($cap.heading)](#\($finalAnchor))\n"
+				else . end
 				| .seen += [ $finalAnchor ]
 			else . end
 		end
```

tianon:
He's also not a fan of my numbered lists, which is easy to fix by replacing `\(.levels[$level]).` with just `-` (and everything else will "just work").

yosifkit (approved):

PR #18660, opened 2025-03-19
Update to Flink 2.0.0

reswqa:
This is the official release of Apache Flink 2.0.0 (previously in preview1), with the following major changes:

1. We support jdk21 in Flink `2.0.0`, this after the preview release, so `preview1` doesn't include it.
2. After discussion, we switch the default tag to `JDK17` instead of `11` as it's the default jdk version for Flink now.

### DIFF
--- a/library/flink
+++ b/library/flink
@@ -3,14 +3,19 @@
 Maintainers: The Apache Flink Project <dev@flink.apache.org> (@ApacheFlink)
 GitRepo: https://github.com/apache/flink-docker.git
 
-Tags: 2.0-preview1-scala_2.12-java17, 2.0-scala_2.12-java17, 2.0-preview1-java17, 2.0-java17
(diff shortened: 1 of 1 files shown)


library/flink
```diff
@@ -3,27 +3,32 @@
 Maintainers: The Apache Flink Project <dev@flink.apache.org> (@ApacheFlink)
 GitRepo: https://github.com/apache/flink-docker.git
 
-Tags: 2.0-preview1-scala_2.12-java17, 2.0-scala_2.12-java17, 2.0-preview1-java17, 2.0-java17
+Tags: 2.0.0-scala_2.12-java21, 2.0-scala_2.12-java21, scala_2.12-java21, 2.0.0-java21, 2.0-java21, java21
```
reswqa, line 6: We support jdk21 in Flink 2.0, this after the preview release, so preview1 doesn't include it.

yosifkit (approved):

PR #18692, opened 2025-03-24
Update ELK images for 9.0.0-rc1

alpar-t:


yosifkit:
I have basically the same comment as the first part of https://github.com/docker-library/official-images/pull/18475#issuecomment-2666997482:

From the failing CI:
 
```

 - elasticsearch:9.0.0-rc1 (FROM docker.elastic.co/ubi9/ubi-minimal:latest) -- completely unsupported base!
 - kibana:9.0.0-rc1 (FROM docker.elastic.co/ubi9/ubi-minimal:latest) -- completely unsupported base!
 - logstash:9.0.0-rc1 (FROM docker.elastic.co/ubi9/ubi-minimal:latest) [amd64, arm64v8]

Read https://github.com/docker-library/official-images#multiple-architectures for more details.
```

In this case, these are referencing images outside Docker Official Images, which is not allowed in Official Images. There are a few Redhat `ubi` images allowed as exceptions listed in https://github.com/docker-library/official-images/tree/master/.external-pins/redhat, e.g. `redhat/ubi9-minimal:latest`.

> - https://github.com/docker-library/official-images/blob/5b59b4e1f2123d308af6681cc08d75a89057589f/README.md#repeatability

library/elasticsearch
```diff
@@ -17,3 +17,8 @@ Tags: 8.17.3
 Architectures: amd64, arm64v8
 GitFetch: refs/heads/8.17
 GitCommit: 204cb29c8473f0353e1e1b9181e9be03ef94be72
+
+Tags: 9.0.0-rc1
+Architectures: amd64, arm64v8
+GitFetch: refs/heads/9.0
+GitCommit: 67dd1114bfe1ab57a2a1898d5903c3198cffa204
```
alpar-t, line 24:
```suggestion
GitCommit: 8fd67af47850632590faeca14ad65734bca0a163
```

alpar-t:
Thanks @yosifkit ! I switched the images to use `redhat/ubi9-minimal:latest` as a base

yosifkit:
I merged https://github.com/docker-library/official-images/pull/18704 and https://github.com/docker-library/official-images/pull/18703 ahead of this and now there is a conflict. Can you rebase and force push? ❤️

tianon:
Any thoughts on https://github.com/docker-library/official-images/pull/18475#issuecomment-2667010290?

> Also, if we're touching these lines anyways, can we please update them with things like `set -e` so they fail properly if/when things go sideways, like we've suggested prior?
> 
> https://github.com/docker-library/official-images/pull/18008#issuecomment-2515633764
> 
> > https://github.com/docker-library/official-images/pull/17158#issuecomment-2223474718
> > 
> > > https://github.com/docker-library/official-images/pull/16747#issuecomment-2102952435 👀 (which would help make sure it fails sooner, such that if it fails on our official build infrastructure that might use aggressive build cache that the cache is "busted" in the appropriate place and the rebuild has a chance at success)
> > > 
> > > To elaborate, it's failing on `RUN tar -zxf /tmp/elasticsearch.tar.gz --strip-components=1` (in the Elasticsearch case), meaning the `RUN curl` line won't be retried on a rebuild attempt on our official infrastructure, so it will just fail indefinitely if we were to have a similar issue with the download there.
> > 
> > > > Friendly reminder: https://github.com/docker-library/official-images/pull/16698#issuecomment-2096609598 (`set -e`/`set -ux` and avoiding `arch`; `curl -fS` would be good too, but likely fail in other ways if `curl` were to fail)
> > 
> > > > > This is good for now, but please do revisit https://github.com/elastic/logstash/pull/16063#discussion_r1577000627 (adding at the very least `set -e` to all semi-colon-using `RUN` lines, if not the full `set -eux` I suggest there).
> > 
> > > > > (We should also apply the same "avoiding `arch` in favor of querying the image userspace environment" logic to these lines...)

(on that last point, the best way to get "userspace architecture" from an RPM-based distribution that I've found is `rpm --query --queryformat='%{ARCH}' rpm`)

--- 

```diff
+LABEL name="Kibana" \
+  maintainer="infra@elastic.co" \
+  vendor="Elastic" \
+  version="9.0.0-rc1" \
+  release="1" \
+  summary="Kibana" \
+  description="Your window into the Elastic Stack."
```

See https://github.com/docker-library/official-images/issues/3540, especially https://github.com/docker-library/official-images/issues/3540#issuecomment-530925319:

And https://github.com/docker-library/official-images/issues/3540#issuecomment-2715158488:

> It's worth updating this to note that our build system now sets several useful and appropriate OCI annotations directly on the images we push: https://oci.dag.dev/?image=hello-world
> 
> ```json
> "annotations": {
>   "com.docker.official-images.bashbrew.arch": "amd64",
>   "org.opencontainers.image.base.name": "scratch",
>   "org.opencontainers.image.created": "2025-01-22T02:27:30Z",
>   "org.opencontainers.image.revision": "a71acd2dfb8c21c8e42a8f926ce37c1ffbf8eae7",
>   "org.opencontainers.image.source": "https://github.com/docker-library/hello-world.git#a71acd2dfb8c21c8e42a8f926ce37c1ffbf8eae7:amd64/hello-world",
>   "org.opencontainers.image.url": "https://hub.docker.com/_/hello-world",
>   "org.opencontainers.image.version": "linux"
> }
> ```

---

```diff
+RUN for iter in 1 2 3 4 5 6 7 8 9 10; do \
```

This (and similar constructions in other images) are weird -- I understand that these package installations have been flaky in the past, but they shouldn't be, and our build system will auto-retry failing builds.  Can you elaborate on why you're still carrying this?

---

```diff
+RUN mkdir /licenses && cp LICENSE.txt /licenses/LICENSE
```

This is harmless but kind of odd -- can you elaborate on what it's for?  (If it's important, would a symlink be sufficient?  If not, can you make it a hard link instead, and created in the same layer as `LICENSE.txt` is first added to the image?)

---

```diff
+COPY env2yaml/env2yaml.go /tmp/go/src/env2yaml/env2yaml.go
+COPY env2yaml/go.mod      /tmp/go/src/env2yaml/go.mod
+COPY env2yaml/go.sum      /tmp/go/src/env2yaml/go.sum
```

This is kind of verbose - it's not a problem, but any reason not to simply `COPY env2yaml/ /tmp/go/src/env2yaml/`?  If you move `WORKDIR` up, that right-hand-side can even be `./` for less duplication.  At the very least, it could get as small as `COPY env2yaml/env2yaml.go env2yaml/go.mod env2yaml/go.sum ./`

It's not ideal that this means the full Go file ends up as part of the build context we review, but that's mostly OK too, since it's pretty small (it just means we might make suggestions in that code too, or ask more strongly that it move later if it grows too much).

---

```diff
+COPY --from=builder-env2yaml /tmp/go/src/env2yaml/env2yaml /usr/local/bin/env2yaml
 COPY config/pipelines.yml config/log4j2.properties config/log4j2.file.properties config/
+COPY config/logstash-full.yml config/logstash.yml
 COPY pipeline/default.conf pipeline/logstash.conf
...
 COPY bin/docker-entrypoint /usr/local/bin/
```

In general, every `COPY` should happen as late in the `Dockerfile` as possible - in this case, all of the `COPY` lines in this stage of this `Dockerfile` should move after the final `RUN`.

tianon:
```diff
+RUN go build
```

It's probably also a good idea to include `-trimpath` here, so that the generated binary is cleaner / more likely to be reproducible in the future (even if the specific path it's built in changes).

tianon:
```diff
+      microdnf update --setopt=tsflags=nodocs -y && \
```

We'd prefer to see package upgrades happen in the base images directly (and the UBI images are updated pretty proactively), so this should be removed.

tianon:
```Dockerfile
# 1. Sync the user and group permissions of /etc/passwd
# 2. Set correct permissions of the entrypoint
# 3. Ensure that there are no files with setuid or setgid, in order to mitigate "stackclash" attacks.
#    We've already run this in previous layers so it ought to be a no-op.
# 4. Replace OpenJDK's built-in CA certificate keystore with the one from the OS
#    vendor. The latter is superior in several ways.
#    REF: https://github.com/elastic/elasticsearch-docker/issues/171
# 5. Tighten up permissions on the ES home dir (the permissions of the contents are handled earlier)
# 6. You can't install plugins that include configuration when running as `elasticsearch` and the `config`
#    dir is owned by `root`, because the installed tries to manipulate the permissions on the plugin's
#    config directory.
RUN chmod g=u /etc/passwd && \
    chmod 0555 /usr/local/bin/docker-entrypoint.sh && \
    find / -xdev -perm -4000 -exec chmod ug-s {} + && \
    chmod 0775 /usr/share/elasticsearch && \
    chown elasticsearch bin config config/jvm.options.d data logs plugins
```

This is all odd as an explicit layer -- modifications to `/usr/share/elasticsearch` should happen in the layer that creates that directory, execute bit on `docker-entrypoint.sh` should be applied in Git already, `/etc/passwd` _should_ already have appropriate permissions, anything with the sticky bit set should be directly from the UBI images.

alpar-t:
@tianon Could we address your changes as a follow up, decoupled form this PR. I'm happy to facilitate that with the maintainers of the images and make sure this happens. Would like to make the new versions of ELK available through dockerhub  as close as possible to the release into other channels. 9.0.0-RC1 is out everywhere else except for official images for several days now.

tianon:
You can see in some of the provided references that I've been poking on a lot of these issues for a very long time, with no apparent progress.  This _is_ intended to be a curated program, and our review is the primary aspect of that curation.  If the `Dockerfile`s cannot be reasonably updated in response to our review, then we might need to rethink whether these images are suitable candidates for the program, but to be very clear that is not the outcome I want to see.

I'm happy to provide more details, examples, etc in order to help progress these changes.

To illustrate, here's a start on a diff for Elasticsearch specifically:

```diff
diff --git a/elasticsearch/Dockerfile b/elasticsearch/Dockerfile
index 2639257..b6aee7e 100644
--- a/elasticsearch/Dockerfile
+++ b/elasticsearch/Dockerfile
@@ -4,6 +4,8 @@
 # Beginning of multi stage Dockerfile
 ################################################################################
 
+# TODO convert this from multi-stage to single-stage (there's nothing here that really needs to be multi-stage)
+
 ################################################################################
 # Build stage 1 `builder`:
 # Extract Elasticsearch artifact
@@ -13,12 +15,7 @@ FROM redhat/ubi9-minimal:latest AS builder
 
 # Install required packages to extract the Elasticsearch distribution
 
-RUN for iter in 1 2 3 4 5 6 7 8 9 10; do \
-      microdnf install -y findutils tar gzip && \
-      exit_code=0 && break || \
-        exit_code=$? && echo "microdnf error: retry $iter in 10s" && sleep 10; \
-    done; \
-    exit $exit_code
+RUN microdnf install -y findutils tar gzip
 
     # `tini` is a tiny but valid init for containers. This is used to cleanly
     # control how ES and any child processes are shut down.
@@ -29,48 +26,48 @@ RUN for iter in 1 2 3 4 5 6 7 8 9 10; do \
     # build. Instead, we check the binary against the published checksum.
     RUN set -eux ; \
         tini_bin="" ; \
-        case "$(arch)" in \
+        arch="$(rpm --query --queryformat='%{ARCH}' rpm)"; \
+        case "$arch" in \
             aarch64) tini_bin='tini-arm64' ;; \
             x86_64)  tini_bin='tini-amd64' ;; \
-            *) echo >&2 ; echo >&2 "Unsupported architecture $(arch)" ; echo >&2 ; exit 1 ;; \
+            *) echo >&2 ; echo >&2 "Unsupported architecture $arch" ; echo >&2 ; exit 1 ;; \
         esac ; \
-        curl --retry 10 -S -L -O https://github.com/krallin/tini/releases/download/v0.19.0/${tini_bin} ; \
-        curl --retry 10 -S -L -O https://github.com/krallin/tini/releases/download/v0.19.0/${tini_bin}.sha256sum ; \
+        curl -f --retry 10 -S -L -O https://github.com/krallin/tini/releases/download/v0.19.0/${tini_bin} ; \
+        curl -f --retry 10 -S -L -O https://github.com/krallin/tini/releases/download/v0.19.0/${tini_bin}.sha256sum ; \
         sha256sum -c ${tini_bin}.sha256sum ; \
         rm ${tini_bin}.sha256sum ; \
         mv ${tini_bin} /bin/tini ; \
         chmod 0555 /bin/tini
 
-RUN mkdir /usr/share/elasticsearch
 WORKDIR /usr/share/elasticsearch
 
-RUN curl --retry 10 -S -L --output /tmp/elasticsearch.tar.gz https://artifacts-no-kpi.elastic.co/downloads/elasticsearch/elasticsearch-9.0.0-rc1-linux-$(arch).tar.gz
-
-RUN tar -zxf /tmp/elasticsearch.tar.gz --strip-components=1
+RUN arch="$(rpm --query --queryformat='%{ARCH}' rpm)" && curl -f --retry 10 -S -L --output /tmp/elasticsearch.tar.gz https://artifacts-no-kpi.elastic.co/downloads/elasticsearch/elasticsearch-9.0.0-rc1-linux-$arch.tar.gz
 
-# The distribution includes a `config` directory, no need to create it
-COPY config/elasticsearch.yml config/
-COPY config/log4j2.properties config/log4j2.docker.properties
-
-#  1. Configure the distribution for Docker
-#  2. Create required directory
-#  3. Move the distribution's default logging config aside
-#  4. Move the generated docker logging config so that it is the default
-#  5. Reset permissions on all directories
-#  6. Reset permissions on all files
-#  7. Make CLI tools executable
-#  8. Make some directories writable. `bin` must be writable because
-#     plugins can install their own CLI utilities.
-#  9. Make some files writable
-RUN sed -i -e 's/ES_DISTRIBUTION_TYPE=tar/ES_DISTRIBUTION_TYPE=docker/' bin/elasticsearch-env && \
+RUN tar -zxf /tmp/elasticsearch.tar.gz --strip-components=1 && \
+# Configure the distribution for Docker
+    sed -i -e 's/ES_DISTRIBUTION_TYPE=tar/ES_DISTRIBUTION_TYPE=docker/' bin/elasticsearch-env && \
+# Create required directory
     mkdir data && \
-    mv config/log4j2.properties config/log4j2.file.properties && \
-    mv config/log4j2.docker.properties config/log4j2.properties && \
+# Reset permissions on all directories
     find . -type d -exec chmod 0555 {} + && \
+# Reset permissions on all files
     find . -type f -exec chmod 0444 {} + && \
+# Make CLI tools executable
     chmod 0555 bin/* jdk/bin/* jdk/lib/jspawnhelper modules/x-pack-ml/platform/linux-*/bin/* && \
+# Make some directories writable. `bin` must be writable because
+# plugins can install their own CLI utilities.
     chmod 0775 bin config config/jvm.options.d data logs plugins && \
-    find config -type f -exec chmod 0664 {} +
+# Make some files writable
+    find config -type f -exec chmod 0664 {} + && \
+# Tighten up permissions on the ES home dir (the permissions of the contents are handled below)
+    chmod 0775 . && \
+# You can't install plugins that include configuration when running as `elasticsearch` and the `config`
+# dir is owned by `root`, because the installed tries to manipulate the permissions on the plugin's
+# config directory.
+    chown 1000:1000 bin config config/jvm.options.d data logs plugins
+
+# The distribution includes a `config` directory, no need to create it
+COPY config/elasticsearch.yml config/log4j2.properties config/
 
 ################################################################################
 # Build stage 2 (the actual Elasticsearch image):
@@ -81,15 +78,9 @@ RUN sed -i -e 's/ES_DISTRIBUTION_TYPE=tar/ES_DISTRIBUTION_TYPE=docker/' bin/elas
 
 FROM redhat/ubi9-minimal:latest
 
-RUN for iter in 1 2 3 4 5 6 7 8 9 10; do \
-      microdnf update --setopt=tsflags=nodocs -y && \
-      microdnf install --setopt=tsflags=nodocs -y \
+RUN microdnf install --setopt=tsflags=nodocs -y \
         nc shadow-utils zip unzip findutils procps-ng && \
-      microdnf clean all && \
-      exit_code=0 && break || \
-        exit_code=$? && echo "microdnf error: retry $iter in 10s" && sleep 10; \
-    done; \
-    exit $exit_code
+      microdnf clean all
 
 RUN groupadd -g 1000 elasticsearch && \
     adduser -u 1000 -g 1000 -G 0 -d /usr/share/elasticsearch elasticsearch && \
@@ -97,34 +88,20 @@ RUN groupadd -g 1000 elasticsearch && \
 
 ENV ELASTIC_CONTAINER true
 
+COPY --from=builder /bin/tini /bin/tini
+
 WORKDIR /usr/share/elasticsearch
 
-COPY --from=builder --chown=0:0 /usr/share/elasticsearch /usr/share/elasticsearch
+COPY --from=builder /usr/share/elasticsearch .
 
-COPY --from=builder --chown=0:0 /bin/tini /bin/tini
+# Replace OpenJDK's built-in CA certificate keystore with the one from the OS
+# vendor. The latter is superior in several ways.
+# REF: https://github.com/elastic/elasticsearch-docker/issues/171
+RUN ln -sf /etc/pki/ca-trust/extracted/java/cacerts jdk/lib/security/cacerts
 
 ENV PATH /usr/share/elasticsearch/bin:$PATH
 ENV SHELL /bin/bash
-COPY bin/docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
-
-# 1. Sync the user and group permissions of /etc/passwd
-# 2. Set correct permissions of the entrypoint
-# 3. Ensure that there are no files with setuid or setgid, in order to mitigate "stackclash" attacks.
-#    We've already run this in previous layers so it ought to be a no-op.
-# 4. Replace OpenJDK's built-in CA certificate keystore with the one from the OS
-#    vendor. The latter is superior in several ways.
-#    REF: https://github.com/elastic/elasticsearch-docker/issues/171
-# 5. Tighten up permissions on the ES home dir (the permissions of the contents are handled earlier)
-# 6. You can't install plugins that include configuration when running as `elasticsearch` and the `config`
-#    dir is owned by `root`, because the installed tries to manipulate the permissions on the plugin's
-#    config directory.
-RUN chmod g=u /etc/passwd && \
-    chmod 0555 /usr/local/bin/docker-entrypoint.sh && \
-    find / -xdev -perm -4000 -exec chmod ug-s {} + && \
-    chmod 0775 /usr/share/elasticsearch && \
-    chown elasticsearch bin config config/jvm.options.d data logs plugins
-
-RUN ln -sf /etc/pki/ca-trust/extracted/java/cacerts /usr/share/elasticsearch/jdk/lib/security/cacerts
+COPY --chmod=0555 bin/docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
 
 EXPOSE 9200 9300
 
@@ -148,16 +125,6 @@ LABEL org.label-schema.build-date="2025-03-20T10:06:19.154801632Z" \
   org.opencontainers.image.vendor="Elastic" \
   org.opencontainers.image.version="9.0.0-rc1"
 
-LABEL name="Elasticsearch" \
-  maintainer="infra@elastic.co" \
-  vendor="Elastic" \
-  version="9.0.0-rc1" \
-  release="1" \
-  summary="Elasticsearch" \
-  description="You know, for search."
-
-RUN mkdir /licenses && cp LICENSE.txt /licenses/LICENSE
-
 # Our actual entrypoint is `tini`, a minimal but functional init program. It
 # calls the entrypoint we provide, while correctly forwarding signals.
 ENTRYPOINT ["/bin/tini", "--", "/usr/local/bin/docker-entrypoint.sh"]
```

breskeby (commented):
@alpar-t  Can we update the maintainer of the elasticsearch container to myself (Rene Groeschke (@breskeby) ?

breskeby:
@tianon Your feedback has been directly addressed in https://github.com/elastic/elasticsearch/pull/126330 which will definitely be part of our next release but not 100% sure we can make it for the 9.0.0 version

tianon:
Nice, welcome!  Will we get some similar attention for Logstash and Kibana? :see_no_evil:  (They have similar issues, but I ran out of steam on just Elasticsearch, as seen in some of the `TODO` comments I left behind :sweat_smile: :heart:)

> not 100% sure we can make it for the 9.0.0 version

This is a big point of confusion for me -- why would your feature freeze apply to your _packaging_?  Why can't the `Dockerfile` be updated out-of-band from the releases?  (Your life as the maintainer of them will frankly be improved by decoupling these and treating them as separate streams of work, with one downstream from the other.)

> https://github.com/elastic/elasticsearch/pull/126330

Is there something that applies these templates in a way that can be previewed and reviewed directly?  That template language looks frankly quite painful to write `Dockerfile`s in (I don't envy you), but it also makes it hard to review them directly and see what the effect will be (which would make it easier to comment on them / continue reviewing).

jbudz:
> Will we get some similar attention for Logstash and Kibana?

:+1: We're taking a look at these too.  Updates to follow.

jsvd:
We're also working on concerns raised for the Logstash image at https://github.com/elastic/logstash/pull/17494

jbudz:
Thanks for all the feedback @tianon.

Is this a hard requirement?  We have these added for openshift certification:

https://docs.redhat.com/en/documentation/red_hat_software_certification/2025/html/red_hat_openshift_software_certification_policy_guide/assembly-requirements-for-container-images_openshift-sw-cert-policy-introduction#con-image-metadata-requirements_openshift-sw-cert-policy-container-images

tianon:
Unfortunately yes, that is a hard requirement here in DOI.

IMO, OpenShift should really consider changing their certification (and should've known better in the first place), as both the relevant Docker documentation _and_ the OCI documentation are fairly clear that label keys "SHOULD" (in the RFC 2119 sense / "we'd make this a MUST but it can't be for various reasons") use something like reverse-DNS notation for labels, and the OCI maintains a list of pre-defined common, standard keys that cover their intended uses:

- https://docs.docker.com/engine/manage-resources/labels/#key-format-recommendations
- https://github.com/opencontainers/image-spec/blob/v1.1.1/config.md#:~:text=creating%20a%20container.-,Labels,-object%2C%20OPTIONAL
- https://github.com/opencontainers/image-spec/blob/v1.1.1/annotations.md#rules

jsvd:
@tianon 👋 I've been working on the changes to Logstash and I'm down to the label issue and there's one concern I'd like to raise. it seems that the image ubi9-minimal also has the non-namespaced labels set, meaning that any image built on it will inherit the label. In Logstash's case we'll now have:

```
❯ docker inspect redhat/ubi9-minimal:latest | jq '.[0].Config.Labels | {name: .name, maintainer: .maintainer, vendor: .vendor}'
{
  "name": "ubi9-minimal",
  "maintainer": "Red Hat, Inc.",
  "vendor": "Red Hat, Inc."
}

❯ docker inspect docker.io/library/logstash:latest | jq '.[0].Config.Labels | {name: .name, maintainer: .maintainer, vendor: .vendor}'
{
  "name": "ubi9-minimal",
  "maintainer": "Red Hat, Inc.",
  "vendor": "Red Hat, Inc."
}
```

I.e. images will still have the non-namespaced labels AND now also show incorrect maintainer, vendor etc. for those that query such labels unknowingly.
If possible I'd prefer for our images to keep overriding those labels, and if upstream ubi9 drops them we can make the change then, the PRs don't seem complicated on elasticsearch, kibana, nor logstash.

yosifkit:
> If possible I'd prefer for our images to keep overriding those labels, and if upstream ubi9 drops them we can make the change then

For now, that seems reasonable to me. Can the other image improvements be applied to the 8.x and 7.x images too?

alpar-t:
Closing this PR as we won't be pursuing a 9.0.0-rc1 release

PR #18697, opened 2025-03-24
REL-3640 SonarQube Server 2025 Release 2

jCOTINEAU:


yosifkit:
The new 2025 release 2 images compared to the `2025.1.1` images are swapping to an older JDK, is that expected?
```diff
-FROM eclipse-temurin:21-jre-noble
+FROM eclipse-temurin:17-jre-noble
```

jCOTINEAU:
> The new 2025 release 2 images compared to the `2025.1.1` images are swapping to an older JDK, is that expected?
> 
> ```diff
> -FROM eclipse-temurin:21-jre-noble
> +FROM eclipse-temurin:17-jre-noble
> ```

Indeed it was not meant, updating PR accordingly.

We would actually like to thank you a lot for the very sharp reviews as usual.

Have a nice day.

Issue #1864, opened 2016-06-21
Do official images get rebuilt automatically?


benbc:
Hello

Do the images in the official library get regularly rebuilt to pick up security updates to the base images?

Thanks
-Ben

yosifkit:
Yes, we do periodic updates on the base images like Debian and Ubuntu which we then rebuild all descendant images.  The base or affected images are also recreated when there is a major vulnerability, like https://github.com/docker-library/official-images/issues/1235.

benbc:
Thanks @yosifkit. Does this happen for all tags, or only those still in the library definition file?

yosifkit:
Only the tags in the library files.

tianon:
(which is the explicit set of "supported tags")

benbc:
Thank you. We have not fully thought through the need to keep rebuilding old images, so we've been removing tags for superseded point releases from the library file despite their still being supported.

benbc:
@yosifkit @tianon Do you, as the library maintainers, have a preference for whether images retain tags for superseded point releases?

tianon:
Our stance is usually one that's best described by example IMO.  If there were a version 1.1.2 and a problem were discovered with it, would there be a 1.1.3 or a 1.1.2.1 or something?  Would there be a new release which fixes the issue?  If the answer is yes, that's what I consider to be "supported" (in the sense that issues with the software itself, not just the Dockerization, will be fixed).

That being said, we're willing to discuss further on a case-by-case basis if there are special considerations that ought to be made. :smile:

dalanmiller:
Another question somewhat related to this. I see that in my public docker hub projects there's a webhook option. Would it be possible to be notified if builds succeed or fail via the webhook?

tianon:
@dalanmiller I think that question is likely more appropriately directed at Docker's support team (via support@docker.com) -- we aren't part of the automated builds team, and can't really speak to the feature set there :innocent:

benbc:
@tianon We (the Neo4j maintainers) have given some more thought to this and to security concerns generally. We're uncomfortable with expecting our users to upgrade to the latest point release of Neo4j in order to get security fixes to the underlying platform.

Would you object to us leaving tags for all point releases of supported versions in our library files so that our users can fix the Neo4j version but still benefit from security patches to the OS etc?

benbc:
@yosifkit, @tianon: Any thoughts on this?

tianon:
[...]

PR #18699, opened 2025-03-24
Update couchbase-server

udkyo:


udkyo:
This change adds actively maintained versions of Couchbase Server which are currently absent.

yosifkit:
I think our definition of "actively maintained" is very different given the conversation on https://github.com/docker-library/official-images/pull/15369 and https://github.com/docker-library/official-images/pull/15134. So, I understand that we may be looking at this list of images a little differently, but I think more versions can be acceptable with a few restrictions: 

1. they are either updated with a new `FROM` or removed from the library file when their base image is no longer supported (e.g., when `ubuntu:20.04` or `22.04` becomes end of life)
2. the number of images is limited (https://github.com/docker-library/official-images/issues/1864)

Without number 1, then the image will not benefit from our periodic rebuilds so it would be effectively as if it were removed, so it should just be removed if it will not be moved to an updated base image. It would also fail the GitHub actions tests on PRs to `docker-library/official-images`. Number 2 helps us manage our build queue so that all of our images benefit in a timely manner to base image updates.

udkyo:
Thanks yosifkit, on the topic of "[actively maintained/supported](https://github.com/docker-library/official-images/pull/15369#issuecomment-1720305383)", I think our definitions may not be too dissimilar. We have some automation producing this file which is only interested in versions of Couchbase Server that are inside their **full maintenance** period (On https://www.couchbase.com/support-policy/EOL/ you can see tables showing GA, End of Full Maintenance and End of Life dates. Once a version reaches the end of its full maintenance period it would be removed promptly)

On point 1, we *shouldn’t* get to a point where we have versions in there on EOL base images, but I've added some logic to ensure we check the base image EOL dates and do not include any tags where the base is EOL just in case (this doesn't change the landscape in this PR, but the guardrails are in place to automatically strip out versions whose base images are EOL should we end up in this specific situation at some point in the future)

On point 2, our situation is similar to https://github.com/docker-library/official-images/issues/1864#issuecomment-235288744 - we have enterprise customers who are running supported point releases of our software and would need security fixes for the base images without being forced to upgrade to a new version of Couchbase Server. These releases are still officially supported and actively maintained by Couchbase, with customers often having valid reasons for staying on specific versions.

I would love to shorten the list, but the versions in this PR are all officially supported and being actively maintained. I believe they should all be on there.

yosifkit (approved):

PR #18701, opened 2025-03-24
Telegraf v1.34.1

srebhan:
Thank you very much!


library/telegraf
```diff
@@ -2,7 +2,7 @@ Maintainers: Sven Rebhan <srebhan@influxdata.com> (@srebhan),
              Dane Strandboge <dstrandboge@influxdata.com> (@DStrand1),
```
tianon, line 2: As something to look at in the future, @DStrand1 isn't a current existing GitHub handle, so this probably needs an update of some kind. :sweat_smile:

tianon (approved):

PR #18705, opened 2025-03-25
Updates for WebSphere and Open Liberty 25.0.0.3 release

wraschke:


yosifkit:
> +    echo "Deprecation notice: IBM expects the last version of the UBI-based Open Liberty container images in Docker Hub ('openliberty/open-liberty') to be 25.0.0.3. To continue to receive updates and security fixes after 25.0.0.3, you must switch to using the images from the IBM Container Registry (ICR). To switch, simply update 'FROM openliberty/open-liberty' in your Dockerfiles to 'FROM icr.io/appcafe/open-liberty'. The same image tags from Docker Hub are also available in ICR. Ubuntu-based Liberty container images will continue to be available from Docker Hub. For more information, see https://ibm.biz/ol-ubi-containers-dh-deprecation"

The deprecation notices only mention the `openliberty/open-liberty` and `ibmcom/websphere-liberty` images and not the images in Docker Official Images, `open-liberty` and `websphere-liberty`. Is this meant to apply in DOI as well or just IBM-built images in the `openliberty/` and `ibmcom/` namespaces on Docker Hub?

wraschke:
@yosifkit when I navigate to that link, I don't see the text you quote. Where again did you copy that excerpt from?

But on topic, the unavailability of post-25.0.0.3 Liberty container images applies only for UBI-based. That notice has no relevance for the Ubuntu-based images you publish on our behalf in Docker Official Images.

I hope this answers your question!

yosifkit:
That notice is in the scripts included in the new images (for example in `open-liberty:25.0.0.3-full-java8-openj9` at `helpers/build/internal/logger.sh`). See the generated diff in the GitHub actions job: https://github.com/docker-library/official-images/actions/runs/14064467925/attempts/1#summary-39383325038

yosifkit:
😕 Wouldn't users see the deprecation warning when `logger.sh` is run or sourced? Oh, is ` logger.sh` (or scripts that call it) not used in the DOI images? Then those scripts shouldn't be included in the DOI build context. If `logger.sh` is used, then it shouldn't have the non-applicable deprecation notice that would confuse users.

leochr:
@yosifkit Thank you for the review. The logger and other scripts are shared between the UBI and Ubuntu/DOI images and the source and build setup wouldn't easily allow them to package different scripts. The deprecation notice explicitly mentions that the deprecation applies only to UBI-based images and the relevant registries. It also states that Ubuntu-based images will continue to be available on Docker Hub. I think that clarifies your concern and shouldn't create confusion for users. The majority of Liberty users use both UBI and Ubuntu-based images, hence the deprecation notice is beneficial to be seen so that they are aware and take the necessary action. Not doing so would result in them running outdated versions/packages and expose themselves to security exploitation. We had some inquiries about the 25.0.0.3 DOI images. Hopefully, the above clears things up to merge PR and publish the images. Thank you

yosifkit (approved):

leochr:
@yosifkit Thank you for approving and merging the PR

PR #18713, opened 2025-03-26
[security] Update ruby

tianon:
Changes:

- https://github.com/docker-library/ruby/commit/0012170: Update 3.2 to 3.2.8
- https://github.com/docker-library/ruby/commit/5331b92: Update 3.1 to 3.1.7


yosifkit (approved):

yosifkit:
As a reminder, and as seen in the diff above, this includes https://github.com/docker-library/ruby/pull/497#issuecomment-2590877455 (less packages in `slim` and `alpine` image variants).

PR #18756, opened 2025-04-01
Desupport SapMachine 11 docker images on ppc64le and arm64v8

RealCLanger:
For SapMachine 11 we now only build x86_64, so other architectures have to be desupported. This was also reported in https://github.com/SAP/SapMachine/issues/1946


tianon (approved):
Nice, thank you!

PR #18781, opened 2025-04-05
Create cmake-single-platform.yml

usahospitals:


yosifkit:
There is no use for `cmake` in this repository. This looks indistinguishable from spam.

PR #18797, opened 2025-04-08
Update Jetty version to 12.0.19

lachlan-roberts:
Update Jetty 12 version from `12.0.18` to `12.0.19`.


tianon (dismissed):

lachlan-roberts:
@tianon I have just updated the JDK23 bases to use JDK24 instead

tianon (approved):

PR #18822, opened 2025-04-10
Update Windows pins (redux)

tianon:


yosifkit (approved):

tianon:
It's interesting that https://support.microsoft.com/en-us/topic/windows-server-containers-for-april-2025-e6f2fda6-6378-4c6b-a892-604fbe354d46 doesn't mention the reverts they've clearly done. :thinking:

tianon:
> IMPORTANT We are aware that the WS2019 floating tags (for example, mcr.microsoft.com/windows/servercore:LTSC2019 or mcr.microsoft.com/windows/servercore:1809) are currently retrieving the 2025.03 B (March 2025) container images instead of the latest 2025.04 B (April 2025) images. We are actively working to resolve this behavior.
> 
> To work around this issue, we recommend using explicit tags—such as the KB number, build number, or SHA256 (for example, mcr.microsoft.com/windows/servercore:10.0.17763.7136)—to pull the latest 2025.04 B container image.

PR #18831, opened 2025-04-11
Update to R 4.5.0 released this morning

eddelbuettel:
Standard update to the new R release made by upstream this morning, and using the Debian (unstable, as made today) package prepared earlier.  

No changes in the container setup or build besides the upgrade from R 4.4.3 to R 4.5.0

### DIFF
--- a/library/r-base
+++ b/library/r-base
@@ -2,8 +2,8 @@ Maintainers: Carl Boettiger <rocker-maintainers@eddelbuettel.com> (@cboettig),
              Dirk Eddelbuettel <rocker-maintainers@eddelbuettel.com> (@eddelbuettel)
 GitRepo: https://github.com/rocker-org/rocker.git
 
-Tags: 4.4.3, latest
+Tags: 4.5.0, latest
 Architectures: amd64, arm64v8, ppc64le, s390x
-GitCommit: 6fcc5d8dad96fa24438df79cb46451c5534c04ae
-Directory: r-base/4.4.3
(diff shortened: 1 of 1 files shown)


library/r-base
```diff
@@ -2,8 +2,8 @@ Maintainers: Carl Boettiger <rocker-maintainers@eddelbuettel.com> (@cboettig),
              Dirk Eddelbuettel <rocker-maintainers@eddelbuettel.com> (@eddelbuettel)
 GitRepo: https://github.com/rocker-org/rocker.git
 
-Tags: 4.4.3, latest
+Tags: 4.5.0, latest
 Architectures: amd64, arm64v8, ppc64le, s390x
-GitCommit: 6fcc5d8dad96fa24438df79cb46451c5534c04ae
-Directory: r-base/4.4.3
+GitCommit: 1c42a9d195fd78b9a04de9eee9c6d0de3699dfce
+Directory: r-base/4.5.0
```
yosifkit, line 8: It looks like you currently only have an `r-base/latest` directory at this commit, not an `r-base/4.5.0` directory.

eddelbuettel: Thank you for eagle eyes. I was asleep at the wheel and `magit` showed me the modified (existing) file and not the (new, via new directory) one I overlooking.  Amended commit.  Will amend PR in a second with new sha1 covering both.

And yay tests!

eddelbuettel:
Amended and force-pushed in our repo, amended sha1 one and force-pushed as well.

yosifkit (approved):

PR #18852, opened 2025-04-14
Add JRuby 10.0.0.0 as latest with JDK/JRE 21 and 24

headius:


headius:
Missed a few things verifying...I'll unmark draft when ready.

headius:
Sorry for the churn, new JDK/JRE24 images required some tweaking.

yosifkit (approved):

PR #18854, opened 2025-04-15
Update ELK images for 8.17.5

sfendourakis:


yosifkit (approved):

yosifkit:
Unless there is a larger change for us to review in one version and not the other (this and https://github.com/docker-library/official-images/pull/18855), please use one PR for simple version bumps to help minimize notifications and the possibility of merge conflicts. Once 7.x is removed (when 9.x is GA?), doing separate PRs is going to have merge conflicts.

PR #18855, opened 2025-04-15
Update ELK images for 8.18.0

alpar-t:


yosifkit:
Unless there is a larger change for us to review in one version and not the other (this and https://github.com/docker-library/official-images/pull/18855), please use one PR for simple version bumps to help minimize notifications and the possibility of merge conflicts. Once 7.x is removed (when 9.x is GA?), doing separate PRs is going to have merge conflicts.

yosifkit:
Can you rebase and force push to clear up the merge conflict that came from #18854?

alpar-t:
Thanks for the suggestion @yosifkit we updated our process to make sure a single PR is sent your way even if we have more than one release in a day

yosifkit (approved):

PR #18861, opened 2025-04-15
feat: release influxdb:3-core and influxdb:3-enterprise

bnpfeife:


bnpfeife:
> I think this is the only part that is a bit unexpected. Since this applies to any argument passed to the image, it would apply to users exploring or extending the image. I would recommend only running the entrypoint for loop when the first argument is the expected influxdb3.

Ah, that makes sense. Thanks for the suggestion. I've revised the entrypoint script.

PR #18899, opened 2025-04-20
eclipse-mosquitto: Bump 1.6 image to alpine 3.21

ralight:


tianon:
As seen in the diff above, there are no `Dockerfile` changes here -- is that intentional? :sweat_smile:

PR #18901, opened 2025-04-21
[Friendica] Image optimizations

nupplaphil:


ne20002:
@nupplaphil 
This will update all images including the 2024.08 which haven't been tested (by me). Is the update to php 8.3 ok for 2024.08?

nupplaphil:
Dropped 2024.08 image :)

ne20002:
Thank you @tianon for the findings.
I'll have a look on the first one tomorrow. 
For the second, yes this change is expected and based on how this is handled in Friendica will only 'affect' new installations.

ne20002:
For the first issue: it should work this way. The set -ex adds tracing output and return value check. The statements are uncritical and should work nonetheless.
I'll add it to the sources anyway so it is available in the next version. Is this ok?

nupplaphil:
Done :)

tianon (approved):

PR #18983, opened 2025-05-06
Update ELK images for 8.17.6, 8.18.1, and 9.0.1

StamatisKourk:


tianon:
Some notes -- I've tried to keep them organized so they're easy to follow / keep straight, but please do feel free to ask questions to clarify further.

Also, I'm very happy to (and think it would be better for both of us) find some way to have us pre-review changes to increase the chances of success at PR time.  As I alluded to in https://github.com/docker-library/official-images/pull/18692#issuecomment-2779757226, it would be _really_ helpful to that end to have some way to pre-review the final `Dockerfile` (because that's what we actually care about) instead of just the templates.

---

# elasticsearch

## blocking questions/comments

```diff
+    arch="$(rpm --query --queryformat='%{ARCH}' rpm)";     case "$(arch)" in \
```

The `$(arch)` in the `case` statement here should be `$arch` or `${arch}` so it uses the variable value (not running a new command).  Also, the whitespacing here makes me think there's probably a glitch in the template (`\` vs `\\`).

---

```diff
+RUN arch="$(rpm --query --queryformat='%{ARCH}' rpm)" && curl -f --retry 10 -S -L --output /tmp/elasticsearch.tar.gz https://artifacts-no-kpi.elastic.co/downloads/elasticsearch/elasticsearch-9.0.1-linux-$(arch).tar.gz
```

Same issue here (`$(arch)` that should be `$arch` or `${arch}`).

---

```diff
+COPY --chmod=0555 bin/docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
+
 RUN chmod g=u /etc/passwd && \
     chmod 0555 /usr/local/bin/docker-entrypoint.sh && \
     find / -xdev -perm -4000 -exec chmod ug-s {} + && \
     chmod 0775 /usr/share/elasticsearch && \
     chown elasticsearch bin config config/jvm.options.d data logs plugins
```

`chmod g=u /etc/passwd` is weird, and the rest of this appears to be changing the permissions and ownership of files copied/installed previously -- they should be adjusted so that they `COPY` or install with the appropriate permissions in the previous layers (since this will make a copy of all the files modified/touched in this new layer).

---

```diff
+RUN mkdir /licenses && ln LICENSE.txt /licenses/LICENSE
```

This is still strange -- what purpose does it serve?  What tooling is this in support of?

---

## non-blocking questions/comments

```diff
+# Build stage 1 `builder`:
```

As discussed previously, I don't see a very compelling reason here for this to be a multi-stage `Dockerfile`.

---

```diff
+    curl -f --retry 10 -S -L -O https://github.com/krallin/tini/releases/download/v0.19.0/${tini_bin}.sha256sum ; \
```

Ideally this checksum value would be embedded directly, not downloaded from the exact same source as the binary being verified.

---

```diff
+ENV SHELL=/bin/bash
```

This is odd -- is this load-bearing somewhere in the Elasticsearch scripts?

---

```diff
+LABEL name="Elasticsearch" \
+  maintainer="infra@elastic.co" \
+  vendor="Elastic" \
+  version="9.0.1" \
+  release="1" \
+  summary="Elasticsearch" \
+  description="You know, for search."
```

As discussed in https://github.com/docker-library/official-images/pull/18692#issuecomment-2797149554, this is not ideal, but acceptable due to the UBI image setting these labels (so these are overrides).

---

Similar to the note about tini (and the notes about kibana and logstash below), it would be ideal if we could embed a checksum for the elasticsearch tarball directly in the `Dockerfile.

---

# kibana

## blocking questions/comments

This new UBI-based 9.0 `Dockerfile` has a lot of the same issues as Elasticsearch, but without some of the in-progress improvements we discussed in https://github.com/docker-library/official-images/pull/18692. :confused:

(`$(arch)` usage, `curl` without `-f`, unnecessary multi-stage, weird retries on package installation, `microdnf update` usage, embedded checksum for raw downloads, etc)

---

# logstash

## blocking questions/comments

```diff
+RUN for iter in {1..10}; do \
+  microdnf update -y && \
+  microdnf install -y procps findutils tar gzip && \
+  microdnf install -y openssl && \
+  microdnf install -y which shadow-utils && \
+  microdnf clean all && \
+  exit_code=0 && break || \
+  exit_code=$? && echo "packaging error: retry $iter in 10s" && \
+  microdnf clean all && sleep 10; \
+  done; \
+  (exit $exit_code)
```

Similar comments as elasticsearch and kibana -- are these retries really necessary?  Also `microdnf update` is a no-go (we'll update the base image proactively from UBI upstream updates and rebuild these images, so updates to packages already installed should come from the base image itself instead).

---

```diff
+  curl -Lo - https://artifacts.elastic.co/downloads/logstash/logstash-9.0.1-linux-x86_64.tar.gz | \
   tar zxf - -C /usr/share && \
```

Is this x86_64-arch-specific artifact really the intended one?  Should this be doing (`rpm -q`-based) architecture detection instead?

Also, `curl` should include at _least_ `-f`/`--fail`, and piping the output of `curl` straight to `tar` is totally unacceptable -- ideally there would be at _least_ an embedded checksum or something verifying the download before extraction (which is something that honestly applies to elasticsearch and kibana too :confused:).

---

## non-blocking questions/comments

```diff
--- /dev/null
+++ b/logstash_9.0.1/env2yaml/env2yaml.go
```

Having this full source code as part of the build context means that it's part of what we'll review and potentially comment/block on.  It seems to be relatively small, and I don't imagine it'll grow much, so this isn't a blocker for us (unless that changes), but is something for you to keep in mind.

alpar-t:
Thanks for the review @tianon , we missed a change-set to Kibana, the discussed improvements should be part of https://github.com/elastic/dockerfiles/commit/a6f325795efcc9c2316ee8e4277b703ffe86484b and we'll be updating this PR to that commit

jsvd:
@tianon a few comments:
1. WRT multi-stage, isn't it a best pratice to curl and unpack the tarball in a separate stage so the tarball and tar doesn't end in the final image layers? do you recommend having a single stage instead? 
2. WRT /license directory, it is a requirement for redhad certification https://docs.redhat.com/en/documentation/red_hat_software_certification/2025/html-single/red_hat_enterprise_linux_software_certification_policy_guide/index#con-image-content-requirements_container-requirements
>  Container images must contain a “licenses” directory. Use this directory to add files containing software terms and conditions for your product and any open source software included in the image.
> Test name: HasLicense | Ensures that customers are aware of the terms and conditions applicable to the software included in the image.

Given the bet on ubi images here, I think docker images that pass the certification checks should be encouraged.

breskeby:
> chmod g=u /etc/passwd is weird, and the rest of this appears to be changing the permissions and ownership of files copied/installed previously -- they should be adjusted so that they COPY or install with the appropriate permissions in the previous layers (since this will make a copy of all the files modified/touched in this new layer).

This is all related to a few specific concerns regarding overall security and permission model in elasticsearch and what we found as a approach a while ago to address those. see https://github.com/elastic/elasticsearch/pull/70635#issuecomment-879099727 for further details. this is probably something we can fix longterm but also requires some investment in discussions with different users and stakeholders as changing this adhoc would be effectively a breaking change I think.

StamatisKourk:
@Tianon, we have updated this PR to the latest hash of the dockerfiles 9.0 branch. We are ready to proceed

tianon:
Can we get the `TINI_CHECKSUM` bits added to `elasticsearch` as well in the next bump?  (I'll reiterate that it'd be really great to get checksums for all the elasticsearch, logstash, and kibana tarballs too :heart:)

> WRT multi-stage, isn't it a best pratice to curl and unpack the tarball in a separate stage so the tarball and tar doesn't end in the final image layers? do you recommend having a single stage instead?

See https://github.com/docker-library/faq#multi-stage-builds -- the short version is that multi-stage builds do have some rough edges for the ways our builds are performed.

I'm OK with keeping the multi-stage builds, but we should work on moving up things like this into the "builder" stage so that when they're copied down, they're copied with the appropriate permissions/ownership already pre-set (because this makes a layered copy of every file whose metadata changes due to how layering works in the container filesystems):

```diff
     chmod 0775 /usr/share/elasticsearch && \
     chown elasticsearch bin config config/jvm.options.d data logs plugins
```

> WRT /license directory, it is a requirement for redhad certification

Oof, ACK, thanks :heart:

> ```diff
>  RUN chmod g=u /etc/passwd && \
>      find / -xdev -perm -4000 -exec chmod ug-s {} + && \
> ```

To this end, I did some checking in the base UBI images with nothing extra but `find` installed:

```console
[root@6b22c7b2ec86 /]# ls -l /etc/passwd
-rw-r--r-- 1 root root 533 Feb  7  2024 /etc/passwd
[root@6b22c7b2ec86 /]# find / -xdev -perm -4000
/usr/bin/newgrp
/usr/bin/gpasswd
/usr/bin/chage
```

Setting the permissions of `/etc/passwd` to be _group_-writable seems like it's a change in the opposite direction of ditching the sticky bit on those executables.  I'm guessing this is probably related to OpenShift and their insistence on being able to modify users at runtime by being part of the root group, which certainly doesn't fill me with warm fuzzies.  The sticky bits can be much more effectively mitigated at runtime with `--security-opt no-new-privileges`, but :shrug:

This isn't a blocker, but it's certainly a "code smell" that makes me uncomfortable (especially as this `find` will be making a copy of all those executables in this layer just to drop the sticky bits).

tianon (approved):

PR #18993, opened 2025-05-07
Update python

tianon:
Changes:

- https://github.com/docker-library/python/commit/4cf0116a: Update 3.14-rc to 3.14.0b1


yosifkit:
Build test failure:
>```console
>pip : The term 'pip' is not recognized as the name of a cmdlet, function, 
>script file, or operable program. Check the spelling of the name, or if a path 
>was included, verify that the path is correct and try again.
>```

>The installer we offer for Windows is being replaced by our new install manager, which can be installed from the Windows Store or our FTP page. [...] The traditional installer will remain available throughout the 3.14 and 3.15 releases.
>
> \- https://www.python.org/downloads/release/python-3140b1/

>Unlike the current Windows Store install, no global pip command is included (the traditional installer also does not include a global pip command, unless the options to modify PATH and to install pip are selected; the first of these is off by default).
>
> \- https://peps.python.org/pep-0773/#global-pip-command

So, the newer "install manager" will no longer have a `pip` command, but the "traditional installer" should still work because we are setting `PrependPath=1` and `Include_pip=1`. 😕 😢

yosifkit (approved):

PR #19000, opened 2025-05-09
Update Semeru release versions

jayasg12:
- Semeru 8/11/17 and 21 release versions are updated. 
- Semeru 23 is removed from list as it reached EOS.

### DIFF
--- a/library/ibm-semeru-runtimes
+++ b/library/ibm-semeru-runtimes
@@ -5,196 +5,158 @@ GitRepo: https://github.com/ibmruntimes/semeru-containers.git
 GitFetch: refs/heads/ibm
 
 #-----------------------------openj9 v8 images---------------------------------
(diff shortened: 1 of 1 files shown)


tianon (approved):

tianon:
We are again failing to rebuild thanks to the Tomcat version (as in https://github.com/docker-library/official-images/pull/17591#issuecomment-2442922629).

Please update with an explicit fallback to archive.apache.org to help prevent this issue in the future (https://github.com/docker-library/official-images/pull/17835#pullrequestreview-2414098355).  You can see an example approach in https://github.com/docker-library/cassandra/blob/1e3d5732f34ceb9e77870d0be9501515f917cc60/5.0/Dockerfile#L107-L118, if you need inspiration, or I'm happy to write an example that's more tailored to `ibm-semeru-runtimes` if you want/need it.

PR #18923, opened 2025-04-24
Update rakudo-star to 2025.04

m-dango:

PR #19011, opened 2025-05-10
Update rakudo-star to 2025.04

AntonOks:


yosifkit:
It seems that whatever was failing in https://github.com/docker-library/official-images/pull/18923 with the missing `git` tool is also assuming something more about its build context that is not true. It is assuming that it is in a `git clone` with a `.git `directory, probably to lookup version/tag or "dirty" checkout info 🤷 (at least that is what I see this behavior most often doing).

```console
+ tar xzf /tmp/tmp.eahjgi/rakudo.tar.gz '--strip-components=1' -C /tmp/tmp.eahjgi/rakudo
+ cd /tmp/tmp.eahjgi/rakudo
+ bash bin/rstar install -p /usr
fatal: not in a git directory
```

tianon:
How would you like to proceed?  Should we comment out the Alpine variant for now until there's a fix so we can at least get the non-Alpine 2025.04 built/published?  Would you rather wait for a fix?

m-dango:
I'll try and take a look in a few hours to see how to proceed, postponing the alpine image update for now may be a good idea.

PR #19063, opened 2025-05-20
Update bash

tianon:
Changes:

- https://github.com/tianon/docker-bash/commit/ff66a22: Update devel to 20250519, commit 870ad4c92b771aafe5b444792a8fa904b2d4a418


tianon:
CI failure is unrelated (it would clear with more retries, but it's really not worth the infra hammering)

PR #19068, opened 2025-05-21
Update Debian (esp. for 12.11)

tianon:


tianon:
An important note for this build (and all builds going forward) is that trixie/testing will no longer include `mips64le`/`mips64el`: https://lists.debian.org/debian-devel-announce/2025/05/msg00004.html

> The Release Team has decided to drop mips64el as a release architecture

yosifkit (approved):

PR #19071, opened 2025-05-21
Updates for the release of 25.0.0.5

wraschke:


tianon (approved):
The new script is weird, but seems harmless enough.

PR #19077, opened 2025-05-22
Update library/liquibase to 4.32.0

liquibot:


yosifkit:
The `ARG`s should stay where they were to be as close to where they are used as is possible.

tianon (approved):

PR #19085, opened 2025-05-22
2025 Q2 maintaince release of MariaDB 10.5.29, 10.6.22, 10.11.13 11.4.7

grooverdan:
Notably skips the 10.11.12/11.4.6 due to a last minute serious regression.

As Ubuntu 20.04 is EOL the 10.6 release includes a 10.6-jammy release. Per #MariaDB/mariadb-docker/issues/553 the next release on 10.6 will be jammy only.

The 10.5 is the last release of its series so this isn't getting the same treatment.


tianon (approved):

grooverdan:
Thanks @tianon

PR #19099, opened 2025-05-24
Aerospike 8.0.0.8

pvinh-spike:


tianon:
```diff
+telemetry &>/dev/null &
```

Unfortunately, starting an unsupervised and unmonitored process in the background is a strict no-go.  If this behavior is important/desirable, I would honestly suggest adding it to the Aerospike process/code directly instead of doing a hacky `curl` in the entrypoint of the container image.

I would also encourage you to strongly consider making any behavior like this opt-out instead of opt-in (https://research.swtch.com/telemetry-opt-in is useful/interesting reading about this and how it went in the Go ecosystem).

pvinh-spike:
Thanks for the info.

PR #19132, opened 2025-05-29
Add JRuby 10.0-dev and 9.4-dev images

headius:
This adds development snapshot images for JRuby 10.0.x and 9.4.x. They will be updated periodically as needed.

If there's a better pattern to follow for pushing development images or snapshots or nightly builds to Hub, please link me to appropriate docs.

### DIFF
--- a/library/jruby
+++ b/library/jruby
@@ -3,7 +3,7 @@ Maintainers: JRuby Admin <admin@jruby.org> (@jruby),
              Thomas E Enebo <tom.enebo@gmail.com> (@enebo)
 GitRepo: https://github.com/jruby/docker-jruby.git
 GitFetch: refs/heads/master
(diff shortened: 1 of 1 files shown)


tianon:
> If there's a better pattern to follow for pushing development images or snapshots or nightly builds to Hub, please link me to appropriate docs.

I think https://github.com/docker-library/faq#can-i-use-a-bot-to-make-my-image-update-prs is as close as we have to official guidance on that.  You might also consider putting snapshots/nightly builds into a separate repository (especially if you want to publish them more often / more proactively or if you want to avoid cluttering an already long/complex tag list in the official image with development tags).

This looks OK to me, but I'll wait to merge so you can decide whether you still want to go ahead with it. :+1:

tianon (approved):

PR #19134, opened 2025-05-29
Ensure common utilities are consistently available in all images

keeganwitt:


yosifkit:
I think this is all ok, but it does make the alpine images a bit bigger:
```console
gradle           8.14.1-jdk21-alpine        206ad621610e   2 seconds ago   598MB

gradle           8.14.1-jdk21-alpine        d78f9860ec40   2 seconds ago   623MB
```

tianon (approved):

PR #19161, opened 2025-05-30
Update hylang

tianon:
Changes:

- https://github.com/hylang/docker-hylang/commit/5c986c2: Add Alpine 3.22 (remove Alpine 3.20)
- https://github.com/hylang/docker-hylang/commit/7d3bb23: Add Alpine 3.22


tianon:
We'll have to restart CI in a while after we have builds. :joy:

PR #19179, opened 2025-06-03
fix(test): allow unavailable jit on postgres 18

reneleonhardt:
Postgres 18 (Debian) doesn't install the JIT anymore to reduce image size.


yosifkit:
Since we are going to keep JIT by default, we don't currently need this, so I'll close for now.

PR #19167, opened 2025-06-01
Update Ubuntu 2025-06-01

ROCKsBot:
Update Ubuntu to the latest published OCI tarballs as of 2025-06-01.

PR #19184, opened 2025-06-03
Update ELK images for 9.0.2

sfendourakis:


jbudz:
There's a few bugs in this due to how we updated the release for 9.0.1.  We're looking into it, I'll circle back with the team and update when the fixes are ready.

tianon (approved):

tianon:
CI failure is https://github.com/docker-library/official-images/pull/19167#issuecomment-2931611620 on versions that aren't being updated here (we won't block on this for now, but we might do so in the future so please do prioritize getting off Ubuntu 20.04, one way or another).

PR #19189, opened 2025-06-03
Disable neurodebian:noble builds for now

tianon:
They're still using a DSA1024 key, which is now explicitly rejected by APT. 😞

See also:

- https://github.com/docker-library/official-images/pull/18548
- https://github.com/neurodebian/neurodebian/issues/97#issuecomment-2936348850


tianon:
wah wah, looks like I should remove focal while I'm here lol

yosifkit (approved):

PR #19199, opened 2025-06-04
Remove notary tags (EOL base image)

tianon:
See also:
- https://github.com/docker/notary-official-images/pull/38


tianon:
We should probably also add a deprecation notice to the image. :thinking:

PR #19211, opened 2025-06-05
[ros] temporarily enable ROS EOL builds to renew APT GPG key

mikaelarguedas:
Similar to https://github.com/docker-library/official-images/pull/15809 https://github.com/docker-library/official-images/pull/15809#issuecomment-1839829310

The snapshots.ros.org APT repo gpg key expired. It has now been renewed: https://discourse.ros.org/t/ros-signing-key-migration-guide/43937/41

Reenabling all ROS image builds to fetch the updated GPG key

---

One exception is ROS noetic that is going EOL and a separate PR to modify that image and any tag building bridge between ROS noetic and another ROS distro, tracked here https://github.com/osrf/docker_images/issues/808

### DIFF
--- a/library/ros
+++ b/library/ros
@@ -2,30 +2,386 @@ Maintainers: Tully Foote <tfoote+buildfarm@osrfoundation.org> (@tfoote)
 GitRepo: https://github.com/osrf/docker_images.git
 
 ################################################################################
+# Release: indigo
+
+########################################
+# Distro: ubuntu:trusty
+
+Tags: indigo-ros-core, indigo-ros-core-trusty
+Architectures: amd64, arm32v7
+GitCommit: 51ba1c72c513b62e8825c3df015ca4d926ac688c
+Directory: ros/indigo/ubuntu/trusty/ros-core
+
+Tags: indigo-ros-base, indigo-ros-base-trusty, indigo
+Architectures: amd64, arm32v7
+GitCommit: 20061b005b245d1b7e23626afd0ea2c39de9db49
+Directory: ros/indigo/ubuntu/trusty/ros-base
+
+Tags: indigo-robot, indigo-robot-trusty
+Architectures: amd64, arm32v7
+GitCommit: 20061b005b245d1b7e23626afd0ea2c39de9db49
+Directory: ros/indigo/ubuntu/trusty/robot
+
+Tags: indigo-perception, indigo-perception-trusty
+Architectures: amd64, arm32v7
+GitCommit: 20061b005b245d1b7e23626afd0ea2c39de9db49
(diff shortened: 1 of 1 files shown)


tianon:
Doesn't EOL mean users shouldn't be using it anymore? (so IMO builds failing because of this is probably a _good_ thing, as it helps users notice that their base image is EOL and they ought to upgrade :thinking:)

tfoote:
@mikaelarguedas Thanks for putting this together. 

@tianon Yes we do want people to move forward. And we have clearly indicated that there will be no new updates. However there are cases where it's highly valuable to have reproducibility of old results. This is especially seen in our academic communities, where you want to be able to reproduce old results. We transition the packages into an alternative not high bandwidth(ala CDN) location snapshots.ros.org from our main repository pakages.ros.org  in the same way as Ubuntu hosts old-releases.ubuntu.com like packages.ubuntu.com. 

We can see the progression of people moving between distros here: https://metrics.ros.org/packages_rosdistro.html There is a bit of a long tail of people using the EOL'd distros but most have moved over. 
image So people know to move over but often have a reason for not moving. 


It's a tradeoff of doing the minimal processing (that is essential for us to do with extra permissions) that will save people from spending a lot of work to rebuild the legacy images themselves. if we can do a single rebuild to get the gpg key updated, it will save many people the effort to take the images fork them, fix them, likely redistribute them. And often it will end up with an individual who will start sharing them through unofficial channels which don't have a good trust mechanism and consequently leave our community at risk of exploitation.  

So if we do the one rebuild, with the new gpg keys, things will remain reproducible and available from a trusted source. And we'll turn the builds back off once we have the rebuilt versions with the minimal gpg key only change.

yosifkit (approved):
So, it looks like these are all the exact Dockerfiles (and build contexts) from the previous time (https://github.com/docker-library/official-images/pull/15809) along with any new EOL versions. Why that is important is that since that PR we have moved to our newer build system\* which uses a hash generated from the parent images' hashes and the hash of the build context to build and tag the image into a staging repo on Docker Hub as a cache to know what is built. So, while the previous builds were not recorded, these will be and when this pull request is repeated in 5 years to update the gpg key for EOL versions, they will not get rebuilt. The hash of the image will already exist in the staging repo and be used as-is because neither the parent images nor the build context changed.

What this means is that the next GPG key update rebuild will require actual changes to the Dockerfiles to update the key (like embedding a checksum of the key or new key ID). As discussed previously, we would not accept contentless changes like `echo "2025-06-06"`.

---

\* build process described [here](https://github.com/docker-library/faq/blob/e4314f481219998c3917731e40c6f17f9122ee98/README.md#an-images-source-changed-in-git-now-what)

---

LGTM

PR #19221, opened 2025-06-08
[ros] migrate noetic to snapshots and disable the one-off rebuild images

mikaelarguedas:
Follow-up of https://github.com/docker-library/official-images/pull/19211

This PR:
- Disable all images that were enabled in #19211 for the one-off rebuild
- Update ROS Noetic to target snapshots repository as Noetic went EOL along its base image ubuntu:focal
- Update the foxy and galactic ROS1/ROS2 bridge to use the final version of ROS Noetic

---

Once merged ans rebuilt, A PR will be submitted to disable the builds for all ROS EOL Distributions (Noetic, Foxy and Galactic)

PR #19233, opened 2025-06-10
[ros] Retire Focal based images

mikaelarguedas:
Now that the images have been rebuilt with the updated GPG key, all Ubuntu Focal based ROS images can be retired

Follow-up of https://github.com/docker-library/official-images/pull/19221#issuecomment-2957046040


mikaelarguedas:
ROS Rolling build failure independant of this PR and adressed in https://github.com/docker-library/official-images/pull/19234

PR #19235, opened 2025-06-10
Add ubi10{,-minimal} to external-pins

grooverdan:
With UBI 10 released on https://hub.docker.com/u/redhat though I'd add these if that's ok?


yosifkit:
"Apply Labels" GitHub action is failing because of an undocumented permissions change with creating new labels: https://github.com/orgs/community/discussions/156181.

Added workaround in https://github.com/docker-library/official-images/pull/19241

grooverdan:
I've repushed to re-trigger after #19241

yosifkit (approved):
Seems fine and matches the digests I get locally when running `./.external-pins/update.sh redhat/ubi10{,-minimal}:latest`

grooverdan:
yep, should of mentioned that's the way I generated it. You do automate things very well.

tianon:
I'd love to move these off pins (and have them be proper images), but perhaps that ship has sailed. :disappointed: :heart:

grooverdan:
> I'd love to move these off pins (and have them be proper images), but perhaps that ship has sailed. 😞 ❤️

It would just mean for images based on those that we'd change the `FROM` line right? If you still want to move, it doesn't sound like a big imposition to change. Its only going to impact current images anyway, or have a I missed something.

tianon:
Yeah, that's correct -- the hard part would be reconvincing the appropriate folks at Red Hat to maintain it. :innocent: :heart:

PR #19245, opened 2025-06-10
Release 11.8 GA and 12.0RC

grooverdan:
Consequently 11.7 is EOL.

Because of Ubuntu 20.04 being EOL these images
where removed. Note ably there stil is 10.6-jammy
but leaving it to Docker Official Images to have
10.6 previous tag pointing to the older focal based image.

Next release will replace 10.6 with the jammy based image in a few months.

Closes #MariaDB/mariadb-docker/issues/647


grooverdan:
Thanks @yosifkit !

PR #19248, opened 2025-06-11
hitch: upgrade to bookworm

gquintard:


yosifkit:
Debian Bookworm will have `ppc64le` and `s390x` at least until it transitions to "LTS" in June 2026. It is Bullseye that no longer has those architectures. You are welcome to add them back now that it is from Bookworm.

PR #19258, opened 2025-06-12
Update couchbase-server

udkyo:
This change updates our images which are currently based on ubuntu:20.04 to 22.04/24.04.


yosifkit (approved):

yosifkit:
Same as https://github.com/docker-library/official-images/pull/16661#issuecomment-2076049220
> reviewer's note: runit install block is identical to the one in https://github.com/docker-library/official-images/pull/16433 :+1: :heart:
>
>(FWIW, `apt-get clean` is a no-op in both the `debian` and `ubuntu` base images because they already do that automatically on every `apt-get install`, but it's harmless to include)

PR #19266, opened 2025-06-13
library/perl: 5.41.12 -> 5.41.13

zakame:
- http://nntp.perl.org/group/perl.perl5.porters/269950
- https://github.com/Perl/docker-perl/pull/185


zakame:
@tianon this should include your fix https://github.com/Perl/docker-perl/pull/184 🙇

zakame:
Thanks @yosifkit ! 🎉 🚀

PR #19273, opened 2025-06-16
Update to Storm 2.8.1 and adds JRE21 image

rzo1:
Update to the latest release of Storm + adds JRE21 runtime flavour.


library/storm
```diff
@@ -4,10 +4,14 @@ Maintainers: The Apache Storm Project <dev@storm.apache.org> (@asfbot),
...
-Directory: 2.8.0
+Tags: 2.8.1, 2.8, latest
+GitCommit: b8be6ef59263919abd3eed95207198d2aca28518
+Directory: 2.8.1
 
-Tags: 2.8.0-jre17, 2.8-jre17
-GitCommit: fefacef2f400d8f437675b11e08bc6098a91c188
+Tags: 2.8.1-jre17, 2.8-jre17
+GitCommit: b8be6ef59263919abd3eed95207198d2aca28518
 Directory: 2.8.0
```
tianon, line 13: I think this was probably supposed to be updated to 2.8.1 also, right?  Also just FYI these two blocks could be combined (the "generic" aliases can be listed next to the more specific ones, `Tags: 2.8.1-jre17, 2.8-jre17, 2.8.1, 2.8, latest` for example).

Also, I don't think I see a 2.8.1 bump in the `Dockerfile` diff -- shouldn't at least https://github.com/apache/storm-docker/blob/b8be6ef59263919abd3eed95207198d2aca28518/2.8.1/Dockerfile#L36 have been updated to 2.8.1?

rzo1: Looks like I forgot to push the 2.8.1 dist change. Thanks for the pointer. Will fix shortly!

rzo1: Should be fixed now.

PR #19318, opened 2025-06-24
Update ELK images for 9.0.3, 8.18.3, 8.17.8 and 7.17.29

StamatisKourk:


tianon:
- elasticsearch:7.17.29 (FROM ubuntu:20.04) -- completely unsupported base!
 - elasticsearch:8.17.8 (FROM ubuntu:20.04) -- completely unsupported base!
 - kibana:7.17.29 (FROM ubuntu:20.04) -- completely unsupported base!
 - logstash:7.17.29 (FROM ubuntu:20.04) -- completely unsupported base!
 - logstash:8.17.8 (FROM ubuntu:20.04) -- completely unsupported base!
 - logstash:8.18.3 (FROM ubuntu:20.04) -- completely unsupported base!

https://github.com/docker-library/official-images/pull/19167#issuecomment-2931611620

> Please either drop the affected tags from the supported list (by removing them from your `library/` file) or update them to a supported release.

StamatisKourk:
@tianon We have some CI failures about logstash 9.0.3, elasticsearch 8.17.8, and kibana 7.17.28. These seem to be coming from your end but can you please advise?

Also regarding upgrading the base image to ubuntu 24.04, we have done that for all tags being tampered in the PR. However, the 7.17.28 ELK images which are already merged, are still being scanned. Given that this is an older version and this is hard for us to upgrade, could we exclude this like we have done in the past? If not, will removing the 7.17.x tag entirely from the library files work? I guess that would remove 7.17.28 from the list of "supported" docker images in docker registry, but it would still be reachable, like all other 7.17.x tags are today (eg. docker pull logstash:7.17.23 still works)

tianon:
Yeah, I can live with that, especially on the promise that resolving this for the 7.x versions is in-progress. :heart: :+1:

---

Looking over the diff, the only place I've got some concerns are in `logstash:9.0.3`:

```diff
-RUN \
+RUN for iter in {1..10}; do \
+  microdnf update -y && \
```

I'm not sure I understand why this has come back in. :sweat_smile:  Are the package installation servers _consistently_ flaky enough to be worth embedding this fiddly retry logic directly in the `Dockerfile`?

```diff
-  arch="$(rpm --query --queryformat='%{ARCH}' rpm)" && \
-  curl -f -Lo logstash.tar.gz "https://artifacts.elastic.co/downloads/logstash/logstash-9.0.2-linux-${arch}.tar.gz" && \
-  tar -zxf logstash.tar.gz -C /usr/share && \
-  rm logstash.tar.gz && \
-  mv /usr/share/logstash-9.0.2 /usr/share/logstash && \
+  curl -Lo - https://artifacts.elastic.co/downloads/logstash/logstash-9.0.3-linux-x86_64.tar.gz | \
+  tar zxf - -C /usr/share && \
+  mv /usr/share/logstash-9.0.3 /usr/share/logstash && \
```

Does 9.0.3 only support x86_64?  (the `library/logstash` entry for it includes `arm64v8` :eyes:)

StamatisKourk:
@tianon We handled the logstash 9.0.3 reverted changes. Indeed they were reverted by our automation and we added them back in. The 2 check failing tags for 7.17.28 kibana and ES, which we are not edited in this PR, should be updated in the future or be removed entirely like logstash 7.17.28.

PR #19334, opened 2025-06-26
version 10.1.0 for TomEE

scriptmonkey:


yosifkit:
With the addition of semicolons in the `RUN` line, I think this needs a `set -e` to make sure that a failing expression stops instead of continues.

tianon:
(high-five :hand: https://github.com/tomitribe/docker-tomee/pull/98#discussion_r2169309024 :joy: :heart:)

PR #19075, opened 2025-05-21
feat: Update influxdb 2.7.11 -> 2.7.12

devanbenz:

PR #19345, opened 2025-06-27
feat: Update alpine tag to 2.7.12

devanbenz:


yosifkit (approved):
LGTM

(tag change was missed in https://github.com/docker-library/official-images/pull/19075)

yosifkit:
If the historical contents of the `influxdb:2.7.11-alpine` image is important and you want to restore it to containing  InfluxDB `2.7.11` instead of its current, incorrect `2.7.12`, then adding a temporary back-fill is ok with us.

Basically, something like this. (commit reference from https://github.com/docker-library/official-images/pull/19075)
```yaml
Tags: 2.7.11-alpine
Architectures: amd64, arm64v8
Directory: influxdb/2.7/alpine
GitCommit: a757de46ff8c42b19e05692e592c153bd52d2fc6
```

PR #19348, opened 2025-06-29
Update Eggdrop to 1.10.1rc1

vanosg:


tianon:
```diff
+  && gpg --keyserver keyserver.ubuntu.com --recv-key E01C240484DE7DBE190FE141E7667DE1D1A39AFF \
```

This invocation of `gpg` should _also_ include `--batch` (`gpg --batch --keyserver ...`), but that's not worth blocking over. :+1:

(the `command -v gpgconf` line can also be dropped completely, but that's also not critical)

tianon (approved):

vanosg:
Thank you for the feedback- will update for the next go!

PR #19357, opened 2025-06-30
maven: java 24 debian

rufusnufus:


yosifkit (approved):

rufusnufus:
@yosifkit Thanks! Would you mind merging PR?

tianon:
I've double checked that the versions match and that all the tags appear to be correct :+1:

PR #19367, opened 2025-07-01
Releasing Neo4j 2025.06.0

ericsporre:
Thank you!


jennyowen:
@yosifkit It looks like the arm builds of this image [are missing from dockerhub](https://hub.docker.com/_/neo4j/tags?name=2025.06)
image

Is this something that resolves itself over time?

tianon:
It appears they're all up now -- we did have a pretty full queue thanks to a Debian rebuild, so yeah, it was probably just a timing thing. :+1: :heart:

PR #19378, opened 2025-07-02
Kapacitor v1.8.0

srebhan:
Thank you!

### DIFF
--- a/library/kapacitor
+++ b/library/kapacitor
@@ -2,14 +2,7 @@ Maintainers: Brandon Pfeifer <bpfeifer@influxdata.com> (@bnpfeife),
              Sven Rebhan <srebhan@influxdata.com> (@srebhan),
              Maya Strandboge <mstrandboge@influxdata.com> (@mstrandboge)
 GitRepo: https://github.com/influxdata/influxdata-docker
-GitCommit: 8b23d6db77384c3fbecc52e880b83c4600fdffc8
-
-Tags: 1.6, 1.6.6
-Architectures: amd64, arm64v8
-Directory: kapacitor/1.6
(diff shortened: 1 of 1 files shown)


tianon:
Two questions:

- did you add `su-exec` on purpose?  I don't see it used anywhere obvious, and it's got a critical parser error in 0.2 that's been waiting for a new release for several years now, so I'd actively recommend against using it :see_no_evil:

- did you want to move the generic `latest` and `alpine` aliases to the new version?

srebhan:
@tianon thanks for the thorough review! 

Version 1.7 was left as the latest because 1.8 removes a deprecated function and breaks the (long abandoned) python 2 support for UDFs. This functionality was deprecated in v1.7.7 to give users some time to migrate but at the same time release the new version to fix some security issues. Clarifying internally if  we should push `latest` now...

For `su-exec` I'm clarifying internally if we really need it or if we can get away with setting the privileges...

tianon:
To clarify my `su-exec` thoughts further - it doesn't make sense for _users_ to invoke `su-exec` manually (they should just use `--user` or similar on their container runtime directly instead), so if it's not used by the scripts or tool in the image, it shouldn't be part of the image (and in the case of `su-exec`, it really shouldn't be part of the image either way due to the aforementioned critical issue).

srebhan:
@tianon I completely understand your thoughts and I fully agree. We are currently working on replacing `su-exec` in the latest images (not only this one but also on others maintained by us). Hope to sort out both issues (the su-exec one and the latest tagging) by tomorrow morning UTC...

tianon (approved):

PR #19385, opened 2025-07-03
feat(rust): Add trixie images

Muscraft:


tianon:
Normally I'd suggest waiting until (closer to) GA to recommend supporting Trixie, especially given the added load, but this is just downloading existing binaries into a new base and Rust only currently supports one version, so the impact is low and it's not default yet, so seems sane to me. :+1: :heart:

tianon (approved):

PR #19403, opened 2025-07-07
Add XWiki 17.4.2 as intermediate LTS version

michitux:
XWiki is introducing a second, intermediate LTS version per year, the latest LTS version can be tracked with the `lts-latest` tags.


michitux:
Looks like the naming `lts-latest` isn't allowed. We'll discuss an alternative naming strategy and then mark the PR as ready for review again.

michitux:
I've removed the labels that contained `lts-latest`, the intermediate LTS version has only labels based on the version number for now until we agree on the naming.

PR #19432, opened 2025-07-09
Update UBI pins

yosifkit:


tianon:
Odd that there's no 9-minimal here :thinking:

tianon (approved):

PR #19439, opened 2025-07-10
Release Neo4j 5.26.9 and 2025.06.2

ericsporre:
Thank you!

### DIFF
--- a/library/neo4j
+++ b/library/neo4j
@@ -3,47 +3,47 @@ Maintainers: Jenny Owen <jenny.owen@neo4j.com> (@jennyowen),
              Tobias Ericsson <tobias.ericsson@neo4j.com> (@neo-tobias)
 GitRepo: https://github.com/neo4j/docker-neo4j-publish.git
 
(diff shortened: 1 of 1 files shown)


yosifkit:
It seems like `neo4j:ubi9` image has the wrong sha256sum. It is downloading the same file as `neo4j:latest` but has a different sum and so fails the test build.
```console
+ curl --fail --silent --show-error --location --remote-name https://dist.neo4j.org/neo4j-community-2025.06.1-unix.tar.gz
+ echo 'b1af4efda350b93500963e330b4df12f0adee1c6ca0137f7b164aa06c8826506  neo4j-community-2025.06.1-unix.tar.gz'
+ sha256sum -c --strict --quiet
sha256sum: WARNING: 1 computed checksum did NOT match
neo4j-community-2025.06.1-unix.tar.gz: FAILED
```

ericsporre:
Hello! Sorry about that hiccup! We managed to sneak in a .2 release which also fixes the sha problem. Thanks!

PR #19449, opened 2025-07-11
release 2.8.10

alcohol:


yosifkit:
Did you mean to remove the `2` and `lts` image tags?
```diff
-composer:2
-composer:lts
```

alcohol:
@yosifkit Yes, the removal of the `lts` tag is intentional. The `2` was not, let me think about that for a moment.

alcohol:
Updated.

tianon (approved):

PR #19467, opened 2025-07-14
Updated nginx debian images to pick up key fixes.

thresheek:
We've introduced two new keys back in 2024 to be able to migrate off the older key.  Last week we actually started using one of the newer keys to sign the packages and metadata, however a bug in the key export logic, meant that only the first key was exported to the keyring, breaking the Dockerfile build.

With
https://github.com/nginx/docker-nginx/commit/0b49b8b12fd214b633114ac16d2dfd65d45ff160 now merged, let's update the references to the fixed base Dockerfiles.


tianon:
Oh, on the topic, I'd suggest adding `--batch` to all `gpg` invocations (the `--batch` flag puts it into "API mode" for lack of a better explanation).  I don't think we need to block this on that addition though, unless you really want to. :+1:

tianon (approved):

thresheek:
Thanks! Let's keep the `--batch` change for the future.

PR #19506, opened 2025-07-21
Update UBI pins

yosifkit:


tianon:
Only UBI8? :thinking: 

Let's give this a little time and see if at least 8-minimal pops up.

tianon (approved):
:shrug: I give up

PR #19514, opened 2025-07-21
Update Debian

tianon:


tianon:
Hmm, that test failure on `oldstable-backports` isn't ideal -- it appears that http://ftp.debian.org/debian/dists/ still has `bullseye-backports` but the `oldstable-backports` symlink has disappeared.  That being said, `bullseye-backports` is extremely EOL now (backports are only officially supported for one year following the next stable release; https://backports.debian.org/Instructions/), so I think we're probably OK to go ahead in spite of that.

I should separately look to remove EOL backports tags from the "supported" list proactively (instead of relying on the archive to remove the folders before I remove the tags), but that's going to take slightly more refactoring than I've got spoons for right at the moment.  (I've also now filed https://bugs.debian.org/1109683 as something that would be really helpful towards that end.)

PR #19539, opened 2025-07-26
feat!: drop support for building SATOSA images on platforms other than linux/amd64 and linux/arm64

xenophonf:
I am unable to determine why [cryptography](https://cryptography.io/) won't build on other platforms.

Closes IdentityPython/satosa-docker#15.


tianon (approved):
Nice, thank you!

PR #19545, opened 2025-07-27
Cleanup legacy test cases

J0WI:


tianon (approved):
Nice!  I spot checked all the version relations here and they all appear correct (the only one I wasn't totally sure of off-hand was the Gem version, but our oldest version is currently 3.4.19 in Ruby 3.2, which is plenty new). :heart:

PR #19553, opened 2025-07-28
Update YOURLS

LeoColomb:
This changeset clarifies the HTTP server virtual host, notably service ports.
It fixes some security context enforcements, for example when used with security admission controllers.

### DIFF
--- a/library/yourls
+++ b/library/yourls
@@ -1,21 +1,21 @@
-# this file is generated via https://github.com/YOURLS/images/blob/d306b06d6b3e94f1441c65385395f7a384a46101/bin/generate-stackbrew-library.sh
+# this file is generated via https://github.com/YOURLS/images/blob/89dfcf9b332685465708b98d3151d39f4f6340a0/bin/generate-stackbrew-library.sh
 
 Maintainers: YOURLS <yourls@yourls.org> (@YOURLS),
              Léo Colombaro <git@colombaro.fr> (@LeoColomb)
-GitRepo: https://github.com/YOURLS/images.git
+GitRepo: https://github.com/YOURLS/containers.git
 GitFetch: refs/heads/dist
 
-Tags: 1.10.1-apache, 1.10-apache, 1-apache, apache, 1.10.1, 1.10, 1, latest
+Tags: 1.10.2-apache, 1.10-apache, 1-apache, apache, 1.10.2, 1.10, 1, latest
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
-GitCommit: d306b06d6b3e94f1441c65385395f7a384a46101
+GitCommit: 89dfcf9b332685465708b98d3151d39f4f6340a0
 Directory: apache
 
-Tags: 1.10.1-fpm, 1.10-fpm, 1-fpm, fpm
+Tags: 1.10.2-fpm, 1.10-fpm, 1-fpm, fpm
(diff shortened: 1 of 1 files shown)


tianon (dismissed):
I'll admit that I really don't understand the purpose/goals/issue-being-fixed by the vhost and port changes here; it's a little weird trying to find the _actual_ change that caused it given the interesting commit-based deployment structure of the YOURLS/containers repo, but I did make it to https://github.com/YOURLS/containers/pull/336 eventually and it's also light on details. :see_no_evil:

I'm happy to merge this as-is if you'd like, but if you'd like us to review/opine on the problem being solved (and whether the port number breaking change is warranted/necessary), I'll need more details on what you're trying to fix. :sweat_smile:

However, to be very clear, the changes look fine and I don't _need_ to understand them if you'd rather just go ahead. :+1:

LeoColomb:
I'd be happy to go ahead, but I can still for sure give some more context (and apologies for it not being clear 😅).

Ports 80 and 443 are outside userland Linux kernel boundaries.
If the container is run as a rootless process in a rather strict environment, it cannot bind these ports.
Kernel capabilities can be requested to gain access (for instance, the [`CAP_NET_BIND_SERVICE` capability](https://man7.org/linux/man-pages/man7/capabilities.7.html#:~:text=CAP_NET_BIND_SERVICE)), but additional capabilities are considered security weaknesses.

Such strict environments become more and more common, both from container engines (Podman, for example) and from orchestrators (OpenShift is super strict about this).

Ports 8080 and 8443 are fairly common for such apps nowadays, and this change just makes everything simpler for everyone with virtually zero impact on usability.
I'm aware the change is relatively breaking from a continuity POV, but we're ready to face it (hopefully 😅).

Edit: I've clarified the change on the release https://github.com/YOURLS/containers/releases/tag/4.0.0

tianon:
Ah, that tracks -- for what it's worth, https://github.com/moby/moby/pull/41030 makes that mostly a moot point for Docker usage, although I believe Kubernetes is still trying to decide how/where/whether to set `net.ip_unprivileged_port_start` by default (https://github.com/kubernetes/kubernetes/issues/102612 -- at least `containerd` 2.0 enables this by default too, so it's already moot there as well and should be in any k8s deployment using c8d 2.0+).

See also https://www.kernel.org/doc/Documentation/networking/ip-sysctl.txt#:~:text=ports.%0A%0A%09Default%3A%20Empty-,ip_unprivileged_port_start,-%2D%20INTEGER%0A%09This%20is for the explicit kernel documentation of this option (that again, the trend is for container runtimes to set to a reasonable number like 0):

```
ip_unprivileged_port_start - INTEGER
	This is a per-namespace sysctl.  It defines the first
	unprivileged port in the network namespace.  Privileged ports
	require root or CAP_NET_BIND_SERVICE in order to bind to them.
	To disable all privileged ports, set this to 0.  They must not
	overlap with the ip_local_port_range.
```

LeoColomb:
Thanks for these links; glad to read it's moving! 👍
That said, even if this change is almost years late now, I'd prefer to go for it.

zjduck0728:
蟹蟹你！我收到啦。

LeoColomb:
Actually, let's keep it on hold for a few days. I'll coordinate a version bump on upstream app so the change can be more visible. 
I'll keep in touch when ready.

LeoColomb:
Here we go, @tianon 😊
With a version bump, so it's a bit less disruptive.
Thanks!

tianon (approved):

PR #19515, opened 2025-07-22
Update ELK images for 9.0.4, 8.18.4 and 8.17.9

ksevasilyeva:

PR #19183, opened 2025-06-03
Update ELK images for 8.17.7

sfendourakis:

PR #19558, opened 2025-07-29
Update ELK images for 9.1.0 and 8.19.0

navyau09:
Update ELK images for 9.1.0 and 8.19.0

### DIFF
--- a/library/elasticsearch
+++ b/library/elasticsearch
@@ -18,7 +18,17 @@ Architectures: amd64, arm64v8
 GitFetch: refs/heads/8.18
 GitCommit: 879005c2065f0c8d78fa6d978ce161fd38734600
 
+Tags: 8.19.0
+Architectures: amd64, arm64v8
+GitFetch: refs/heads/8.19
+GitCommit: a01d05345bb7ef65ee8fe2ca4713cc0efa98cf5b
+
 Tags: 9.0.4
 Architectures: amd64, arm64v8
 GitFetch: refs/heads/9.0
 GitCommit: 8052f28559ff55a203e9ce99ecb48251bbc67079
+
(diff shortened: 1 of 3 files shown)


navyau09:
@jbudz Could you please take a look on the failing test for 8.19.0? Thanks

tianon:
- elasticsearch:7.17.28 (FROM ubuntu:20.04) -- completely unsupported base!
 - kibana:7.17.28 (FROM ubuntu:20.04) -- completely unsupported base!

It's been almost two months now since https://github.com/docker-library/official-images/pull/19184#issuecomment-2945374539 (https://github.com/docker-library/official-images/pull/19167#issuecomment-2931611620, https://github.com/docker-library/official-images/pull/19183#issuecomment-2936541454, https://github.com/docker-library/official-images/pull/19318#issuecomment-3001306868, https://github.com/docker-library/official-images/pull/19515#issuecomment-3103632815).  Can we get a status update on the "get 7.x off Ubuntu 20.04" efforts? (or finally commit and remove it from the library file / supported list)

tianon:
```diff
+  curl -f --retry 8 -S -L -O "https://github.com/krallin/tini/releases/download/${TINI_VERSION}/${TINI_BIN}.sha256sum" ; \
```

This is odd, given it's unused and the checksum is embedded above it and verified below it, but it's also pretty harmless.

tianon (approved):

tianon:
We should also remove any minor versions that are EOL/unsupported, but that can happen separately.  Looking at https://endoflife.date/elasticsearch it appears only 9.x, 8.x, and 7.x are the supported granularities, but y'all admittedly would know much better than I or even endoflife.date. :smile:

PR #19562, opened 2025-07-29
Update UBI pins

yosifkit:


tianon:
_Everything_ except `ubi10` (non-minimal)?  sus.

tianon (dismissed):

tianon:
(Approving but might wait to merge to see if we pick up a ubi10.)

yosifkit:
And the patience is rewarded (or the timestamp says that it's been there all along) https://github.com/docker-library/official-images/commit/57c3555fb4f2d5a475e73baf9e5e8c1a385352f1

tianon (approved):

PR #19567, opened 2025-07-30
Update to Flink 2.1.0

lsyldliu:
This is the official release of Apache Flink 2.1.0:
- Add new released 2.1.0.
- Remove unsupported 1.19.3.

### DIFF
--- a/library/flink
+++ b/library/flink
@@ -3,17 +3,32 @@
 Maintainers: The Apache Flink Project <dev@flink.apache.org> (@ApacheFlink)
 GitRepo: https://github.com/apache/flink-docker.git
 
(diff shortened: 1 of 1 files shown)


library/flink
```diff
@@ -3,17 +3,32 @@
 Maintainers: The Apache Flink Project <dev@flink.apache.org> (@ApacheFlink)
```
tianon, line 3: As a note to take back for a future PR, `@ApacheFlink` here is supposed to be a GitHub handle, and this isn't -- it would be _ideal_ if we could get a list of human beings that we can actually @-notify on PRs here, but at the very least this should be updated to be a real GitHub handle.

tianon (approved):

PR #19570, opened 2025-07-30
SonarQube latest editions set to 2025.3

carminevassallo:
Dear team,

We are aware of a licensing issue with SonarQube Server 2025.4, which was released yesterday. So we are proposing this PR to make sure that the latest edition tags (e.g., developer, enterprise, datacenter-app, datacenter-search) point to 2025.3.1 and that those are listed as the supported versions on the page.

Could you please treat it with urgency? Thanks for your help!

### DIFF
--- a/library/sonarqube
+++ b/library/sonarqube
@@ -4,24 +4,24 @@ GitRepo: https://github.com/SonarSource/docker-sonarqube.git
 Architectures: amd64, arm64v8
 Builder: buildkit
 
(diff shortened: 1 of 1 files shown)


tianon (approved):

carminevassallo:
Thanks @tianon for the fast reaction! How much time does it take to get the corresponding dockerhub updated?

tianon:
It's updated. :+1:

PR #19542, opened 2025-07-27
eclipse-temurin: add ubi10-minimal and removed deprecated ubuntu focal images

YanivKunda:
Adding ubi10-minimal images for eclipse-temurin -

The external-pins for the the new base images have been added:
https://github.com/docker-library/official-images/pull/19235
And the Dockerfiles for all versions have been merged:
https://github.com/adoptium/containers/pull/770

Removing deprecated ubuntu focal images in https://github.com/adoptium/containers/pull/772

PR #19579, opened 2025-08-01
eclipse-temurin: July PSU updates

sxa:
The full set of platforms which are included in our containers have now been released, so we need to get the images created and published to dockerhub for the new release.

@gdams A second pair of eyeballs on this as more than just a rubber stamp would be appreciated as we have a few changes of levels in here and the removal of focal/addition of alpine 3.22 makes this less trivial than it usually is.

### DIFF
--- a/library/eclipse-temurin
+++ b/library/eclipse-temurin
@@ -7,610 +7,681 @@ GitFetch: refs/heads/main
 Builder: buildkit
 
 #------------------------------v8 images---------------------------------
-Tags: 8u452-b09-jdk-alpine-3.20, 8-jdk-alpine-3.20, 8-alpine-3.20
+Tags: 8u462-b08-jdk-alpine-3.20, 8-jdk-alpine-3.20, 8-alpine-3.20
 Architectures: amd64
-GitCommit: 4890fb4638e68c037c00c4a28c7587e0b5c96fcf
+GitCommit: fc54f27893bb7c1ffb1d7eb82f2d22d7605e57bc
 Directory: 8/jdk/alpine/3.20
 
-Tags: 8u452-b09-jdk-alpine-3.21, 8-jdk-alpine-3.21, 8-alpine-3.21, 8u452-b09-jdk-alpine, 8-jdk-alpine, 8-alpine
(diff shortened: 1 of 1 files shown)


gdams:
Looks good to me!

sxa:
@gdams Do you know what's going on with that long comment that's causing the check to fail? Looks like an error code from something has got in here

kuanyili:
My understanding is that the bot failed to post a diff log comment here in the PR because the diff log is too large.

`eclipse-temurin` is currently the largest definition file in the library thus exceeded the limit.

The PR could actually be fine though.

sxa:
> My understanding is that the bot failed to post a diff log comment here in the PR because the diff log is too large.
> 
> `eclipse-temurin` is currently the largest definition file in the library thus exceeded the limit.
> 
> The PR could actually be fine though.

Thanks. Hopefully it's just that and this shouldn't block. I'll tag @tianon too for confirmation since that it's "just" that. If this is likely to recur in the future for the same reason then we may need to consider adjusting the check.

sxa:
FYI @yosifkit as you [commented on another PR](https://github.com/docker-library/official-images/pull/19542#pullrequestreview-3064231166) which this will supercede. Any thoughts on the comment above regarding the the PR check failure? It would be good to get this in now (as you can see from the other PRs peopel have put in place about this it seems in high demand ;-) )

yosifkit:
The diff output being too long for a GitHub comment is usually fine. It becomes problematic if it happens every time (since that increases the time for us to review it). It does still show up in the job summary, but since that doesn't stick around long term, I had the bot make a gist of it.

yosifkit (approved):

tianon:
FYI affected maintainers for the Ubuntu Focal EOL/removal:

- `solr`: @asfbot @shalinmangar @dsmiley @janhoy @houstonputman (`solr:8.11.4` + `-slim`)
- `spark`: @ApacheSpark (`spark:3.5.6-scala2.12-java11-ubuntu`; @HyukjinKwon @yaooqinn @LuciferYang - please also add at least one human to `library/spark` for notifications like this)

PR #19587, opened 2025-08-03
Update photon image

gerrit-photon:
Update photon image


tianon (approved):
The `/mnt/cdrom` flip-flop brings me joy, especially inside a container image :joy: :heart:

PR #19593, opened 2025-08-04
Redis release 8.2.0

Peter-Sh:
This is 8.2.0 release of redis.

@adamiBs or @adobrzhansky please approve that the release and changes in maintainers list are accepted since I'm not there yet.


adobrzhansky:
Thanks @Peter-Sh, LGTM

adamiBs:
Thanks @Peter-Sh LGTM

yosifkit (approved):

tianon:
It looks like 6.2 might be EOL now?  Probably worth removing those entries, if that's accurate.

Peter-Sh:
> It looks like 6.2 might be EOL now? Probably worth removing those entries, if that's accurate.

@tianon Thanks! I asked about it, and the information I got is that we're extending support for 6.2 until the end of the year.

PR #19629, opened 2025-08-08
maridadb: 2025 Q3 release

grooverdan:
Notabely: MariaDB-10.6 is now based on jammy (24.04) due to Focal 20.04 EOL

12.0 is now latest.

With UBI10 being stable, the 12.0+ series is based on UBI10 for the ubi series.

The ubi images have now a FIP enable by default
profile for connections. SQL functions like MD5/SHA1 are still available as they are explictly requested of the OpenSSL library

### DIFF
--- a/library/mariadb
+++ b/library/mariadb
@@ -1,4 +1,4 @@
-# this file is generated via https://github.com/MariaDB/mariadb-docker/blob/9bc98d6905a26282e6209da20970d9d4b055a384/generate-stackbrew-library.sh
+# this file is generated via https://github.com/MariaDB/mariadb-docker/blob/3ca7af6aaaf6284d8e7c238bf114894cbfef37d0/generate-stackbrew-library.sh
 
 Maintainers: Daniel Black <daniel@mariadb.org> (@grooverdan),
              Daniel Bartholomew <dbart@mariadb.com> (@dbart),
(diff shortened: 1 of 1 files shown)


tianon:
Generally speaking, FIPS compliance is not something we have much support _or_ sympathy for within DOI.

However, if you want to move forward with this, I would suggest making that `sed` more defensive -- as is, minor format changes to the files being modified might cause you to no longer be modifying them in the ways you are expecting to, and nothing will catch that during build.  For example, you could follow it up by checking that the lines you've added _did_ get added successfully, or you could find a command to query from `openssl` itself that reports whether the "FIPS mode" is enabled.

grooverdan:
> Generally speaking, FIPS compliance is not something we have much support _or_ sympathy for within DOI.

Is it exceeding the minimal principles of DOI and/or FIPs giving a bad user experience (whether they know it or not)? If you'd rather not answer publicly that's fine.

> However, if you want to move forward with this,

[MDEV-32368](https://jira.mariadb.org/browse/MDEV-32368) was a sort of request for it. The achievement of previously was suboptimal.

> I would suggest making that `sed` more defensive -- as is, minor format changes to the files being modified might cause you to no longer be modifying them in the ways you are expecting to, and nothing will catch that during build. For example, you could follow it up by checking that the lines you've added _did_ get added successfully, or you could find a command to query from `openssl` itself that reports whether the "FIPS mode" is enabled.

Done.

yosifkit (approved):

tianon:
> > Generally speaking, FIPS compliance is not something we have much support or sympathy for within DOI.
>
> Is it exceeding the minimal principles of DOI and/or FIPs giving a bad user experience (whether they know it or not)?

Yeah, definitely some of both -- in general, FIPS compliance is really only meaningful in regulated environments, and government standards like that are orthogonal (and in some cases counter) to the principles of Free and Open Source Software (which is the primary focus of the DOI program).

PR #19633, opened 2025-08-08
Update UBI pins

yosifkit:


tianon:
Updates so frequent, and still no `ubi8-minimal` :thinking:

tianon (approved):
buildah still be wildin :shrug:

PR #19641, opened 2025-08-11
Update SapMachine dockerfiles

RealCLanger:
This updates the source dockerfiles for SapMachine.

It basically contains two fixes:
1. Trigger build of alpine based images by inserting dependency to nss lib. This should remove the vulnerability.
2. Fix entry point for JRE images

### DIFF
--- a/library/sapmachine
+++ b/library/sapmachine
@@ -13,12 +13,12 @@ Directory: dockerfiles/24/ubuntu/24_04/jdk-headless
 
 Tags: jre, jre-ubuntu, 24-jre, 24-jre-ubuntu, 24.0.2-jre, 24.0.2-jre-ubuntu, jre-ubuntu-noble, jre-ubuntu-24.04, 24-jre-ubuntu-noble, 24-jre-ubuntu-24.04, 24.0.2-jre-ubuntu-noble, 24.0.2-jre-ubuntu-24.04
 Architectures: amd64, arm64v8, ppc64le
-GitCommit: 7c561e5b3845d2d64659cef30adf35abfe9b1407
+GitCommit: 051d757d530f478f382196a670e172a334c3f3cc
 Directory: dockerfiles/24/ubuntu/24_04/jre
 
 Tags: jre-headless, jre-headless-ubuntu, 24-jre-headless, 24-jre-headless-ubuntu, 24.0.2-jre-headless, 24.0.2-jre-headless-ubuntu, jre-headless-ubuntu-noble, jre-headless-ubuntu-24.04, 24-jre-headless-ubuntu-noble, 24-jre-headless-ubuntu-24.04, 24.0.2-jre-headless-ubuntu-noble, 24.0.2-jre-headless-ubuntu-24.04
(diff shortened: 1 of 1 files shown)


tianon:
> Trigger build of alpine based images by inserting dependency to nss lib. This should remove the vulnerability.

```diff
-    apk add sapmachine-21-jdk=21.0.8-r0
+    apk add nss sapmachine-21-jdk=21.0.8-r0
```

Is this somehow exploitable from SapMachine?  My read of the vulnerability in question was that the vector was extremely thin, and certainly didn't seem to warrant going out of our way for an image rebuild, and we generally do not allow "no-op" changes to `Dockerfile`s like this, preferring instead to encourage the base image maintainers to update if something is serious enough to warrant rebuilds.

RealCLanger:
> > Trigger build of alpine based images by inserting dependency to nss lib. This should remove the vulnerability.
> 
> ```diff
> -    apk add sapmachine-21-jdk=21.0.8-r0
> +    apk add nss sapmachine-21-jdk=21.0.8-r0
> ```
> 
> Is this somehow exploitable from SapMachine? My read of the vulnerability in question was that the vector was extremely thin, and certainly didn't seem to warrant going out of our way for an image rebuild, and we generally do not allow "no-op" changes to `Dockerfile`s like this, preferring instead to encourage the base image maintainers to update if something is serious enough to warrant rebuilds.

Well, CVE-2025-6965 has a severity of 7.2 (High), so we thought it would warrant an update. It's hard to see how exploitable it would be from a JDK. The issue is within sqlite-libs and this package is not part of the base image but comes through the apk dependencies of the sapmachine package.

I can certainly remove the noop-change in the dockerfile. Is there a way to request a forced rebuild for certain tags without modifying the dockerfile?

yosifkit:
> I can certainly remove the noop-change in the dockerfile. Is there a way to request a forced rebuild for certain tags without modifying the dockerfile?

Not currently. We want to provide periodic rebuilds for all DOI, but we haven't been able to prioritize it yet.

One issue is that this temporary fix must stay in the Dockerfile almost indefinitely (or at least until each next JDK bump). If our build system isn't specifically aware of a rebuild (our future plan), and this is ever reverted while the base image still hasn't been updated, then the old image build would be reused (since the build context and parent match).

RealCLanger:
> > I can certainly remove the noop-change in the dockerfile. Is there a way to request a forced rebuild for certain tags without modifying the dockerfile?
> 
> Not currently. We want to provide periodic rebuilds for all DOI, but we haven't been able to prioritize it yet.
> 
> One issue is that this temporary fix must stay in the Dockerfile almost indefinitely (or at least until each next JDK bump). If our build system isn't specifically aware of a rebuild (our future plan), and this is ever reverted while the base image still hasn't been updated, then the old image build would be reused (since the build context and parent match).

OK, I was planning to remove the noop change with the next version bump. But then I guess we'll have to live with the vulnerability until then. Users will have to do apk upgrade to get rid of it in the meanwhile...

PR #19643, opened 2025-08-11
Update Jetty to 12.0.24

lachlan-roberts:
- Update Jetty Version to `12.0.24`
- Changes to regenerate the `jetty.start` file after the Jetty version is updated (see https://github.com/jetty/jetty.docker/pull/262).


yosifkit (approved):
Seems fine to me.

Side note: Perhaps it should also regenerate `jetty.start` files that weren't created with a version in them (i.e. any created before this update)? Would that be too disruptive?

lachlan-roberts:
> Side note: Perhaps it should also regenerate jetty.start files that weren't created with a version in them (i.e. any created before this update)? Would that be too disruptive?

I did consider doing it like that, I don't see any reason why it would cause disruption.

Added commit https://github.com/jetty/jetty.docker/commit/68752e005359f51f5ba72091fb4daf006db482e0#diff-d7b7947711a74a36a923288e4113502560ff35b37898e37fe3691343031be87b and will push an update here once all the tests pass.

lachlan-roberts:
@yosifkit done

tianon (approved):

PR #19645, opened 2025-08-11
Update gcc

yosifkit:
Changes:

- https://github.com/docker-library/gcc/commit/915af5c: Update 15 to debian trixie
- https://github.com/docker-library/gcc/commit/280306a: Update 14 to debian trixie


yosifkit:
Hold till after the `debian:trixie` release image build.

tianon (approved):

PR #19581, opened 2025-08-01
Update Ubuntu 2025-08-01

ROCKsBot:
Update Ubuntu to the latest published OCI tarballs as of 2025-08-01.

PR #19653, opened 2025-08-11
Reject riscv64 for images from Ubuntu 25.10+

yosifkit:
RVA23 is not supported on our builders because there is no hardware available. When we have a way to build a riscv64 image with RVA23 profile, we can reconsider this.

- https://canonical.com/blog/rivos-and-canonical-partner-to-deliver-scalable-risc-v-solutions-in-data-centers
- https://bugs.launchpad.net/ubuntu/+source/ubuntu-release-upgrader/+bug/2111715
- https://github.com/docker-library/official-images/pull/19581

```console
$ ./naughty-from.sh buildpack-deps
 - buildpack-deps:questing-curl (FROM ubuntu:questing) [riscv64]
DOI cannot support risvc64 on Ubuntu 25.10+ since it uses RVA23 and hardware does not exist yet
```

### DIFF
--- a/naughty-from.sh
+++ b/naughty-from.sh
@@ -11,8 +11,30 @@ externalPinsDir="$(dirname "$BASH_SOURCE")/.external-pins"
 declare -A externalPinsArchesCache=(
 	#[img:tag]='["arch","arch",...]' # (json array of strings)
 )
+
+# custom error message (for why Ubuntu 25.10+ does not support riscv64)
+message=''
+
 _is_naughty() {
 	local from="$1"; shift
+
+	# DOI cannot support riscv64 on Ubuntu 25.10+ since it uses RVA23 and hardware does not exist yet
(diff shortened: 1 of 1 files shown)


naughty-from.sh
```diff
@@ -11,9 +11,33 @@ externalPinsDir="$(dirname "$BASH_SOURCE")/.external-pins"
...
 	#[img:tag]='["arch","arch",...]' # (json array of strings)
 )
+
+# custom error message (for why Ubuntu 25.10+ does not support riscv64)
+message=''
+
 _is_naughty() {
 	local from="$1"; shift
 
+	# DOI cannot support risvc64 on Ubuntu 25.10+ since it uses RVA23 and hardware does not exist yet
```
tianon, line 21: "risvc64" (should be "riscv64")

PR #19660, opened 2025-08-12
Update php

yosifkit:
Changes:

- https://github.com/docker-library/php/commit/2ab91e25: Merge pull request https://github.com/docker-library/php/pull/1596 from endelwar/trixie
- https://github.com/docker-library/php/commit/686f9529: Add Debian Trixie (and remove Bullseye)

PR #19725, opened 2025-08-19
Update YOURLS

LeoColomb:
Image release: https://github.com/YOURLS/containers/releases/tag/4.1.0
Reference: https://github.com/docker-library/official-images/pull/19660#issuecomment-3198610715


tianon (approved):
Thank you!

PR #19727, opened 2025-08-19
TomEE 10.1.1

scriptmonkey:


tianon:
There aren't any `Dockerfile` changes here :sweat_smile:

scriptmonkey:
Why else would my tests pass the 1st time.  lol.

Seems I neglected to update the commit hash.

PR #19731, opened 2025-08-19
Update Monica and remove mips64le tag

asbiin:
Update Monica to remove `mips64le` tag.
See https://github.com/docker-library/official-images/pull/19660#issuecomment-3198610715


tianon (approved):
Thanks!

PR #19747, opened 2025-08-24
library/perl: 5.43.1 -> 5.43.2

zakame:
- https://github.com/Perl/docker-perl/pull/193
- http://nntp.perl.org/group/perl.perl5.porters/270238


zakame:
Thanks @yosifkit @tianon 🎉

PR #19748, opened 2025-08-24
disable gazebo classic as focal is EOL

mikaelarguedas:
This disable builds for Gazebo 11 that is based on Ubuntu Focal

New gazebo images will be released in the future so the library will be updated https://github.com/gazebo-release/gz-sim8-release/issues/8


tianon (approved):
Thank you!

PR #19750, opened 2025-08-25
[varnish] debian - fresh and old moved to trixie

skyscooby:
fresh and old debian images now based on trixie.
stable stays on bookworm due to legacy build constraints.


gquintard:
Thank you @skyscooby !

PR #19757, opened 2025-08-26
feat(rust): Support riscv64 on Trixie

Muscraft:


Muscraft:
It looks like the CI failure was due to a transient network error. I ran into the [same error in docker-rust's CI](https://github.com/rust-lang/docker-rust/actions/runs/17250071779/job/48950255678), and [rerunning the job fixed it](https://github.com/rust-lang/docker-rust/actions/runs/17250071779).

PR #19758, opened 2025-08-27
InfluxDB3 Core and Enterprise v3.4.0

hiltontj:


praveen-influx:
We are in the process of re-releasing, could you please not merge until we can confirm images have been rebuilt correctly? This PR still stands, just need time for rebuilding the image. Thanks!

hiltontj:
Closing as we wait on re-release. We will open a new PR when the build is finished.

tianon:
FWIW, moving to "draft" state is probably better/cleaner than closing (especially as you can then force push / un-draft when the release is ready).

hiltontj:
Okay, thank you @tianon we can follow that advice if we run into this scenario again. https://github.com/docker-library/official-images/pull/19762 has superceded this though.

tianon:
Oof, I should've checked closed - @yosifkit is too fast :joy: :heart:

Merged in https://github.com/docker-library/official-images/pull/19762

PR #19768, opened 2025-08-27
Fedora Update 20250827

Phuong-Le-NN:


cverna:
One small thing to update but otherwise LGTM, thanks @Phuong-Le-NN

Phuong-Le-NN:
I see. I have just added it.
I didn't notice that before, thank you a lot!

yosifkit (approved):

PR #19675, opened 2025-08-12
Update golang

yosifkit:
Changes:

- https://github.com/docker-library/golang/commit/46370e0: Merge pull request https://github.com/docker-library/golang/pull/567 from infosiftr/go1.25
- https://github.com/docker-library/golang/commit/87f680b: Update to 1.25 (GA)
- https://github.com/docker-library/golang/commit/bab6314: Merge pull request https://github.com/docker-library/golang/pull/564 from infosiftr/trixie
- https://github.com/docker-library/golang/commit/86aa8ce: Add Debian Trixie (drop Bullseye)

PR #19773, opened 2025-08-28
Update ELK images

StamatisKourk:
## Description

Update ELK images for 8.18.6, 8.19.3, 9.0.6 and 9.1.3 releases

### DIFF
--- a/library/elasticsearch
+++ b/library/elasticsearch
@@ -18,10 +18,10 @@ Architectures: amd64, arm64v8
 GitFetch: refs/heads/8.19
 GitCommit: 2ed23b0bc3f4c094097cc7ec75bcd359abd0018f
 
-Tags: 9.0.5
+Tags: 9.0.6
 Architectures: amd64, arm64v8
 GitFetch: refs/heads/9.0
-GitCommit: e69a6a63c299a6ac53e838174d1f451392ff2bdc
+GitCommit: 5b377e63667793a5584e72272da6699da2b22b44
 
 Tags: 9.1.3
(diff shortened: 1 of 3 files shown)


StamatisKourk:
Hey, it is not clear why this is failing the CI. I see:
```

 - logstash:9.0.6 (FROM golang:1.23) [amd64, arm64v8]
 - logstash:9.1.3 (FROM golang:1.23) [amd64, arm64v8]
```
Does this refer to the use of `golang:1.23` as the base image in the logstash Dockerfile? This was there for 9.1.2 version as well, so could you please explain what exactly is the problem? Thank you 🙇🏻

donoghuc:
@StamatisKourk I believe this would have started failing with https://github.com/docker-library/official-images/pull/19675/files I have filed https://github.com/elastic/logstash/issues/18063 to update the logstash images.

yosifkit:
> Does this refer to the use of `golang:1.23` as the base image in the logstash Dockerfile?

Yes. Since the release of go `1.25`, the `1.23.x` release has reached end of life in the Go project and so also reached end of life in Docker Official Images. As such, we require moving to a supported go release (`1.24` or `1.25`) as soon as possible.

tianon:
See also https://github.com/docker-library/official-images/pull/19675#issuecomment-3198576477
(if there are more folks that should receive pings like that, please add their GitHub handles to the relevant `Maintainers:` lines)

StamatisKourk:
@yosifkit as checks are successful now, I think this is ready for a review.

tianon (approved):

PR #19785, opened 2025-08-29
Replace `libssl3` with `libssl3t64` for spiped

TimWolla:
see TimWolla/docker-spiped#60


tianon:
Thanks for carrying this all the way through!

TimWolla:
> Thanks for carrying this all the way through!

My fault for locking down the repository that much :wink: I intentionally sign my commits for DOI, because I'm part of a relevant supply chain there.

Thank you for doing the research regarding the breakage.

PR #19822, opened 2025-09-07
Update to Eggdrop v1.10.1 **STABLE**

vanosg:


tianon:
I think you need to rebase on our latest `master` (and perhaps in the future open your PR branches from our HEAD instead of your fork's) so that the delete of 1.10.1rc1 actually takes. :eyes:

tianon:
For this PR, the easy mode might be something like `git fetch https://github.com/docker-library/official-images.git HEAD: && git reset --hard FETCH_HEAD && git cherry-pick 4d57a65b0c1d8c18325e8955538e0e18373c140e && git push --force` (since 4d57a65b0c1d8c18325e8955538e0e18373c140e is the only commit you actually want here, and it should apply cleanly on our HEAD).

tianon:
Friendly ping? :sweat_smile:

Also https://github.com/docker-library/official-images/pull/20400#issuecomment-3987803149 (dropping or updating off the now-EOL `alpine:3.19` :eyes:)

PR #19844, opened 2025-09-09
Update tianon buildkit pins

yosifkit:


tianon:
```diff
-			"created_by": "EXPOSE map[2375/tcp:{} 2376/tcp:{}]",
+			"created_by": "EXPOSE &{[{{103 0} {103 0}}] <nil>}",
```

I want to die. 😑

tianon:
We should probably dig into that `EXPOSE` presentation and/or reproduce on latest buildkit and file a bug. :disappointed:

tianon:
It's not fixed in these images, but it _is_ reported and being handled upstream, so I'm satisfied.

PR #19868, opened 2025-09-12
Update clojure to remove git from lein variants

cap10morgan:
As requested by the community: https://github.com/Quantisan/docker-clojure/issues/258


tianon:
Heh, the Levenshtein distance between `2.12.0` in the tags and `1.12.0` in the `Dockerfile` made me do a double take and almost ask if the tags should be updated, but I cross checked again and noticed my mistake. :joy:

tianon:
Ah, and `1.12.2.1565` - y'all are really going through it with these similar version numbers, huh? :sweat_smile:

cap10morgan:
> Ah, and `1.12.2.1565` - y'all are really going through it with these similar version numbers, huh? :sweat_smile:

Haha yeah. Clojure gets fun with those because it's a language that's a Java library. Just look at our tags! 🤣

PR #19892, opened 2025-09-17
Update amazoncorretto for Corretto 25 Release

mmm-choi:
Adding 25 images for GA.


tianon (approved):
This looks fine to me :+1:

While reviewing it, I noticed that `latest` still points at JDK 8 :sweat_smile:  Is that intentional?  Maybe something worth considering updating?

(If you want a suggestion, I'd suggest having it always point to the highest LTS release, but it's ultimately up-to-y'all.)

Rudometov:
thank you for suggestion, that's a good one.

latest pointing to 8 is something we cannot change easily without breaking a lot of folks unfortunately.

PR #19904, opened 2025-09-18
Use the new YAML anchors support to DRY up our bashbrew version

tianon:
This still isn't *ideal* because I'd rather have `uses:` be the dynamic value embedding a variable of bashbrew version, but parsing the version out of *that* value isn't horrible so if this works, I'll try to be satisfied.

https://github.blog/changelog/2025-09-18-actions-yaml-anchors-and-non-public-workflow-templates/

### DIFF
--- a/.github/workflows/.bashbrew/action.yml
+++ b/.github/workflows/.bashbrew/action.yml
@@ -8,11 +8,14 @@ runs:
   using: 'composite'
   steps:
 
-    # these two version numbers are intentionally as close together as I could possibly get them because no matter what I tried, GitHub will not allow me to DRY them (can't have any useful variables in `uses:` and can't even have YAML references to steal it in `env:` or something)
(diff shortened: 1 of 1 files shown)


tianon:
lmao

```
Error: /home/runner/work/official-images/official-images/./.github/workflows/.bashbrew/action.yml: Anchors are not currently supported. Remove the anchor 'bb'
Error: System.ArgumentException: Unexpected type '' encountered while reading 'action manifest root'. The type 'MappingToken' was expected.
   at GitHub.DistributedTask.ObjectTemplating.Tokens.TemplateTokenExtensions.AssertMapping(TemplateToken value, String objectDescription)
   at GitHub.Runner.Worker.ActionManifestManager.Load(IExecutionContext executionContext, String manifestFile)
Error: Failed to load /home/runner/work/official-images/official-images/./.github/workflows/.bashbrew/action.yml
```

PR #19908, opened 2025-09-19
2.8.12

alcohol:


tianon:
We should dig into the 1.10.27 failure (maybe that's just fixing or excluding the test?  maybe it's a sign of something bigger/elsewhere?  I'm happy to help with that with more Composer guidance on what the failure means :smile:) but I don't think that needs to block this. :+1:

PR #19909, opened 2025-09-19
Gradle 9.1.0

cobexer:


cobexer:
@tianon This is the first time I update images so please take that into account while reviewing... to avoid mistakes beginners make =)

tianon (approved):
If there's something wrong with it, it's more subtle than I'm catching right now! :smile: :heart: :+1:

(It looks fine to me :bow:)

PR #19915, opened 2025-09-22
TomEE 10.1.2

jungm:


jungm:
quick side note: what would the process look like for me to get added as a maintainer? I'm a PMC on the Apache TomEE project and also have commit access to the tomitribe/docker-tomee repository so IMO it would only make sense

tianon:
> quick side note: what would the process look like for me to get added as a maintainer? I'm a PMC on the Apache TomEE project and also have commit access to the tomitribe/docker-tomee repository so IMO it would only make sense

That's just updating https://github.com/docker-library/official-images/blob/1edf5df19e7390ba2c84789885b9f6ec8cec7c2b/library/tomee#L1 :sweat_smile:

(that's `Name <email> (@GitHub-handle)`)

tianon (approved):

jungm:
Oh, yeah makes sense! was in a hurry earlier and didn't see it 😅

PR #19918, opened 2025-09-22
feat: Remove influxdb 1.10 from supported tags

devanbenz:
After internal discussions we have decided to remove 1.10 from our supported dockerhub tags.


tianon (approved):
:heart: :+1:

PR #19944, opened 2025-09-24
Update: release InfluxDB Enterprise v1.11.9

vlastahajek:


tianon:
Should the non-enterprise be updated too?

vlastahajek:
No. Only Enterprise.

tianon (approved):

PR #19948, opened 2025-09-25
Update Spark 3.5 from 3.5.6 to 3.5.7

peter-toth:
This PR proposes to update Spark 3.5 to its latest 3.5.7.

### DIFF
--- a/library/spark
+++ b/library/spark
@@ -41,42 +41,42 @@ Architectures: amd64, arm64v8
 GitCommit: 4bd1dbce94797b5b387b784db6b378069a8b6328
 Directory: ./4.0.0/scala2.13-java17-python3-r-ubuntu
 
-Tags: 3.5.6-scala2.12-java17-python3-ubuntu, 3.5.6-java17-python3, 3.5.6-java17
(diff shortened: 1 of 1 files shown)


peter-toth:
seems like `11-jre-focal` is not supported any more: https://github.com/docker-library/official-images/commit/b80d1ee97f02902c6105525f2c89a5e642953524#diff-d4316a94e552843bdf794aee0bbe42c4e73790b19f2ade819a361ac5fdc0b9f2L211

tianon:
yep, see https://github.com/docker-library/official-images/pull/19167#issuecomment-2931611620 -- I'll reiterate that without any humans listed in `Maintainers:`, we don't have a way to notify y'all of things like this, so it would be great (and helpful to you) if you could add some even if they' just `Name (@GitHub-handle)` (ie, no email)

peter-toth:
@tianon , this is ready for review. I upgraded the images from `11-jre-focal` to `11-jammy` in https://github.com/apache/spark-docker/commit/2ebf694ad45fee6f4beeeb4204bcdb01d73c988f.

PR #19953, opened 2025-09-25
rakudo-star: Add trixie as "latest"

AntonOks:


library/rakudo-star
```diff
@@ -4,7 +4,12 @@ Maintainers: Moritz Lenz <moritz.lenz@gmail.com> (@moritz),
              Anton Oks <anton.oks@gmx.de> (@AntonOks)
 GitRepo: https://github.com/rakudo/docker
 
-Tags: latest, 2025.08.1, bookworm
+Tags: trixie, 2025.08.1-trixie, latest
+Architectures: amd64, arm64v8
+GitCommit: 8c213b2fb89d60eb631068eecba27dab9a0989bc
```
yosifkit, line 9: It looks like this is one commit off and should be https://github.com/rakudo/docker/commit/c40c9a549916b02b0c6c8d5cda7dbd444c094278 instead in order to pick up the `trixie` variant.

AntonOks: @yosifkit Thanks for the catch!

yosifkit (approved):

PR #19956, opened 2025-09-25
eclipse-temurin: add JDK25 images

gdams:


j3t:
We are currently using the jammy version of the JDK 21 builds. Is there a reason why the jammy version is not longer built for JDK 25 images?

gdams:
> We are currently using the jammy version of the JDK 21 builds. Is there a reason why the jammy version is not longer built for JDK 25 images?

this seems to be an oversight, Jammy is still supported untill 2027. I'll get this added today

g0ddest:
Hey, @gdams , latest tag has gone, was it intended?

tianon:
@g0ddest can you be more specific?  It's right here:

https://github.com/docker-library/official-images/blob/e02cf7e37d0d7fd8bfab1764cabd995bd223bd90/library/eclipse-temurin#L577-L581

PR #19972, opened 2025-09-28
Gradle 9: Java 25

cobexer:


tianon:
`jdk-25-and-25` seems like kind of a weird alias - should that variant be removed for now until it's meaningful again?  Or perhaps just any tag aliases where the two version numbers are the same, so we have `lts-and-current` but not `25-and-25`?  I'm happy to help with an implementation of what I mean, if that's useful.

---

```diff
@@ -26,9 +26,15 @@ RUN set -o errexit -o nounset \
     && apt-get install --yes --no-install-recommends \
         # common utilities
         make \
+        # https://github.com/adoptium/containers/issues/814
+        curl \
+        wget \
         \
         # Dockerfile dependencies
         unzip \
+        wget \
+        # common utilities \
+        curl \
         \
         # VCSes
         brz \
```

This is kind of odd :sweat_smile: are these duplicated intentionally?  (it's harmless, but want to double check :+1:)

cobexer:
@tianon 
> jdk-25-and-25 seems like kind of a weird alias

I agree that the alias is sort of weird, but I expect users of that alias to generate it based on variables like:

```Dockerfile
ARG LTS=21
ARG CURRENT=24
FROM gradle:jdk-${LTS}-and-${CURRENT}-alpine
# ...
```

```bash
# used like:
$ ~> docker build --build-arg LTS=25 --build-arg CURRENT=25 .
```


Thus I'd like to keep them for consistency, and in a few months (March 2026 - Java 26) they make more sense again.

> are these duplicated intentionally

Nope, fully harmless but thanks for catching this - I updated the Dockerfiles to all consistently list some (not too much scope creep) of the base utilities while not affecting the resulting images (thus UBI and Corretto stay with `curl-minimal`).

While both the above are harmless, I noticed weird diffs for `jdk24-corretto-al2023` and `jdk24-graal-noble` that I took care of:
Since Java 24 images no longer exist for Gradle 9+ but both listed ones still exist for Gradle 8, users would receive "updates" for their images that would have unexpectedly downgraded Gradle to the previous major. I made sure we don't generate the tags (the ones without Gradle version)

PR #19975, opened 2025-09-29
Update CrateDB to 6.0.2

matriv:


tianon (approved):
Looks solid -- just to make sure, 5.9 and 5.10 are both still actively supported, right? :+1:

(If not, they should be removed, but if so, all's well.)

matriv:
> Looks solid -- just to make sure, 5.9 and 5.10 are both still actively supported, right? 👍
> 
> (If not, they should be removed, but if so, all's well.)

Yes they're still supported, but soon we will remove 5.9. We always have in mind to not leave unnecessary versions in the file and put pressure on your CI. Thank you for the reminder though!

PR #19982, opened 2025-09-29
Update Debian

tianon:


tianon:
```diff
+var/cache/debconf/tmp.ci/
```

This was interesting, but https://bugs.debian.org/1092667 makes it sound like it's very expected. :+1:

Seeing unstable get smaller is neat:

```diff
-			"size": 49657990
+			"size": 48376965

...

-			"size": 30020111
+			"size": 29010248
```

yosifkit (approved):

PR #19983, opened 2025-09-29
chore: update influxdb core/enterprise to 3.5.0

hiltontj:


yosifkit:
The diff is missing any Dockerfile changes compared to the current versions. It looks like they still have the previous release: https://github.com/influxdata/influxdata-docker/blob/34a7d250c910f02fce342ff85b73d0cf630140c6/influxdb/3.5-core/Dockerfile#L19

hiltontj:
Ah, sorry @yosifkit I missed that, thank you for catching - I will update upstream. For now I will put this in draft.

hiltontj:
@yosifkit - this is ready again.

PR #19985, opened 2025-09-30
Enable Semeru 25 Open Edition container Images

jayasg12:


tianon (approved):
Is 24 EOL now too? :eyes:

PR #19995, opened 2025-10-01
gitsha bump for Neo4j 2025.09.0

jennyowen:
There was a minor a problem with the https://dist.neo4j.org/neo4j-community-2025.09.0-unix.tar.gz and https://dist.neo4j.org/neo4j-enterprise-2025.09.0-unix.tar.gz tars, so we had to re-upload a new version with a fix. 

The neo4j download URI remains the same, but the SHA is different, so the docker image source needs an update otherwise it will fail the next time a rebuild is scheduled. Hope that's ok!


yosifkit (approved):

yosifkit:
> The neo4j download URI remains the same, but the SHA is different, so the docker image source needs an update otherwise it will fail the next time a rebuild is scheduled. Hope that's ok!

A PR like this is totally fine and thanks for the explanation.

tianon:
Actively encouraged, even!  We really appreciate y'all proactively making sure our rebuilds of your image don't fail. :sparkling_heart:

PR #19984, opened 2025-09-29
Update tests to Debian Trixie

J0WI:

PR #20003, opened 2025-10-02
Update JRuby top JDK to 25

headius:
* 9.4 gains JDK 25 and JRE 25 images
* 10 moves the JDK and JRE 24 images to JDK and JRE 25

### DIFF
--- a/library/jruby
+++ b/library/jruby
@@ -3,7 +3,7 @@ Maintainers: JRuby Admin <admin@jruby.org> (@jruby),
              Thomas E Enebo <tom.enebo@gmail.com> (@enebo)
 GitRepo: https://github.com/jruby/docker-jruby.git
 GitFetch: refs/heads/master
-GitCommit: 73b84171f2eefb713ff4c4420a9cd46b19f7df42
+GitCommit: a8ffc83b6e262b7171f26a07284053dc3de9af6b
 
 Tags: latest, 10, 10.0, 10.0.2, 10.0-jre, 10.0-jre21, 10.0.2-jre, 10.0.2-jre21, 10.0.2.0, 10.0.2.0-jre, 10.0.2.0-jre21
 Architectures: amd64, arm64v8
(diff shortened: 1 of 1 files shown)


headius:
Wow is something seriously wrong with the Jammy images for Eclipse Temurin JDK? I have not changed the 9.3 Dockerfiles in a very long time and now they're failing... but only on this PR?

Everything passed for all images in a build test here: https://github.com/jruby/docker-jruby/actions/runs/18198545800

headius:
There appears to have been an incident earlier today on the Ubuntu servers, but I don't know why that would affect my builds now.

Only the Ubuntu "Noble" images passed, with two builds still apparently stuck trying to fetch the upstream images. I don't think these failures are my fault.

yosifkit:
Yeah, the test failures are from some over aggressive changes in #19984 and should be fixed by #20006. I'll retrigger these builds once the tests are updated.

yosifkit:
Close/reopen to retrigger with updated tests

yosifkit:
Ah now that there are fewer noisy failures, we can see `curl` failures. It seems to be because the `eclipse-temurin` image doesn't include it in jdk25 (well, 22+): https://github.com/adoptium/containers/issues/630 https://github.com/adoptium/containers/issues/814

headius:
Ok, I will add curl to the list of install packages.

headius:
@yosifkit Thanks for the help!

yosifkit (approved):

library/jruby
```diff
@@ -53,6 +53,14 @@ Tags: 9.4-jre21, 9.4.14-jre21, 9.4.14.0-jre21
...
 
+Tags: 9.4-jdk25, 9.4.14-jdk25, 9.4.14.0-jdk25
+Architectures: amd64, arm64v8
+Directory: 9.4/jdk25
+
+Tags: 9.4-jre25, 9.4.14-jre25, 9.4.14.0-jre25
+Architectures: amd64, arm64v8
+Directory: 9.4/jre25
+
 Tags: 9.3, 9.3.15, 9.3-jre, 9.3-jre8, 9.3.15-jre, 9.3.15-jre8, 9.3.15.0, 9.3.15.0-jre, 9.3.15.0-jre8
```
tianon: JRuby 9.3 is EOL, right? https://github.com/jruby/jruby/wiki/Roadmap

If so, we should remove all the 9.3 tags from this file. :bow: :heart:

headius:
@tianon Ah, good to know that the old ones won't go away. I will file an issue to to get the eol versions out of there.

PR #20052, opened 2025-10-08
Update UBI pins

yosifkit:


tianon:
Switching from OCI to Docker media types is sus - that feels like something we should probably report somewhere :thinking:

tianon:
Oof, https://access.redhat.com/articles/4238681#:~:text=If%20you%20have%20problems%20or%20enhancement%20requests%20for%20UBI%2C%20enter%20a%20bug%20at%20https%3A//issues.redhat.com%20under%20the%20RHEL%20project.
~~(I don't have an account there, and I'm not sure I want one :sob:)~~ scratch that, apparently I do lolsob

tianon:
https://issues.redhat.com/browse/RHEL-119910

tianon:
> https://issues.redhat.com/browse/RHEL-119910

I've seen an inconsistent ability to access my own public ticket on this (which feels ironic), so I'm going to copy the relevant bits here, especially https://issues.redhat.com/browse/RHEL-119910?focusedId=28231298&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-28231298

> [Jiri Popelka](https://issues.redhat.com/secure/ViewProfile.jspa?name=jpopelka%40redhat.com) added a comment - [2025/10/13 8:38 AM](https://issues.redhat.com/browse/RHEL-119910?focusedId=28231298&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-28231298)
> From an [announcement](https://redhat-internal.slack.com/archives/C06RBBBGY11/p1758912421398329):
> 
> When products moved to Konflux, the images produced changed from using the Docker v2-2 spec to the OCI spec. We have recently learned of some unintended consequences. While OCI spec is widely supported by modern registry implementations the problem is our documentation currently states “[You can use any container registry that supports Docker v2-2](https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/disconnected_environments/connected-to-disconnected)”. We know that at least [one customer has been affected by this](https://issues.redhat.com/browse/OCPBUGS-61042), requiring the change.
> 
> We are going to ensure that the OCP 4.20 documentation is updated to indicate that support for Docker v2-2 spec is deprecated and that OCP 4.21 will require mirror registries to provide support for OCI v1 spec. We need to move to the OCI spec in the near future to support other initiatives and all major registry implementations support OCI spec in their latest versions as of late 2024 and most added support by 2023.
> 
> All Konflux images built and supported in OCP 4.20 and earlier MUST make changes to ensure that your images are compatible with the v2-2 spec. We have updated the pipeline’s default format for container images as part of KONFLUX-10189.

PR #20062, opened 2025-10-09
 Update Jetty versions to 12.0.28 & 12.1.2

lachlan-roberts:
###  Update Jetty versions
- `12.0.27` -> `12.0.28`
- `12.1.1` -> `12.1.2`

### DIFF
--- a/library/jetty
+++ b/library/jetty
@@ -119,15 +119,15 @@ Architectures: amd64, arm64v8
 Directory: eclipse-temurin/12.0/jre17
 GitCommit: d809b5668ca6c063cd913153dd0f5935f1276a4b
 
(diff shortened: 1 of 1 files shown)


yosifkit:
From the failing test:
> Invalid 'FROM' + 'Architectures' combinations detected:
>
> - jetty:12.0.28-jdk24 (FROM eclipse-temurin:24-jdk) -- completely unsupported base!
> - jetty:12.0.28-jdk24-alpine (FROM eclipse-temurin:24-jdk-alpine) -- completely unsupported base!

This is because `jdk24` reached end of life, but now `jdk25` is released (and is an LTS version). Please remove these images or update them to a supported base (like `eclipse-temurin:25-jdk`). I think `amazoncorretto` will drop their jdk24 soon and they also have jdk25 images.

lachlan-roberts:
@yosifkit sorry for the delay, I have now updated the images to use JDK25 instead of JDK24.

PR #20064, opened 2025-10-09
Updates for the release of WebSphere and Open Liberty 25.0.0.10

gkertasef:


tianon:
Looks like builds are probably failing due to `LIBERTY_URL=${LIBERTY_URL:-$(wget -q -O - https://public.dhe.ibm.com/ibmdl/export/pub/software/websphere/wasdev/downloads/wlp/index.yml | grep -E "^\s*kernel:.*${LIBERTY_VERSION}\.zip" | sed -n 's/\s*kernel:\s//p' | tr -d '\r' )}` failing - it would be _really_ helpful if that value could be pre-calculated and embedded in the `Dockerfile` directly instead of calculated during build (for many reasons, the current obscure build failure being just one).

yosifkit:
Ah, I think the removal of `.3` images was missed when `.9` was added (https://github.com/docker-library/official-images/pull/19857) since that is what normally happens on the PRs for the quarterly (3/6/9/12) releases.

leochr:
Thank you for investigating the failure. 

We have a new requirement to support quarterly releases longer starting with 25.0.0.3, which is the reason 25.0.0.3 wasn't removed when 25.0.0.9 was added. But as evident from the error seen in the build, that's currently not possible. The index file/DHE repo need to change. To unblock the critical 25.0.0.10 release, I suggest that we remove 25.0.0.3 references as part of this PR and handle that separately in a different PR (once index file/DHE is sorted out).

@gkertasef @wraschke Could you please add a commit to remove 25.0.0.3 for now? Thank you.

gkertasef:
I have removed 25.0.0.3 images.

yosifkit:
Would it be important to prevent users from breaking themselves if they had whitespace in their `SERVICEABILITY_NAMESPACE` or `HOSTNAME`?
```diff
   SERVICEABILITY_FOLDER="/serviceability/$SERVICEABILITY_NAMESPACE/$HOSTNAME/logs"
-  mkdir -p $SERVICEABILITY_FOLDER
+  mkdir -p "$SERVICEABILITY_FOLDER"
   rm /liberty/logs
-  ln -s $SERVICEABILITY_FOLDER /liberty/logs
+  ln -s "$SERVICEABILITY_FOLDER" /liberty/logs
```

leochr:
@yosifkit SERVICEABILITY_NAMESPACE or HOSTNAME will be automatically set by Kubernetes or Liberty Operator (or some orchestrator/automation), hence the chance of user error is very slim. Thanks

leochr:
Few users inquired when the 25.0.0.10 images will be available. Could this PR be merged please? Thank you.

yosifkit:
I would still highly recommend defense in depth on your shell scripts, even if the values come from automation.

PR #20068, opened 2025-10-09
Update python

yosifkit:
Changes:

- https://github.com/docker-library/python/commit/e9399041: Merge pull request https://github.com/docker-library/python/pull/1088 from infosiftr/race
- https://github.com/docker-library/python/commit/54a65f4f: Bump 3.10 and 3.11
- https://github.com/docker-library/python/commit/00c4cce6: Update 3.9 to 3.9.24, setuptools 79.0.1
- https://github.com/docker-library/python/commit/e4ab0fe5: Update 3.12 to 3.12.12

\* setuptools bump in Python [3.9](https://github.com/python/cpython/pull/135397), [3.10](https://github.com/python/cpython/pull/135398), and [3.11](https://github.com/python/cpython/pull/135396) https://github.com/python/cpython/issues/135374

### DIFF
--- a/library/python
+++ b/library/python
@@ -99,126 +99,126 @@ Directory: 3.13/windows/windowsservercore-ltsc2022
 Constraints: windowsservercore-ltsc2022
 Builder: classic
 
-Tags: 3.12.11-trixie, 3.12-trixie
-SharedTags: 3.12.11, 3.12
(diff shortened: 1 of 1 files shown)


yosifkit:
We should probably wait for 3.10.19 and 3.11.14 that were also released today. It seems our automation just looked a little too early.

PR #20072, opened 2025-10-10
Updated swipl devel to 9.3.32

JanWielemaker:


tianon:
I was hoping this would be the fix for https://github.com/docker-library/official-images/pull/16167#issuecomment-3363537774 but apparently not :disappointed:

What this means concretely is that if we merge this, it will fail to build. :sob:

JanWielemaker:
I think I messed up.  Will make a new PR.

PR #20075, opened 2025-10-11
Updated swipl devel to 9.3.32

JanWielemaker:
Updates RocksDB package as well.    This updates RocksDB itself and hopefully fixes https://github.com/docker-library/official-images/pull/16167#issuecomment-3363537774.

### DIFF
--- a/library/swipl
+++ b/library/swipl
@@ -1,7 +1,7 @@
 Maintainers: Jan Wielemaker <jan@swi-prolog.org> (@JanWielemaker),
              Dave Curylo <dave@curylo.org> (@ninjarobot)
 GitRepo: https://github.com/SWI-Prolog/docker-swipl.git
-GitCommit: 69b22e9613a29ff6079ee9783502afe91e7b16b5
+GitCommit: d4411aa88faba036ab47f95f744f77b4fec6749d
 Architectures: amd64, arm32v7, arm64v8
 
 Tags: latest, 9.3.32


yosifkit:
Unfortunately, the rocksdb bump wasn't enough; it is still getting build errors (I think related to `avx512`): https://doi-janky.infosiftr.net/job/wip/job/test-pr/157/console

tianon:
Looks like maybe https://github.com/facebook/rocksdb/blob/04c085a3faab360747615fb9975495bfa5a3360a/INSTALL.md?plain=1#L23-L31 is relevant?

> * By default the binary we produce is optimized for the CPU you're compiling on
> (`-march=native` or the equivalent). To build a binary compatible with the most
> general architecture supported by your CPU and compiler, set `PORTABLE=1` for
> the build, but performance will suffer as many operations benefit from newer
> and wider instructions. In addition to `PORTABLE=0` (default) and `PORTABLE=1`,
> it can be set to an architecture name recognized by your compiler. For example,
> on 64-bit x86, a reasonable compromise is `PORTABLE=haswell` which supports
> many or most of the available optimizations while still being compatible with
> most processors made since roughly 2013.

JanWielemaker:
Thanks.  Looked at another recent RocksDB Dockerfile.   They pass `PORTABLE=1`, so I think that is the next thing to try.   Pushed a commit for that.   Note this is relatively complicated as the Prolog package builder creates a clean environment, so we have to hook into the package builder.

Fingers crossed ...

yosifkit:
`swipl:9.3.32` (aka `latest`) works! And `swipl:9.2.9` (`stable`) will need the change too.

JanWielemaker:
> `swipl:9.2.9` (`stable`) will need the change too.

I expect 9.4.0 (or maybe 10.0.0, haven't decided yet) to be out in a couple of weeks.    That will fix this issue :smile:

Issue #20039, opened 2025-10-07
Retiring Unit


thresheek:
Hello!

The development on the Unit project sadly, ceased, and it will no longer receive any feature or security updates.  In fact, the code repo is soon to be archived.

Thus, I would like to understand a procedure to stop rebuilding the official images for Unit.

We would also need to alter the Hub page via the `docs` repo to let users know it's now unsupported in any way.

Thanks!

yosifkit:
The procedure to stop rebuilds is to empty out the `library/unit` file save for the `Maintainers` and `GitRepo` fields (e.g. https://github.com/docker-library/official-images/pull/17094 and https://github.com/docker-library/official-images/pull/17276).

For the Hub pages, we have functionality for a `deprecated.md` that'll show prominently at the top of the description (e.g. https://github.com/docker-library/docs/pull/2205 and https://hub.docker.com/_/centos). Feel free to word it as you wish. As seen in that PR we also recommend adding a "DEPRECATED" to the `README-short.md` so that it also shows up in search pages on Docker Hub.

tianon:
We're also happy to help with any of that, if you need or want it. 🖤

thresheek:
Thank you for a detailed explanation (as usual!)

@tianon I think it makes sense for me to do since it was me who introduced the image in the Library in the first place :) Full circle, all that...

thresheek:
Thank you, I think it's all done now.

PR #20083, opened 2025-10-13
Retire Unit images.

thresheek:
Refs: https://github.com/docker-library/official-images/issues/20039


tianon:
:cry: :saluting_face:

PR #20097, opened 2025-10-15
Update UBI pins

yosifkit:


tianon:
More media type switching (https://github.com/docker-library/official-images/pull/20052#issuecomment-3398410452), but this is also cute:

```diff
-			"name": "ubi9-minimal",
+			"name": "ubi9/ubi-minimal",
```

tianon (approved):

PR #20106, opened 2025-10-17
Update Semeru containers to pick tomcat version v9.0.111

jayasg12:


yosifkit:
I think this was superseded by https://github.com/docker-library/official-images/pull/20184. So, this can be closed.

PR #20115, opened 2025-10-19
Add clickhouse tests

GSokol:
Tests https://github.com/ClickHouse/ClickHouse/pull/88042 and another Clickhouse image.


tianon:
Instead of adding more tests here, I'd suggest updating the implementation in https://github.com/ClickHouse/ClickHouse/blob/66fca799cc057332c959ad1e60fe42bfac88fb84/ci/jobs/docker_server.py#L283 to make use of the "multiple config files" functionality in our test wrapper: https://github.com/docker-library/official-images/tree/master/test#alternate-config-files

https://github.com/docker-library/bashbrew/blob/e6e3a3dc22d66aa7a7c9057fca86c35a45b167f6/scripts/github-actions/generate.sh#L113-L114

https://github.com/docker-library/bashbrew/issues/46#issuecomment-1152567694

Then, additional tests can live in the directory of the alternate config file (and y'all can maintain them directly to your heart's content).

(We're trying to shrink the set of tests we run and maintain directly here, especially tests that only apply to a single image, so that we only have things that are generic or apply to multiple implementations of a given thing like Java.)

GSokol:
What do you think, @Felixoid?

I like @tianon 's suggestion to override the test config with extra file, rather then patching it with `sed`, like in ClickHouse done.

Also I would leave `basic` and `initdb` here (similar to `mysql` tests), but move specific `override-tcp-port` to ClickHouse repo.

Felixoid:
There are quite a few topics here =)

> Instead of adding more tests here, I'd suggest updating the implementation in https://github.com/ClickHouse/ClickHouse/blob/66fca799cc057332c959ad1e60fe42bfac88fb84/ci/jobs/docker_server.py#L283 to make use of the "multiple config files" functionality in our test wrapper: https://github.com/docker-library/official-images/tree/master/test#alternate-config-files

My final aim was to protect the official and our images from obvious errors, so we'd have something similar to https://github.com/docker-library/official-images/blob/0db5856b7b2b4eb7fb98ec3baa720c72a7ea0e50/test/config.sh#L205-L208

My thoughts were that the place for tests would be very organic here.

What is the destiny of tests for other DBs like percona, mariadb, postgres, and others?

GSokol:
@tianon, @Felixoid, left only simple tests here, moved `clickhouse-override-tcp-port` test to ClickHouse repo PR.

tianon:
Ah right, to maintain these for both [`clickhouse`](https://github.com/ClickHouse/docker-library) _and_ [`clickhouse/clickhouse`](https://github.com/ClickHouse/ClickHouse) you'd have to copy them, do some submodule hacks, or clone them during CI in one place or the other, hmm.

Given that https://github.com/ClickHouse/docker-library only contains releases, it would probably be pretty reasonable for it to clone and test them from https://github.com/ClickHouse/ClickHouse at either latest `HEAD` or the latest release during commits there in GHA, right?

(either way it would make sense to add some kind of CI on https://github.com/ClickHouse/docker-library to at least do a basic `docker build`, and that'd be a pretty natural place to inject tests like this)

tianon:
> What is the destiny of tests for other DBs like percona, mariadb, postgres, and others?

For percona, mariadb, mongodb, and mysql, those tests apply to multiple images (mysql tests across all the mysql variations and mongodb tests across the percona mongodb implementation too).

For postgresql, I'd ultimately like to move those tests to https://github.com/docker-library/postgres instead, but there's some complexity there that we haven't been able to prioritize untangling.

In general, our test harness/framework is "good enough", but we're not particularly happy with it.

**Edit:** see https://github.com/docker-library/ghost/pull/384 for a concrete example of that (harder to move an existing test that's using reusable bits than it is to green-field a new test)

GSokol:
@tianon, so you do not want Clickhouse tests here at all?

tianon:
Ideally I would like to avoid that, yes, but I'm sympathetic to the integration struggles and so this isn't a hard block (just a request that we discuss all the possibilities before we commit for sure to _this_ particular shape). :+1:

GSokol:
@Felixoid, looks like https://github.com/ClickHouse/docker-library is fully auto-generated. Could CI run the tests on the new image definition before publishing it there?

GSokol:
Moved the tests clickhouse main repo (https://github.com/ClickHouse/ClickHouse/pull/88882).

Felixoid:
Thanks, having all the tests in `ClickHouse/ClickHouse` is looking natural to me, given that we have all the default tests from this repo + image-specific tests in our.

GSokol:
@tianon, @yosifkit, now we are struggling to using the test harness with our custom tests: we faced the problem of building cross-platform images with `docker-build.sh` (see: [the logs](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=85663&sha=3bc3ae87df2866115d2790c302c8b4ab32788f84&name_0=PR&name_1=Docker%20server%20image&name_1=Docker%20server%20image)).

---

Not pretty sure, that's the proper place to discuss, but what do you think if I extend `docker-build.sh`, making it `docker inspect` / `pull` step **optional** and `docker build` -- **overridable**?

---

For sure, I would do it in a separate PR, just want to have your opinion in advance.

tianon:
Oof, yeah, that's a helper that hasn't really adapted to BuildKit / `docker buildx`, and the ways that it changes the way things work under the hood.

Your simplest answer might be to set `DOCKER_DEFAULT_PLATFORM` on your emulated build job, although that's generally not recommended. :sweat_smile:

Felixoid:
We got rid of image rebuilding in favor of `-v` binding, and now the tests work fine with emulation too. Thanks!

PR #20156, opened 2025-10-24
library/perl: 5.43.2 -> 5.43.4

zakame:
- https://github.com/Perl/docker-perl/pull/199
- http://nntp.perl.org/group/perl.perl5.porters/270412


zakame:
Thanks @yosifkit 🙇

PR #20185, opened 2025-10-30
Official Convertigo 8.3.10 release!

nicolas-albert:
Thx!


yosifkit:
> Keys SHOULD be named using a reverse domain notation - e.g. `com.example.myKey`

Image label names should adhere to the image spec: https://github.com/opencontainers/image-spec/blob/v1.1.1/annotations.md. I think `org.opencontainers.image.authors` is the closest pre-defined key to maintainer.

We have some annotations that we populate during the build: https://github.com/docker-library/official-images/issues/3540#issuecomment-2715158488

tianon:
(FWIW, I'd personally suggest just removing this metadata entirely :heart: :eyes:)

nicolas-albert:
Label dropped! 👍

PR #20186, opened 2025-10-30
Haskell: bump ghc 9.10 to 9.10.3 + refreshed Dockerfiles

Totto16:
See https://github.com/haskell/docker-haskell/pull/166

Also since the last commit a new image `9.10.3-bookworm` and `9.10.3-slim-bookworm` was added, chnaging the `9.10` tag to point to bookworm wouldn't be a good idea, so I didn't do that

### DIFF
--- a/library/haskell
+++ b/library/haskell
@@ -12,6 +12,16 @@ Architectures: amd64, arm64v8
 GitCommit: d478ceb8ff33f6208db87008e48cd084fd9747a1
 Directory: 9.12/slim-bookworm
 
+Tags: 9.10.3-bookworm, 9.10-bookworm
+Architectures: amd64, arm64v8
+GitCommit: d478ceb8ff33f6208db87008e48cd084fd9747a1
+Directory: 9.10/bookworm
+
+Tags: 9.10.3-slim-bookworm, 9.10-slim-bookworm
+Architectures: amd64, arm64v8
+GitCommit: d478ceb8ff33f6208db87008e48cd084fd9747a1
+Directory: 9.10/slim-bookworm
+
 Tags: 9.10.3-bullseye, 9.10-bullseye, 9-bullseye, bullseye, 9.10.3, 9.10
 Architectures: amd64, arm64v8
 GitCommit: d478ceb8ff33f6208db87008e48cd084fd9747a1


Totto16:
> Since you've updated commit hashes for the rest of GHCs, I'd suggest to mention it in the PR description/title, something along the lines of "refreshed Dockerfiles".
> 
> Also `latest` are subject to eventual upgrade anyway, so it's fine by me to move it forward every so often

I just did the same as https://github.com/docker-library/official-images/pull/19733, which also updated all commit ids 😓

tianon:
Unfortunately the nature of our build system requires that we parse all our `Dockerfile`s to determine build ordering, and that parsing does not support the BuildKit heredocs syntax.

As a meta-point, updating the commit for everything is totally fine and reasonable for image maintainers; I think the point was just that it should be noted that the PR does more than _just_ the version bump in the PR description. :smile:

If it's easier to maintain that way, you can make that even easier by dropping all the individual `GitCommit:` lines and specifying a single `GitCommit:` line up at the top in what we call the "global entry" right next to `GitRepo:` instead. :eyes:

(For repositories we maintain, we generally automate the generation of this file to help us avoid typos in things like version numbers. :sweat_smile:)

Totto16:
@tianon @develop7 So do I need to change something here, or is it fine as is?

> Also latest are subject to eventual upgrade anyway, so it's fine by me to move it forward every so often

So would it be better to adjust the labels `9.10.3, 9.10` to the bookworm image or not?

It would be neat to at least get 9.10.3 available, so that the stackage [lts-24.12](https://www.stackage.org/lts-24.12) and upwards can be used 😄

yosifkit:
> Unfortunately the nature of our build system requires that we parse all our Dockerfiles to determine build ordering, and that parsing does not support the BuildKit heredocs syntax.

> So do I need to change something here, or is it fine as is?

The Dockerfiles need to be changed because our build system does not support Dockerfiles with heredoc syntax (e.g. `RUN <<EOT`).

develop7:
@Totto16 I've reverted heredocs in https://github.com/haskell/docker-haskell/pull/167 and regenerated `Dockerfile`s, so all you got to do is update this PR with new commit ID (`3a8f59b4ded7ad43be51fd61bace7be4298ad568`)

Totto16:
> @Totto16 I've reverted heredocs in [haskell/docker-haskell#167](https://github.com/haskell/docker-haskell/pull/167) and regenerated `Dockerfile`s, so all you got to do is update this PR with new commit ID (`3a8f59b4ded7ad43be51fd61bace7be4298ad568`)

Oh, that was the problem, in the comment from @tianon  I didn't really understand, that
> Unfortunately the nature of our build system requires that we parse all our Dockerfiles to determine build ordering, and that parsing does not support the BuildKit heredocs syntax.

Was saying, that the upstream dockerfiles were "wrong", my bad 😓

yosifkit:
```diff
-        *) echo >&2 "error: unsupported architecture '$ARCH'" ; exit 1 ;; \
+        *) echo >&2 "error: unsupported architecture '$ARCH'" exit 1 ; \
```
This change is incorrect. The new version will no longer exit and instead just echo the "exit 1". Also, the double semicolon is needed to end the case match correctly.

develop7:
@yosifkit my bad, will fix soon

develop7:
@Totto16 fixed in https://github.com/haskell/docker-haskell/commit/d478ceb8ff33f6208db87008e48cd084fd9747a1 please update the PR

yosifkit:
```diff
     GHC_URL="https://downloads.haskell.org/~ghc/$GHC/ghc-$GHC-$ARCH-deb12-linux.tar.xz"; \
-    # sha256 from https://downloads.haskell.org/~ghc/$GHC/SHA256SUMS
+    # sha256 from https://downloads.haskell.org/~ghc/$GHC/SHA256SUMS \
     case "$ARCH" in \
         'aarch64') \
-            GHC_SHA256='bee95bc91a621d8a2e9a9d86dac28ff839605e87316518dae12c779709bd58f1'; \
+            GHC_SHA256='052789dfe7f6fba6dc3822de0da272e8a5bd358c37adae17d8e82cff39bc1008'; \
+            GHC_URL='https://downloads.haskell.org/~ghc/9.10.3/ghc-9.10.3-aarch64-deb11-linux.tar.xz'; \
```
Just a reminder for myself that using the `deb10` tar in `deb11` or `deb11` in `deb12` is on purpose when there isn't a build available for that suite (and is still [noted](https://github.com/haskell/docker-haskell/blob/d478ceb8ff33f6208db87008e48cd084fd9747a1/9.10/slim-bookworm.yaml#L18-L20)).

```diff
-            # Note: the 'aarch64-deb11' (bullseye) variant of images doesn't exist in https://downloads.haskell.org/~ghc/9.6.7/
-            # But deb10 (buster) images should be binary compatible with deb11, so using deb10 instead
-            GHC_URL="https://downloads.haskell.org/~ghc/$GHC/ghc-$GHC-$ARCH-deb10-linux.tar.xz"; \
             GHC_SHA256='3cfa843687856de304a946dbe849a497c4fdad021f0275628b8ca7b55ccf8082'; \
+            GHC_URL='https://downloads.haskell.org/~ghc/9.6.7/ghc-9.6.7-aarch64-deb10-linux.tar.xz'; \
```

yosifkit (approved):

PR #20201, opened 2025-10-31
Update python

yosifkit:
Changes:

- https://github.com/docker-library/python/commit/38aab6a1: Update 3.9 to 3.9.25

This should be the final 3.9 release; it is end of life now: https://devguide.python.org/versions/


tianon:
> This should be the final 3.9 release; it is end of life now: https://devguide.python.org/versions/

Additionally confirmed via https://www.python.org/downloads/ :muscle:

> end of life, last release was [3.9.25](https://www.python.org/downloads/release/python-3925/)

PR #20216, opened 2025-11-04
Redis: Update to 8.4-rc1

adobrzhansky:


tianon:
FWIW, it's probably worth looking at upgrading to Trixie in the future :+1:

tianon (approved):

PR #20219, opened 2025-11-04
Updates for release 25.0.0.11

wraschke:
And restore entries for 25.0.0.3 images; these are again mentioned in https://[public.dhe.ibm.com/ibmdl/export/pub/software/websphere/wasdev/downloads/wlp/index.yml](https://public.dhe.ibm.com/ibmdl/export/pub/software/websphere/wasdev/downloads/wlp/index.yml).

### DIFF
--- a/library/open-liberty
+++ b/library/open-liberty
@@ -52,6 +52,36 @@ Directory: releases/latest/full
 Architectures: amd64, arm64v8, ppc64le, s390x
 File: Dockerfile.ubuntu.openjdk17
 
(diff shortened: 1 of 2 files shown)


tianon:
https://github.com/docker-library/official-images/pull/20064#issuecomment-3387096387

> it would be _really_ helpful if that value could be pre-calculated and embedded in the `Dockerfile` directly instead of calculated during build (for many reasons, the current obscure build failure being just one)

PR #20224, opened 2025-11-05
Updated swipl devel to 9.3.34

JanWielemaker:


yosifkit:
https://github.com/docker-library/official-images/pull/20075#issuecomment-3400764871

Does it look like the next `stable` will be out soon (or an update to the `9.2.9` image)? We get failing builds when attempting to rebuild `swiple:stable` (`9.2.9`) after a Debian update (like https://github.com/docker-library/official-images/pull/20212) that slow down our overall rebuild process. So, we'd like to have the image fixed as soon as possible.

JanWielemaker:
> Does it look like the next `stable` will be out soon (or an update to the `9.2.9` image)?

That is the plan.    Open issues are bothering though :cry:    Notably a number of features that would be nice to have but are not ready for prime time.    Fixing 9.2.9 is too much wasted effort.    I'll keep an eye on delaying the immature libraries, adding them somewhat later to the stable series as adding a library doesn't break anything.

PR #20273, opened 2025-11-12
Update UBI pins

yosifkit:


tianon:
Huh, that looks like a rollback from the dates, but the versions are increasing. :shrug:

PR #20307, opened 2025-11-17
Gradle 9.2.1

cobexer:


cobexer:
Diff LGTM

Note: for those that look very closely: UBI 10 images will not yet appear because of: https://github.com/gradle/docker-gradle/issues/374

yosifkit (approved):

PR #20315, opened 2025-11-18
Update Redis to 8.4.0

adobrzhansky:
Note: we haven't yet updated the bookworm to trixie since trixie is missing the `linux/mips64le` architecture


yosifkit:
> Note: we haven't yet updated the bookworm to trixie since trixie is missing the linux/mips64le architecture

Trixie will never have `mips64le`; it has been dropped as a release architecture for Trixie (https://github.com/docker-library/official-images/pull/19068#issuecomment-2896649696) and further has been dropped from future Debian releases by being removed from Debian Unstable as well (https://github.com/docker-library/official-images/pull/20212).

Edit: But Trixie does now have `riscv64` as a supported architecture.

yosifkit (approved):

PR #20328, opened 2025-11-20
update influxdb3 to 3.7.0

mgattozzi:
This commit bumps influxdb3 to our latest image 3.7.0

### DIFF
--- a/library/influxdb
+++ b/library/influxdb
@@ -6,7 +6,7 @@ Maintainers: Brandon Pfeifer <bpfeifer@influxdata.com> (@bnpfeife),
              Trevor Hilton <thilton@influxdata.com> (@hiltontj),
              Wayne Warren <wwarren@influxdata.com> (@waynr)
 GitRepo: https://github.com/influxdata/influxdata-docker
-GitCommit: 90a822e13f120254eff09740f8b16a2b2b275cc8
+GitCommit: 222eeaf138af2b69f346b8579816dcab39f00935
 
 Tags: 1.12, 1.12.2
 Architectures: amd64, arm64v8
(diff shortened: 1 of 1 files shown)


tianon:
```diff
+# Unset environment variables; splitting on whitespace
+# Usage: INFLUXDB3_UNSET_VARS="HOST FOO BAR"
+if [[ -n "${INFLUXDB3_UNSET_VARS:-}" ]]; then
+    read -ra vars <<< "${INFLUXDB3_UNSET_VARS}"
+    for var in "${vars[@]}"; do
+        unset "$var" || { echo "Error: Failed to unset variable '$var' (may be readonly)"; exit 1; }
+    done
+fi
```

This is kind of a wild feature addition - can you say more?  Are there cases where setting a variable to the empty value isn't enough?  I found https://github.com/influxdata/influxdata-docker/pull/838, and it links to https://github.com/influxdata/influxdb_pro/pull/1650, but the latter is private so we're a bit light on public details that I can find.

hiltontj:
@philjb originally made the change, and may comment, but the original description from that PR is:

> Allows unsetting environment variables in Docker containers via INFLUXDB3_UNSET_VARS, addressing the limitation that Docker Compose cannot unset variables already present in the image.
> 
> The script will fail if attempting to unset a readonly variable, providing feedback for misconfiguration.

philjb:
> Are there cases where setting a variable to the empty value isn't enough?

Yes, for influxdb3; setting an env var to an empty value isn't enough. The db will read that as meaning that `""` is the value the user wants to use for that env var config. To disable some features that are enabled in the image definition via `ENV` directives, the env var will need to be unset. This change to the entrypoint script allows power users to do that.

yosifkit (approved):

PR #20340, opened 2025-11-21
Add debian-latest tag

sithlord48:
- Add a `latest-slim` tag to the debian image. This new tag should point to the latest version of the `slim` release (currently trixie-slim)


tianon:
This is a duplicate of https://github.com/debuerreotype/docker-debian-artifacts/issues/205 (also https://github.com/debuerreotype/docker-debian-artifacts/issues/88) :sweat_smile:

I think `debian:stable-slim` is what you're looking for (or at least, as close to it as we currently/plan to support).

PR #20352, opened 2025-11-25
library/perl: 5.43.4 -> 5.43.5

zakame:
- https://github.com/Perl/docker-perl/pull/200
- http://nntp.perl.org/group/perl.perl5.porters/270513


zakame:
Thanks @yosifkit ! 🙇

PR #20362, opened 2025-11-25
Update Postfixadmin

J0WI:


yosifkit:
🤔 While I am pretty sure that copying composer from the composer image is architecture-agnostic, none of the build system or image definition support that. So, adding the `composer` image to the hierarchy means that `arm32v5` can't be used since it isn't available on the `composer:2` image

yosifkit (approved):

PR #20366, opened 2025-11-26
Add MistServer 3.11

junojense:
# Checklist for Review

	- repo under https://github.com/DDVTECH (https://github.com/DDVTECH/mistserver is upstream)
	- "unlicense" (public domain)
	- service
	- https://github.com/docker-library/docs/pull/2640
-	[ ] passes current tests? any simple new tests that might be appropriate to add, [assuming this image has any other images it might be interchangeable with](https://github.com/docker-library/official-images/tree/HEAD/test#what-tests-belong-here)?

## Extra Information
I am associated with upstream, the forked repository is contained in the official MistServer organisation. Our software is public domain, and is "licensed" under the [unlicense](https://opensource.org/license/unlicense). The current image on Docker hub builds off alpine and can be found [here](https://hub.docker.com/r/ddvtech/mistserver). The tests seem to be passing. I have also created a [documentation PR](https://github.com/docker-library/docs/pull/2640) which passes the markdown formatter.

### DIFF
--- a/library/mistserver
+++ b/library/mistserver
@@ -1,0 +1,15 @@
+Maintainers: Jaron Viëtor <jaron.vietor@ddvtech.com> (@Thulinma),
+             Marco van Dijk <marco.van.dijk@ddvtech.com> (@stronk-dev),
+             Carina van der Meer <carina.van.der.meer@ddvtech.com> (@thoronwen),
+             Balder Viëtor <balder.vietor@ddvtech.com> (@Rokamun),
+             Ramkoemar Bhoera <ramkoemar.bhoera@ddvtech.com> (@ramkoemar),
+             Juno Jense <unit-stamp-sled@duck.com> (@junojense)
+GitRepo: https://github.com/DDVTECH/mistserver-docker-builder.git
+GitFetch: refs/heads/main
+GitCommit: a0d4de02e3f9c3044a652cf6cbd56d4aba8c0315
+Builder: buildkit
+
+Tags: latest, 3.9.2
+Architectures: amd64, arm64v8
+Directory: 3.9.2
+File: Dockerfile.mistserver
--- a/library/mistserver
+++ b/library/mistserver
@@ -6,10 +6,9 @@ Maintainers: Jaron Viëtor <jaron.vietor@ddvtech.com> (@Thulinma),
              Juno Jense <unit-stamp-sled@duck.com> (@junojense)
 GitRepo: https://github.com/DDVTECH/mistserver-docker-builder.git
 GitFetch: refs/heads/main
-GitCommit: a0d4de02e3f9c3044a652cf6cbd56d4aba8c0315
-Builder: buildkit
+GitCommit: 6ad8faed85bf59bc95173fd9832f1867df55a278
 
-Tags: latest, 3.9.2
+Tags: latest, 3.11.1
 Architectures: amd64, arm64v8
-Directory: 3.9.2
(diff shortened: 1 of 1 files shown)


junojense:
Using a Docker-specific repository now that contains Dockerfiles for different versions, instead of directly including the source leading to jobs failing. The PR should be ready for review.

junojense:
Hi again, can this PR be looked at?

tianon:
I've taken a first pass at review, and these are my very high-level notes from it:

At first glance, the multi-stage build doesn't seem to be one of the cases in https://github.com/docker-library/faq#multi-stage-builds (using `apk add --virtual` and something like `scanelf` to find the runtime deps and keep them afterwards would probably work really well).

That's a lot of `ARG` - are all these really meaningful? Should their values be documented better? (Our build tooling won't ever use them, so they're mostly just noise here.)

Explicitly using `Builder: buildkit` isn't really necessary (the `TARGETPLATFORM` value it seems to be used to get isn't something we recommend relying on, preferring instead to see userspace/runtime detection based off things like `apk --print-arch`).

Building mbedtls from source is a little eyebrow-raising -- Alpine contains the same version 3.6.5, is there a reason it isn't used? (https://pkgs.alpinelinux.org/package/v3.23/main/x86_64/mbedtls-dev)

Having such _long_ single-line `RUN` lines is really hard to read/review -- those should be split up with `\` and newlines.

Using `FROM alpine` with no specific version is going to have a large tendency to break when the "square wheel" rolls, so we highly recommend pinning to an explicit release of Alpine like `FROM alpine:3.23` so that you can more directly control/manage the rolling of that wheel (and the breakage that may or may not come from it).

Using `LABEL` is not recommended, as the inheritance behavior of it is really poor; see https://github.com/docker-library/official-images/issues/3540, especially https://github.com/docker-library/official-images/issues/3540#issuecomment-2715158488 (our build system automatically adds appropriate similar annotations).

Using `HEALTHCHECK` is also not something I'd personally recommend setting on the image (https://github.com/docker-library/faq#healthcheck) -- perhaps this command should be included in the documentation instead, so that users who want its behavior can get it easily?  (in either a healthcheck in Docker or an appropriate probe in Kubernetes, for example, which won't read `HEALTHCHECK` metadata anyways)

Ideally the download of the source code would have some kind of verification, but that's a little hard with the source code tarball being generated by GitHub directly. 😔

junojense:
Thanks for the feedback @tianon, all points have been addressed in the updated Dockerfile.

- Got rid of the multi-stage build, we now use `scanelf` to find run-time deps
- Significantly reduced `ARG` usage, the remaining arguments are...
    - `MIST_DEBUG` sets the MistServer logging level to production instead of development
    - `MIST_OPTS` space separated compile options as specified [here](https://docs.mistserver.org/mistserver/installation/compile#compile-flags)
- Removed `Builder: buildkit` in favour of `apk --print-arch`
- Removed `HEALTHCHECK`, `LABEL`, improved the formatting of `RUN`, and pinned the alpine version

To elaborate on building mbedtls from source; we require DTLS support which is not present in mbedtls official builds / releases. Though, the retrieved source code is now at least verified with its SHA-256 checksum.

As for verification of MistServer itself, our newest release (3.11.1) includes [downloads](https://mistserver.org/download) for the source code with matching SHA-256 checksum that we check against in the new Dockerfile. 

As everything mentioned has been addressed, the PR should be ready for further review.

PR #20375, opened 2025-11-28
Gradle: Add UBI10 images and improve all UBI tags

cobexer:
* Add UBI10 images to Gradle 9 Java 17 and Java 21
* Fix all images to be tagged with `-ubi9` or `-ubi10` instead of `-ubi-minimal`


cobexer:
Diff LGTM!

The copy detection on the bottom picked weird candidates for renames, but the added diff there makes sense for the tag it's in. Thus I think this is correct.

tianon:
Slowly creeping towards the combinatorial matrix-from-hell that is https://github.com/docker-library/tomcat/pull/299 :joy: :sob: :heart:

PR #20379, opened 2025-12-01
Update UBI pins

yosifkit:


tianon:
10-minimal but not 10? :thinking:

PR #20400, opened 2025-12-03
alpine: add 3.23.0 as latest, remove EOL 3.19

ncopa:

PR #20408, opened 2025-12-04
Drop Fluentd v1.16 alpine image

kenhys:
alpine 3.19 had alreay reached EOL.

See https://alpinelinux.org/releases/

NOTE: Fluentd v1.16 alpine image will not follow newer alpine image, so just drop it.


kenhys:
See https://github.com/docker-library/official-images/pull/20400#issuecomment-3608482434

tianon (approved):

PR #20466, opened 2025-12-10
add InfluxDB 2.8.0 and update latest tag

gwossum:
Add InfluxDB 2.8.0 and 2.8 tags. Update latest tag to 2.8.0.

### DIFF
--- a/library/influxdb
+++ b/library/influxdb
@@ -6,7 +6,7 @@ Maintainers: Brandon Pfeifer <bpfeifer@influxdata.com> (@bnpfeife),
              Trevor Hilton <thilton@influxdata.com> (@hiltontj),
              Wayne Warren <wwarren@influxdata.com> (@waynr)
 GitRepo: https://github.com/influxdata/influxdata-docker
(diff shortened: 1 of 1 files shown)


yosifkit:
Now that it appears in the diff, the `gosu` version is a bit out of date. `1.19` is the [latest](https://github.com/tianon/gosu/releases). Besides the perception of security fixes, the binaries are a little smaller. Do you want to change that now?

gwossum:
I'll add a commit to take gosu to version 1.19.

tianon:
Is 2.7 still actively maintained with the release of 2.8?  (if not, it should also be removed)

gwossum:
> Is 2.7 still actively maintained with the release of 2.8? (if not, it should also be removed)

There are no plans for another 2.7 release, but I wanted to keep it open until 2.8 gets more testing from end-users. If 2.8.0 doesn't have any issues, I expect removal of 2.7 when 2.8.1 or 2.8.2 is released.

PR #20521, opened 2025-12-18
Update ELK images

navyau09:
Update ELK images


jsvd:
@tianon indeed :) but our app is already in Java, and this golang launcher was only a source of (mostly false positive) CVE findings at this point.

PR #20527, opened 2025-12-18
[security] Update php

tianon:
Changes:

- https://github.com/docker-library/php/commit/43c273b5: Update 8.5-rc
- https://github.com/docker-library/php/commit/9ed36f43: Update 8.5 to 8.5.1
- https://github.com/docker-library/php/commit/a43ea9bd: Update 8.3-rc
- https://github.com/docker-library/php/commit/4ec8d1a2: Update 8.3 to 8.3.29


tianon:
(8.2 is also in progress)

yosifkit:
> (8.2 is also in progress)

It'll be another hour or so before the update job completes, so we'll get these versions started in the DOI build pipeline and open another PR once it is ready.

PR #20552, opened 2025-12-24
alpine: bump edge snapshot to 20251224

ncopa:


tianon (approved):
```diff
-alpine:20250108
+alpine:20251224
 alpine:edge
```

Whew, that's quite the gap :sweat_smile:  Thanks for the update!

PR #20558, opened 2025-12-26
Update ruby

tianon:
Follow up to https://github.com/docker-library/official-images/pull/20555 (not sure if that wasn't generated via the script or if there's a glitch we should look into, cc @harshalbhakta)

### DIFF
--- a/library/ruby
+++ b/library/ruby
@@ -9,27 +9,27 @@ Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, ppc64le, riscv64, s390x
 GitCommit: 12f8e3ca39458c5168b8736a47a68fcc7a047a02
 Directory: 4.0/trixie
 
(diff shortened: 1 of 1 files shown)


LaurentGoderre (approved):

harshalbhakta:
> Follow up to #20555 (not sure if that wasn't generated via the script or if there's a glitch we should look into, cc @harshalbhakta)

I could not find any PR created by the bot so made changes manually be referring to old PRs.

tianon:
> I could not find any PR created by the bot so made changes manually be referring to old PRs.

Ah, that makes sense -- for future reference, @docker-library-bot doesn't ever make PRs directly.  It creates branches at https://github.com/docker-library-bot/official-images/branches/all and we review them and open them manually, especially because (as you can see in that list) we'll often review the changes and decide _not_ to open the PR.

However/additionally, the contents of `library/ruby` are canonically generated by the `generate-stackbrew-library.sh` script in the `docker-library/ruby` repository (as noted in the comment at the top of the file 😅). :+1:

PR #20562, opened 2025-12-29
influxdb: Update meta and data tags with no version to point to latest

devanbenz:


tianon:
FWIW, these were removed (what looks like intentionally) in https://github.com/docker-library/official-images/pull/15039 -- just to be sure, adding them back is intentional too, right?  (explicit cc @bnpfeife)

bnpfeife:
@tianon Yup, adding the tags back is intentional. In https://github.com/docker-library/official-images/pull/15039, I removed the 1.8 images to minimize CI costs (since that PR adds the 1.11 images). Since 1.9 and 1.10 were already released, I ended up believing that those tags were intentionally pinned at 1.8.

PR #20599, opened 2026-01-08
Update PSMDB release tags

surbhat1595:


vorsel:
Hi @tianon 
We are going to release PSMDB 6.0 with CVE fix next week.

PR #20649, opened 2026-01-15
Update rabbitmq

tianon:
Changes:

- https://github.com/docker-library/rabbitmq/commit/18c229a: Update 4.2 to 4.2.2
- https://github.com/docker-library/rabbitmq/commit/0f723db: Update 4.2 to 4.2.1


tianon:
oop, that burger has nothing inside

PR #20686, opened 2026-01-21
Update to Flink 2.2.0

ruanhang1993:


yosifkit:
There is the possibility that this change might break users relying on `root`, but the entrypoint of the image wasn't doing anything with it (like `chown` or `chmod` that is often done for usability in other images).

The main users I would think to be temporarily broken by this will be those installing packages or other customizations in an image `FROM flink:[version]` that can just add `USER root` before and `USER flink` after their customizations.

Issue #20653, opened 2026-01-15
[RFC] varnish packages


gquintard:
Happy new year everyone!

I'm seeking some guidance regarding https://github.com/varnish/docker-varnish. So far, we've just build creating a Varnish packages from source, and then bluntly `make install`ing extra vmods (plugins), and it works well.

However, going forward, we want to serve our community better and provide easily installable packages for a bunch of out-of-core vmods and other tools. So I've set up https://github.com/varnish/all-packager. It really just a packaging pipeline that uses `github actions` and some [metadata](https://github.com/varnish/all-packager/blob/main/pkg.env) to build and (soon) push packages automatically to a repository.

The big benefit is that users can easily create a `Dockerfile` and just `apt-get update; apt-get install` extra vmods that weren't in the base image. And for sanity's sake, I want the vmods in the official image to be properly packaged too, which will prevent issues when a user tries to install a vmod package over the `make install`ed one.

And so, I have a menu of options:
- just have the `Dockerfile` install all the packages created and published by `all-packager`, but I know it's not the preferred way as we want as much reproducibility and auditability as possible
- I can git clone `all-packager` and do everything from it, and the scripts there will take care of all the source downloads and checksum verification, but that's possibly a lot of code that's not local
- I could either snapshot the `all-packager` commit, or use a `git submodule` (would the pipeline support it), but that feels like a big duplication of code

I'm thinking I'm not the first one having this challenge, so I'm coming to you for some advice :-)

tianon:
Our stance is usually that if you (as upstream) are publishing official pre-built packages, and recommend that users should use them, then those are what you should install in your official image too. :smile:

(I think that's essentially your first option?)

What we ask is that you use pinning to make sure that when we build `varnish:1.2.3`, it either includes Varnish version 1.2.3 _or_ fails to build (and if there's PGP verification involved, that the keys are fetched in a secure and verified way, which I'm happy to say more about if needed). 👍

gquintard:
awesome, that makes things A LOT easier then. I will set that up!

tianon:
Happy to help or consult further in any way that might be helpful to you! 🫡❤️

PR #20688, opened 2026-01-21
varnish: use prebuilt packages

gquintard:
following the discussion in #20653 (thanks again for the prompt advice!) this PR pivots the `fresh` `debian` image to use https://varnish.org/docs/install-guide/install-debian-ubuntu/ (built using https://github.com/varnish/all-packager)

### DIFF
--- a/library/varnish
+++ b/library/varnish
@@ -1,33 +1,33 @@
-# this file was generated using https://github.com/varnish/docker-varnish/blob/351d45e8503d1345369fb3beb43d8dac4383757c/populate.sh
+# this file was generated using https://github.com/varnish/docker-varnish/blob/9d19c601591be47c263b2390b1ac1701b6065f55/populate.sh
 Maintainers: Guillaume Quintard <guillaume.quintard@gmail.com> (@gquintard)
 GitRepo: https://github.com/varnish/docker-varnish.git
 
 Tags: fresh, 8.0.0, 8, 8.0, latest
-Architectures: amd64, arm32v7, arm64v8, i386, ppc64le, s390x
+Architectures: amd64, arm64v8
 Directory: fresh/debian
-GitCommit: 68b666606150f9aa637a23d4358906eb6a409252
+GitCommit: 8bac2b1f964998f3867aa5e338fedf25a5b9ff1c
 GitFetch: refs/heads/main
 
 Tags: fresh-alpine, 8.0.0-alpine, 8-alpine, 8.0-alpine, alpine
-Architectures: amd64, arm32v7, arm64v8, i386, ppc64le, s390x
+Architectures: amd64, arm64v8
 Directory: fresh/alpine
-GitCommit: 68b666606150f9aa637a23d4358906eb6a409252
+GitCommit: 9d19c601591be47c263b2390b1ac1701b6065f55
 GitFetch: refs/heads/main
 
 Tags: old, 7.7.3, 7.7
-Architectures: amd64, arm32v7, arm64v8, i386, ppc64le, s390x
+Architectures: amd64, arm64v8
 Directory: old/debian
-GitCommit: 68b666606150f9aa637a23d4358906eb6a409252
+GitCommit: b341feabd8feeafaa2d25d35fdf44eb27a2eaaa1
 GitFetch: refs/heads/main
 
 Tags: old-alpine, 7.7.3-alpine, 7.7-alpine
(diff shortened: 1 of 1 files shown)


gquintard:
sorry, that was painful, `git` is hard...

tianon:
```diff
+COPY varnish.pub.asc /etc/varnish/
```

You might find https://github.com/docker-library/faq#openpgp--gnupg-keys-and-verification to be helpful information :sweat_smile:

I'm happy to help write something different here, if that would be useful to you (or point to other examples of doing similar things).

---

```diff
+    apt update; \
```

As noted by the warning it spits out when used like this, `apt` is not yet suitable/intended for programmatic usage, so this should still be `apt-get`.

---

```diff
+    cat /etc/varnish/varnish.pub.asc | gpg --dearmor | tee /etc/apt/keyrings/varnish.gpg > /dev/null; \
```

If the file ends in `.asc`, APT will happily use it without dearmoring (I personally prefer to make all my APT keyrings be `.pgp.asc` even if I create them as explained in the FAQ in my first note here, because they're just so much easier to deal with than binary file keyrings are).

---

```diff
+    apt-get install -y --no-install-recommends {varnish,varnish-dev,varnish-modules,vmod-fileserver,vmod-reqwest,vmod-rers}=${VARNISH_VERSION} varnish adduser libgetdns10t64 netbase; \
```

The bash-ism here is very cute :joy: (very minor note of caution that `SHELL` is an inherited property, so anyone doing `FROM varnish:xxx` might end up meeting up with Hyrum's Law in some way)

The explicit added packages on the end (`adduser libgetdns10t64 netbase`) are somewhat eyebrow-raising, and make me wonder if the dependencies on your upstream packages are missing something?

gquintard:
Hey! Thanks for the feedback!

The quick ones first:
- `apt`: yup, my bad. It's upsetting that I got it right a few lines above. 
-  dearmoring: I wasn't aware of that, I'll switch to pure text, for now...
- `SHELL`: this is a remnant of when I was trying to build the packages in the `Dockerfile`, it's not needed anymore (even though `source` feels a lot cleaner than `.`, eh), I'm happy to get rid of it
- bash-ism: with `SHELL` gone, it's out too. To be fair, Irely on it so often I never realized it was a a `bash` specific feature...
- `adduser`: that's me being dense, I switched from `useradd` to avoid deps and forgot a step
- `libgetdns10t64` and `netbase`: they are dependencies for `vmod-dynamic` which is not packaged. We're just going to survive with those until March, and I'll get rid of the vmod altogether for the next release. Thanks to your comment,  I've been able to revisit the package list and removed a bunch of no longer needed dependencies

noooooow, for the painful point: the key. I do remember initially that we used a key server, but I remember no details from it, only the lingering PTSD :-) No excuse though, I should have RTFM. I will look into it, but I will happily take all the help/pointers you will offer.

gquintard:
@tianon, how firm are you on the key? I will work on it, but it's probably take some time

tianon:
You've still got one `SHELL []` in there that I don't think you meant to keep :eyes:

For the key, I think an acceptable compromise would be a checksum (for now) -- this maybe isn't ideal forever, given that the key likely has an expiration date so will need to be cycled eventually even if only to update the expiration (but maybe it's fine and you just keep on top of it because that's probably once a year at the most often and more likely once every two), but it's Good Enough that gives a chain of provenance back to the original artifact:

```dockerfile
	curl -fL -o /etc/apt/keyrings/varnish.pub.asc 'https://packages.varnish-software.com/varnish/varnish.pub.asc'; \
	echo 'eea1110e5aa0afc0e8cfe872f0f5bbf0778d95e811227e19835c02378dd03ff2 */etc/apt/keyrings/varnish.pub.asc' | sha256sum --strict --check -; \
```

gquintard:
I think I'm somewhat happy with this. It's currently using the ubuntu key servers and I want to move away from them eventually, but it can wait a bit

tianon:
Two minor notes:

- `apt-transport-https` has been transitional / unnecessary for quite a long time now (there is no supported version of Debian where it isn't, in fact :sweat_smile:)

- this leaves `~root/.gnupg` around with the key in it, which is harmless, but maybe something you want to clean up?

neither of these are critical and I'm happy with merging as-is if you want :+1:

gquintard:
duly noted, let merge it as-is. I've sorted out the key stuff, plus a few other extra changes, but let's make it a new PR if that's ok with you

PR #20694, opened 2026-01-22
Groovy 5.0.4

keeganwitt:


yosifkit:
Since the `groovy:4` image actually contains groovy 5.0.3, let us know if you'd like us to do something with it. We could delete that tag from Docker Hub, or rebuild it (just re-add some of the `4.0.27-jdk21-noble` section of the `library/groovy` file), or just leave it.

PR #20699, opened 2026-01-22
Redis: Update to 8.6-rc1

Peter-Sh:
Hi Docker team!

Could you please take a look and approve this PR for Redis? We’re about to publish the 8.6 release candidate.

Release commit: a6fe2110712dbfa89f0fb933ca1d34d0771fa35d
Release tag: v8.6-rc1
Compare: https://github.com/redis/docker-library-redis/compare/v8.6-rc1^1...v8.6-rc1

@adamiBs @yossigo @adobrzhansky @maxb-io @dagansandler @Peter-Sh

### DIFF
--- a/library/redis
+++ b/library/redis
@@ -6,6 +6,18 @@ Maintainers: Adam Ben Shmuel <adam.ben-shmuel@redis.com> (@adamiBs),
              Petar Shtuchkin <petar.shtuchkin@redis.com> (@Peter-Sh)
 GitRepo: https://github.com/redis/docker-library-redis.git
 
+Tags: 8.6-rc1, 8.6-rc1-trixie
+Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, riscv64, ppc64le, s390x
+GitCommit: 275b71068b3065fcf794d08f908b61548f451795
+GitFetch: refs/tags/v8.6-rc1
+Directory: debian
+
+Tags: 8.6-rc1-alpine, 8.6-rc1-alpine3.23
+Architectures: amd64, arm32v6, arm32v7, arm64v8, i386, ppc64le, riscv64, s390x
+GitCommit: 275b71068b3065fcf794d08f908b61548f451795
+GitFetch: refs/tags/v8.6-rc1
+Directory: alpine
+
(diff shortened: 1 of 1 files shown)


tianon:
```diff
+ARG REDISEARCH_VERSION
+ARG REDISJSON_VERSION
+ARG REDISBLOOM_VERSION
+ARG REDISTIMESERIES_VERSION
+ARG CUSTOM_BUILD=false
```

Are all of these really necessary here?  They won't be used by the DOI builds, and the conditionals that use them do make the `Dockerfile` flow harder to follow, so I think we'd prefer to see these be part of your templating for your own builds (and removed for the `Dockerfile`s proposed to us for review / official building).

---

```diff
+# add the LLVM APT repository key once (vendored in this repository)
+COPY llvm-snapshot.gpg.key /etc/apt/trusted.gpg.d/apt.llvm.org.asc
```

Unfortunately this is not an acceptable way to embed PGP keys in builds (see https://github.com/docker-library/faq#openpgp--gnupg-keys-and-verification, but also my next point).

---

```diff
+	echo "deb http://apt.llvm.org/trixie/ llvm-toolchain-trixie-21 main" \
+			> /etc/apt/sources.list.d/llvm-toolchain.list; \
+		apt-get update; \
```

Looking at https://apt.llvm.org/, these builds are all _nightly_ builds of LLVM, not official releases (and as far as I've found, official binary releases of LLVM don't seem to exist), so I would strongly discourage using them for your official/stable builds.

Peter-Sh:
Hi @tianon!

Thanks for the feedback. I believe I’ve addressed all the issues you raised.

The custom build arguments have been moved out of the official Dockerfiles; they’re now conditionally included only in our internal workflows.

For Debian, we use Clang 19 as provided by the distribution.

Regarding Clang: if we eventually need a newer version to build the project, what would you recommend? What are the accepted approaches for this use case?

yosifkit:
>Regarding Clang: if we eventually need a newer version to build the project, what would you recommend? What are the accepted approaches for this use case?

I'd generally not recommend using a Clang version newer than that available in the newest stable Linux distributions because packagers will have a hard to impossible time shipping Redis on many Linux distributions.

PR #20713, opened 2026-01-26
Update bonita, add 2025.2-u3

abirembaut:


yosifkit:
It seems strange to add security via the pro-grade library when it hasn't had a release in over 10 years, but maybe my perspective is different, and the library is bug free and feature complete.

If `prograde.policy` is to increase security when running bonita, shouldn't it be included in the `bundle-tomcat-10.4.3.zip` so that non-docker users can also benefit from the policy?

abirembaut:
@yosifkit thank you very much for your feedback.

I totally agree that adding pro-grade may not be the best option, but it is what we currently have working in our other images. We are working toward finding another solution to prepare for Java 24, in which the Security Manager is removed.
However, this is a significant change that will require time, so the decision was made to secure the image with what we already have (and has proven reliable) in our other images.

Regarding including the policy in the non-Docker bundle: we initially focused on securing the Docker image, which has a more controlled environment. The non-Docker bundle is typically deployed in more varied setups where users have full control over JVM configuration. That said, this is a valid point: we can include our policy in the Tomcat bundle the same way we do in the Docker image. I'll target this for the next minor version.

PR #20724, opened 2026-01-27
Release SonarQube Server 2026.1.0

carminevassallo:
Dear team,

we would like to release new version of SonarQube Server: 2026.1.0

p.s.
We did also open [a related PR](https://github.com/docker-library/docs/pull/2650) on the `docker-library/docs` repo.

### DIFF
--- a/library/sonarqube
+++ b/library/sonarqube
@@ -5,22 +5,22 @@ Builder: buildkit
 
 Tags: 2026.1.0-developer, 2026.1-developer, developer, 2026-lta-developer
 Directory: commercial-editions/developer
-GitCommit: 3e2728fcf20f2f5420cdb87104bb557650b0d3ba
+GitCommit: a067605a9abfd0b56a9e19a63cc2636d5cd7a4e5
 GitFetch: refs/heads/master
 
 Tags: 2026.1.0-enterprise, 2026.1-enterprise, enterprise, 2026-lta-enterprise
 Directory: commercial-editions/enterprise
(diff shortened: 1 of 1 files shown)


tianon:
Switching from the JRE to the JDK images feels pretty heavy -- does the new version have runtime features that require the full JDK? :thinking:

Similarly, it seems weird to add `wget` while `curl` is already installed, is that intentional?  What's the use case for including both in an image designed for running a server process?

carminevassallo:
Hi @tianon,

Thanks for your comments!

> Switching from the JRE to the JDK images feels pretty heavy -- does the new version have runtime features that require the full JDK? 🤔

This change in our base image is required by an update of an internal dependency (ElasticSearch). We added a specific note about this in our [release notes](https://docs.sonarsource.com/sonarqube-server/server-update-and-maintenance/release-notes#java-requirements-for-sonarqube-server-runtime).

> Similarly, it seems weird to add wget while curl is already installed, is that intentional? What's the use case for including both in an image designed for running a server process?

We removed the `wget` installation.

When looking at the diff, you might wonder why we are adding `2026-lta-*` tags. This is because 2026.1 is gonna be our 2026 LTA release (as announced [here](https://www.sonarsource.com/products/sonarqube/sonarqube-server-20261-lta-is-coming/)).

I hope this addresses your question, please let me know if you have more. We would need to release our images asap.

yosifkit (approved):

PR #20728, opened 2026-01-27
varnish: update to 8.0.0-4

gquintard:
Also, add a few vmods, and clean up some cruft noticed in https://github.com/docker-library/official-images/pull/20688

### DIFF
--- a/library/varnish
+++ b/library/varnish
@@ -1,17 +1,17 @@
-# this file was generated using https://github.com/varnish/docker-varnish/blob/9d19c601591be47c263b2390b1ac1701b6065f55/populate.sh
+# this file was generated using https://github.com/varnish/docker-varnish/blob/3096b1557dce299636ef1eb770aec8b9ea9ce401/populate.sh
 Maintainers: Guillaume Quintard <guillaume.quintard@gmail.com> (@gquintard)
 GitRepo: https://github.com/varnish/docker-varnish.git
 
 Tags: fresh, 8.0.0, 8, 8.0, latest
 Architectures: amd64, arm64v8
 Directory: fresh/debian
-GitCommit: 8bac2b1f964998f3867aa5e338fedf25a5b9ff1c
+GitCommit: 3096b1557dce299636ef1eb770aec8b9ea9ce401
 GitFetch: refs/heads/main
 
 Tags: fresh-alpine, 8.0.0-alpine, 8-alpine, 8.0-alpine, alpine
 Architectures: amd64, arm64v8
 Directory: fresh/alpine
-GitCommit: 9d19c601591be47c263b2390b1ac1701b6065f55
+GitCommit: d80cb1ac78089dfa29cb2148c3b3cc5e196757c6
 GitFetch: refs/heads/main
 
 Tags: old, 7.7.3, 7.7


gquintard:
@tianon, thanks again for your time and patience, I really, really appreciate.

This adds a few things, but notably:
- cleans up `~root/.gnupg`
- don't install `apt-transport-https`
- switches the key server to keys.openpgp.org

tianon:
Nice!  I don't love the committed `index.html` file, but it is absolutely better than HTML embedded in VCL, so it's fine (it's straightforward, easy to review, and unlikely to change very much over time).

We'll need to keep an eye on that `rm -r ~/.gnupg` line because GnuPG has a pretty nasty little race condition with `dirmngr` dying that we might see in the builds periodically, but I don't mind rolling the dice and seeing how it does. :+1:

gquintard:
thanks!

For in `html` file, I could externalize it to `varnish/toolbox`, but it's pretty specific to the VCL at the moment. If it stays here, I'll try to keep the edits batched, and to a minimum.

For the `rm`, is there a better solution?

tianon:
We usually add `gpgconf --kill all` and `rm -f` as seen in the new documentation @yosifkit just cleaned up over in https://github.com/docker-library/faq#how-can-i-use-a-keys-file-for-verifying-pgp-signatures but it's also possible that it's OK as-is - we're OK with trying it as-is if you'd prefer to see how it goes and only adjust if/when we see it hitting the race. :+1:

gquintard:
perfectly fine with me. I'm refactoring the code a bunch these next few weeks, I'll make sure to switch to the cleaner `gpgconf --kill all` in the next PR. (I promise to not bother you daily though)

tianon:
Heh, `varnish:fresh` is now failing to build on arm64 :sweat_smile:

```console
+ apt-get update
Hit:1 http://deb.debian.org/debian trixie InRelease
Hit:2 http://deb.debian.org/debian trixie-updates InRelease
Hit:3 http://deb.debian.org/debian-security trixie-security InRelease
Hit:4 https://packages.varnish-software.com/varnish/debian trixie InRelease
Reading package lists...
+ apt-get install -y --no-install-recommends varnish=8.0.0-4~trixie varnish-dev=8.0.0-4~trixie varnish-modules=8.0.0-4~trixie vmod-cfg=8.0.0-4~trixie vmod-digest=8.0.0-4~trixie vmod-fileserver=8.0.0-4~trixie vmod-geoip2=8.0.0-4~trixie vmod-jq=8.0.0-4~trixie vmod-querystring=8.0.0-4~trixie vmod-redis=8.0.0-4~trixie vmod-reqwest=8.0.0-4~trixie vmod-rers=8.0.0-4~trixie vmod-uuid=8.0.0-4~trixie libgetdns10t64 netbase
Reading package lists...
Building dependency tree...
Reading state information...
Package vmod-cfg is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Version '8.0.0-4~trixie' for 'vmod-cfg' was not found
```

gquintard:
let me get the naughty stick

gquintard:
that one was actually on my, I forgot that we aren't packaging `vmod-cfg` on `arm`. Fixed in #20740, with the gpg` cleanup (no `gpgconf` installed, so I skipped it)

Is there some CI framework I can leverage to also build on `arm`, or should I set it up myself?

tianon:
I haven't used them myself for various reasons, but GitHub Actions supports arm64 now: https://github.blog/news-insights/product-news/arm64-on-github-actions-powering-faster-more-efficient-build-systems/

https://github.com/actions/partner-runner-images#available-images

gquintard:
indeed (we use them to build the `varnish` packages), but as the strategy in our `actions` is defined by `docker-library/bashbrew/v0.1.12/scripts/github-actions/generate.sh`, I'm thinking I need to post process that output locally to duplicate all entries and change the `os` field?

tianon:
Ahh, I see -- yeah, there might be something inspirational in https://github.com/docker-library/bashbrew/pull/80 and/or https://github.com/docker-library/bashbrew/pull/81. :thinking:

PR #20740, opened 2026-01-28
varnish: no vmod-cfg on arm

gquintard:


yosifkit:

```diff
+    rm -rf "$GNUPGHOME"; \
```
I don't see a corresponding setting or creation of `GNUPGHOME`. I think this is harmless (since deleting the empty string is fine) but probably needs a creation half to work as expected `GNUPGHOME="$(mktemp -d)"; export GNUPGHOME;`.

```diff
     apt-get install -y --no-install-recommends \
 				varnish=${VARNISH_VERSION} \
@@ -46,6 +50,7 @@ RUN set -ex; \
 				vmod-reqwest=${VARNISH_VERSION} \
 				vmod-rers=${VARNISH_VERSION} \
 				vmod-uuid=${VARNISH_VERSION} \
+				${EXTRA_VMODS} \
 				libgetdns10t64 \
 				netbase; \
```
I'd expect to see a line like this in there: `-				vmod-cfg=${VARNISH_VERSION} \`

gquintard:
Thanks, good point.

I'm a bit confused about `GNUPGHOME` though. I was previously clearing `~/.gnupg`, but the docs mention `GNUPGHOME`. Should I `mktemp` and export to force `gnupg` to use a temporary directory instead of using `~/.gnupg`?

yosifkit:
`~/.gnupg` is the default if `GNUPGHOME` isn't set. You can stick with deleting `~/.gnupg` or creating, setting, and cleaning up `GNUPGHOME`. The image's environment doesn't have `GNUPGHOME` set by default, so `rm -rf "$GNUPGHOME"` is just `rm -rf ""` and doesn't clean up `~/.gnupg`.

gquintard:
understood, I've reverted back to just `rm`ing `~/.gnupg`

PR #20746, opened 2026-01-29
GHC 9.14

develop7:
Added overdue GHC 9.14


develop7:
https://github.com/docker-library/official-images/actions/runs/21486894425/job/61898633474?pr=20746 failed because of rate limiting of Github API

develop7:
> Hopefully you're looking at an upgrade to Debian Trixie sometime in the near future? 😅

https://gitlab.haskell.org/ghc/ci-images/-/merge_requests/178 upstream sure did, but depends on binary distributions availability really. Quite certain it's coming.

> It's probably also worth considering dropping 9.6 and 9.8, given that (unless I've misunderstood upstream) they likely won't be getting any more releases?

that's true, will do

PR #20753, opened 2026-01-29
eclipse-mosquitto 2.1.0

ralight:
It's intentional not to have moved the `2` tag yet.

### DIFF
--- a/library/eclipse-mosquitto
+++ b/library/eclipse-mosquitto
@@ -1,9 +1,12 @@
 Maintainers: Roger Light <roger@atchoo.org> (@ralight)
-GitRepo: https://github.com/eclipse/mosquitto.git
-GitCommit: ff1187fd9c74ae3a7ba0097e7933828bdcdbce71
+GitRepo: https://github.com/eclipse-mosquitto/mosquitto.git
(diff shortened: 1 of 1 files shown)


tianon:
```diff
+if [ "$PUID" = "" ]; then
+	PUID="mosquitto"
+fi
+if [ "$PGID" = "" ]; then
+	PGID="mosquitto"
+fi
```

I was confused about this because I don't see anything that uses these except the `chown`, but then I found https://github.com/eclipse-mosquitto/mosquitto/commit/95e2c4d54396bdc8c7e714262595cb82aff1501e where they were implemented upstream in 2.1.

It's definitely better for users to use `--user` or `user:` to implement this privilege separation (as it's significantly more secure), but this is fine.

(I just got tripped because these are the same variable names used in other Docker-based solutions so they looked like a mistake without the context of the upstream change. :sweat_smile:)

ralight:
@tianon Ok, good. Mosquitto will always drop privileges if it starts as root. Handling PUID/PGID just allows easier control of what it drops to. If `user:` is already set to something non-root then it won't use PUID/PGID anyway. I think this covers both cases fine.

PR #20766, opened 2026-02-02
Update Semeru OE versions with the Q1 release.

jayasg12:
BUMP Semeru JDK/JRE versions to 8u482,11.0.30,17.0.18,21.0.10,25.0.2

### DIFF
--- a/library/ibm-semeru-runtimes
+++ b/library/ibm-semeru-runtimes
@@ -5,136 +5,136 @@ GitRepo: https://github.com/ibmruntimes/semeru-containers.git
 GitFetch: refs/heads/ibm
 
 #-----------------------------openj9 v8 images---------------------------------
-Tags: open-8u472-b08-jdk-jammy, open-8-jdk-jammy
+Tags: open-8u482-b08-jdk-jammy, open-8-jdk-jammy
 Architectures: amd64, ppc64le, s390x, arm64v8
-GitCommit: ea7c829c68017a394b768f9bc4bbb807242ec3bd
+GitCommit: 5242482eaa80dc302bdbcaccee52b8d0b9d04988
 Directory: 8/jdk/ubuntu/jammy
 File: Dockerfile.open.releases.full
 
-Tags: open-8u472-b08-jdk-noble, open-8-jdk-noble
-SharedTags: open-8u472-b08-jdk, open-8-jdk
+Tags: open-8u482-b08-jdk-noble, open-8-jdk-noble
+SharedTags: open-8u482-b08-jdk, open-8-jdk
 Architectures: amd64, ppc64le, s390x, arm64v8
-GitCommit: ea7c829c68017a394b768f9bc4bbb807242ec3bd
+GitCommit: 5242482eaa80dc302bdbcaccee52b8d0b9d04988
 Directory: 8/jdk/ubuntu/noble
 File: Dockerfile.open.releases.full
 
-Tags: open-8u472-b08-jre-jammy, open-8-jre-jammy
+Tags: open-8u482-b08-jre-jammy, open-8-jre-jammy
 Architectures: amd64, ppc64le, s390x, arm64v8
-GitCommit: ea7c829c68017a394b768f9bc4bbb807242ec3bd
+GitCommit: 5242482eaa80dc302bdbcaccee52b8d0b9d04988
 Directory: 8/jre/ubuntu/jammy
 File: Dockerfile.open.releases.full
 
-Tags: open-8u472-b08-jre-noble, open-8-jre-noble
-SharedTags: open-8u472-b08-jre, open-8-jre
+Tags: open-8u482-b08-jre-noble, open-8-jre-noble
+SharedTags: open-8u482-b08-jre, open-8-jre
 Architectures: amd64, ppc64le, s390x, arm64v8
-GitCommit: ea7c829c68017a394b768f9bc4bbb807242ec3bd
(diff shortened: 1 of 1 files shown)


yosifkit:
The Java packages are being bumped, but the corresponding tags are unchanged, for example:
```diff
-ENV JAVA_VERSION jdk-11.0.29+7_openj9-0.56.0
+ENV JAVA_VERSION jdk-11.0.30+7_openj9-0.57.0
```
```yaml
 Tags: open-11.0.29_7-jdk-jammy, open-11-jdk-jammy
```

---

```diff
-    sleep 5; \
+    sleep 20; \
     "${INSTALL_PATH_TOMCAT}"/bin/shutdown.sh -force; \
     sleep 15; \
```

Is there an alternative to just adding more sleep that would be able to track the expected goal that running tomcat provides and checking for that? It seems like this is trying to give tomcat enough time to start listening so perhaps checking for the port to open or a successful `curl` (in a limited sleep loop) would be a better check.

```sh
success=;
for i in $(seq 1 20); do
  if curl ...; then
    success=1
    break
  fi
  sleep 1
done
if [ -z "$success" ]; then
  echo >&2 "no successful after 20 retries"
  exit 1
fi
```

jayasg12:
@yosifkit Thank you for catching the missing release tag update. I’ve resolved it and pushed the changes.

jayasg12:
We’ll work on an alternative approach for adding the Tomcat sleep time and will raise a separate PR once the change is ready and tested. Let's not wait for this PR merge for that please. Thanks !!

tianon:
Any further updates on the possibility of publishing pre-generated profiles instead of generating them during build, like we've talked about previously?

jayasg12:
@tianon Sorry, I don’t recall the previous discussion. If you can point me to the issue where this was discussed, I can review it and take the necessary action.
But should this PR wait for that change  ? :)

tianon:
See https://github.com/docker-library/official-images/pull/8781#issuecomment-710689261 for one of the oldest places it's been discussed, and https://github.com/docker-library/official-images/pull/10666#issuecomment-923331547 for where improving it has been discussed (slightly) more recently.

To reiterate, if the "same" cache is acceptable for various users of this Docker image, why wouldn't it be similarly acceptable for various users of the upstream distribution, or alternatively, why couldn't it be pre-calculated in a way that targets the Docker image but is calculated/hosted separately?  Then you'd be free to run whatever tools you want and do so for however long you want to calculate a suitable cache without affecting the downstream build speed/complexity.

Hopefully that makes it clear, but to be explicit, the reason this is relevant to _this_ PR is because it's been a hanging chad for quite a long time now, and this change actively makes the offending block of code marginally worse (a block of code we had reservations about accepting at all back in 2020 and it's still essentially the same approach today).

mstoodle:
> because it's been a hanging chad for quite a long time now, and this change actively makes the offending block of code marginally worse (a block of code we had reservations about accepting at all back in 2020 and it's still essentially the same approach today).

I'm afraid it's not so simple, which is why nobody has resolved your concern in the intervening 5+ years. On top of the challenge I mentioned the first time we discussed it in 2021, when we populate a cache with code, outside of containers we optimize that code for the exact CPU version you're running on (as the Java JIT compiler does for every other method it compiles when running Java applications). Inside containers, for portability reasons, we instead generate code for a very old CPU version. Your suggested solution can have a fairly dramatic impact on startup outside of containers because the "universal" prepopulated cache can only work in containers is if that code is generated for the older processor. But outside of containers, such a cache would only actually deliver an improvement for the very first (single) cold startup run while penalizing all subsequent starts because otherwise they would (and do today) use the best optimized code. You're asking me to consider making millions of Semeru deployments worse because (for some reason) running one specific command and waiting for it to do some work inside the container build step makes you unhappy. The byproduct of running that command, however, helps the startup performance for both cold and subsequent runs for all downstream containers. In short, it has value, even if it's a "little" unconventional. I think I like the idea to use the availability of the server port to decide to shut the tomcat process down (thanks for mentioning it @yosifkit we'll try to improve that for a subsequent release, though we'll need to augment with some kind of timeout so it doesn't hang there if something untoward happens to the server). We actually do something similar when prepopulating caches for some of our performance demos, so I'm hopeful that should help cut down the build time overhead of this prepopulation step across all the Semeru containers you build. But I (still) don't see a way to avoid populating the cache that doesn't hurt some large class of our stakeholders.

@tianon Hopefully that gives more context for why the "hanging chad" hasn't been fixed. It's not that we haven't thought about it. This release fixes security vulnerabilities that have already been published by Oracle. Could you please stop holding our release hostage now that you have an answer to a question you could have chosen to ask any number of different non-impactful ways at any point during the past 5 years?

tianon:
For context, I _have_ raised this several times, and it has been a source of instability during the build -- downloading Tomcat and running it during build is flaky.

I can totally understand not making this part of Semeru proper, that makes perfect sense, but if you were to pre-calculate this for Docker, you could run whatever you want and pre-generate a cache specifically for Docker that provides a potentially better experience than just running Tomcat during build can offer (and that would avoid the flakiness/issues with Tomcat being part of the build downstream).

tianon (approved):

mstoodle:
Thanks @tianon. We'll take a look at your suggestion, though I'm not super enthusiastic to add another package type per platform, it wouldn't be the first time and clearly this would address a pain point for you. In the meantime, thanks for putting up with us.

PR #20787, opened 2026-02-03
varnish: update to 8.0.0-5, vmod-cfg on arm

gquintard:
I talked with upstream and we now have `vmod-cfg` packages for `arm` too

### DIFF
--- a/library/varnish
+++ b/library/varnish
@@ -1,11 +1,11 @@
-# this file was generated using https://github.com/varnish/docker-varnish/blob/58af075d7a6979394b8099c31b7e66572dc741c3/populate.sh
+# this file was generated using https://github.com/varnish/docker-varnish/blob/9a68f156ac2fa5d6c2d08ebca5cd80fa7f5a4623/populate.sh
 Maintainers: Guillaume Quintard <guillaume.quintard@gmail.com> (@gquintard)
 GitRepo: https://github.com/varnish/docker-varnish.git
 
 Tags: fresh, 8.0.0, 8, 8.0, latest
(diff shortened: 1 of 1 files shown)


tianon:
a mug with a yellow square that says "We love to see it"

gquintard:
by the way, in case this helps someone, I've added testing on arm runners too: https://github.com/varnish/docker-varnish/commit/a9d3e6d23665b74bbe0464bd1de57991cd45c25d

It's ugly, but it gets the job done

yosifkit:
Weirdly, the arm runner images don't seem to have a `latest`, so I think that needs an Ubuntu version number: `(.matrix.include[] | .os = "ubuntu-24.04-arm")`

Maybe also a `| .name += " (arm64)"` would help differentiate the arm builds from the AMD64 ones.

gquintard:
good point, that explains why my job was hanging for so long. I do get another issue now, which is weird since I've tested the `Dockerfile` in AWS...

yosifkit:
Ah, that probably needs something like `.runs.build |= gsub( "--platform[= ]linux/[^ ]+"; "--platform linux/arm64")` since the generate script sets the [platform flag](https://github.com/docker-library/bashbrew/blob/170b589d0472871a2e831de88e63ef2eb7e02093/scripts/github-actions/generate.sh#L91).

gquintard:
yes, thank you! I was just coming to that conclusion too.

I blindly applied your changed, and it works! https://github.com/varnish/docker-varnish/actions/runs/21652624947

PR #20789, opened 2026-02-03
Create images for all supported Groovy versions

keeganwitt:


yosifkit:
It looks like all of the Alpine-based images are using Alpine 3.23 except for `3.0.25-jdk11-jammy` that is `FROM eclipse-temurin:11-jdk-alpine` (Alpine 3.22). Is that intended?

keeganwitt:
> It looks like all of the Alpine-based images are using Alpine 3.23 except for `3.0.25-jdk11-jammy` that is `FROM eclipse-temurin:11-jdk-alpine` (Alpine 3.22). Is that intended?

Nope. Good catch!

PR #20800, opened 2026-02-04
eclipse-mosquitto 2.1.1

ralight:
Update to alpine 3.23
Replace obsolete labels with oci versions
Move `2` tag to the 2.1 line


ralight:
This also removes the in-image build of libwebsockets for the 1.6 and 2.0 images and uses the version packaged by alpine instead.

PR #20804, opened 2026-02-05
January 2026 PSU updates

smlambert:
January 2026 updates for JDK versions 8, 11, 17, 21, 25

### DIFF
--- a/library/eclipse-temurin
+++ b/library/eclipse-temurin
@@ -7,721 +7,720 @@ GitFetch: refs/heads/main
 Builder: buildkit
 
 #------------------------------v8 images---------------------------------
-Tags: 8u472-b08-jdk-alpine-3.20, 8-jdk-alpine-3.20, 8-alpine-3.20
+Tags: 8u482-b08-jdk-alpine-3.20, 8-jdk-alpine-3.20, 8-alpine-3.20
 Architectures: amd64
-GitCommit: 38127871aaedc393f8bd5abbc0b910b2a7569824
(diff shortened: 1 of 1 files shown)


smlambert:
@gdams - check failed on Diff Comment, is this typical?

gdams:
> @gdams - check failed on Diff Comment, is this typical?

yeah I think it's because the diff is longer than the allowed GitHub comment length.

jiekang:
For more context, the job is trying to have a bot comment on this PR with a diff of the changes, for example:

https://github.com/docker-library/official-images/pull/20245#issuecomment-3503527817

As George says, the diff for this PR is now larger than the maximum GitHub comment length.

jiekang:
All the Test PR jobs passed so I don't think there are any issues with the changes themselves.

tianon:
FWIW, decreasing the size of the support matrix *would* generally be a sane idea (for example, Alpine 3.20 and 3.21, while not technically EOL, are not especially "supported" either, and https://hub.docker.com/_/eclipse-temurin#using-a-different-base-image even documents for users how to create for themselves a combination y'all don't explicitly support).

In full transparency, the diff comment being "too big" isn't a blocker by itself, but I'd be lying if I said I didn't consider it akin to a [code smell](https://en.wikipedia.org/wiki/Code_smell), and it definitely slows down the review.

smlambert:
Thanks @tianon - I will raise it in our next meeting to discuss a 'trimming activity'.

tianon (approved):

PR #20894, opened 2026-02-21
Update openjdk

yosifkit:
Changes:

- https://github.com/docker-library/openjdk/commit/9a62d7b: Update 27 to 27-ea+10

### DIFF
--- a/library/openjdk
+++ b/library/openjdk
@@ -3,7 +3,7 @@
 Maintainers: Tianon Gravi <admwiggin@gmail.com> (@tianon),
              Joseph Ferguson <yosifkit@gmail.com> (@yosifkit)
 GitRepo: https://github.com/docker-library/openjdk.git
-GitCommit: 651902f8cc1223e50230df473943fa6101cef483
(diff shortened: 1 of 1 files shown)


tianon (approved):

tianon:
> failed to connect to the docker API at npipe:////./pipe/docker_engine; check if the path is correct and if the daemon is running: open //./pipe/docker_engine: The system cannot find the file specified.

I've been seeing this more and more often with Windows workers on GHA :upside_down_face:

gdams:
> failed to connect to the docker API at npipe:////./pipe/docker_engine;

@tianon we've also been seeing it unfortunately. There's an upstream bug here: https://github.com/actions/runner-images/issues/13729

I wonder if adding the workaround to the CI scripts in the short term is a good solution?

```yml
      - name: Docker issue workaround
        id: docker-workaround
        run: |
          $dockerState = Get-Service -Name "docker"
          if ($dockerState.Status -ne "Running") {
            Start-Service -Name "docker"
          }
```

PR #20897, opened 2026-02-21
[Friendica] New version 2026.01

ne20002:
The long awaited new version of Friendica :)


n816-Life:
Thanks for the release. Longly awaited.

PR #20908, opened 2026-02-23
Update Debian

tianon:


yosifkit:
```diff
+usr/lib/udev/
+usr/lib/udev/rules.d/
+usr/lib/udev/rules.d/61-apt.rules
```

This part stands out as something new in forky/testing sid/unstable. It looks like https://salsa.debian.org/apt-team/apt/-/commit/31cbb48a36060e59e5e05a718553334c74e20286 is the commit to introduce it. Since the images don't usually run systemd (or udev), I think it is harmless.

yosifkit (approved):

tianon:
```diff
+usr/share/pixmaps/debian-logo.bmp
```

This is amusing too, but also harmless and likely expected. 

👍

PR #20929, opened 2026-02-27
Almalinux auto-update - 20260227 15:48:24

almalinuxautobot:
This is an auto-generated commit. Any concern or issues, please contact or email AlmaLinux OS Foundation cloud-infra@almalinux.org (@AlmaLinux)

### DIFF
--- a/library/almalinux
+++ b/library/almalinux
@@ -1,24 +1,26 @@
-# This file was generated on https://github.com/AlmaLinux/container-images/actions/runs/21467784587
+# This file was generated on https://github.com/AlmaLinux/container-images/actions/runs/22493082393
 Maintainers: The AlmaLinux OS Foundation <cloud-infra@almalinux.org> (@AlmaLinux)
 GitRepo: https://github.com/AlmaLinux/container-images.git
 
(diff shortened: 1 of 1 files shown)


andrewlukoshko:
This PR introduces RISC-V containers for AlmaLinux Kitten 10

tianon:
Do you happen to know which profile baseline it's using?

Our build system is a sifive unmatched, so we can definitely support _this_ PR, but I'm trying to just be certain we don't need to update the bits we added for Ubuntu 25.10+ (where they target a profile for which hardware doesn't exist yet so we have to block any images `FROM` it using `riscv64` since they'll all fail to build with SIGILL :sweat_smile: :see_no_evil:).  I'm guessing y'all didn't do that, but better safe than sorry! :bow:

andrewlukoshko:
@tianon we're using RV64GC

yosifkit (approved):

yosifkit:
That is the same as Debian, so LGTM.
> The Debian port uses RV64GC as the hardware baseline
>
> - https://wiki.debian.org/RISC-V

tianon:
Thank you for confirming! :heart:

PR #20939, opened 2026-03-01
Update JRuby 10.0-dev and add 10.1-dev

headius:


headius:
The builds of JRuby 10.0-dev and 10.1-dev failed due to a new transitive dependency of activesupport, installed by the `ruby-gems` test script to verify Ruby images.

activesupport 8.1.2 depends on minitest 6.0.2 which recently started depending on the prism gem.

The prism gem currently only exists as a C library (or C extension for CRuby) and therefore requires C build tools to be present on the host system. Because JRuby has mostly JVM-based libraries (even in published gems), we do not include C build tools as part of the image.

Options:

* Remove activesupport from the libraries installed for verification by the `ruby-gems` test.
* Add C build tools to the JRuby images (we really would rather not)

FWIW, just installing these libraries is not much verification, especially on JRuby where they literally just download and unpack into the gem directory. Removing activesupport would not reduce the test coverage by much.

headius:
Also FWIW, we are working with the Prism folks to publish a pure-JVM version of the gem that will eliminate the need for build tools, but that's still WIP.

yosifkit:
Sorry for the delay. I'm working on updating our tests to work better with jruby and perhaps dived deeper than I needed to. I'll have a quick fix up today and work on a deeper fix later.

yosifkit:
Quick question, the updated 10.0-dev images are failing because they are missing `syslog`. Is it expected that the next jruby 10.0 won't have syslog? From https://github.com/jruby/jruby/pull/9198, it seems like it is supposed to, but perhaps something is missing in the container install?

```console
Ignoring syslog-0.4.0 because its extensions are not built. Try: gem pristine syslog --version 0.4.0
./container.rb:177: warning: syslog is not part of the default gems starting from Ruby 3.4.0. Install syslog from RubyGems.


require 'syslog' failed: cannot load such file -- syslog
org/jruby/RubyKernel.java:1193:in 'require'
/opt/jruby/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:139:in 'require'
./container.rb:180:in 'block in <main>'
org/jruby/RubyArray.java:2093:in 'each'
./container.rb:177:in '<main>'

	'ruby-standard-libs' [5/10]...failed
```

headius:
@yosifkit I believe syslog was moved to a bundled gem (i.e. preinstalled in the release) for Ruby 3.4 (JRuby 10), and removed altogether for Ruby 4.0 (JRuby 10.1).

tianon:
Right, so it's unexpected that it would fail on JRuby 10.0 with that module missing :sweat_smile:  Did something go wrong with JRuby 10.0 that might cause that?

(We've got Ruby 4.0 images where it still works, so I don't think it's removed yet if they're planning to :bow:)

headius:
Apologies. I had syslog confused with the logger gem, which was slated for removal in 4.0 but I guess they decided not to do that.

I'll look into the syslog error on JRuby 10.0-dev. There were some changes recently but it should still be there (it's using the official sources now rather than our local copy).

tianon:
Would you like us to go ahead with https://github.com/docker-library/official-images/pull/20958 and get this in now as-is, or would you rather dig into that first? :eyes:

headius:
I can reproduce the same issue with both JRuby 10.0.x and CRuby 3.4 by trying to load `syslog` from `bundle exec` when it is not part of the bundle.

```
[] tmp $ cx 3.4 bundle install
Fetching gem metadata from https://rubygems.org/.
Resolving dependencies...
Bundle complete! 1 Gemfile dependency, 1 gem now installed.
Use `bundle info [gemname]` to see where a bundled gem is installed.
[] tmp $ cx 3.4 bundle exec ruby -rsyslog -e 1
warning: syslog was loaded from the standard library, but is not part of the default gems starting from Ruby 3.4.0.
You can add syslog to your Gemfile or gemspec to silence this warning.
/Users/headius/.rubies/ruby-3.4.7/lib/ruby/3.4.0/bundled_gems.rb:82:in 'Kernel.require': cannot load such file -- syslog (LoadError)
	from /Users/headius/.rubies/ruby-3.4.7/lib/ruby/3.4.0/bundled_gems.rb:82:in 'block (2 levels) in Kernel#replace_require'
```

I'll look at your PR now to see if that's what's happening here.

tianon:
I think we're just doing `require 'syslog'` :sweat_smile:

headius:
@tianon I'm still trying to sort out what all your verification scripts do, but 10.0-dev should indeed work with a simple `require 'syslog'`:

```
$ jruby -e 'require "syslog"; Syslog.open.info("hello"); puts :ok'
ok
```

headius:
> Ignoring syslog-0.4.0 because its extensions are not built.

You should also not be seeing this error after https://github.com/jruby/jruby/pull/9198. 🤔

headius:
I have reproduce the issue with the 10.0.4.0-SNAPSHOT dist tarball. Please hold.

headius:
There's a bug in our assembly of the dist artifact and it appears to be excluding some files needed to load syslog (triggering the missing extension warning). I am working on a fix and will re-push this when it works its way into the dev build.

headius:
I have the issue fixed.

We will release JRuby 10.0.4.0 tomorrow so I will just make sure the dev images come along with that.

Thanks for your help! I'm glad we found this before the 10.0.4.0 release!

PR #20944, opened 2026-03-02
Updated stable alpine-slim hashes for nginx

thresheek:
Incorporates changes making sure nginx.org repo is used when building stable images.

W/o the changes due to a recent nginx version bump in Alpine upstream, wrong binaries could be pulled, as caught by CI.

### DIFF
--- a/library/nginx
+++ b/library/nginx
@@ -30,7 +30,7 @@ Directory: mainline/alpine-perl
 
 Tags: 1.29.5-alpine-slim, mainline-alpine-slim, 1-alpine-slim, 1.29-alpine-slim, alpine-slim, 1.29.5-alpine3.23-slim, mainline-alpine3.23-slim, 1-alpine3.23-slim, 1.29-alpine3.23-slim, alpine3.23-slim
 Architectures: arm64v8, arm32v6, arm32v7, ppc64le, s390x, i386, amd64, riscv64
-GitCommit: ffe72978e08c5b0dacecd604e528f6d0741a9ae5
+GitCommit: e2ce6fac3776b07c746ec4e0d26ab07cb45c3cec
 Directory: mainline/alpine-slim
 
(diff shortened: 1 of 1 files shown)


tianon:
I'm trying to parse out what's happening in these new lines -- as far as I can tell, you've got some kind of metapackage you're querying dependencies from with `apk query` to pre-install them at the right versions?  Maybe you could embed some comments in the `Dockerfile` itself to explain it to casual onlookers?  I think it would probably also benefit from being broken up a little (both for readability _and_ for reliability, because if your `apk query` fails for any reason, `set -e` won't trigger and will happily pass an empty block to `apk add` which may or may not fail).

thresheek:
Fair enough, this was actually a question my colleague asked me during the review of the changes in https://github.com/nginx/docker-nginx/pull/1029 (that went behind the scenes though, in DMs...).

The problem this PR was trying to solve is that currently Alpine ships `nginx=1.28.5-r1` which happens to be the same version/release as the binary from nginx.org.  Apparently `apk` would prefer system repos (maybe due to a higher `provider_priority`?  but I don't want priority wars just like we have with `rpm`...), so the unexpected binary will get installed.  This will happen now if an `alpine` base image gets updated.

My idea was to maybe just tag the repos as we document on https://nginx.org/en/linux_packages.html#Alpine, but that proved to be problematic for further `apk` invocations as the package manager relies on tagged repos to be present in the `/etc/apk/repositories` file.

So indeed, I implemented the following idea:
 - `apk` will only operate on a single repository with the expected `nginx` packages when doing an actual install
 - `apk` will pre-install the required dependencies based on the query combining both upstream and nginx.org repos.

I will add some comments the Dockerfiles to explain this a bit.

Splitting the complex `apk add`/`apk query` invocation makes sense as well, although I believe the next `apk add` invocation will rightfully fail if any of those fail.

Thank you!

thresheek:
Added comments & refactored the `apk query / apk add` parts.  Thank you!

PR #20947, opened 2026-03-02
Update photon image

gerrit-photon:
Update photon image


tianon:
Is an update from 2026-01-16 to 2026-02-14 expected?  Should that be something closer to today? :sweat_smile:

(It's OK if not, and this is clearly an upgrade so I'm going to go ahead and merge, but wanted to raise just in case it's unexpected)

PR #20951, opened 2026-03-03
Official Convertigo 8.4.0 release!

nicolas-albert:
Thx!


tianon:
```diff
-FROM tomcat:9-jdk21-temurin-jammy
+FROM tomcat:9-jdk25-temurin
```

This is OK if it's what you really want to do, but I want to note that this is putting yourself on a "square wheel" which is likely going to cause inevitable breakage when the default changes in roughly a month or so -- for repos we maintain, we usually choose to explicitly manage this wheel rolling so that we get to control more precisely when those breaking changes happen and we don't end up with breaking builds that block our updates.

---

```console
	'convertigo-hello-world' [4/4]...failed
```

Is there something in this test that needs to be updated? :sweat_smile: (is it still a useful test?  is this a real failure?)

https://github.com/docker-library/official-images/blob/6536bc3bd6aa7370f956ae784db422b8156dfec3/test/tests/convertigo-hello-world/run.sh

PR #20963, opened 2026-03-04
Update JRuby 10.0 to 10.0.4.0 and 10.1-dev to latest

headius:
Should go more smoothly this time.


headius:
10.0-dev failed because I hadn't pushed a new snapshot. Will fix that.

headius:
The syslog error should not be happening. I even created a CI job in JRuby to test this. Not sure what's going on.

headius:
Confirmed the dev snapshot is still throwing that syslog error but I don't know why yet.

headius:
I believe I found the issue. Deploying from a local repository that's already been built with JRuby 10.1 ends up putting some files in the wrong place. I'll do a clean deploy and it should correct it.

headius:
Green!

PR #20967, opened 2026-03-04
Use tmpfs for Nextcloud tests

J0WI:
I don't know GitHub Actions/Azure internals but maybe this makes the tests a bit snappier and more reliable. Is there a way to test this before merging?

### DIFF
--- a/library/nextcloud
+++ b/library/nextcloud
@@ -47,3 +47,4 @@ Tags: 33.0.0-fpm-alpine, 33.0-fpm-alpine, 33-fpm-alpine, fpm-alpine
 Architectures: amd64, arm32v6, arm32v7, arm64v8, i386, ppc64le, riscv64, s390x
 GitCommit: 53d81594c8050ba84ab7410860b355493f753346
 Directory: 33/fpm-alpine
+


yosifkit:
> Is there a way to test this before merging?

When we need to, we usually test them via a jenkins job. But to test on GHA, I think an extra commit adding a newline to `library/nextcloud` would be enough to run them. And then we can link to that run and drop out the extra commit.

yosifkit:
Added a commit to force builds to run. ~Now just waiting for GitHub actions to run (https://www.githubstatus.com/incidents/g5gnt5l5hf56).~ 

Success: https://github.com/docker-library/official-images/actions/runs/22730228423

Is it faster or more reliable? 🤷 Most of the test runs are a few seconds faster from the last runs in https://github.com/docker-library/official-images/pull/20923, but a few are not.

tianon:
Yeah, looks like it is faster on average, but _really_ marginally (and probably depends a lot on the tiny instances GHA gives us).

J0WI:
I leave it up to you to whether you think it's worth to merge or just close it.

tianon:
Yeah, thanks - I think it's interesting, and it was worth trying for sure, but I don't think the difference is dramatic enough to be worth merging. :heart:

PR #20981, opened 2026-03-06
[security] Update golang

tianon:
Changes:

- https://github.com/docker-library/golang/commit/c8592aa: Update 1.26 to 1.26.1
- https://github.com/docker-library/golang/commit/f51096b: Update 1.25 to 1.25.8


tianon:
CI is close enough -- we have proof that the Windows binaries are fine, so any glitches from here have to be handled separately anyhow.

PR #20986, opened 2026-03-06
Releasing Neo4j 2026.02.2

jennyowen:
Thank you!


tianon:
If you want an even lighter-weight alternative to `su-exec`, for your uses, `setpriv` should be sufficient: https://manpages.debian.org/trixie/util-linux/setpriv.1.en.html

(specifically, you'll want at least `--reuid`, `--regid`, and either `--clear-groups` or `--init-groups`)

PR #20994, opened 2026-03-08
library/perl: 5.42.0 -> 5.42.1

zakame:
- http://nntp.perl.org/group/perl.perl5.porters/270767
- https://github.com/Perl/docker-perl/pull/204


zakame:
Thanks @tianon ! 🚀

PR #21000, opened 2026-03-09
Update golang

yosifkit:
Changes:

- https://github.com/docker-library/golang/commit/5d0ebaa: Update tip to tip-20260308, commit b7d1c5887561ee5502b6c04baaa27ead13b63971

### DIFF
--- a/library/golang
+++ b/library/golang
@@ -112,23 +112,23 @@ Directory: 1.25/windows/nanoserver-ltsc2022
 Builder: classic
 Constraints: nanoserver-ltsc2022, windowsservercore-ltsc2022
 
-Tags: tip-20260301-trixie, tip-trixie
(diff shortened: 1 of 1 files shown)


yosifkit:
Note: These tip tags (`tip-20260308*`) will not have arm32 builds because they don't finish compiling. See https://github.com/golang/go/issues/78053. The next tip tags should be fine.

tianon:
I have also confirmed that https://github.com/golang/go/commit/97d643045888d1af2004410365a00bafe1713f47 builds successfully for `linux/arm/v7` (`docker buildx build --pull --platform linux/arm/v7 --build-arg GOLANG_COMMIT=97d643045888d1af2004410365a00bafe1713f47 tip/trixie`), so we should indeed be fine next week :+1:

PR #21038, opened 2026-03-13
update influxdb 1.12 tags to version 1.12.3

gwossum:
- Update influxdb 1.12 tags to version 1.12.3. 
- Add influxdb 1.12.3 tags.


yosifkit:
```Dockerfile
ENV INFLUXDB_VERSION=1.12.3-c1.12.3
# INFLUXDB_PR is for suffixes like -1
ENV INFLUXDB_PR=
ENV INFLUXDB_PV=${INFLUXDB_VERSION}
```
The `influxdb:data-alpine` Dockerfile doesn't have the same suffix concatenation as the other images: `ENV INFLUXDB_PV=${INFLUXDB_VERSION}${INFLUXDB_PR}`. This doesn't affect the image, so can be adjusted later.

I also noticed that a few sets of `curl` flags were given as `-fssLO` instead of `-fsSLO` in the meta and data images (including alpine). It also shouldn't affect the image but should be corrected; it just won't show the error on failure.

I don't see these as blocking this PR.

PR #21053, opened 2026-03-17
Update for Corretto releases: 26.0.0.35.2

satyenme:
Update amazoncorretto for latest release. For more details, see the following:
* [corretto/corretto-docker repo](https://github.com/corretto/corretto-docker/commit/main)
* [corretto/corretto-26 release](https://github.com/corretto/corretto-26/releases/tag/26.0.0.35.2)


tianon:
As a point of feedback, I would suggest that at some point you'll want to replace `uname -m` with something like `rpm --query --queryformat='%{ARCH}' rpm` so that it's querying the _userspace_ architecture instead of the kernel architecture, because those don't necessarily have to match.

Additionally, capturing that in a variable before running another command would be good so that the error status propagates more correctly (ie, `... && foo="$(captured)" && some command using "$foo" && ...` instead of `... && some command using "$(captured)" && ...`

Issue #16625, opened 2024-04-17
An issue with nginx 1.25.5 manifest.


oxpa:
Hello,

I'm trying to pull nginx:1.25.5 image from docker.io but get an error:
```
:~$ docker pull docker.io/nginx:1.25.5
1.25.5: Pulling from library/nginx
no matching manifest for linux/amd64 in the manifest list entries
```
At the same time, I can see that the build was successful : https://doi-janky.infosiftr.net/job/multiarch/job/amd64/job/nginx/
and the specific image can be pulled with:
```
:~$ docker pull docker.io/amd64/nginx:1.25.5
1.25.5: Pulling from amd64/nginx
13808c22b207: Already exists
c46d32988911: Pull complete
b9d060e68bbc: Pull complete
c028eb95b065: Pull complete
e74ab6743f66: Pull complete
4a666f159bd0: Pull complete
297a65ba6f04: Pull complete
Digest: sha256:9e6b127264e7cd5a52b366323ced6367d4e88f6c4cbd218ba3caf477690d8300
Status: Downloaded newer image for amd64/nginx:1.25.5
docker.io/amd64/nginx:1.25.5
```

Which tells me that the image is in the registry but there may be a problem with a manifest.

Is this something I can fix on my side?

Kind regards,
Oxpa

whalelines:
That image is very recent. It may take a few hours for all os/architecture images to be built and made available.

oxpa:
Fair enough.
Though it usually just doesn't list a new tag in docker.io/library/nginx until it's all built. Or maybe I was just lucky.
Thank you for the update, I'll wait for the issue to resolve itself and close this one.

tianon:
Yep, when the tag canonically is supposed to exist, but there are no builds yet, an empty list is often pushed.  This will almost always only show up for a very short window, but it is normal and expected. :+1:

PR #21062, opened 2026-03-19
update varnish to 9.0.0

gquintard:
I've cut out non-debian, non-fresh images as they need a few adjustments to align with the new world order, but I wanted to provide a fresh option quickly.

### DIFF
--- a/library/varnish
+++ b/library/varnish
@@ -1,33 +1,9 @@
-# this file was generated using https://github.com/varnish/docker-varnish/blob/9a68f156ac2fa5d6c2d08ebca5cd80fa7f5a4623/populate.sh
+# this file was generated using https://github.com/varnish/docker-varnish/blob/9549601f65e81660417432a6963cdb9a41da3caf/populate.sh
 Maintainers: Guillaume Quintard <guillaume.quintard@gmail.com> (@gquintard)
 GitRepo: https://github.com/varnish/docker-varnish.git
 
-Tags: fresh, 8.0.0, 8, 8.0, latest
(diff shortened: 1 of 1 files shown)


tianon:
Hmm, are you sure you want to _delete_ them all?

As a reminder, removing tags here will remove them from the "Supported" section on the Hub readme (and will prevent us from spending cycles rebuilding them on the official build servers), but the tags will still be available to users who want them.

We're fine either way, but I want to make sure you understand the implications of doing so. :sweat_smile:

gquintard:
Yes, let's go for that for now. Kubecon is coming next week and I need that image out :-)

I'll fix the other images early next week with a couple more upgrades

gquintard:
uh: https://hub.docker.com/layers/library/varnish/9.0.0/images/sha256-eb37f58646a901dc7727cf448cae36daaefaba79de33b5058dab79aa4c04aefb

The images seem to have landed, but I can't pull them:

```
$ docker pull varnish:9.0.0
9.0.0: Pulling from library/varnish
no matching manifest for linux/amd64 in the manifest list entries
```

tianon:
That's somewhat expected -- that means the system knows the builds should exist, but they don't yet (it'll be updated as the builds actually complete).

References with slightly more detailed descriptions:

- https://github.com/docker-library/golang/issues/550#issuecomment-2654671051
- https://github.com/docker-library/official-images/issues/16625#issuecomment-2062242608
- https://github.com/debuerreotype/docker-debian-artifacts/issues/243#issuecomment-3109588047

gquintard:
ah, great! no need to worry then, thank you!

PR #21064, opened 2026-03-19
Update ELK official images for 8.19.13, 9.2.7, 9.3.2

alina-bacalete:
Update ELK official images for `8.19.13`, `9.2.7` and `9.3.2`

### DIFF
--- a/library/elasticsearch
+++ b/library/elasticsearch
@@ -13,7 +13,7 @@ Architectures: amd64, arm64v8
 GitFetch: refs/heads/9.2
 GitCommit: 6894c1c3452a1129471711450e77cae505aa075c
 
-Tags: 9.3.1
+Tags: 9.3.2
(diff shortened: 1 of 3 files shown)


tianon:
```diff
+# Symlink sh to bash to override dash which doens't respect dotted env vars
+RUN ln -sf bash /bin/sh && for iter in 1 2 3 4 5 6 7 8 9 10; do \
```

FWIW, this is kind of dangerous, as any update to the `dash` package will revert this change, so I'd recommend doing an explicit `dpkg-divert` if you want this change to persist (but this isn't a hard blocker, so I'm going to merge as-is for now).

tianon (approved):

alina-bacalete:
cc @breskeby for the above comment

breskeby:
Ack. Taking note

PR #21098, opened 2026-03-24
[nats] Releases v2.12.6, v2.11.15

neilalexander:
Also remove v2.10.29 as no longer supported.

More info on releases here:

- https://github.com/nats-io/nats-server/releases/tag/v2.12.6
- https://github.com/nats-io/nats-server/releases/tag/v2.11.15


tianon:
In the future, you might find some part of https://github.com/docker-library/official-images#security-releases interesting/useful (it's possible it doesn't apply to this release, but in general this is the best suggestions for making sure our eyes are appropriately on security releases more proactively than normal).

PR #21125, opened 2026-03-27
Update Percona tags and add 8.4 as default for 8

jpadmin:


yosifkit:
cc `percona` image maintainers: @EvgeniyPatlan @hors @vorsel @adivinho @surbhat1595. Is this correct for adding the `8.4` version to Docker Official Images?

jpadmin:
@yosifkit, I can see that several recent versions are missing there — the latest as of now should be 8.4.7, but [this file](https://github.com/percona/percona-docker/blob/main/percona-server-8.4/Dockerfile-dockerhub) is still at 8.4.3.

tianon:
ping @EvgeniyPatlan @hors @vorsel @adivinho @surbhat1595

yosifkit:
Apparently 8.4 is still "in progress". Closing since it is not ready yet. (https://github.com/docker-library/official-images/pull/21510#issuecomment-4511989703)

PR #21223, opened 2026-04-12
Add PeakURL

Abd-Ur-Rehman:
This proposes `peakurl` as a new Docker Official Image.

PeakURL is an upstream-maintained, MIT-licensed, self-hosted URL shortener and branded links platform.

Upstream repos:
- image repo: https://github.com/PeakURL/containers
- application repo: https://github.com/PeakURL/PeakURL

Initial details for this proposal:
- proposed image name: `peakurl`
- initial tags: `1.0.13`, `1.0`, `1`, `latest`
- initial architectures: `amd64`, `arm64v8`
- library entry is generated from: `PeakURL/containers` commit `9b7b8bfce70b6ec8471452a808b44947cd5dda52`

Why this image is ready for review:
- builds from the official published PeakURL release archive, not from a live source checkout
- the DOI build context is version-pinned in `1.0/`
- the release archive is verified by SHA256 during the Docker build
- DOI metadata is kept in `versions.json`, with `update-versions-json.sh` to keep the pinned directory and manifest source in sync
- the image is built from the official `php:8.4-apache-bookworm` base image
- the runtime interface is intentionally simple and shell-compatible

Operational note:
- the upstream project currently also publishes `peakurl/peakurl` on Docker Hub while this DOI proposal is under review
- this proposal is for the official Docker image name `peakurl`

Docs PR:
- docker-library/docs#2665


tianon:
Hey, thanks for putting this together! Before I get into the Dockerization details, I want to surface a concern upfront so we're not spending a lot of effort before we figure out whether it's a fit. 🙇

I have some questions about the "popularity" aspect -- we have to have a minimal popularity requirement to keep our limited available time focused on things that will benefit a larger set of users. The normal metrics we check are things like GitHub stars (while not perfect, they do serve as *a* metric of user traffic and interest -- users do tend to star repositories that they use or enjoy). Looking over https://github.com/PeakURL, `PeakURL/PeakURL` was created just a few weeks ago and currently has 3 stars and no forks; `PeakURL/containers` is only a few days old.

To be clear, I don't mean to be discouraging -- I think this looks like a genuinely useful self-hosted tool and Expat/MIT-licensed to boot 🎉 -- but I'm questioning whether *this* is the right moment for it to be in the Official Images program. The program tends to work best for projects that already have a demonstrated community and maintainership track record.

I also see you've already got `peakurl/peakurl` on Docker Hub, which is great -- but with the image being so new, the pull count is still in the low hundreds, and that's exactly the kind of organic traction we'd want to see growing before bringing something into the Official Images program. I'd suggest letting things develop a bit more and then revisiting once there's more to point to. 🙏 ❤️

---

That said, since you put the work in, figured I'd write up a few things I noticed in the diff anyway -- mostly just for your reference. 😊

---

```diff
+ARG VCS_REF=
+ARG BUILD_DATE=
[...]
+LABEL org.opencontainers.image.revision="${VCS_REF}"
+LABEL org.opencontainers.image.created="${BUILD_DATE}"
```

We don't actively recommend labels in general (https://github.com/docker-library/official-images/issues/3540), but if you were going to keep them, these two would be ones to drop -- `$VCS_REF` and `$BUILD_DATE` default to empty, so they'd always produce empty labels in any normal build (users aren't going to know to pass them as `--build-arg` at build time).

---

```diff
+    if [ -n "$PEAKURL_RELEASE_SHA256" ]; then echo "$PEAKURL_RELEASE_SHA256  /tmp/peakurl.zip" | sha256sum -c -; fi; \
```

This would be a blocker worth cleaning up: the checksum verification being conditional on `$PEAKURL_RELEASE_SHA256` being non-empty means it can be silently bypassed with `--build-arg PEAKURL_RELEASE_SHA256=`. Since the `ARG` already has the correct hash as its default, the `if` wrapper isn't really buying anything -- the verification should be unconditional so a missing or emptied hash fails the build rather than skipping the check:

```sh
    echo "$PEAKURL_RELEASE_SHA256  /tmp/peakurl.zip" | sha256sum -c -; \
```

---

```diff
+COPY docker-entrypoint.sh /usr/local/bin/peakurl-entrypoint

+RUN chmod +x /usr/local/bin/peakurl-entrypoint
```

We've seen issues with permissions not round-tripping correctly through certain graph drivers when a `chmod` happens in a separate layer from the `COPY`. The fix is to commit the executable bit to Git directly -- `chmod +x docker-entrypoint.sh` locally and commit that mode change, then `COPY` will pick it up natively and the extra `RUN` goes away.

---

```diff
+if [ ! -e "$APP_ROOT/install.php" ]; then
+    tar -C "$SOURCE_ROOT" -cf - . | tar -C "$APP_ROOT" -xf -
+fi

+chown -R www-data:www-data "$APP_ROOT"
```

The runtime file-copy on first start (when `install.php` isn't present) feels like it's there primarily to support mounting `/var/www/html` as a persistent volume -- is that the intended usage pattern? If most deployments won't use a persistent volume there, it might be simpler to just have the app live in `/var/www/html` in the image directly and skip the `/usr/src/peakurl` staging directory.

The unconditional `chown -R` on the whole app directory at every container start is also worth thinking about -- on a meaningful number of files that could add noticeable startup time, and since the `RUN` block in the `Dockerfile` already sets the ownership correctly, it'd be safe to skip for non-volume deployments at least. 😊

Abd-Ur-Rehman:
Thanks for the candid feedback.

That’s fair on the GitHub-side signals. The GitHub repos are still new, so the stars/forks are low right now. One extra data point I wanted to mention is that the Docker Hub image is currently at 2,909 pulls as of April 21, 2026, so the image usage seems to be ahead of the GitHub metrics.

Part of the reason I opened this proposal early is that we do plan to support PeakURL for the long term, and I wanted to start aligning the image work with that. There’s also a practical usability side to it: `docker pull peakurl` / `image: peakurl` is a much cleaner default for users than `peakurl/peakurl`.

I also went ahead and cleaned up the technical issues you pointed out:
- checksum verification is now unconditional
- the entrypoint executable bit is committed in Git instead of being set in a separate `chmod` layer
- the empty `VCS_REF` / `BUILD_DATE` labels were removed

On the runtime pattern you asked about: yes, the `/var/www/html` mount is intentional in the current image design. The goal is to support a persistent full app tree there, so a fresh empty mount is initialized from the bundled release on first start, and existing installs keep their persisted app files after that. I also tightened the startup ownership behavior so the recursive `chown` now only runs when that first-time initialization happens, not on every container start.

If you still think it’s too early for DOI on program-fit grounds, that’s completely fine. In that case we’ll keep publishing `peakurl/peakurl` directly for now and revisit later once there’s a longer public usage and maintenance track record. I mainly wanted to add the current Docker Hub usage signal, clarify the intended runtime pattern, and note that the container review items have been addressed on our side.

tianon:
Thank you for understanding :bow:  Let's indeed wait and revisit later :+1:

PR #21233, opened 2026-04-13
Update photon image

gerrit-photon:
Update photon image

### DIFF
--- a/library/photon
+++ b/library/photon
@@ -5,19 +5,19 @@ GitRepo: https://github.com/vmware/photon-docker-image.git
 Directory: docker
 
 
-Tags: 5.0, 5.0-20260214, latest
+Tags: 5.0, 5.0-20260411, latest
 Architectures: amd64, arm64v8
-GitFetch: refs/heads/x86_64/5.0-20260214
-GitCommit: 7daa2733c6c3f04e47fb27a7e582f57918838c26
+GitFetch: refs/heads/x86_64/5.0-20260411
+GitCommit: 0b1afbddf1a388a3defcc49553c879007622daf8
 arm64v8-GitFetch: refs/heads/aarch64/5.0-20260214
 arm64v8-GitCommit: d3f853a6fcd2d23251b22890d736ab044fc2b384
 
-Tags: 4.0, 4.0-20260215
+Tags: 4.0, 4.0-20260418
 Architectures: amd64, arm64v8
-GitFetch: refs/heads/x86_64/4.0-20260215
-GitCommit: dee98791f59633825e0d754921cdec17aef0e3e0
-arm64v8-GitFetch: refs/heads/aarch64/4.0-20260215
-arm64v8-GitCommit: 4eb5e937e3e597033f274799a4136152807db7eb
+GitFetch: refs/heads/x86_64/4.0-20260418
+GitCommit: ecd8187ea8c6a2b5b45a406a11386af410351218
+arm64v8-GitFetch: refs/heads/aarch64/4.0-20260418
+arm64v8-GitCommit: c101e445fa25e8737907772c58332af4f051dc1c
 
 Tags: 3.0, 3.0-20250316
 Architectures: amd64, arm64v8
--- a/library/photon
+++ b/library/photon
@@ -5,19 +5,19 @@ GitRepo: https://github.com/vmware/photon-docker-image.git
 Directory: docker
 
 
-Tags: 5.0, 5.0-20260214, latest
+Tags: 5.0, 5.0-20260610, latest
 Architectures: amd64, arm64v8
-GitFetch: refs/heads/x86_64/5.0-20260214
-GitCommit: 7daa2733c6c3f04e47fb27a7e582f57918838c26
-arm64v8-GitFetch: refs/heads/aarch64/5.0-20260214
-arm64v8-GitCommit: d3f853a6fcd2d23251b22890d736ab044fc2b384
+GitFetch: refs/heads/x86_64/5.0-20260610
+GitCommit: e521d6719ba85414811c93e7fbbb8d1ce5425ef7
+arm64v8-GitFetch: refs/heads/aarch64/5.0-20260610
+arm64v8-GitCommit: a0f7a2be9f8056b140f447d060690fb907009088
 
-Tags: 4.0, 4.0-20260215
+Tags: 4.0, 4.0-20260524
 Architectures: amd64, arm64v8
-GitFetch: refs/heads/x86_64/4.0-20260215
-GitCommit: dee98791f59633825e0d754921cdec17aef0e3e0
-arm64v8-GitFetch: refs/heads/aarch64/4.0-20260215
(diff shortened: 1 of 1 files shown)


library/photon
```diff
@@ -1,16 +1,14 @@
...
 
 
-Tags: 5.0, 5.0-20260214, latest
-Architectures: amd64, arm64v8
-GitFetch: refs/heads/x86_64/5.0-20260214
-GitCommit: 7daa2733c6c3f04e47fb27a7e582f57918838c26
-arm64v8-GitFetch: refs/heads/aarch64/5.0-20260214
-arm64v8-GitCommit: d3f853a6fcd2d23251b22890d736ab044fc2b384
+Tags: 5.0, 5.0-20260411, latest
+Architectures: amd64
```
tianon, line 9: Is dropping `arm64v8` support intentional?

library/photon
```diff
@@ -5,19 +5,19 @@ GitRepo: https://github.com/vmware/photon-docker-image.git
...
 
-Tags: 5.0, 5.0-20260214, latest
+Tags: 5.0, 5.0-20260411, latest
 Architectures: amd64, arm64v8
-GitFetch: refs/heads/x86_64/5.0-20260214
-GitCommit: 7daa2733c6c3f04e47fb27a7e582f57918838c26
+GitFetch: refs/heads/x86_64/5.0-20260411
+GitCommit: 0b1afbddf1a388a3defcc49553c879007622daf8
 arm64v8-GitFetch: refs/heads/aarch64/5.0-20260214
 arm64v8-GitCommit: d3f853a6fcd2d23251b22890d736ab044fc2b384
```
yosifkit, line 13: This does not include any changes for `arm64v8` but is just tagging the old `arm64v8` image (`5.0-20260214`) as `5.0-20260411`.

michellew-vmware:
Is that possible we provide different docker tag for amd64 and arm64v8 in 5.0? @tianon @yosifkit

tianon:
To be clear, you want the `5.0` and `latest` generic tags to still point to the older arm64 build, but the more specific tag to be available and have amd64-only?

I'm not sure we've got a *clean* way we could accomplish this with the current framework. :thinking:

I want to be clear that I *strongly* discourage this, and that if you go forward with it that it be something extremely temporary -- if your intent is to no longer support arm64 long-term, that's fine and should result in it being dropped (and that's reasonable), but if this is an expression of an infra issue or similar, hopefully that gets resolved in a reasonable amount of time; maybe something like this?

```yaml
#
# TEMPORARY HACKS TO SUPPORT NEWER 5.0 BUILDS FOR AMD64 WHILE ARM64 LAGS BEHIND FOR <reasons here>
#

Tags: 5.0, latest
Architectures: amd64, arm64v8
GitFetch: refs/heads/x86_64/5.0-20260411
GitCommit: 0b1afbddf1a388a3defcc49553c879007622daf8
arm64v8-GitFetch: refs/heads/aarch64/5.0-20260214
arm64v8-GitCommit: d3f853a6fcd2d23251b22890d736ab044fc2b384

Tags: 5.0-20260411
Architectures: amd64
GitFetch: refs/heads/x86_64/5.0-20260411
GitCommit: 0b1afbddf1a388a3defcc49553c879007622daf8

Tags: 5.0-20260214
Architectures: amd64, arm64v8
GitFetch: refs/heads/x86_64/5.0-20260214
GitCommit: 7daa2733c6c3f04e47fb27a7e582f57918838c26
arm64v8-GitFetch: refs/heads/aarch64/5.0-20260214
arm64v8-GitCommit: d3f853a6fcd2d23251b22890d736ab044fc2b384

#
# END TEMPORARY HACKS FOR 5.0
#
```

michellew-vmware:
If it looks good to you guys, please help to merge it.

aarch64 docker publish will be delayed as expected for now.

Thank you for your time and attention,
Michelle

tianon:
The structure here doesn't match what I outlined. With `5.0-20260411` bundled into the same stanza as `5.0`/`latest`, `photon:5.0-20260411` on arm64 delivers the February 14 build -- not an April 11 build. The date tag is misleading.

The fix seems simple -- move `5.0-20260411` into its own `Architectures: amd64` stanza -- but that creates a follow-on problem: the "obvious" next move to preserve arm64 access for the old date would be an arm64-only `5.0-20260214` stanza, and that would clobber the amd64 builds for that tag. You need the full `amd64, arm64v8` stanza for `5.0-20260214`, pointing to both arch builds from that date, to handle it correctly.

And you can't simply drop arm64 from the `5.0`/`latest` stanza entirely -- it's clear from your comments that keeping those generic tags multi-arch is the whole point. That's exactly what makes the three-stanza structure necessary 😅 -- and why the comments explaining all this are essential. Without them, whoever touches this file next will "simplify" it and break something.

The structure I'd want to see (same as what I proposed earlier, with the current commit hashes):

```yaml
#
# TEMPORARY HACKS TO SUPPORT NEWER 5.0 BUILDS FOR AMD64 WHILE ARM64 LAGS BEHIND FOR <reasons here>
#

Tags: 5.0, latest
Architectures: amd64, arm64v8
GitFetch: refs/heads/x86_64/5.0-20260411
GitCommit: 0b1afbddf1a388a3defcc49553c879007622daf8
arm64v8-GitFetch: refs/heads/aarch64/5.0-20260214
arm64v8-GitCommit: d3f853a6fcd2d23251b22890d736ab044fc2b384

Tags: 5.0-20260411
Architectures: amd64
GitFetch: refs/heads/x86_64/5.0-20260411
GitCommit: 0b1afbddf1a388a3defcc49553c879007622daf8

Tags: 5.0-20260214
Architectures: amd64, arm64v8
GitFetch: refs/heads/x86_64/5.0-20260214
GitCommit: 7daa2733c6c3f04e47fb27a7e582f57918838c26
arm64v8-GitFetch: refs/heads/aarch64/5.0-20260214
arm64v8-GitCommit: d3f853a6fcd2d23251b22890d736ab044fc2b384

#
# END TEMPORARY HACKS FOR 5.0
#
```

Still looking for context on the root cause -- is this a temporary lag where arm64 builds are behind, or is arm64 support for 5.0 being wound down? That changes what the right long-term answer looks like. 🤔

michellew-vmware:
we are working on fix aarch64 build break... hope we can ship a aarch64 photon docker image next week.

No action is required in your side for now. Keep in touch.

Thanks,
Michelle

michellew-vmware:
Please help to review and publish it.

Thanks,
Michelle

yosifkit (approved):

michellew-vmware:
Thank you for taking care of this. Finally...

Michelle

yosifkit:
It looks like the arm image is still incorrect: https://github.com/docker-library/official-images/issues/21661.

And this can be seen in the 0-byte tar file in the build context: https://github.com/vmware/photon-docker-image/blob/aarch64/5.0-20260610/docker/photon-rootfs-5.0-06c74ffdf.aarch64.tar.gz

PR #21240, opened 2026-04-14
Updated stable nginx to 1.30.0.

oxpa:
Nginx stable 1.30.0 release today, so stable images now use it.
Hopefully - no other changes.

### DIFF
--- a/library/nginx
+++ b/library/nginx
@@ -38,37 +38,37 @@ Architectures: amd64, arm64v8
 GitCommit: 71081b25390771f6b1275ddf7c73c965f304493f
 Directory: mainline/alpine-otel
 
-Tags: 1.28.3, stable, 1.28, 1.28.3-trixie, stable-trixie, 1.28-trixie
+Tags: 1.30.0, stable, 1.28, 1.30.0-trixie, stable-trixie, 1.28-trixie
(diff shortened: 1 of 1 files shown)


library/nginx
```diff
@@ -38,37 +38,37 @@ Architectures: amd64, arm64v8
 GitCommit: 71081b25390771f6b1275ddf7c73c965f304493f
 Directory: mainline/alpine-otel
 
-Tags: 1.28.3, stable, 1.28, 1.28.3-trixie, stable-trixie, 1.28-trixie
+Tags: 1.30.0, stable, 1.28, 1.30.0-trixie, stable-trixie, 1.28-trixie
```
yosifkit:
```suggestion
Tags: 1.30.0, stable, 1.30, 1.30.0-trixie, stable-trixie, 1.30-trixie
```

Looks like some of the `1.28*` tags were missed in the update (on this line and the other lines).

thresheek: Yeah this needs more changes on the `docker-nginx` repo side.

oxpa: Fixed the simple way for now, sorry about this.
I'll have a look to fix it in a way that would prevent this from happening in future.

tianon (approved):

PR #21242, opened 2026-04-14
Update Windows pins

yosifkit:


tianon:
https://support.microsoft.com/en-us/topic/windows-server-container-for-april-2026-90645550-16f4-4524-a482-f3f697181e4b

tianon (approved):

PR #21247, opened 2026-04-15
Update haproxy

tianon:
Changes:

- https://github.com/docker-library/haproxy/commit/89c2946: Update 3.4 to 3.4-dev9


tianon:
This fails to build on all our 32bit architectures :sob:

I filed https://github.com/haproxy/haproxy/issues/3334 upstream about it :+1:

PR #21251, opened 2026-04-16
Update ALT Linux - 20260416

FNDenisovna:


tianon (approved):
Removing (deprecated) `MAINTAINER`; nice 👍

---

```diff
+LABEL org.opencontainers.image.authors="Alexey Shabalin <shaba@altlinux.org>,Nadezhda Fedorova <fedor@altlinux.org>" \
+org.opencontainers.image.licenses="ALT-Container or GPLv3" \
```

Since you're touching the `LABEL` block anyway, worth mentioning: we don't actively recommend using `LABEL` -- the inheritance behavior is problematic (child images built `FROM` this one pick up all the baked-in labels, which is almost never what downstream users want). The build system already sets OCI annotations on the final image manifest automatically, so the in-image `LABEL` metadata is largely redundant. See https://github.com/docker-library/official-images/issues/3540#issuecomment-2715158488 for what gets set for you. Not a blocker -- the `org.opencontainers.image.*` keys you're using are the right namespace if you'd prefer to keep them -- but it's worth considering whether the whole `LABEL` block could just be removed entirely.

PR #21262, opened 2026-04-16
Update mongo

tianon:
Changes:

- https://github.com/docker-library/mongo/commit/86029db: Update 8.2-rc
- https://github.com/docker-library/mongo/commit/2709443: Update 8.2 to 8.2.7
- https://github.com/docker-library/mongo/commit/bcd9989: Update 7.0-rc

### DIFF
--- a/library/mongo
+++ b/library/mongo
@@ -31,57 +31,30 @@ GitCommit: d7a18fed30efc943c53cf73003afc638a5ba8f59
 Directory: 8.3-rc/windows/nanoserver-ltsc2022
 Constraints: nanoserver-ltsc2022, windowsservercore-ltsc2022
 
-Tags: 8.2.7-rc0-noble, 8.2-rc-noble
-SharedTags: 8.2.7-rc0, 8.2-rc
+Tags: 8.2.7-noble, 8.2-noble, 8-noble, noble
+SharedTags: 8.2.7, 8.2, 8, latest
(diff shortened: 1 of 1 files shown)


tianon:
"claude my eyes right out" reviewed this for me (based on a large and growing corpus of my own review notes collected from ~13 years of history):

> ```diff
> -ENV MONGO_DOWNLOAD_SHA256=bf630d9b09585aa31e7b71e027a1593ef9d1c4242776d9111a4a75903e3ef8c1
> +ENV MONGO_DOWNLOAD_SHA256=1f04282b3a57ede7afef7aef29be177e0ea66a4f36a166e0acdec4bfd7d45eca
> ```
> 
> The download URL for the 7.0-rc MSI didn't change, but the checksum did. If the same URL is now serving a different binary than it was before, that's a repeatability concern — anyone who built this image previously got a different artifact under the same version string. Was the MSI re-signed or rebuilt at the same URL? Even a brief comment in the Dockerfile (or a note here in the PR) explaining what changed would help future reviewers understand this was intentional rather than an error. 👀
> 
> ---
> 
> The 8.2.6 → 8.2.7 bump looks clean, and retiring the 8.2-rc now that 8.2.7 is stable is exactly right. 👍

It's very correct about the installer digest changing being an orange flag, but it's not the first time this has happened and so it doesn't actually read as *unusual* to me in this specific case (even though it normally would).

PR #21267, opened 2026-04-17
Chronograf v1.11.1

srebhan:
Thank you!

### DIFF
--- a/library/chronograf
+++ b/library/chronograf
@@ -1,14 +1,7 @@
 Maintainers: Brandon Pfeifer <bpfeifer@influxdata.com> (@bnpfeife),
              Sven Rebhan <srebhan@influxdata.com> (@srebhan)
 GitRepo: https://github.com/influxdata/influxdata-docker
-GitCommit: 3dc2d064944d033907d137e1af8069d35852de94
-
-Tags: 1.7, 1.7.17
-Architectures: amd64, arm32v7, arm64v8
-Directory: chronograf/1.7
-
-Tags: 1.7-alpine, 1.7.17-alpine
-Directory: chronograf/1.7/alpine
+GitCommit: 34242807de431aa0d6b7a01ca7ff88074f20e996
 
 Tags: 1.8, 1.8.10
 Architectures: amd64, arm32v7, arm64v8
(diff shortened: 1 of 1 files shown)


tianon (approved):

yosifkit:
Just a heads up that the `arm32v7` builds of Chronograf 1.11.1 fail to build because the detached signature file is missing. The build fails at verifying the signature, but the real error is that it got a 404, so I'd also suggest adding a `-f` to the [curl commands](https://github.com/influxdata/influxdata-docker/blob/b4a8f55d08b5360dad0c315211468045fd6a0fae/chronograf/1.11/Dockerfile#L26-L27) so that it would more clearly show a 404 failure instead of a signature failure. It'll then bail earlier when the download fails rather than trying to use the 404 page as a signature.

Build log ([full](https://doi-janky.infosiftr.net/job/meta/job/arm32v7/job/build/99578/console)):
```console
+ gpg --batch --verify chronograf_1.11.1_armhf.deb.asc chronograf_1.11.1_armhf.deb
gpg: WARNING: unsafe permissions on homedir '/root/.gnupg'
gpg: no valid OpenPGP data found.
gpg: the signature could not be verified.
```

Test with curl:

```console
root@d7cccd251d4f:/# curl -sSL0 https://dl.influxdata.com/chronograf/releases/chronograf_1.11.1_armhf.deb.asc
root@d7cccd251d4f:/# cat chronograf_1.11.1_armhf.deb.asc
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message><Key>chronograf/releases/chronograf_1.11.1_armhf.deb.asc</Key><RequestId>0PJG3H1HERHYYFEG</RequestId><HostId>uNi3STgsecUhe2MXLAtryM9vPSaHOLzkrspWsinga8G3WwkyYuOxdpkCmJGAeX9eHLaDy32yZnRvWA2QwnZGn+l4GvzWjFvD</HostId></Error>
root@d7cccd251d4f:/# curl -fsSLO https://dl.influxdata.com/chronograf/releases/chronograf_1.11.1_armhf.deb.asc
curl: (22) The requested URL returned error: 404
root@d7cccd251d4f:/# echo $?
22
```

tianon:
ping @bnpfeife @srebhan on the build failure above -- https://dl.influxdata.com/chronograf/releases/chronograf_1.11.1_armhf.deb.asc is 404 :sweat_smile:

PR #21271, opened 2026-04-18
Update ghost

tianon:
Changes:

- https://github.com/docker-library/ghost/commit/1bf8746: Update to 6.30.0, move from yarn to pnpm (https://github.com/docker-library/ghost/pull/467)

### DIFF
--- a/library/ghost
+++ b/library/ghost
@@ -4,12 +4,12 @@ Maintainers: Tianon Gravi <admwiggin@gmail.com> (@tianon),
              Joseph Ferguson <yosifkit@gmail.com> (@yosifkit),
              Austin Burdine <austin@acburdine.me> (@acburdine)
 GitRepo: https://github.com/docker-library/ghost.git
-GitCommit: 949d90ab1cf3a7bcb7b2dd510752679b641689f7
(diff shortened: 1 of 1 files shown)


tianon:
"claude my eyes right out" reviewed this for me (based on a large and growing corpus of my own review notes collected from ~13 years of history):

> ```diff
> +	corepack enable; \
> ```
> ```diff
> -	gosu node yarn cache clean; \
> +	gosu node pnpm store prune; \
> ```
> 
> The Yarn → pnpm migration looks clean overall. One thing worth a look: when corepack intercepts the first `pnpm` invocation during the Ghost install step, it downloads the pnpm binary to a local cache (`/home/node/.cache/node/corepack/` or wherever `COREPACK_HOME` resolves). The cleanup sequence sweeps the pnpm content-addressable store (`pnpm store prune`) and the npm caches, but doesn't appear to touch the corepack binary cache itself — which could be leaving a few MiB of downloaded tooling sitting in the final image. Worth confirming whether that directory exists post-install and whether it needs to be removed in the same `RUN` layer. Not a blocker. 👀

PR #21275, opened 2026-04-20
maven: update to 3.9.15 and add eclipse-temurin 26

emouty:


tianon:
````diff
+ARG MAVEN_VERSION=3.9.15
+FROM maven:${MAVEN_VERSION}-eclipse-temurin-17 AS maven_upstream
````

Because our build system has to be able to query/calculate inter-image dependencies, we unfortunately do not support parameterized `FROM` values. This applies whether the substitution covers the whole reference or just part of the tag -- `maven:${MAVEN_VERSION}-eclipse-temurin-17` is as much of a hard blocker as `FROM $BASE_IMAGE` would be.

The intent here (deduplicate the tag string across three `COPY --from=` lines) is good, and using a named intermediate stage is a perfectly reasonable way to achieve it; the `FROM` itself just needs to be a literal:

```dockerfile
FROM maven:3.9.15-eclipse-temurin-17 AS maven_upstream

FROM amazoncorretto:11-al2023
# ...
COPY --from=maven_upstream ${MAVEN_HOME} ${MAVEN_HOME}
COPY --from=maven_upstream /usr/local/bin/mvn-entrypoint.sh /usr/local/bin/mvn-entrypoint.sh
```

This applies to every `Dockerfile` in the PR that uses the new pattern -- all of the 3.x and 4.0.0-rc-5 variant `Dockerfile`s will need the literal tag in the `FROM ... AS maven_upstream` line. 😅

emouty:
Hello @tianon, I have updated https://github.com/carlossg/docker-maven to use literal tags in FROM on Dockerfiles and this PR to target the latest commit

tianon (approved):
Nice, thanks!

---

As a note for other reviewers, the removal of the explicit `settings-docker.xml` copy is fine:

````diff
-COPY --from=maven:3.9.14-eclipse-temurin-17 /usr/share/maven/ref/settings-docker.xml /usr/share/maven/ref/settings-docker.xml
````

`settings-docker.xml` lives inside `${MAVEN_HOME}` (`/usr/share/maven/ref/`), so it is already covered by the preceding `COPY --from=maven_upstream ${MAVEN_HOME} ${MAVEN_HOME}` -- the explicit copy was always redundant, and cleaning it up is welcome.

PR #21287, opened 2026-04-21
Odoo: update 17.0-19.0 to release 20260421

d-fence:
Hello,

here are the latest updates of Odoo supported versions.

Thanks

### DIFF
--- a/library/odoo
+++ b/library/odoo
@@ -1,16 +1,16 @@
 Maintainers: Christophe Monniez <moc@odoo.com> (@d-fence)
 GitRepo: https://github.com/odoo/docker
-GitCommit: b8b9102bf64e67d7baeaa410ff506dbe16fa8dfe
+GitCommit: 22d1fa4c8006452832196232dbc1d106deac4d9a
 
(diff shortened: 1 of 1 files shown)


tianon (approved):
Two pre-existing notes (neither introduced by this PR, but worth a cleanup on a future bump):

---

```diff
 RUN curl -o odoo.deb -sSL http://nightly.odoo.com/${ODOO_VERSION}/nightly/deb/odoo_${ODOO_VERSION}.${ODOO_RELEASE}_all.deb \
     && echo "${ODOO_SHA} odoo.deb" | sha1sum -c - \
```

The `curl` uses `-sSL` without `-f`/`--fail`, so an HTTP error response (404, 500, etc.) gets silently downloaded as `odoo.deb` and the build only fails later at the `sha1sum` check -- which does catch it, but `-f` would surface the real error immediately. Worth adding `-f` alongside `-sSL` next time.

---

The checksum verification uses `sha1sum` rather than a stronger algorithm like `sha256sum`. Worth considering upgrading if the upstream nightly builds ever start publishing SHA-256 checksums alongside the `.deb`.

PR #21298, opened 2026-04-21
Update postgres

yosifkit:
Changes:

- https://github.com/docker-library/postgres/commit/2353f03: Merge pull request https://github.com/docker-library/postgres/pull/1409 from infosiftr/mountpoint-fallback
- https://github.com/docker-library/postgres/commit/3b6b5fc: Fix Alpine missing `/var/lib/postgresql/data` mounts


tianon (approved):
> This BusyBox `mountpoint` workaround looks correct -- field 5 of `/proc/self/mountinfo` is indeed the canonical mount point path, so the `awk` fallback cleanly catches the bind-mount case that BusyBox's dev/ino check misses. The `{ ... }` grouping is a clean way to fold the two-part check into a single `&&` condition without a subshell. The inline comments explaining the why are a nice touch too. 👍

("claude my eyes right out")

PR #21299, opened 2026-04-22
Aerospike 8.1.2.0

pvinh-spike:

### DIFF
--- a/library/aerospike
+++ b/library/aerospike
@@ -4,14 +4,14 @@ Maintainers: Lucien Volmar <lucien@aerospike.com> (@volmarl),
              Kevin Porter <kporter@aerospike.com> (@kportertx),
              Thomas Huelbert <thuelbert@aerospike.com> (@thomasaerospike)
 
-Tags: ee-8.1.1.1, ee-8.1.1.1_1
+Tags: ee-8.1.2.0, ee-8.1.2.0_1
 Architectures: amd64, arm64v8
 GitRepo: https://github.com/aerospike/aerospike-server.docker.git
-GitCommit: 8b5917e1d0dba5b3d7f7ea8839a6bbe74f609b1b
-Directory: enterprise/ubuntu24.04
+GitCommit: 5626486c69e17bb91e0829550f06e41926720e6c
+Directory: releases/8.1/enterprise/ubuntu24.04
 
-Tags: ce-8.1.1.1, ce-8.1.1.1_1
+Tags: ce-8.1.2.0, ce-8.1.2.0_1
 Architectures: amd64, arm64v8
 GitRepo: https://github.com/aerospike/aerospike-server.docker.git
-GitCommit: 8b5917e1d0dba5b3d7f7ea8839a6bbe74f609b1b
-Directory: community/ubuntu24.04
+GitCommit: 5626486c69e17bb91e0829550f06e41926720e6c
+Directory: releases/8.1/community/ubuntu24.04


yosifkit:
Unfortunately, the nature of our build system requires that we parse all our `Dockerfiles` to determine build ordering, and that parsing does not support the BuildKit heredocs syntax. So, any heredocs in the Dockerfiles need to be changed to a more classic `RUN` line.
```dockerfile
RUN set -eux; \
apt-get update; \
...
```

tianon:
Building on yosifkit's comment above:

```diff
+# syntax=docker/dockerfile:1
```

```diff
+RUN bash <<'AEROSPIKE_INSTALL'
```

The `# syntax=docker/dockerfile:1` directive enables the BuildKit frontend, and the heredoc requires it -- both are BuildKit-only features the DOI build system doesn't support. But beyond the tooling issue, the heredoc signals that the install logic has grown more complex than belongs in a `Dockerfile` -- the fix isn't just to inline the script differently, it's to simplify what the `RUN` step is actually doing. Several of the concerns below point in that direction.

---

```diff
+COPY static/tini/as-tini-static-amd64 static/tini/as-tini-static-arm64 /opt/aerospike-tini/
```

```diff
+ARG TARGETARCH=
+RUN case "${TARGETARCH:-$(uname -m)}" in \
+        amd64 | x86_64) arch=amd64 ;; \
+        arm64 | aarch64) arch=arm64 ;; \
```

These are connected concerns. Vendoring pre-built binary blobs into the packaging repo means there's no way to verify build provenance or reproducibility the way there is for an artifact fetched at build time and verified against a SHA256. That vendored approach also drives the arch-detection complexity here: `TARGETARCH` is a BuildKit automatic `ARG` that won't be populated in the DOI build system, so this always falls back to `uname -m` -- which reads the host kernel's architecture rather than the image's userspace architecture, and is very strongly discouraged.

The cleaner approach is to fetch `as-tini-static` from a tagged release URL with a SHA256 at build time, the same way the main server package is handled. That also eliminates the `TARGETARCH`/`uname -m` fallback entirely -- the download step can select the right URL based on `dpkg --print-architecture`, which is already used correctly for the main install.

---

The `AEROSPIKE_INSTALL` block is 250+ lines of inline shell, duplicated essentially verbatim between the CE and EE `Dockerfile`s, doing things that `apt-get` and `dpkg` would ordinarily handle directly. Even setting the heredoc aside as a hard blocker, the content has a number of issues the rewrite would need to address:

```diff
+    dpkg --configure -a || echo "WARNING: dpkg --configure -a reported errors (LDAP deps)" >&2
```

`dpkg --configure -a` is for finishing configuration of packages left half-configured after an interrupted install -- running it speculatively as a recovery step everywhere implies broken state is expected. If `apt-get install` succeeded, there is nothing to reconfigure.

---

```diff
+        if ! apt-get install -y --no-install-recommends "${cpkg}"; then
+            echo "Note: ${cpkg} available but not installable"
+        fi
```

The `|| echo "WARNING: ..."` and `if ! ...; then echo "Note: ..."; fi` patterns throughout this script actively defeat `set -Eeuo pipefail`. The whole point of strict mode is that failures are fatal -- wrapping every command in an error suppressor means silent failures are baked in from the start. Either a package is required (fail if it doesn't install) or it isn't needed (don't install it).

---

```diff
+    ldconfig || echo "WARNING: ldconfig failed" >&2
```

`apt-get install` runs post-install scripts including `ldconfig` as part of normal package installation. Running it again manually is redundant -- and `|| echo WARNING` here makes it less reassuring, not more. 😅

---

```diff
+function _install_tools_from_tgz() {
+    ar -x aerospike/aerospike-tools*.deb --output aerospike/pkg
+    tar xf aerospike/pkg/data.tar.xz -C aerospike/pkg/
+    find aerospike/pkg/opt/aerospike/bin/ -exec chown root:root {} +
+    mv aerospike/pkg/etc/aerospike/astools.conf /etc/aerospike
```

This manually reimplements what `dpkg -i aerospike-tools*.deb` does, without dependency handling and with hand-rolled path manipulation. If a `.deb` is in hand, install it -- `apt-get install ./aerospike-tools*.deb` handles dependencies automatically.

There is also a concrete bug in this function:

```diff
+    if [ -f 'aerospike/pkg/opt/aerospike/bin/asinfo' ]; then
+        mv aerospike/pkg/opt/aerospike/bin/asinfo /usr/lib/asadm/
+    fi
+    ln -snf /usr/lib/asadm/asinfo /usr/bin/asinfo
```

The `ln` is unconditional. If `asinfo` is not present in the package, the `mv` is skipped but the symlink is still created -- leaving `/usr/bin/asinfo` pointing at nothing.

---

```diff
+    rm -rf /opt/aerospike/bin
+
+    _install_tools_from_tgz
```

Deleting a directory the package manager just installed into encodes a fragile assumption: if the server package ever relocates those binaries, `rm -rf` becomes a silent no-op and the cleanup breaks invisibly. The right approach here is dpkg excludes -- a `path-exclude` entry in `/etc/dpkg/dpkg.cfg.d/` tells `dpkg` not to extract those paths in the first place, so the package manager's own state stays accurate and the post-install deletion isn't needed.

---

```diff
+    _try=1
+    while [ "${_try}" -le 3 ]; do
+        if apt-get install -y --no-install-recommends \
+            binutils xz-utils ca-certificates curl procps; then
+            break
+        fi
```

```diff
+command -v curl >/dev/null 2>&1 || {
+    echo "ERROR: curl not found" >&2
+    exit 1
+}
```

A retry loop on `apt-get install`, followed by a check that the installed package actually landed -- this implies the build environment is expected to be broken. If `apt-get install` exits 0, `curl` is installed. If it fails, the build should stop immediately rather than retry and second-guess itself. The arm64-specific special-casing is also worth questioning: what specific known failure mode is this working around?

---

```diff
+if [ "${AEROSPIKE_LOCAL_PKG:-0}" = "1" ]; then
```

This entire branch is not just dead code -- it is permanently inaccessible. `AEROSPIKE_LOCAL_PKG` is never declared as an `ARG` in the `Dockerfile`. Docker silently drops undeclared `--build-arg` values and they never reach the `RUN` environment. `${AEROSPIKE_LOCAL_PKG:-0}` always evaluates to `0`, period.

---

```diff
+ARG AEROSPIKE_COMPAT_LIBS="0"
```

```diff
+    for cpkg in libssl1.1 "${curl_pkg}"; do
```

`AEROSPIKE_COMPAT_LIBS` is documented as being for Aerospike 7.2 compatibility on Ubuntu 24.04. It defaults to `0`, and even if it were activated, the first package it would try to install is `libssl1.1` -- which does not exist in Ubuntu 24.04's repos. This is legacy dead code from the template and can be removed entirely.

---

```diff
+export DEBIAN_FRONTEND=noninteractive
```

`DEBIAN_FRONTEND=noninteractive` is very rarely necessary -- the `debconf` bugs that originally motivated this workaround have been fixed (https://bugs.debian.org/929417, https://bugs.debian.org/1061480). Just omit it. If a specific package genuinely requires it, the right form is to apply it inline for that one invocation only: `DEBIAN_FRONTEND=noninteractive apt-get install -y tzdata`.

tianon:
Here's a simplified example, to help illustrate:

```dockerfile
FROM ubuntu:24.04

RUN set -eux; \
    apt-get update; \
        ca-certificates \
        procps \
    ; \
    apt-get dist-clean

RUN set -eux; \
    apt-get update; \
    apt-get install -y --no-install-recommends curl; \
    \
    dpkgArch="$(dpkg --print-architecture)"; \
    case "$dpkgArch" in \
        amd64) \
            tiniUrl='https://github.com/aerospike/tini/releases/download/1.0.1/as-tini-static'; \
            tiniSha='d1f6826dd70cdd88dde3d5a20d8ed248883a3bc2caba3071c8a3a9b0e0de5940'; \
            pkgLink='https://download.aerospike.com/artifacts/aerospike-server-community/8.1.2.0/aerospike-server-community_8.1.2.0_tools-13.0.0_ubuntu24.04_x86_64.tgz'; \
            pkgSha='bef8a6acf697837d037a8608b913d61ee01e4d03c22a986eafe5d56289aaa91e'; \
            ;; \
        arm64) \
            tiniUrl='https://github.com/aerospike/tini/releases/download/1.0.1/as-tini-static-arm64'; \
            tiniSha='1c398e5283af2f33888b7d8ac5b01ac89f777ea27c85d25866a40d1e64d0341b'; \
            pkgLink='https://download.aerospike.com/artifacts/aerospike-server-community/8.1.2.0/aerospike-server-community_8.1.2.0_tools-13.0.0_ubuntu24.04_aarch64.tgz'; \
            pkgSha='f8859425a8a6a4a3cca8dd3728dd154c45ba9f42529f079b1182b5d428d0c54c'; \
            ;; \
        *) echo >&2 "error: unsupported architecture '$dpkgArch'"; exit 1 ;; \
    esac; \
    \
    curl -fL -o /usr/bin/as-tini-static "$tiniUrl"; \
    echo "$tiniSha */usr/bin/as-tini-static" | sha256sum --strict --check -; \
    chmod +x /usr/bin/as-tini-static; \
    \
    mkdir -p /tmp/aerospike; \
    curl -fL -o /tmp/aerospike/pkg.tgz "$pkgLink"; \
    echo "$pkgSha */tmp/aerospike/pkg.tgz" | sha256sum --strict --check -; \
    tar -xzf /tmp/aerospike/pkg.tgz --strip-components=1 -C /tmp/aerospike; \
    \
        /tmp/aerospike/aerospike-server-*.deb \
        /tmp/aerospike/aerospike-tools*.deb \
    ; \
    \
    mkdir -p /etc/aerospike /licenses /var/log/aerospike /var/run/aerospike; \
    cp /tmp/aerospike/LICENSE /licenses/; \
    \
    rm -rf /tmp/aerospike; \
    apt-get purge -y --auto-remove curl; \
    apt-get dist-clean

COPY aerospike.template.conf /etc/aerospike/aerospike.template.conf

EXPOSE 3000 3001 3002

COPY entrypoint.sh /entrypoint.sh

STOPSIGNAL SIGTERM

ENTRYPOINT ["/usr/bin/as-tini-static", "-r", "SIGUSR1", "-t", "SIGTERM", "--", "/entrypoint.sh"]
CMD ["asd"]
```

pvinh-spike:
Hi @yosifkit & @tianon,  Thanks for the very useful feedbacks.  I'm closing this one to open the new pr.     Very much appreciated.

PR #21323, opened 2026-04-24
Aerospike 8.1.2.0

pvinh-spike:

### DIFF
--- a/library/aerospike
+++ b/library/aerospike
@@ -4,14 +4,14 @@ Maintainers: Lucien Volmar <lucien@aerospike.com> (@volmarl),
              Kevin Porter <kporter@aerospike.com> (@kportertx),
              Thomas Huelbert <thuelbert@aerospike.com> (@thomasaerospike)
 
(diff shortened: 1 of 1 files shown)


tianon:
This is a much cleaner approach than the previous attempts -- the simplified install logic closely matches the reference example from #21299. 👍

One small difference from that example worth noting -- the cleanup here:

```diff
+    apt-mark auto curl; \
+    apt-get autoremove -y --purge; \
```

works fine, though `apt-get purge -y --auto-remove curl;` is a single operation that does the same thing more directly. Not a blocker. 👍

---

Still outstanding from #21147:

```diff
+Tags: ee-8.1.2.0, ee-8.1.2.0_1
```

The `aerospike:ee` proprietary license concern hasn't been addressed yet. We're not going to act on that today, but it would be good to start that conversation -- happy to set up a meeting with folks from Docker to discuss alternative programs (DVP especially) if that'd be helpful. ❤️

tianon (approved):

pvinh-spike:
Thanks a lot.

PR #21326, opened 2026-04-25
azul-zulu: add alpine variation

azul-publisher:
Hello, we would like to extend our Azul Zulu image offering with Alpine-based variants, in addition to the existing Debian images. All Alpine Dockerfiles follow the same structure across versions — the only difference between them is the specific Zulu package installed via apk add.

### DIFF
--- a/library/azul-zulu
+++ b/library/azul-zulu
@@ -7,8 +7,24 @@ Maintainers: Azul Devops Team <devops@azul.com> (@azul-publisher),
              Geertjan Wielenga <gwielenga@azul.com> (@geertjanw)
 GitRepo: https://github.com/AzulSystems/azul-zulu-images.git
 GitFetch: refs/heads/main
-GitCommit: 53819781a6977fc2027c49f716184cf4694e1c3c
+GitCommit: 1144a5ee00eb560430eedeac7ad88d3e93e39f44
 
+
+Tags: 8.94-8.0.492-jdk-alpine3.23, 8-alpine3.23, 8-jdk-alpine, 8-jdk-alpine3.23
+Architectures: amd64, arm64v8
+Directory: 8/jdk/alpine
+
+Tags: 8.94-8.0.492-jdk-headless-alpine3.23, 8-headless-alpine, 8-headless-alpine3.23
+Architectures: amd64, arm64v8
+Directory: 8/jdk-headless/alpine
+
+Tags: 8.94-8.0.492-jre-alpine3.23, 8-jre-alpine, 8-jre-alpine3.23
+Architectures: amd64, arm64v8
+Directory: 8/jre/alpine
+
+Tags: 8.94-8.0.492-jre-headless-alpine3.23, 8-jre-headless-alpine, 8-jre-headless-alpine3.23
+Architectures: amd64, arm64v8
+Directory: 8/jre-headless/alpine
 
 Tags: 8.94-8.0.492-jdk-debian13, 8, 8-jdk, 8-jdk-debian, 8-jdk-debian13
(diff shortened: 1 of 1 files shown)


tianon:
```diff
+    wget -P /etc/apk/keys/ https://cdn.azul.com/public_keys/alpine-signing@azul.com-5d5dc44c.rsa.pub && \
```

This downloads the APK signing key and places it into APK's trusted keystore with no verification that it is actually the expected key. HTTPS protects the transport, but that's not the same thing -- a compromised CDN or a redirect can silently swap in a different key, which APK would then trust to sign any package installed from the Azul repo. This needs to download the key first, verify it matches an expected fingerprint or SHA256 hash (the same way we require for GPG keys and RPM keys in other images), and only then place it in `/etc/apk/keys/`.

---

```diff
+    apk update && \
+    apk add --no-cache zulu11-jdk-headless=11.0.31-r3 && \
```

The `apk update` here is redundant -- `--no-cache` implies updating the package lists.

---

```diff
+RUN set -eux && \
+    wget -P /etc/apk/keys/ ... && \
+    echo "..." | tee -a /etc/apk/repositories && \
+    apk update && \
+    apk add --no-cache zulu11-jdk-headless=11.0.31-r3 && \
+    java -version
```

You might consider using `;` rather than `&&` as the command delimiter -- given `set -eux` is already there, the `;` form makes the error-propagation behaviour a little more explicit and reads a bit cleaner:

```dockerfile
RUN set -eux; \
    wget -P /etc/apk/keys/ ...; \
    echo "..." | tee -a /etc/apk/repositories; \
    apk add --no-cache zulu11-jdk-headless=11.0.31-r3; \
    java -version
```

---

```diff
+CMD ["jshell"]
```

Just to confirm, the headless JDK packages still ship `jshell`, right?  (it would usually survive the headless package trimming, but just in case)

bydga:
Thanks @tianon for the feedback, we processed all your comments. Yes, the jshell is available in the headless, no need to differentiate that 🙂 
```
docker run -it azul.internal/devops/zulu-ca-temp-build/alpine-17.64-17.0.18-r3-jdk-headless-arm64:878819
Unable to find image 'azul.internal/devops/zulu-ca-temp-build/alpine-17.64-17.0.18-r3-jdk-headless-arm64:878819' locally
878819: Pulling from devops/zulu-ca-temp-build/alpine-17.64-17.0.18-r3-jdk-headless-arm64
f4d1e965a6ff: Pull complete
Digest: sha256:1568500da5581f500f2bf21f7a4388a7bc2586da20b4d0089fa09d9e8e961abd
Status: Downloaded newer image for azul.internal/devops/zulu-ca-temp-build/alpine-17.64-17.0.18-r3-jdk-headless-arm64:878819
/ # jshell
Apr 29, 2026 8:23:41 PM java.util.prefs.FileSystemPreferences$1 run
INFO: Created user preferences directory.
|  Welcome to JShell -- Version 17.0.18
|  For an introduction type: /help intro

jshell>
```

tianon (approved):
Glad to see the previous feedback addressed -- the key verification flow is much cleaner now. 👍

One small note on the new Alpine Dockerfiles:

```diff
+     echo "6c6393d4755818a15cf055a5216cffa599f038cd508433faed2226925956509a  /tmp/azul-signing.pub" | sha256sum -c -; \
```

The `*`-prefixed binary mode is the conventional format for verifying downloaded files -- worth switching from the double-space form:

```sh
echo "6c6393d4755818a15cf055a5216cffa599f038cd508433faed2226925956509a */tmp/azul-signing.pub" | sha256sum -c -
```

(Not a blocker, just a minor hardening note.)

PR #21327, opened 2026-04-26
archlinux: Release 20260426.0.520785

archlinux-github:

---

### DIFF
--- a/library/archlinux
+++ b/library/archlinux
@@ -5,18 +5,23 @@ Maintainers: Santiago Torres-Arias <santiago@archlinux.org> (@SantiagoTorres),
              Justin Kromlinger <hashworks@archlinux.org> (@hashworks)
 GitRepo: https://gitlab.archlinux.org/archlinux/archlinux-docker.git
 
-Tags: latest, base, base-20260419.0.517065
-GitCommit: 25354c172306956092edf933ad385f23975ebedd
-GitFetch: refs/tags/v20260419.0.517065
+Tags: latest, base, base-20260426.0.520785
+GitCommit: 2c315ee0bfb72104b4754026f53daa9761dd9c1e
+GitFetch: refs/tags/v20260426.0.520785
 File: Dockerfile.base
 
-Tags: base-devel, base-devel-20260419.0.517065
-GitCommit: 25354c172306956092edf933ad385f23975ebedd
-GitFetch: refs/tags/v20260419.0.517065
+Tags: base-devel, base-devel-20260426.0.520785
+GitCommit: 2c315ee0bfb72104b4754026f53daa9761dd9c1e
+GitFetch: refs/tags/v20260426.0.520785
 File: Dockerfile.base-devel
 
-Tags: multilib-devel, multilib-devel-20260419.0.517065
-GitCommit: 25354c172306956092edf933ad385f23975ebedd
-GitFetch: refs/tags/v20260419.0.517065
+Tags: multilib-devel, multilib-devel-20260426.0.520785
+GitCommit: 2c315ee0bfb72104b4754026f53daa9761dd9c1e
+GitFetch: refs/tags/v20260426.0.520785
 File: Dockerfile.multilib-devel
+
+Tags: repro, repro-20260426.0.520785
+GitCommit: 2c315ee0bfb72104b4754026f53daa9761dd9c1e
+GitFetch: refs/tags/v20260426.0.520785
+File: Dockerfile.repro
 


tianon:
```diff
+Tags: repro, repro-20260426.0.520785
+GitFetch: refs/tags/v20260426.0.520785
+GitCommit: 2c315ee0bfb72104b4754026f53daa9761dd9c1e
+File: Dockerfile.repro
```

The `repro` variant is a great idea 👍 -- but a Dockerfile-based build can only go partway. `rm -f /var/cache/ldconfig/aux-cache` handles one specific non-deterministic file, and the normalized `created` timestamp is a good thought, but the `RUN ldconfig && sed ...` layer will still carry non-deterministic mtimes on anything it touches -- so the layer digest will still vary between builds.

The right tool for this in DOI is `Builder: oci-import`. You'd need a dedicated branch (e.g. `dist-repro`) that gets force-pushed on each release containing the [OCI image layout (`blobs/`, `oci-layout`, `index.json`)](https://github.com/opencontainers/image-spec/blob/v1.1.1/image-layout.md) -- using a tag isn't workable because OCI layouts contain the actual image blobs, and storing those at every tagged commit accumulates binary history in git that makes cloning the repo untenable over time. With a force-pushed branch, only the current release's blobs are in the tree at any given time. The library entry would look something like:

```
Tags: repro, repro-20260426.0.520785
GitFetch: refs/heads/dist-repro
GitCommit: 2c315ee0bfb72104b4754026f53daa9761dd9c1e
Builder: oci-import
File: index.json
```

[See how `debian` does this](https://github.com/docker-library/official-images/blob/a5b03e51da25f39284b60cab5bf45a261f233b3b/library/debian) with its `dist-amd64` branch in `docker-debian-artifacts`. With `oci-import`, DOI validates blob digests on ingest and publishes the image content verbatim -- the layer and config digests from your build are preserved end-to-end through the full publish pipeline. No `Dockerfile.repro` needed at all.

tianon:
I'm also curious about the relationship between the `repro` and `base` rootfs tarballs -- are they built from the same package set and differ only in how they're produced (i.e. `repro` is just `base` built reproducibly), or do they actually diverge in content?

Antiz96:
Hello @tianon,

I am the person who worked on the `repro` variant. Thanks for the review!
Please, note that I may not be *fully* aware of how things works on your side for the publishing process :angel: 

> The repro variant is a great idea 👍 -- but a Dockerfile-based build can only go partway. rm -f /var/cache/ldconfig/aux-cache handles one specific non-deterministic file, and the normalized created timestamp is a good thought, but the RUN ldconfig && sed ... layer will still carry non-deterministic mtimes on anything it touches -- so the layer digest will still vary between builds.

We are building the repro image with the `--source-date-epoch=${SOURCE_DATE_EPOCH} --rewrite-timestamp` docker build options, which appears to be sufficient to normalize mtimes here. For what it's worth, we were able to achieve the same digest between builds already (comparing the output of `docker inspect --format '{{.Digest}}' <image>`). Am I missing something?

> The right tool for this in DOI is Builder: oci-import. You'd need a dedicated branch (e.g. dist-repro) that gets force-pushed on each release containing the [OCI image layout (blobs/, oci-layout, index.json)](https://github.com/opencontainers/image-spec/blob/v1.1.1/image-layout.md) -- using a tag isn't workable because OCI layouts contain the actual image blobs, and storing those at every tagged commit accumulates binary history in git that makes cloning the repo untenable over time. With a force-pushed branch, only the current release's blobs are in the tree at any given time. The library entry would look something like:
> 
> ```
> Tags: repro, repro-20260426.0.520785
> GitFetch: refs/heads/dist-repro
> GitCommit: 2c315ee0bfb72104b4754026f53daa9761dd9c1e
> Builder: oci-import
> File: index.json
> ```
> 
> [See how debian does this](https://github.com/docker-library/official-images/blob/a5b03e51da25f39284b60cab5bf45a261f233b3b/library/debian) with its dist-amd64 branch in docker-debian-artifacts. With oci-import, DOI validates blob digests on ingest and publishes the image content verbatim -- the layer and config digests from your build are preserved end-to-end through the full publish pipeline. No Dockerfile.repro needed at all.

That's interesting, thanks for the information! I indeed see the benefit of publishing a prebuilt image / OCI layout (if I understood correctly) in that context. That’s quite a different workflow from what we currently have however, so I’ll need to take a closer look at how this could fit into our pipeline.

> I'm also curious about the relationship between the repro and base rootfs tarballs -- are they built from the same package set and differ only in how they're produced (i.e. repro is just base built reproducibly), or do they actually diverge in content?

Yes, they are built from the same package set, `repro` aims to be `base` built reproducibly.

tianon:
> We are building the repro image with the `--source-date-epoch=${SOURCE_DATE_EPOCH} --rewrite-timestamp` docker build options, which appears to be sufficient to normalize mtimes here.

This is exactly the crux of the problem - our pipeline will _not_ build with those options enabled, so this variant will not be reproducible here as-is. :sweat_smile:

tianon:
> Yes, they are built from the same package set, `repro` aims to be `base` built reproducibly.

So what's the use case for publishing both? :sweat_smile:

Antiz96:
> > We are building the repro image with the `--source-date-epoch=${SOURCE_DATE_EPOCH} --rewrite-timestamp` docker build options, which appears to be sufficient to normalize mtimes here.
> 
> This is exactly the crux of the problem - our pipeline will _not_ build with those options enabled, so this variant will not be reproducible here as-is. 😅

Ah okay, I see. Then I understand the need for the `Builder: oci-import` setup! Thanks for the clarifications :smile:
I'll see if we can exclude the `repro` tag from those release PRs and only push it to our own namespace while we evaluate the switch to such a setup.

> > Yes, they are built from the same package set, `repro` aims to be `base` built reproducibly.
> 
> So what's the use case for publishing both? 😅

I forgot to mention one crucial difference between the two images. To ensure reproducibility, the local pacman keyring is stripped out from the `repro` image (as keys in the keyring are *obviously* non-deterministic). This means that pacman is not usable out of the box with this image. Users are expected to regenerate the keyring first before being able to install / update packages through pacman (by running `pacman-key --init && pacman-key --populate archlinux`). While waiting to find a suitable solution to this technical constraint, we are therefore providing this reproducible image under a dedicated tag as a first milestone (so we are not shipping such a breaking change in our `base` image).

hashworks:
I guess publishing a repro tag from a tarball in the official DockerHub repo doesn't make much sense if it differs from `:repro` images from other sources.

> I'm also curious about the relationship between the repro and base rootfs tarballs -- are they built from the same package set and differ only in how they're produced (i.e. repro is just base built reproducibly), or do they actually diverge in content?

To clarify here, the `repro` tag uses the [state in the ArchLinux Archive of the previous day](https://archive.archlinux.org/repos/2026/04/25/). This means while base will probably contain the same packages as repro, they might be a bit newer. So if anything base could be built upon repro, but it doesn't really make sense for the Arch use case due to daily rebuilds (in repositories we control directly).

Antiz96:
> I guess publishing a repro tag from a tarball in the official DockerHub repo doesn't make much sense if it differs from :repro images from other sources.

I agree, let's exclude the `repro` tag from the official Docker Hub library releases and just keep it in our own namespace for now. We can still revise / reconsider later once we are ready to switch to a `Builder: oci-import` based setup as mentioned by @tianon :)

hashworks:
@tianon feel free to close this for now (if someone with access to the `archlinux-github` technical user isn't faster :smile: )

PR #20765, opened 2026-02-02
Update Jetty to 12.0.32 & 12.1.6, and remove EOL versions

lachlan-roberts:
- Update Jetty version `12.0.31` -> `12.0.32`.
- Update Jetty version `12.1.5` -> `12.1.6`.
- Remove EOL versions `9.4.x`, `10.0.x`, `11.0.x` which will receive no more public releases (see https://github.com/jetty/jetty.project/issues/13918).
- Update "latest" JDK to 25.

PR #21331, opened 2026-04-27
Release of GeoNetwork 4.4.10 and 4.2.15

juanluisrp:


tianon (dismissed):
```diff
+ENV GN_DOWNLOAD_MD5=79cd931ac7fc0be044ec1f2dd44f3f96
```

As a heads-up: MD5 is cryptographically broken for integrity verification purposes -- SHA256 (and/or better on top, PGP signature verification if GeoNetwork provides it) would be a significant improvement. This isn't something you introduced in this PR, and I'm not going to block a version bump over it, but I'd love to see it addressed in a follow-up. 🙇

tianon:
Oh, we've got some unsupported base images: :sweat_smile:

 - geonetwork:4.2.15 (FROM jetty:9-jre8) -- completely unsupported base!
 - geonetwork:4.4.10 (FROM jetty:9-jre11) -- completely unsupported base!

https://github.com/docker-library/official-images/pull/20765#issuecomment-3987807359

juanluisrp:
@tianon we have replaced the unsupported Jetty images by Tomcat 9 ones.

tianon:
Thanks!  Since we're the ones who pushed the Jetty -> Tomcat migration, a couple of things worth confirming before this lands so users aren't caught off guard:

**Running as root.** The previous images ran as the `jetty` user; the new images have no `USER` instruction and the base `tomcat` image runs as root. That's a concrete breaking change: existing persistent volumes with `jetty`-owned files will hit permission errors on upgrade, and anyone building `FROM geonetwork:4.x` expecting a non-root user will be surprised. Is this intentional? Just want to make sure you're aware so you know what to tell users when they run into it.

**Jetty -> Tomcat path and env changes.** A few things downstream users may have relied on are now gone:

- `$JETTY_HOME` / `$JETTY_BASE` env vars
- `/var/lib/jetty/webapps/geonetwork/` (4.2) -- now `/usr/local/tomcat/webapps/geonetwork/`
- `CMD ["java", "-jar", "/usr/local/jetty/start.jar"]`

Anyone `COPY`-ing to those paths in a `FROM geonetwork:4.x` image, or referencing those env vars in scripts, silently breaks. Probably unavoidable given the base swap, but worth confirming you've considered it and plan to document it.

---

Both `Dockerfile`s would benefit from `set -eux` at the top of the non-trivial `RUN` blocks -- it propagates failures from `curl`, the `md5sum` check, and `unzip` properly instead of relying on `&&` chaining:

```diff
-RUN apt-get -y update && \
-    apt-get -y install --no-install-recommends \
-        curl \
-        unzip && \
-    rm -rf /var/lib/apt/lists/* && \
-    mkdir -p ${DATA_DIR} && \
-    mkdir -p /usr/local/tomcat/webapps/geonetwork
+RUN set -eux; \
+    apt-get update; \
+    apt-get install -y --no-install-recommends \
+        curl \
+        unzip \
+    ; \
+    rm -rf /var/lib/apt/lists/*; \
+    mkdir -p "${DATA_DIR}"; \
+    mkdir -p /usr/local/tomcat/webapps/geonetwork
```

---

In `geonetwork_4.2/docker-entrypoint.sh`:

```diff
+export CATALINA_OPTS="${JAVA_OPTIONS}"
```

`JAVA_OPTIONS` isn't defined -- looks like a leftover from the Jetty entrypoint that translated `JAVA_OPTS` into what Jetty expected. With Tomcat, `catalina.sh` reads `JAVA_OPTS` from the environment directly, so this line can just be removed.

---

One question about `tomcat/server.xml` in both variants: the HTTP connector sets `maxPostSize="500000"` (~488 KiB), which is about a quarter of Tomcat's default (2 MiB). For a GIS application handling KML, GeoJSON, or shapefile uploads, that seems like it could be tight -- is that intentional, and does it cover all the upload paths in GeoNetwork's UI?

---

FYI: `tomcat:9-jre8` and `tomcat:9-jre11` carry over the same "no OS version pinned" pattern your Jetty tags were using, so no new concern there -- just noting the pattern continues.

juanluisrp:
Thanks for the thorough review @tianon. We've pushed fixes for the actionable items (commit [6f8d309](https://github.com/geonetwork/docker-geonetwork/commit/6f8d30999bbc22e58ca54784251b567c21673dbc) in docker-geonetwork):

  - `set -eux`: applied to all RUN blocks in both Dockerfiles.
  - `CATALINA_OPTS="${JAVA_OPTIONS}"`: removed from the 4.2.15 entrypoint. JAVA_OPTIONS was Jetty's internal variable. Tthe old entrypoint translated JAVA_OPTS into it, so users always set JAVA_OPTS, which Tomcat reads directly with no translation needed.
  - `maxPostSize="500000"`: dropped from both `server.xml` files, restoring Tomcat's default (2 MiB).

On the remaining points:

  - **Running as root:** Yes, intentional. The Tomcat base image runs as root and we haven't overridden that. I guess that making it to run as the same user that the Jetty base image does would break other things.
  - **Jetty -> Tomcat breaking changes:** Also intentional and unavoidable given the base image swap. We can call out the `$JETTY_HOME/$JETTY_BASE` env vars, the old webapps path, and the `CMD` change explicitly in the documentation.
  - **Image OS version pin:** Should we modify this and pin to a specific OS version?

juanluisrp:
Hi @tianon,

Just wanted to bump this and ask for a re-review since the latest changes from last week.

tianon:
None of these are blockers so I'm going to go ahead and merge:

---

On the OS version pinning question -- yes, I'd recommend it. Both `tomcat:9-jre8` and `tomcat:9-jre11` currently resolve to `noble` (Ubuntu 24.04 LTS), so the versioned equivalents today are `tomcat:9-jre8-temurin-noble` and `tomcat:9-jre11-temurin-noble`. Pinning avoids the square wheel when the unversioned tag eventually points to a new OS release (or even a different Java vendor, which isn't planned but is possible).

---

```diff
+ENV GN_DOWNLOAD_MD5=79cd931ac7fc0be044ec1f2dd44f3f96
```

(same for the 4.4.10 `GN_DOWNLOAD_MD5` value)

Not introduced by this PR, but since the Dockerfiles are being revised anyway -- MD5 is cryptographically broken and isn't something we'd want to rely on for download verification. It appears that the GeoNetwork project does publish `.sha256` files alongside their SourceForge releases, so this might be a straightforward upgrade? PGP signature verification on top of that would be even better if they publish a KEYS file, but swapping from MD5 to SHA256 would be a significant upgrade. 👀

---

Soft note on both entrypoints -- the `KB_URL` and `ES_*` values get interpolated directly into `sed` replacement strings using `#` as the delimiter:

```bash
sed -i "s#http://localhost:9200#${ES_PROTOCOL:=http}://${ES_HOST}:${ES_PORT:=9200}#g" ...
sed -i "s#http://localhost:5601#${KB_URL}#g" ...
```

If any of those values contains a `#`, or an unescaped `&` (which sed treats as "the matched text"), the substitution will produce garbage or fail. Expected inputs are URLs and hostnames so this is unlikely to matter in practice, but worth flagging. 👀

tianon (approved):

PR #21335, opened 2026-04-28
Update bonita, add 2026.1

akantcheff:


tianon:
```diff
-  permission java.lang.RuntimePermission "getenv.*";
```

I noticed `getenv.*` was removed from the `deny codeBase "file:${java.io.tmpdir}/-"` block in `prograde.policy` -- I'm curious about the reasoning there. 🤔

In container environments, env vars are the standard way to pass sensitive configuration (passwords, tokens, etc.), so relaxing a policy that previously blocked temp-directory bytecode from reading them is worth a second look. Is this intentional for 11.0.0?

I tried finding the justification in the commits, but https://github.com/bonitasoft/bonita-distrib/commit/d6f12e5666e1a3ada7b72093aee635ece2eb7972 was the best I could find - I can't find what that's actually merging _from_ to see if there's more detail.

abirembaut:
Hello @tianon ,

Thank you for spotting that.

This change to the default policy is intentional, as the initial policy we introduced in January was too restrictive for some of our Java connectors that need to read environment variables to work. Reading environment variables from Groovy scripts (which is the attack vector this policy aims to protect against) is still blocked by the policy.
As the policy can be configured as explained in our documentation, it is still possible to reinforce this if preferred when not using the Java connectors that use env variables. We simply did not want to block connector usage by default.

tianon (approved):

PR #21338, opened 2026-04-28
Update openjdk

tianon:
Changes:

- https://github.com/docker-library/openjdk/commit/d75506b: Update 27 to 27-ea+19


tianon:
> Looks good to me -- clean EA build bump with consistent checksums across all variants. 👍

PR #21351, opened 2026-04-30
[nats] Release v2.14.0

neilalexander:
Release info: https://github.com/nats-io/nats-server/releases/tag/v2.14.0


library/nats
tianon, line 71: Is the 2.11 series still actively supported? :eyes:  (if not, these should be removed)

tianon (approved):

PR #21352, opened 2026-04-30
Update ELK official images for 8.19.15 and 9.3.4

ninalee12:
Update ELK official images for `8.19.15` and `9.3.4`

### DIFF
--- a/library/elasticsearch
+++ b/library/elasticsearch
@@ -13,7 +13,7 @@ Architectures: amd64, arm64v8
 GitFetch: refs/heads/9.2
 GitCommit: f365a5d7a0c03b5cb6df3cad5031fe5e2489f800
 
-Tags: 9.3.3
+Tags: 9.3.4
 Architectures: amd64, arm64v8
 GitFetch: refs/heads/9.3
-GitCommit: a1615a8e16daafa082073ae28ddf389bcbf6845c
+GitCommit: 82d52284ccbfe049efc893d110b35ccde03235fa
--- a/library/elasticsearch
+++ b/library/elasticsearch
@@ -3,10 +3,10 @@ GitRepo: https://github.com/elastic/dockerfiles.git
 Directory: elasticsearch
 Builder: buildkit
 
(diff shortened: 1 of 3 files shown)


tianon (approved):
```diff
+LABEL org.label-schema.build-date="2026-04-28T13:06:49.648073236Z" \
```

As a heads-up for a future cleanup: the `org.label-schema.*` keys are deprecated on their own merits (https://github.com/label-schema/label-schema.org#readme) -- I'd suggest dropping them entirely rather than carrying them alongside the OCI equivalents. Not a blocker on this bump. 😅

---

```diff
+RUN curl --retry 10 -S -L --output /tmp/elasticsearch.tar.gz https://artifacts-no-kpi.elastic.co/downloads/elasticsearch/elasticsearch-8.19.15-linux-$(arch).tar.gz
```

The 8.x `elasticsearch` `curl` is missing `--fail` / `-f` -- the 9.x variant already has it (`curl -f --retry 10 -S -L`). Without `--fail`, an HTTP error response (404, 500, etc.) is written silently to the output file and the build proceeds as though the download succeeded. Worth syncing the two variants on this. 👀

---

```diff
+RUN curl -Lo - https://artifacts.elastic.co/downloads/logstash/logstash-8.19.15-linux-$(arch).tar.gz | \
+  tar zxf - -C /usr/share && \
```

Two things on the 8.x `logstash` download, both worth matching to what the 9.x variant already does:

- Same missing `--fail` as above.
- Piping `curl` directly into `tar` means `curl`'s exit code is lost -- only `tar`'s propagates. A 404 or a truncated download that `tar` can still partially extract won't fail the build. The 9.x `logstash` already downloads to a file first (`--output logstash.tar.gz`) before extracting -- that's the right pattern.

Not blocking on the version bump itself, but these would be good to clean up to match the 9.x behavior. 🙇

breskeby:
@tianon thanks for the feedback. the elasticsearch team will look into applying your suggestions. Will keep you posted.

PR #21362, opened 2026-05-01
feat: Update influxdb2 image to 2.9.0, remove 2.7.x

devanbenz:


yosifkit:
This has conflicts since I merged https://github.com/docker-library/official-images/pull/21361 first. Can you rebase and force push?

devanbenz:
> This has conflicts since I merged #21361 first. Can you rebase and force push?

I've updated the PR to resolve MC

yosifkit:
```diff
--- a/influxdb_alpine/Dockerfile
+++ b/influxdb_alpine/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.23
+FROM alpine:3.21
```

The previous `influx:alpine` image (`2.8.0-alpine`) moved to Alpine 3.23 with https://github.com/docker-library/official-images/pull/21361, but the new `influx:alpine` image (`2.9.0-alpine`) is still Alpine 3.21. Do you want to keep it that way?

devanbenz:
> ```diff
> --- a/influxdb_alpine/Dockerfile
> +++ b/influxdb_alpine/Dockerfile
> @@ -1,4 +1,4 @@
> -FROM alpine:3.23
> +FROM alpine:3.21
> ```
> 
> The previous `influx:alpine` image (`2.8.0-alpine`) moved to Alpine 3.23 with #21361, but the new `influx:alpine` image (`2.9.0-alpine`) is still Alpine 3.21. Do you want to keep it that way?

Good call out, I was still using an older template where we had alpine at 3.21, I've adjusted this pr to reflect the newer alpine version.

yosifkit (approved):

tianon:
```diff
+      amd64) echo '3c947a8dcd88856a32c172081db091c38059394fb57a15fa43871f6d046427e1  /usr/local/bin/dasel' ;; \
+      arm64) echo 'a128c5554c53e6e4af880700adba1d212ce651db208da1592fb1cae0e959cbc6  /usr/local/bin/dasel' ;; \
```

Minor note -- the `*`-prefixed binary mode is the conventional format for verifying downloaded binaries, rather than double-space:

```sh
amd64) echo '3c947a8dcd88856a32c172081db091c38059394fb57a15fa43871f6d046427e1 */usr/local/bin/dasel' ;; \
arm64) echo 'a128c5554c53e6e4af880700adba1d212ce651db208da1592fb1cae0e959cbc6 */usr/local/bin/dasel' ;; \
```

Not a blocker.

---

There's quite a bit of shell scripting here beyond the `Dockerfile`s themselves. Since InfluxData directly maintains InfluxDB, I'd ask the question more directly than I would for a community-packaged image: the ideal for Official Images is few or no shell scripts -- just a `Dockerfile` and whatever the application ships. The packaging layer has narrower error handling, a limited test surface, and bugs tend to be silent; all of this is better handled natively. The examples below feel like reasonable upstream feature requests that, if implemented, would chip away at that goal:

- `DOCKER_INFLUXDB_INIT_PASSWORD_FILE` / `DOCKER_INFLUXDB_INIT_USERNAME_FILE` / `DOCKER_INFLUXDB_INIT_ADMIN_TOKEN_FILE` -- the `_FILE` env-var pattern for reading secrets from mounted files is a common container primitive that many daemons support natively. If `influxd` or the `influx` CLI supported these directly, this shell logic could go away.
- `set_init_resource_ids` -- parsing `influx user/org/bucket list --hide-headers | cut -f 1` to extract IDs is fragile; if `influx setup` returned structured output (JSON) with the created IDs, this post-setup scraping wouldn't be needed.
- `influxd::config::get` -- reading config values via dasel to detect `http-bind-address` / `tls-cert` / `tls-key` exists because `influxd` doesn't expose these at startup in a scriptable way; native env-var overrides or a `--print-config` flag would eliminate this entirely.

Not blocking, and I realize this is a longer-term ask -- just wanted it on record. 😅

If the scripts do stay, a couple of notes contingent on that:

```diff
+  dasel -i "${config_format}" 'get("'"${primary_key}"'")' < "${INFLUXD_CONFIG_PATH}" 2>/dev/null || \
```

The `'get("'"${primary_key}"'")' ` pattern embeds the shell variable directly into the dasel expression by concatenating out of the single-quoted string -- same injection class as the previous `dasel -r/-s` form. For the `influxd::config::get` callers this is low risk since all call sites use hard-coded key names, but `init_influxd` uses the same pattern with `init_bind_addr`, which derives from `INFLUXD_INIT_PORT` -- a user-settable env var.

Dasel v3 actually resolves `$VARNAME` expressions by calling `os.Getenv()` as a fallback, so you can avoid the shell concatenation entirely:

```sh
INFLUXD_CONFIG_KEY="${primary_key}" dasel -i "${config_format}" 'get($INFLUXD_CONFIG_KEY)' < "${INFLUXD_CONFIG_PATH}" 2>/dev/null
```

No matter what `primary_key` contains, it never touches the expression string.

---

And a small consistency nit: the Alpine `cleanup_influxd` uses `rm -rf -- "${BOLT_PATH}"` (end-of-options `--`) while the Debian variant omits it. Worth aligning.

PR #21366, opened 2026-05-03
archlinux: Release 20260503.0.523481

archlinux-github:

---

### DIFF
--- a/library/archlinux
+++ b/library/archlinux
@@ -5,18 +5,18 @@ Maintainers: Santiago Torres-Arias <santiago@archlinux.org> (@SantiagoTorres),
              Justin Kromlinger <hashworks@archlinux.org> (@hashworks)
 GitRepo: https://gitlab.archlinux.org/archlinux/archlinux-docker.git
 
-Tags: latest, base, base-20260419.0.517065
-GitCommit: 25354c172306956092edf933ad385f23975ebedd
-GitFetch: refs/tags/v20260419.0.517065
+Tags: latest, base, base-20260503.0.523481
+GitCommit: 89b8bd477ecb2ad67f5069083b8255162a26cf7d
+GitFetch: refs/tags/v20260503.0.523481
 File: Dockerfile.base
 
-Tags: base-devel, base-devel-20260419.0.517065
-GitCommit: 25354c172306956092edf933ad385f23975ebedd
-GitFetch: refs/tags/v20260419.0.517065
+Tags: base-devel, base-devel-20260503.0.523481
+GitCommit: 89b8bd477ecb2ad67f5069083b8255162a26cf7d
+GitFetch: refs/tags/v20260503.0.523481
 File: Dockerfile.base-devel
 
-Tags: multilib-devel, multilib-devel-20260419.0.517065
-GitCommit: 25354c172306956092edf933ad385f23975ebedd
-GitFetch: refs/tags/v20260419.0.517065
+Tags: multilib-devel, multilib-devel-20260503.0.523481
+GitCommit: 89b8bd477ecb2ad67f5069083b8255162a26cf7d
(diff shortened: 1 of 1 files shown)


yosifkit:
This change is a bit strange. Are there times that `ldconfig` or this `sed` would fail that would be reasonable to allow that failure into the final image?
```diff
 RUN ldconfig && \
-    sed -i '/BUILD_ID/a VERSION_ID=20260419.0.517065' /etc/os-release
+    sed -i '/BUILD_ID/a VERSION_ID=20260503.0.523481' /etc/os-release && \
+    true
```

christian-heusel:
cc @antiz96 regarding https://gitlab.archlinux.org/archlinux/archlinux-docker/-/commit/4b15f9a1a11d11474cad201521962dc6192630ea

Antiz96:
> This change is a bit strange. Are there times that `ldconfig` or this `sed` would fail that would be reasonable to allow that failure into the final image?
> 
> ```diff
>  RUN ldconfig && \
> -    sed -i '/BUILD_ID/a VERSION_ID=20260419.0.517065' /etc/os-release
> +    sed -i '/BUILD_ID/a VERSION_ID=20260503.0.523481' /etc/os-release && \
> +    true
> ```

Hello @yosifkit,

Failures from `ldconfig` or `sed` would still cause the build to properly exit on error here as the `true` statement is only executed on success (`[...] **&&** true`), it should not allow failure into the final image.

For the context, this change has been introduced for our new "repro" tag image, which aims to provide a reproducible version of the image (this "repro" tag is not included in these official Docker Hub releases at the moment): In order to avoid non-determinism, we are deleting the ldconfig auxiliary cache file from the "repro" image. We therefore updated the script that generates our Dockerfiles for each tag to:

- Delete the file for the repro tag image (`&& rm -f /var/cache/ldconfig/aux-cache`)
- Do nothing for other tags (`&& true`)

This could be done differently technically speaking, it was just the straightforward way we came up with to integrate this condition in our existing script. I get that this change might be a bit confusing at first glance though, but since we are using an "AND" logical operator here, it should be harmless.

However, if this is considered a blocking point on your side for this release (or if you generally prefer this to be done in a more elegant way on our side), please let me know and will look at a way to not introduce the "&& true" part in here. :slightly_smiling_face:

tianon:
It's not explicitly blocking but it's definitely a code smell / pause :smile:

My gut impression would be that maybe it's sane to remove the `ldconfig` cache all the time?  Is the downside of running without that cache significant?

I'll go ahead and merge this now, but please do work on ways to clean that up (especially if it cleans up the divergence against repro / brings us closer to those being the same or interdependent :eyes:).

tianon (approved):

Antiz96:
> It's not explicitly blocking but it's definitely a code smell / pause 😄
> 

Fair enough

> My gut impression would be that maybe it's sane to remove the `ldconfig` cache all the time? Is the downside of running without that cache significant?
> 

I don't think there's much downside to it. It's simply that, when we decided to implement the repro image as a separate tag (as a first milestone), the idea was to *not* introduce any change for the other already existing tags in the process. But maybe this can be revised now :)

> I'll go ahead and merge this now, but please do work on ways to clean that up (especially if it cleans up the divergence against repro / brings us closer to those being the same or interdependent 👀).

Will do! I'll make a change to either get rid of the confusing "&& true" part or remove the `ldconfig` cache for every tags. Thanks!

Antiz96:
Patch submitted and merged on our side. The `ldconfig + sed` RUN command will now only be expanded in the repro tag's Dockerfile to include the aux-cache cleanup. The Dockerfile for the other tags will remain unchanged, the next release will therefore have the "&& true" part dropped!

Thanks for the feedback :hugs:

PR #21372, opened 2026-05-04
[security] Update httpd

yosifkit:
Changes:

- https://github.com/docker-library/httpd/commit/80149d9: Update to 2.4.67


reijnden:
Hi there,
do we have a timeline on when the 2.4.67 docker image will be available. This is a critical security issue.
Thanks

tianon (approved):

PR #21380, opened 2026-05-05
eclipse-temurin: Update tags and Git commits for multiple images

gdams:
Noting that Ubuntu Resolute becomes the default "latest" distro now that it's GA.

### DIFF
--- a/library/eclipse-temurin
+++ b/library/eclipse-temurin
@@ -310,7 +310,7 @@ Directory: 17/jdk/alpine/3.21
 
 Tags: 17.0.19_10-jdk-resolute, 17-jdk-resolute, 17-resolute
 SharedTags: 17.0.19_10-jdk, 17-jdk, 17
-Architectures: amd64, arm32v7, arm64v8, ppc64le, riscv64, s390x
+Architectures: amd64, arm32v7, arm64v8, ppc64le, s390x
 GitCommit: df6138afaf1b564116e895b0acd51d70e11cd996
 Directory: 17/jdk/ubuntu/resolute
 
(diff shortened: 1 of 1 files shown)


gdams:
> DOI cannot support riscv64 on Ubuntu 25.10+ since it uses RVA23 and hardware does not exist yet

@tianon is this still correct? I see the `ubuntu` image ships a `riscv64` build of 26.04?

yosifkit:
>> DOI cannot support riscv64 on Ubuntu 25.10+ since it uses RVA23 and hardware does not exist yet
>
>@tianon is this still correct? I see the `ubuntu` image ships a `riscv64` build of 26.04?

Yes, this is still accurate. We currently only build on actual hardware and do not yet have access to an RVA23-enabled RISC-V server. We can build the Ubuntu image itself because it doesn't require running any binaries from the image.

gdams:
> > > DOI cannot support riscv64 on Ubuntu 25.10+ since it uses RVA23 and hardware does not exist yet
> > 
> > 
> > @tianon is this still correct? I see the `ubuntu` image ships a `riscv64` build of 26.04?
> 
> Yes, this is still accurate. We currently only build on actual hardware and do not yet have access to an RVA23-enabled RISC-V server. We can build the Ubuntu image itself because it doesn't require running any binaries from the image.

okay I've backed out the riscv64 changes for now. Do you have an ETA on when this will be resolved?

tianon:
> okay I've backed out the riscv64 changes for now. Do you have an ETA on when this will be resolved?

Unfortunately, Ubuntu made the choice to target an ISA that doesn't really exist yet -- I get that they have constraints that make that make sense for them, but this is what can only be described as "the expected downstream effect" (ie, unfortunately no, we don't have an ETA because we don't control the hardware vendors and there aren't many "cloud" riscv64 providers we can lean on so the logistics are even more complicated).

Using Debian instead is an alternative, but one with other side effects you'd need to weigh carefully for your users.

tianon (approved):

PR #21383, opened 2026-05-05
[SECURITY] Redis seurity fixes in 6.x, 7.x and 8.x series

Peter-Sh:
Hi Docker team!

Please update Redis to the newest versions that include security fixes. Various CVEs are addressed in Redis and its modules, including RCE vulnerabilities.

https://github.com/redis/redis/releases/tag/6.2.22
https://github.com/redis/redis/releases/tag/7.2.14
https://github.com/redis/redis/releases/tag/7.4.9
https://github.com/redis/redis/releases/tag/8.2.6
https://github.com/redis/redis/releases/tag/8.4.3
https://github.com/redis/redis/releases/tag/8.6.3

Note: For the 6.x and 7.x series, specifically for Docker:
* We are using the new repository layout, which is branch/tag-based, as it is for 8.x.
* `gosu` has been replaced with `setpriv` in the entrypoint.

### DIFF
--- a/library/redis
+++ b/library/redis
@@ -18,40 +18,40 @@ GitCommit: fa43b354211879f28d1773ae94e740bd534eb6ef
 GitFetch: refs/tags/v8.8-m03
 Directory: alpine
 
-Tags: 8.6.2, 8.6, 8, 8.6.2-trixie, 8.6-trixie, 8-trixie, latest, trixie
+Tags: 8.6.3, 8.6, 8, 8.6.3-trixie, 8.6-trixie, 8-trixie, latest, trixie
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, riscv64, ppc64le, s390x
-GitCommit: 8c81e2a44cae9258294718d767ce594e5cbbf20e
-GitFetch: refs/tags/v8.6.2
+GitCommit: 70233d27f12d7df70a1146d9348c6a70c31eebcc
+GitFetch: refs/tags/v8.6.3
 Directory: debian
 
-Tags: 8.6.2-alpine, 8.6-alpine, 8-alpine, 8.6.2-alpine3.23, 8.6-alpine3.23, 8-alpine3.23, alpine, alpine3.23
+Tags: 8.6.3-alpine, 8.6-alpine, 8-alpine, 8.6.3-alpine3.23, 8.6-alpine3.23, 8-alpine3.23, alpine, alpine3.23
 Architectures: amd64, arm32v6, arm32v7, arm64v8, i386, ppc64le, riscv64, s390x
-GitCommit: 8c81e2a44cae9258294718d767ce594e5cbbf20e
-GitFetch: refs/tags/v8.6.2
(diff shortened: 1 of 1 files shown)


library/redis
```diff
@@ -18,40 +18,40 @@ GitCommit: fa43b354211879f28d1773ae94e740bd534eb6ef
...
-Tags: 8.2.5-alpine, 8.2-alpine, 8.2.5-alpine3.22, 8.2-alpine3.22
+Tags: 8.2.6-alpine, 8.2-alpine, 8.2.6-alpine3.22, 8.2-alpine3.22
 Architectures: amd64, arm32v6, arm32v7, arm64v8, i386, ppc64le, riscv64, s390x
-GitCommit: d42260abdb8f76db20151682a71e0e154a8147f3
-GitFetch: refs/tags/v8.2.5
+GitCommit: ae2372898bd403acf24f8044fd72968cedc3f370
+GitFetch: refs/tags/v8.2.6
 Directory: alpine
 
 Tags: 8.0.6, 8.0, 8.0.6-bookworm, 8.0-bookworm
```
yosifkit, line 57: Given that `8.0.x` has reached end of life and is not updated with this security fix, I'd suggest that the `8.0.6*` images be removed from the library file (in another PR, no need to delay this security fix).

yosifkit (approved):

tianon (approved):
```diff
-#!/bin/sh
+#!/bin/bash
```

The `bookworm` variants switched to `#!/bin/bash`, but the `alpine` counterparts kept `#!/bin/sh` -- and the actual code in both is identical. The new `&&`-separated condition is plain POSIX, so `#!/bin/bash` isn't buying anything here. The inconsistency looks unintentional; `#!/bin/sh` works for both. 🤔

---

```diff
+SETPRIV="/usr/bin/setpriv --reuid redis --regid redis --clear-groups"
...
+	exec $SETPRIV "$0" "$@"
```

`$SETPRIV` is defined once and used once, and the unquoted expansion relies on word-splitting to split the flags correctly. This works, but inlining the command is cleaner and removes the implicit dependency on word-splitting:

```sh
exec /usr/bin/setpriv --reuid redis --regid redis --clear-groups "$0" "$@"
```

Peter-Sh:
@yosifkit @tianon Thanks for the quick merge and feedback!

Regarding 8.0.x series, good point, I'll check it with the team and mark as EOL.

Regarding entrypoint changes, I think both came from the 8.x (where entrypoint is more complex), would also fix it.

One question: do you accept PRs modifying entrypoint/Dockerfile/library without actual software version change?

Because 6.x and 7.x would get a new version only in case of CVE found in the core.

yosifkit:
> One question: do you accept PRs modifying entrypoint/Dockerfile/library without actual software version change?

Of course! We're happy to have fixes/updates applied to the entrypoint script and Dockerfiles.

tianon:
We honestly prefer that to having to block security fixes on unacceptable unrelated changes, for example :smile:

PR #21417, opened 2026-05-08
[ros] add ROS Lyrical beta images

mikaelarguedas:
This PR is to create a first version of the docker images for the ROS Lyrical release (the next LTS version of ROS)

The official release will take place on May 23rd at which point a new PR updating the Dockerfiles and the latest tag will be made

Cross-referencing:
https://github.com/osrf/docker_images/pull/877
https://discourse.openrobotics.org/t/lyrical-luth-test-and-tutorial-party-instructions/54427

### DIFF
--- a/library/ros
+++ b/library/ros
@@ -68,6 +68,28 @@ Directory: ros/kilted/ubuntu/noble/perception
 
 
 ################################################################################
+# Release: lyrical
+
+########################################
+# Distro: ubuntu:resolute
+
+Tags: lyrical-ros-core, lyrical-ros-core-resolute
+Architectures: amd64, arm64v8
+GitCommit: 0d65786f3d9bb10d55dc28ccd5b87da204240d1a
+Directory: ros/lyrical/ubuntu/resolute/ros-core
+
+Tags: lyrical-ros-base, lyrical-ros-base-resolute, lyrical
+Architectures: amd64, arm64v8
+GitCommit: 0d65786f3d9bb10d55dc28ccd5b87da204240d1a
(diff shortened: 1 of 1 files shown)


tianon (approved):
```diff
+RUN curl -L -s -o /tmp/ros2-testing-apt-source.deb https://github.com/ros-infrastructure/ros-apt-source/releases/download/1.2.0/ros2-testing-apt-source_1.2.0.resolute_all.deb \
```

`curl` here is missing `--fail` (`-f`) -- without it, an HTTP error response (404, 500, etc.) gets written to the temp `.deb` file and `curl` exits 0. The build will still eventually fail at the `sha256sum` step, but with a confusing hash-mismatch message rather than a clear download error at the right place. Adding `-f` makes the failure surface immediately at the download step. 👀

---

```diff
+    && echo "da9261ca7c06244da1528e0ede44018f7bb2e24a8a077eb0202f70706b578546 /tmp/ros2-testing-apt-source.deb" | sha256sum --strict --check \
```

Minor note -- not a blocker: the conventional format for verifying downloaded binaries is the `*`-prefixed binary mode rather than the space-only text-mode form. The preferred form is:

```sh
echo "da9261ca7c06244da1528e0ede44018f7bb2e24a8a077eb0202f70706b578546 */tmp/ros2-testing-apt-source.deb" | sha256sum --strict --check
```

The `*` form explicitly marks this as a binary-mode check, which is the right choice for a downloaded `.deb` -- text mode exists for files where line-ending normalization matters, which isn't relevant here.

PR #21419, opened 2026-05-08
Add ParadeDB image

isaacvando:
Adds [ParadeDB](https://github.com/paradedb/paradedb) as an official image. Thank you for reviewing!

Docs PR: https://github.com/docker-library/docs/pull/2671

### DIFF
--- a/library/paradedb
+++ b/library/paradedb
@@ -1,0 +1,23 @@
+Maintainers: Isaac Van Doren <isaac@paradedb.com> (@isaacvando),
+             Philippe Noël <phil@paradedb.com> (@philippemnoel),
+             Ming Ying <ming@paradedb.com> (@rebasedming)
+Architectures: amd64, arm64v8
+GitRepo: https://github.com/paradedb/paradedb.git
+GitFetch: refs/heads/main
+GitCommit: d5b3c63b4d325e1aa4b75cee2c4770bf5c18d2ba
+
+Tags: latest, 0.23.4, 0.23.4-pg18, v0.23.4, v0.23.4-pg18, 18-v0.23.4, pg18
+Directory: docker
+File: Dockerfile.official-18
+
+Tags: 0.23.4-pg17, v0.23.4-pg17, 17-v0.23.4, pg17
+Directory: docker
+File: Dockerfile.official-17
+
+Tags: 0.23.4-pg16, v0.23.4-pg16, 16-v0.23.4, pg16
+Directory: docker
+File: Dockerfile.official-16
+
+Tags: 0.23.4-pg15, v0.23.4-pg15, 15-v0.23.4, pg15
+Directory: docker
+File: Dockerfile.official-15
--- a/library/paradedb
+++ b/library/paradedb
@@ -4,20 +4,8 @@ Maintainers: Isaac Van Doren <isaac@paradedb.com> (@isaacvando),
 Architectures: amd64, arm64v8
 GitRepo: https://github.com/paradedb/paradedb.git
 GitFetch: refs/heads/main
-GitCommit: d5b3c63b4d325e1aa4b75cee2c4770bf5c18d2ba
+GitCommit: a5a8228fa1ce7e40d5c3f32d9bf9995e8a278e99
 
-Tags: latest, 0.23.4, 0.23.4-pg18, v0.23.4, v0.23.4-pg18, 18-v0.23.4, pg18
+Tags: latest, 0.25.3, 0.25.3-pg18, v0.25.3, v0.25.3-pg18, 18-v0.25.3, pg18
 Directory: docker
 File: Dockerfile.official-18
-
-Tags: 0.23.4-pg17, v0.23.4-pg17, 17-v0.23.4, pg17
-Directory: docker
-File: Dockerfile.official-17
-
-Tags: 0.23.4-pg16, v0.23.4-pg16, 16-v0.23.4, pg16
-Directory: docker
-File: Dockerfile.official-16
-
-Tags: 0.23.4-pg15, v0.23.4-pg15, 15-v0.23.4, pg15
-Directory: docker
-File: Dockerfile.official-15


isaacvando:
Hi @tianon @yosifkit, could one of you take a look at this PR? Thanks for your help!

tianon:
Please, don't merge into this PR -- we'll review the PR when we can.  If you wish to update it, please rebase on our branch, then force push (which should also remove all the errant merge commits).  Generally, this change adds a brand new file, so there shouldn't be any conflicts and thus very little reason to update/rebase unless you have something about the PR itself you need to update (or there's some new test, etc we merged into our branch that you'd like the CI to test, but that's pretty rare too).

Beyond that, please, please be patient -- our `new-image` queue takes a lot more maintainer energy to review / process, so it tends to be a lower priority but you _are_ in queue and we'll get to it eventually. :heart:

philippemnoel:
> Please, don't merge into this PR -- we'll review the PR when we can. If you wish to update it, please rebase on our branch, then force push (which should also remove all the errant merge commits). Generally, this change adds a brand new file, so there shouldn't be any conflicts and thus very little reason to update/rebase unless you have something about the PR itself you need to update (or there's some new test, etc we merged into our branch that you'd like the CI to test, but that's pretty rare too).
> 
> Beyond that, please, please be patient -- our `new-image` queue takes a lot more maintainer energy to review / process, so it tends to be a lower priority but you _are_ in queue and we'll get to it eventually. ❤️

Hi! Thank you for the kind message, and really sorry about the noise. If you want, we can open a new PR that doesn't have all these commits. I went ahead and force-pushed as you suggested, and our `main` is now exactly one commit ahead of yours. We won't be touching it until you're ready to review it.

Thanks again for taking a look and we'll sit tight. Thank you 🙏

philippemnoel:
Hi @tianon @yosifkit. Just wanted to follow up on this PR. We'd still appreciate your review, and have tried our best to implement all of the best practices your team recommends to save you review time.

tianon:
```diff
+Tags: 0.23.4-pg15, v0.23.4-pg15, 15-v0.23.4, pg15
+Directory: docker
+File: Dockerfile.official-15
+
+Tags: 0.23.4-pg16, v0.23.4-pg16, 16-v0.23.4, pg16
```

Four PostgreSQL major versions at once is a lot for a new image submission -- for reviewability it would be easier to start with just the most recent (pg18) and add older versions once that's in good shape. It's less disruptive (for you and for us) to add them later than to revisit four Dockerfiles for every concern that comes up (however, I do think it's manageable at this scale). 😅

---

```diff
+LABEL maintainer="ParadeDB - https://paradedb.com" \
```

The bare `maintainer` key is a hard blocker -- label keys need to be namespaced per the OCI image spec (https://github.com/opencontainers/image-spec/blob/v1.0.1/annotations.md). The predefined OCI key for maintainer contact is `org.opencontainers.image.authors`.

That said, we don't actively recommend `LABEL` at all -- labels are inherited by any downstream image built `FROM` this one, which is almost never desired. See https://github.com/docker-library/official-images/issues/3540 (especially https://github.com/docker-library/official-images/issues/3540#issuecomment-2715158488 for what the build system already sets on its own) for context. The `org.opencontainers.image.description` and `org.opencontainers.image.source` keys are fine by name -- `maintainer` either needs to become `org.opencontainers.image.authors` or `LABEL` gets dropped entirely.

(Minor aside: "Simple, Elastic-quality search for Postgres" reads like a tagline -- a more neutral factual description would be more appropriate there, though also not a blocker.)

---

```diff
+SHELL ["/bin/bash", "-o", "pipefail", "-c", "-e"]
```

`SHELL` is one of those instructions that's inherited by child images -- anyone who does `FROM paradedb:pg18` will find their `RUN` blocks unexpectedly executing under bash with `-e` and `pipefail` enforced, without having set that themselves. That's a real footgun that's hard to diagnose when it breaks something.

Worth noting too: looking at the actual `RUN` blocks, there are no bash-specific features being used -- `case`, `$()`, and `&&` are all plain POSIX sh. So there's no particular reason to be on bash here at all. The right approach is to drop `SHELL` entirely and put `set -eux` at the top of each `RUN` block (and then use `;` at the end of each line instead of `&&` so the failure behavior works consistently and correctly); the scope of the setting is then explicit and self-contained rather than silently inherited.

---

```diff
+FROM postgres:15-trixie AS paradedb
```

The `AS paradedb` is dead code -- this is a single-stage build with no `COPY --from=` anywhere. Can just be dropped.

---

```diff
+    echo "${checksum}  /tmp/pg_search.deb" | sha256sum -c - && \
```

Two minor hardening notes (neither a blocker): the `*`-prefixed binary mode is the conventional format for verifying downloaded binaries, and since this is Debian-based, GNU coreutils `sha256sum` supports `--strict` which exits non-zero on malformed input rather than silently skipping:

```sh
echo "${checksum} */tmp/pg_search.deb" | sha256sum --strict --check -
```

---

```diff
+    && apt-get install -y --no-install-recommends ca-certificates \
```

`ca-certificates` was already installed in the first `RUN` block -- redundant here. 👀

---

```diff
+    postgresql-15-postgis-$POSTGIS_VERSION_MAJOR \
+    postgresql-15-postgis-$POSTGIS_VERSION_MAJOR-scripts \
```

The comment above this block says these come from the pgdg-archive specifically so "pinned extension versions remain available" -- but then PostGIS itself has no version pin, unlike pgvector (`0.8.1-2.pgdg13+1`), pg-cron (`1.6.7-2.pgdg13+1`), and pg-ivm (`1.13-1.pgdg13+1`) which are all pinned. Looks like an oversight.

---

```diff
+    && rm /etc/apt/sources.list.d/pgdg-archive.list \
```

This seems strange -- why remove the pgdg-archive sources file from the image? If the source was trusted enough to install from, removing it afterwards just makes life harder for anyone extending the image -- they'd need to re-add it before they can install anything else from the archive. The right cleanup is discarding the index cache, not the configuration.

(Related: since these Dockerfiles target trixie, `apt-get dist-clean` is available and handles the cache cleanup more carefully than `rm -rf /var/lib/apt/lists/*` -- it preserves important cryptographic state. Worth switching.)

---

```diff
+    update-ca-certificates
```

`apt-get install ca-certificates` already runs `update-ca-certificates` via the package postinst -- this explicit call at the end of the `RUN` block is redundant.

---

```diff
+WORKDIR /
```

The postgres base image doesn't set a custom `WORKDIR`, so the default is already `/`. None of the `RUN` blocks change the working directory, and the subsequent `COPY` uses an absolute destination path anyway. This is a no-op.

---

```diff
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends ca-certificates curl && \
```

None of the `RUN` blocks start with `set -eux`. The `SHELL` instruction was covering `-e` and `pipefail`, but as noted above, `SHELL` should be dropped entirely -- which means the recommended pattern throughout is:

```dockerfile
RUN set -eux; \
    apt-get update; \
    apt-get install -y --no-install-recommends ca-certificates curl; \
    ...
```

---

```diff
+for DB in template1 paradedb "$POSTGRES_DB"; do
+  echo "Loading ParadeDB extensions into $DB"
+  psql -d "$DB" <<-'EOSQL'
+    CREATE EXTENSION IF NOT EXISTS pg_search;
+    CREATE EXTENSION IF NOT EXISTS pg_ivm;
+    CREATE EXTENSION IF NOT EXISTS vector;
+    CREATE EXTENSION IF NOT EXISTS postgis;
+    CREATE EXTENSION IF NOT EXISTS postgis_topology;
+    CREATE EXTENSION IF NOT EXISTS fuzzystrmatch;
+    CREATE EXTENSION IF NOT EXISTS postgis_tiger_geocoder;
+    CREATE EXTENSION IF NOT EXISTS pg_stat_statements;
+EOSQL
+done
```

Loading all of this into `template1` means every database created in this container -- by the user, by an application, by anything -- gets pg_search, pgvector, pg_ivm, PostGIS, PostGIS topology, fuzzystrmatch, TIGER geocoder, and pg_stat_statements pre-created whether they wanted them or not. That's a very heavy and opinionated default, especially for something whose core value is a search extension.

With that said, I'm uncertain whether this image is a good fit for the Docker Official Images. `pg_search` is a Postgres extension -- there's no "ParadeDB" process to run, just Postgres with an extension pre-installed and some initialization done at first startup. The image is essentially `FROM postgres` plus a few apt packages, some config surgery, and an init script.

What gives me the most pause is the PostGIS bundle specifically -- `postgis`, `postgis_topology`, `postgis_tiger_geocoder` are a full GIS stack that doesn't have an obvious connection to a search extension. Is all of this actually required by or for pg_search, or is it more "everything ParadeDB happens to ship together"? Understanding that would help me understand whether the image is offering something focused and necessary vs. something that would be better served by documentation and a reference Dockerfile users can adapt.

Not trying to close the door here -- just being honest about what I'm thinking as I look at this. 🙇

isaacvando:
Thank you for the thoughtful review! I pushed a new version that addresses all of your feedback.

> Loading all of this into template1 means every database created in this container -- by the user, by an application, by anything -- gets pg_search, pgvector, pg_ivm, PostGIS, PostGIS topology, fuzzystrmatch, TIGER geocoder, and pg_stat_statements pre-created whether they wanted them or not. That's a very heavy and opinionated default, especially for something whose core value is a search extension.

I agree, the extra extensions don't make sense for an official image. I removed all of them except for ParadeDB (pg_search) and pgvector (which is required by ParadeDB). I still included the extensions in template1 as the vast majority of ParadeDB users want it enabled on every DB they create. This is a much lighter default and seems like the right choice for users of the image. 

Calling the extension `pg_search` was some suboptimal naming on our part; we are planning on changing that to `paradedb` in the future.

> With that said, I'm uncertain whether this image is a good fit for the Docker Official Images. pg_search is a Postgres extension -- there's no "ParadeDB" process to run, just Postgres with an extension pre-installed and some initialization done at first startup. The image is essentially FROM postgres plus a few apt packages, some config surgery, and an init script.

ParadeDB is a full search & analytics database with a dedicated storage layer, query planning, etc. The fact that it's implemented as a Postgres extension is largely an implementation detail. Users of databases other than Postgres often switch to ParadeDB for the combo of search and transactional features it offers; that is to say, they don't view it as a _only_ a Postgres extension, but really a distinct DB.

ParadeDB solves the same kinds of problems as other official images projects like Elasticsearch and Solr. There are also other official images that don't have their own dedicated process. For example, Kong does its own configuration and then [runs nginx](https://github.com/Kong/docker-kong/blob/d1ed165bfeb6304815bc537b3e8ff387fa0b350a/docker-entrypoint.sh#L67) as the entrypoint.

Given that users view ParadeDB primarily as "search and transactions in a single database" and that there's no other official image with that kind of value-prop, it seems like a good candidate for inclusion. Curious to hear your thoughts! 😄

PR #21427, opened 2026-05-09
fix: Revert: Update influxdb2 image to 2.9.0, remove 2.7.x (#21362)

devanbenz:
This reverts commit ca8da91c636fc9a06033324b792c4c43bee7635f.


yosifkit:
I understand that the `2.9.0` update has a few bugs but are you certain that you want to add back `2.7.x`?

Alternatively, this is just drop `2.9.0` and move the default tags to `2.8.x` since `2.8.x` is unchanged in the upstream [commit range](https://github.com/influxdata/influxdata-docker/compare/1d9bc3961bee14268e069bf16767a014d2f62c2a...74f0633592894e9245520b947e1b89ccfaa14ced) .

PR #21461, opened 2026-05-13
latest sec patches

alcohol:


alcohol:
The error (failing test) is not familiar to me, anything I should be concerned about?

yosifkit:
Test failure is just that the test was updated to exclusively test composer `2.x` (https://github.com/docker-library/official-images/commit/e069fdc1bbb16a2338dbfbe70a86a2b9a2260acc). Rather than changing the test, I think we can ignore the failure since the `1.10.28` version seems to be a one-off release for a security fix. It will only be kept in DOI long enough to be built and pushed to Docker Hub, right?

(I was going to say that we could have the tests skip that one for version 1, but we never implemented support for that 😅)

alcohol:
@yosifkit your assumption is correct, technically 1.10 is no longer maintained. But security patches are still applied.

yosifkit (approved):

PR #21483, opened 2026-05-15
Update mongo

yosifkit:
Changes:

- https://github.com/docker-library/mongo/commit/d944b46: Merge pull request https://github.com/docker-library/mongo/pull/757 from infosiftr/eol


yosifkit:
All built and deployed/published downstream so these can be removed again. 👍

PR #21498, opened 2026-05-19
Updates for Liberty release 26.0.0.5

wraschke:


leochr:
Not entirely sure why `override-cmd` test is failing. 

The only significant change in 26.0.0.5 is to auto-generate a random LTPA password if user hasn't already specified one (via PR https://github.com/OpenLiberty/ci.docker/pull/719/changes). A message is output using echo to indicate the auto-generation. I wonder whether the test is checking for a particular message and doesn't consider/like this message.

yosifkit:
`override-cmd` is our rough test to codify https://github.com/docker-library/official-images/tree/97d51ca5b57c48744f0c30bbdb4d186756bc267c#consistency. It is testing, "does `docker run -it --rm open-liberty:[tag] [some binary from the image]` work?".

In the specific case of the test, it runs echo from the image and compares to an expected output (so extra output from the entrypoint script will break it). While not outputting extra messages from the entrypoint script will fulfill the test, the goal of the test is to encourage creating images that don't run extra unrelated launch stuff if the user asked for `bash` or some other tool from the image that isn't the main purpose of the image (`/opt/ol/wlp/bin/server` and `/opt/ibm/wlp/bin/server`).

leochr:
@yosifkit Thank you for taking a look and providing additional context. 

The Liberty startup script is used as entrypoint as it handles various setup for our enterprise customers for their production workloads (i.e. keystore and truststore setup, loading cache, etc). Hence, we won't be able to change that as that'll break users. Users can override the entrypoint if they need to. 

The override commands from the test are still run. The test must check whether the message from the override command is present in the output instead of looking to exactly match the message. Could the test please be updated?

tianon:
If you echo to stderr `echo >&2 ...` instead of stdout it should pass the test.

Alternatively (and our preference, as noted by yosifkit above), the entrypoint could check whether the command is a "Liberty" command before performing extra operations like generating a password.

wraschke:
@yosifkit @tianon the checks are passing now. Can this PR be merged?

yosifkit (approved):

PR #21524, opened 2026-05-23
Apache CouchDB 3.5.2

nickva:
Due to an issue with Java 21 installing and running on s390x architecturem we're not releasing an s390x arch flavor for this release. It's just x86_64 and arm. In the future we may return support for it.

Also, replace Joan with Ronny Berndt (our PMC) in maintainers list. Joan has not been active for about 5 years and Ronny made this Docker release and regularly participates in release creation and testing.

### DIFF
--- a/library/couchdb
+++ b/library/couchdb
@@ -1,18 +1,18 @@
 # see https://couchdb.apache.org/
 # also the announce@couchdb.apache.org mailing list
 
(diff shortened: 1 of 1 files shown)


janl:
+1

yosifkit (approved):

PR #21535, opened 2026-05-26
Update golang

yosifkit:
Changes:

- https://github.com/docker-library/golang/commit/1725ebf: Update tip to commit 4b37aa2bc11486fff53217fb0584d3ce2099b548
- https://github.com/docker-library/golang/commit/c60076d: Update tip to tip-20260523, commit 24ec7d2120d3d16b79cb2a7009a3e8025e4a2ce4

### DIFF
--- a/library/golang
+++ b/library/golang
@@ -112,23 +112,23 @@ Directory: 1.25/windows/nanoserver-ltsc2022
 Builder: classic
 Constraints: nanoserver-ltsc2022, windowsservercore-ltsc2022
 
-Tags: tip-20260517-trixie, tip-trixie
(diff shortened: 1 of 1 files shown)


tianon:
oof, looking at the two commits, I _think_ golang/go@24ec7d2120d3d16b79cb2a7009a3e8025e4a2ce4 is more correct than golang/go@4b37aa2bc11486fff53217fb0584d3ce2099b548 :sob:

need to look into what that script is doing

tianon:
Ah nope, it's a weird merge that looks fishy as hell, but it's legit and correct and the latter commit is the newer one (on the master branch upstream).

PR #21636, opened 2026-06-10
update eclipse-temurin to use latest entrypoint

trecloux:
add support CA certs mounted in subdirectories


yosifkit:
CC maintainers: @gdams @sxa, Are you ready for this change?

tianon:
This is a great example of the exact flavor of feature creep I was worried about originally and why I pushed back on the feature in the beginning. :sweat_smile:

It's not a bad feature / tool, but it's especially and particularly wild for it to be part of the `eclipse-temurin` images directly.

gdams:
@tianon would it be possible to get this merged please? I agree that the scope creep is concerning but we need to address the security fixes for the ecosystem.

tianon:
Maybe revert the script changes upstream so the security fixes can get in unencumbered? :sweat_smile:

(this PR is just the script changes, not the security fixes)

PR #21664, opened 2026-06-12
Update photon image

gerrit-photon:
Update photon image

### DIFF
--- a/library/photon
+++ b/library/photon
@@ -5,12 +5,12 @@ GitRepo: https://github.com/vmware/photon-docker-image.git
 Directory: docker
 
 
-Tags: 5.0, 5.0-20260610, latest
+Tags: 5.0, latest
 Architectures: amd64, arm64v8
 GitFetch: refs/heads/x86_64/5.0-20260610
 GitCommit: e521d6719ba85414811c93e7fbbb8d1ce5425ef7
-arm64v8-GitFetch: refs/heads/aarch64/5.0-20260610
-arm64v8-GitCommit: a0f7a2be9f8056b140f447d060690fb907009088
+arm64v8-GitFetch: refs/heads/aarch64/5.0-20260214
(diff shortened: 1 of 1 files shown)


michellew-vmware:
@yosifkit @tianon , 

could you please help to merge this to revert the bad aarch64 photon docker image 20260610?

There is some build break in aarch64 and we will fix it and publish correct docker images later this week or early next week.

Thanks,
Michelle

library/photon
```diff
@@ -5,12 +5,12 @@ GitRepo: https://github.com/vmware/photon-docker-image.git
 Directory: docker
 
 
-Tags: 5.0, 5.0-20260610, latest
+Tags: 5.0, latest
```
yosifkit, line 8: Did you want to also fix the `5.0-20260610` tag to keep just the amd64 image? Adding another section for just that tag and architecture would work:

```suggestion
Tags: 5.0-20260610
Architectures: amd64
GitFetch: refs/heads/x86_64/5.0-20260610
GitCommit: e521d6719ba85414811c93e7fbbb8d1ce5425ef7

Tags: 5.0, latest
```

michellew-vmware: amd64 docker image is good...

yosifkit: Yeah, this is the same fetch and commit from below for amd64. It is adding it as a separate tag (`5.0-20260610`) with just amd64. This will fix the tag on [Docker Hub](https://hub.docker.com/_/photon/tags?name=5.0-20260610) so this no longer has the broken `arm64v8` image in it.

michellew-vmware: thank you for your suggestion..  tag(5.0-20260610) is not very important for us for now.

And we will publish fixed aarch64 image soon and recover the 5.0-20260610 for both amd64 and aarch64.

Thanks,
Michelle

yosifkit (approved):

library/photon
michellew-vmware: Thanks @yosifkit

PR #21672, opened 2026-06-15
Update Ubuntu 2026-06-15

ROCKsBot:
Update Ubuntu to the latest published OCI tarballs as of 2026-06-15.

PR #21698, opened 2026-06-20
Aerospike 8.1.2.3

pvinh-spike:

### DIFF
--- a/library/aerospike
+++ b/library/aerospike
@@ -4,14 +4,14 @@ Maintainers: Lucien Volmar <lucien@aerospike.com> (@volmarl),
              Kevin Porter <kporter@aerospike.com> (@kportertx),
              Thomas Huelbert <thuelbert@aerospike.com> (@thomasaerospike)
 
-Tags: ee-8.1.2.2, ee-8.1.2.2_1
+Tags: ee-8.1.2.3, ee-8.1.2.3_1
 Architectures: amd64, arm64v8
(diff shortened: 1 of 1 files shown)


yosifkit:
```diff
+  apt-get install -y --no-install-recommends \
+    libgnutls30t64=3.8.3-1.1ubuntu3.6 \
+    libssl3t64=3.0.13-0ubuntu3.11 \
+    openssl=3.0.13-0ubuntu3.11 \
+  ; \
```

We recommend against package upgrades within official-images. When package upgrades are applied in a dependent image, it duplicates content of the base image, making the image larger than necessary. It also only delays the inevitable "there are outdated packages". The Official Images build pipeline rebuilds on a base image update, so we make periodic base image updates to then fully rebuild all dependent images (e.g., the [Debian](https://github.com/docker-library/official-images/pulls?q=is%3Apr+label%3Alibrary%2Fdebian) and [Ubuntu](https://github.com/docker-library/official-images/pulls?q=is%3Apr+label%3Alibrary%2Fubuntu) images are updated a least every 30 days). That way all images `FROM` them can benefit from updated packages.

>

I've pinged the Ubuntu image maintainers to let them know that an update is desired for `22.04` and `24.04` (https://github.com/docker-library/official-images/pull/21672#issuecomment-4774261577).

pvinh-spike:
Thanks for the advise.

PR #21630, opened 2026-06-09
alpine: add 3.24.0 as latest, remove EOL 3.20

ncopa:

PR #21702, opened 2026-06-22
Chronograf v1.11.4

srebhan:
Thank you!

### DIFF
--- a/library/chronograf
+++ b/library/chronograf
@@ -1,7 +1,7 @@
 Maintainers: Brandon Pfeifer <bpfeifer@influxdata.com> (@bnpfeife),
              Sven Rebhan <srebhan@influxdata.com> (@srebhan)
 GitRepo: https://github.com/influxdata/influxdata-docker
-GitCommit: 6496967c862c32efb249d610db48658d1bf7221d
+GitCommit: 61dfafef8cb60c34525c13ffcf48124f3f2cf08f
 
 Tags: 1.8, 1.8.10
 Architectures: amd64, arm32v7, arm64v8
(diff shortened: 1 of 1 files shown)


yosifkit:
I see one check failing due to:
```

 - chronograf:1.8-alpine (FROM alpine:3.20) -- completely unsupported base!
 - chronograf:1.9-alpine (FROM alpine:3.20) -- completely unsupported base!
```

There was notification of the Alpine 3.20 end of life in https://github.com/docker-library/official-images/pull/21630#issuecomment-4663497876.

Please either update the 1.8 & 1.9 images to use `alpine:3.21` or newer; or drop them if they are no longer supported.

srebhan:
@yosifkit I need some advice. We currently keep the container versions in lockstep with the software version so users know what they get. If we now change the base image we will get two different containers (with different hashes) into the wild depending on when the user pulled the image. This is problematic, isn't it, as setups are no longer reproducible because you get a different container.

Do you see a way around it without removing old versions or give-up on the version lockstep?

yosifkit:
> we will get two different containers (with different hashes) into the wild depending on when the user pulled the image

That is already the case in official images since we "Actively rebuild for updates and security fixes". We rebuild images when their parent image is updated in order to provide system package updates. If users need reproducible images, then they should pull by image digest and not tags.

These are all currently the same image (https://oci.dag.dev/?image=chronograf%3A1.9-alpine):
- `chronograf:1.9-alpine`, `chronograf:1.9.4-alpine`
- `chronograf@sha256:0f80cd3823ff72714fb0be9429588b58a6cd474733e0369bc7f9181346b1d5cd`
- `chronograf:1.9-alpine@sha256:0f80cd3823ff72714fb0be9429588b58a6cd474733e0369bc7f9181346b1d5cd`
  - in this case the tag is ignored and is just for humans
- `chronograf@sha256:f621eed42ec67e2c0af874c13b22cf7c5eca4a2f92496e0e777e83e0d0e54a18`
  - this one skips the image index and is directly the image manifest for linux/amd64

>container versions in lockstep with the software version

This is still the case; the version of `chronograf` is locked to be the same as the image tags. It is just the software around it that is updated (like `ca-certificates` and base image packages). The git history of the files in https://github.com/docker-library/repo-info/tree/master/repos/chronograf (remote and local directories) can show the rebuilds that have happened for specific tags over time. We scrape that data from the registry API and the built images themselves.

---

If you want unchanging/historical tags, then they should be removed from the `library/` file. We don't delete tags from Docker Hub, so they will remain pullable.

tianon (approved):

srebhan:
Thanks a lot @tianon for merging and @yosifkit for your advice!

PR #21710, opened 2026-06-23
library/odoo Update title and intro in activity view documentation

hyperare80-pixel:


yosifkit:
Indistinguishable from spam.

PR #5149, opened 2018-12-06
Update for latest CouchDB (2.3.0)

wohali:
This is a security update, so we have dropped the 2.2.0 version.

Note that we are now installing from our `.deb` installer, which simplifies the Dockerfile, speeds image build, and is also saving us about 53MB of space.

We've got updated docs too, so I'll submit a PR for https://github.com/docker-library/docs/tree/master/couchdb in the morning.

PR #21714, opened 2026-06-23
Apache CouchDB 3.5.2.1

nickva:
This is packaging only update:

 * Erlang 26 update https://www.erlang.org/patches/OTP-26.2.5.21 which fixes a few CVEs (CVE-2026-42789, CVE-2026-42790)

 * Nouveau not starting fix systemd https://github.com/apache/couchdb/issues/6041

### DIFF
--- a/library/couchdb
+++ b/library/couchdb
@@ -4,15 +4,15 @@
 Maintainers: Ronny Berndt <ronny@apache.org> (@big-r81), Jan Lehnardt <jan@apache.org> (@janl), Nick Vatamaniuc (@nickva)
 GitRepo: https://github.com/apache/couchdb-docker
 GitFetch: refs/heads/main
-GitCommit: adcb4f73e8a3c6fb0e901ab88cab50d71f02937a
+GitCommit: a76020b130d5d2e2a0e54f9437d90d35652b8707
 
-Tags: latest, 3.5.2, 3.5, 3
+Tags: latest, 3.5.2.1, 3.5.2, 3.5, 3
 Architectures: amd64, arm64v8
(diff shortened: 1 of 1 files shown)


big-r81:
+1

yosifkit:
I didn't see it mentioned in previous pull requests, and I don't think is a blocker since it has been there for [almost 8 years](https://github.com/docker-library/official-images/pull/5149), but it seems odd to have `--allow-downgrades --allow-remove-essential --allow-change-held-packages` when installing the `couchdb` and `couchdb-nouveau` packages. Are there specific instances where these are necessary when creating the image? I don't think the `debian:trixie-slim` parent image would have any installed packages for which this is necessary.

nickva:
@yosifkit good spot, they are likely unnecessary. We've carried them since https://github.com/apache/couchdb-docker/pull/118 via copy-and-paste. Those were likely from a recommendation for unattended upgrades recipe `DEBIAN_FRONTEND=noninteractive apt-get install -y --force-yes couchdb` and `force-yes` being deprecated it was replaced by the three fine grain permissions. Also potentially at some point couchdb was bundled in some os base package and we needed to override that version.

nickva:
@yosifkit we updated the release and removed the extra deb flags, thanks again for spotting those!

yosifkit (approved):

yosifkit:
You may want to evaluate if `DEBIAN_FRONTEND=noninteractive` is needed as well. It has been a while since I've seen places where it is necessary.

tianon:
Indeed, it's long-since fixed in `debconf` itself.

PR #21759, opened 2026-06-30
Add new XWiki intermediate LTS

tmortagne:


yosifkit:
```console
Incorrectly formatted 'latest' tags detected:
  xwiki:latest-lts-mysql-tomcat xwiki:latest-lts-mysql xwiki:latest-lts xwiki:latest-lts-postgres-tomcat xwiki:latest-lts-postgres xwiki:latest-lts-mariadb-tomcat xwiki:latest-lts-mariadb
```
Unfortunately, tags containing `latest` are not allowed (https://github.com/docker-library/official-images/tree/61faa135b499620c908397a4c632d3152ac621b2#tags-and-aliases).

tmortagne:
Right, I forgot that about that, removing them.

library/xwiki
```diff
@@ -17,6 +17,22 @@ Architectures: amd64, arm64v8
 Directory: 18/mariadb-tomcat
 GitCommit: 1aa7d0452f0df468f77f3a70be2f89d2af3f41ef
 
+# Intermediate LTS
+Tags: 18.4, 18.4.2, 18.4-mysql-tomcat, 18.4.2-mysql-tomcat
+Architectures: amd64, arm64v8
+Directory: 18/mysql-tomcat
```
yosifkit:
```diff
diff --git a/xwiki_latest/Dockerfile b/xwiki_18.4.2-mysql-tomcat/Dockerfile
similarity index 100%
copy from xwiki_latest/Dockerfile
copy to xwiki_18.4.2-mysql-tomcat/Dockerfile
```
The generated diff is showing that all these new versions are the same as the `18.5` Dockerfiles. ~Is there a different directory or commit that the 18.4 images should be pointing at?~ Oh, I see there is an [18.4 directory](https://github.com/xwiki/xwiki-docker/tree/master/18.4). In that case, these all need an adjustment like this:

```suggestion
Directory: 18.4/mysql-tomcat
```

tmortagne: Seems I'm really not used to release new intermediate LTS. Should be OK now.

yosifkit (approved):

PR #21827, opened 2026-07-12
Added Mageia version 10 and updated other images

juanluisbaptiste:
Added Mageia version 10 and updated other images

### DIFF
--- a/library/mageia
+++ b/library/mageia
@@ -1,8 +1,16 @@
 Maintainers: Juan Luis Baptiste <juancho@mageia.org> (@juanluisbaptiste)
 GitRepo: https://github.com/juanluisbaptiste/docker-brew-mageia.git
 
-Tags: 9,latest
-GitCommit: 2dcaddf8f7d1cdab109b6c1e6259068acaeb72a9
+Tags: 10,latest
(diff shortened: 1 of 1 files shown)


library/mageia
```diff
@@ -1,16 +1,24 @@
...
 
-Tags: 9,latest
-GitCommit: 2dcaddf8f7d1cdab109b6c1e6259068acaeb72a9
+Tags: 10,latest
+GitCommit: 954689fd6baa6e905d443306879ebd8c7020ea26
+GitFetch: refs/heads/dist
+Architectures: amd64, arm32v7, arm64v8
+amd64-Directory: dist/9/x86_64
+arm32v7-Directory: dist/9/armv7hl
+arm64v8-Directory: dist/9/aarch64
```
tianon: I think you probably meant to put `10` in here? :eyes:


```suggestion
amd64-Directory: dist/10/x86_64
arm32v7-Directory: dist/10/armv7hl
arm64v8-Directory: dist/10/aarch64
```

juanluisbaptiste: Ahhhhhh I didn't saw that...

We have a saying here that says "All that remains of eagerness is weariness"

I have to have it more in mind 🤣 

Fixed.

yosifkit (approved):

PR #21845, opened 2026-07-14
Updates for the release of WebSphere and Open Liberty 26.0.0.7

wraschke:


yosifkit:
Not a blocker, but I would suggest checking on the file mode changes; it seems odd to be adding executability to many of these files (like `xml`).

```diff
old mode 100644
new mode 100755
```

yosifkit (approved):

PR #21853, opened 2026-07-16
New Docker image for Silverpeas 6.4.7

mmoqui:
Update silverpeas Docker library to take into account Silverpeas 6.4.7

### DIFF
--- a/library/silverpeas
+++ b/library/silverpeas
@@ -2,10 +2,10 @@
 Maintainers: Miguel Moquillon <miguel.moquillon@silverpeas.org> (@mmoqui)
 GitRepo: https://github.com/Silverpeas/docker-silverpeas-prod.git
 
-Tags: 6.4.7, latest
+Tags: 6.4.7, 6.4, latest
 GitCommit: 143ba37b837ac7ac3f455d7cc4d873ada394e770
 GitFetch: refs/heads/6.4.x
 
-Tags: 6.3.6-jammy
+Tags: 6.3.6, 6.3
 GitCommit: fb25885d1cd43172693b3acf79e9fac056a9db34
 GitFetch: refs/heads/6.3.x
--- a/library/silverpeas
+++ b/library/silverpeas
@@ -3,8 +3,8 @@ Maintainers: Miguel Moquillon <miguel.moquillon@silverpeas.org> (@mmoqui)
 GitRepo: https://github.com/Silverpeas/docker-silverpeas-prod.git
 
 Tags: 6.4.7, latest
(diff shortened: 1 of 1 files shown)


library/silverpeas
```diff
@@ -2,10 +2,10 @@
...
 GitRepo: https://github.com/Silverpeas/docker-silverpeas-prod.git
 
-Tags: 6.4.6, latest
-GitCommit: b916cf7a89c24cac188b39cd31e1128671e67653
+Tags: 6.4.7, latest
+GitCommit: 143ba37b837ac7ac3f455d7cc4d873ada394e770
 GitFetch: refs/heads/6.4.x
 
-Tags: 6.3.6
+Tags: 6.3.6-jammy
```
tianon: Can you elaborate on what you're looking to accomplish with this change? 😅

mmoqui: Docker image for Silverpeas 6.3.6 was based upon a deprecated version of Ubuntu. So, each time a fix version of Silverpeas 6.4 is released and a Docker image built for it, the PR here failed because the Docker image of the previous minor version of Silverpeas (6.3) didn't pass your expectations (because based upon a deprecated version of Ubuntu and, although already existing, the image is recreated in your process, so it passes again the different  checkpoints). Then, because in our policy tags shouldn't be overridden, we have created a new tag here for a Docker image of Silverpeas 6.3.6 based upon Ubuntu Jammy (and named as such so that we know at a glance the reason of such a tag).

tianon: That sounds like conflating Git tags (and their associated policy) with Docker image tags. :sweat_smile:

One of the [core purposes of this program is _actively rebuilding_ image tags](https://github.com/docker-library/official-images#what-are-official-images), which necessarily requires that they be re-built/re-pushed/overwritten over time.

To make this slightly more concrete, this change means that `silverpeas:6.3.6` (which has previously _already_ been updated to `jammy` prior to this PR) will become stagnant / never be updated again, while `silverpeas:6.3.6-jammy` will be freshly created and start receiving rebuilds/updates (and doesn't match a pattern users are expecting from `silverpeas`).

Honestly, we'd normally recommend even _more_ aliases on a build like this, like `Tags: 6.3.6, 6.3, 6` so that users can pin to the level of change they're willing to accept (if they want immutable builds, that's why every piece of container tooling and the underlying protocols are all designed to support by-digest pulling/pinning that uses a cryptographic digest).

mmoqui: Ok I will change that

yosifkit (approved):

PR #21855, opened 2026-07-16
lightstreamer: matrix refactor + -base variant

gfinocchiaro:
Refactor of the entire matrix. Highlights:

- All previously-published tag names remain valid — no `docker pull` will 404.
- Every previously-published tag continues to use Eclipse Temurin as its JDK.
- Directory layout changed to `<mm>/<runtime>/temurin-<os>/`, which lets us publish explicit-OS tag variants (`-noble`, `-jammy`) instead of relying on Temurin's floating default.
- All pre-existing tags now resolve to `temurin-noble` (previously they used Temurin's unpinned base). Same JDK, forward-compatible OS pin.
- New `-base` edition (`base`, `<ver>-<runtime>-temurin-<os>-base`, etc.) provides a minimal image with factory config stripped, intended for downstream to layer their own conf/adapters.
- Retains the `-temurin` (no-OS) alias family for backwards compatibility.
 - Adds Gianluca Finocchiaro (@gfinocchiaro) to the `Maintainers:` list.

### DIFF
--- a/library/lightstreamer
+++ b/library/lightstreamer
@@ -3,322 +3,162 @@ Maintainers: Lightstreamer Server Development Team <support@lightstreamer.com> (
              Gianluca Finocchiaro <gianluca.finocchiaro@lightstreamer.com> (@gfinocchiaro)
 GitRepo: https://github.com/Lightstreamer/Docker.git
 
-Tags: 6.0.3-jdk8-temurin-jammy-base, 6.0-jdk8-temurin-jammy-base
+Tags: 6.0.3-jdk8, 6.0-jdk8, 6.0.3-jdk8-temurin, 6.0-jdk8-temurin, 6.0.3, 6.0
 Architectures: amd64, arm64v8
-GitCommit: 156c401231708a23043ac9a06238569e10a3d406
-Directory: 6.0/jdk8/temurin-jammy-base
+GitCommit: ca38a7d50a3c1a6e50fdc2cfb53e480673f356f9
+Directory: 6.0/jdk8
 
-Tags: 6.0.3-jdk8-temurin-jammy, 6.0-jdk8-temurin-jammy
+Tags: 6.0.3-jdk8-base, 6.0-jdk8-base, 6.0.3-jdk8-temurin-base, 6.0-jdk8-temurin-base, 6.0.3-base, 6.0-base
 Architectures: amd64, arm64v8
-GitCommit: 156c401231708a23043ac9a06238569e10a3d406
-Directory: 6.0/jdk8/temurin-jammy
+GitCommit: ca38a7d50a3c1a6e50fdc2cfb53e480673f356f9
+Directory: 6.0/jdk8-base
 
-Tags: 6.0.3-jdk8-temurin-noble-base, 6.0-jdk8-temurin-noble-base, 6.0.3-jdk8-base, 6.0-jdk8-base, 6.0.3-jdk8-temurin-base, 6.0-jdk8-temurin-base, 6.0.3-base, 6.0-base
+Tags: 6.1.0-jdk8, 6.1-jdk8, 6.1.0-jdk8-temurin, 6.1-jdk8-temurin, 6-jdk8, 6-jdk8-temurin, 6.1.0, 6.1, 6
 Architectures: amd64, arm64v8
-GitCommit: 156c401231708a23043ac9a06238569e10a3d406
-Directory: 6.0/jdk8/temurin-noble-base
+GitCommit: ca38a7d50a3c1a6e50fdc2cfb53e480673f356f9
+Directory: 6.1/jdk8
 
-Tags: 6.0.3-jdk8-temurin-noble, 6.0-jdk8-temurin-noble, 6.0.3-jdk8, 6.0-jdk8, 6.0.3-jdk8-temurin, 6.0-jdk8-temurin, 6.0.3, 6.0
+Tags: 6.1.0-jdk8-base, 6.1-jdk8-base, 6.1.0-jdk8-temurin-base, 6.1-jdk8-temurin-base, 6-jdk8-base, 6-jdk8-temurin-base, 6.1.0-base, 6.1-base, 6-base
 Architectures: amd64, arm64v8
-GitCommit: 156c401231708a23043ac9a06238569e10a3d406
-Directory: 6.0/jdk8/temurin-noble
+GitCommit: ca38a7d50a3c1a6e50fdc2cfb53e480673f356f9
+Directory: 6.1/jdk8-base
 
-Tags: 6.1.0-jdk8-temurin-jammy-base, 6.1-jdk8-temurin-jammy-base, 6-jdk8-temurin-jammy-base
+Tags: 7.0.3-jdk8, 7.0-jdk8, 7.0.3-jdk8-temurin, 7.0-jdk8-temurin
 Architectures: amd64, arm64v8
-GitCommit: 156c401231708a23043ac9a06238569e10a3d406
-Directory: 6.1/jdk8/temurin-jammy-base
+GitCommit: ca38a7d50a3c1a6e50fdc2cfb53e480673f356f9
+Directory: 7.0/jdk8
 
-Tags: 6.1.0-jdk8-temurin-jammy, 6.1-jdk8-temurin-jammy, 6-jdk8-temurin-jammy
+Tags: 7.0.3-jdk8-base, 7.0-jdk8-base, 7.0.3-jdk8-temurin-base, 7.0-jdk8-temurin-base
 Architectures: amd64, arm64v8
-GitCommit: 156c401231708a23043ac9a06238569e10a3d406
-Directory: 6.1/jdk8/temurin-jammy
+GitCommit: ca38a7d50a3c1a6e50fdc2cfb53e480673f356f9
+Directory: 7.0/jdk8-base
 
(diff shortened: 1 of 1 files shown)


gfinocchiaro:
The `diff-pr` bot's auto-comment failed with `Body is too long (maximum is 65536 characters)` because this PR is a matrix refactor that expands the tag set from 83 to 338. Quick summary for reviewers, complementing the PR description above:

- **83 → 338 unique tags** (net +255); **16 → 64 image blocks** (+48).
- **Every one of the 83 currently-published tags is preserved.** Verified by regenerating the manifest and comparing the tag list against the current DOI-published `library/lightstreamer` — no tags dropped.
- **All previously-published tags continue to use Eclipse Temurin**; the only change to them is an explicit `-noble` OS pin on tags that previously used the unpinned Temurin base.
- **New content**: fully-qualified `-noble`/`-jammy` tag variants, plus a new `-base` edition (`base`, `<ver>-<runtime>-temurin-<os>-base`) intended for downstream customization.
- **`-temurin` (no-OS) alias family retained** for back-compat.
- Third maintainer added (@gfinocchiaro) alongside the existing two.

Happy to split into smaller PRs if maintainers prefer (e.g. refactor first, then `-base` edition separately). Local `bashbrew --library "$(pwd)" build lightstreamer` passes green on the current tip.

tianon:
Ooof, this has some major echos of https://github.com/docker-library/tomcat/pull/299 which you might find interesting (I honestly don't have a good answer to the general problem -- the Java ecosystem is in a very dark place)

gfinocchiaro:
Thanks for the pointer — genuinely useful context, and it confirms there isn't a clean general answer to the matrix-explosion problem. 

Given that, I've taken the pragmatic route on this PR: dropped the OS variants entirely and let upstream Temurin pick its own per-JDK default OS. The `-base` edition addition — the actual user-visible motivation for the whole change — is retained, and every currently-published tag is preserved unchanged.

Happy to revisit the OS-variant discussion in a later PR if there's ever a general fix.

yosifkit:
I'm not sure I understand the need for a multistage build here. It seems like it is only using `buildpack-deps:bookworm-curl` as a place to download and verify the lightstreamer tar via `curl` and `gpg`. The `eclipse-temurin` images already include both [`curl`](https://github.com/adoptium/containers/issues/255) and [`gnupg`](https://github.com/adoptium/containers/blob/165aa69451210b2ad7496bfc10d551a06ad1de85/docker_templates/ubuntu.Dockerfile.j2#L17). What is the gain from using an extra stage?

gfinocchiaro:
Actually, the mentioned tools are not included in the latest images:

```
docker run eclipse-temurin:25-jdk curl
/__cacert_entrypoint.sh: line 114: exec: curl: not found
```

```
docker run eclipse-temurin:25-jdk gpg
/__cacert_entrypoint.sh: line 114: exec: gpg: not found
```

gfinocchiaro:
Please let us know if you would prefer us to switch back to a single-stage build.

yosifkit:
> Please let us know if you would prefer us to switch back to a single-stage build.

Yeah, that would be appreciated; we try to avoid multi-stage builds in most cases. See also our FAQ entry: https://github.com/docker-library/faq/tree/7e5a82e9cb6e131b9a50d6532ab1b9cae95b0b30#multi-stage-builds

gfinocchiaro:
> > Please let us know if you would prefer us to switch back to a single-stage build.
> 
> Yeah, that would be appreciated; we try to avoid multi-stage builds in most cases. See also our FAQ entry: https://github.com/docker-library/faq/tree/7e5a82e9cb6e131b9a50d6532ab1b9cae95b0b30#multi-stage-builds

Perfect, just reverted.

tianon:
```diff
+LABEL maintainer="Lightstreamer Server Development Team <support@lightstreamer.com>"
```

Every `Dockerfile` in the diff has this. We don't actively recommend `LABEL` in official images -- the build system already sets OCI annotations on the final published image (see https://github.com/docker-library/official-images/issues/3540#issuecomment-2715158488 for what that looks like; you might find you don't need any label at all). If you do want to keep it, the key name is the blocker: a bare `maintainer` key is the same problem as the deprecated `MAINTAINER` instruction. Either drop the `LABEL` entirely -- the `Maintainers:` field in the library file already captures this -- or rename the key to the proper OCI-namespaced form:

```dockerfile
LABEL org.opencontainers.image.authors="Lightstreamer Server Development Team <support@lightstreamer.com>"
```

This applies to every `Dockerfile` in the diff, so ideally it's a one-line fix in the template.

---

```diff
+FROM eclipse-temurin:8-jdk
```

(same pattern for `11-jdk`, `17-jdk`, `21-jdk`, `25-jdk` throughout)

None of the `FROM` lines pin to a specific OS release -- `eclipse-temurin:8-jdk` currently resolves to `8-jdk-resolute` (Ubuntu 25.10), but that alias will silently roll forward when the default changes. That's the square wheel: the base OS update happens under you rather than being a deliberate choice. Tags like `eclipse-temurin:8-jdk-noble` or `eclipse-temurin:8-jdk-jammy` let you control when you move between Ubuntu releases.

These `FROM` tags are unchanged from the previous Dockerfiles, so I wouldn't hold this PR over it -- worth cleaning up eventually, but not a new concern introduced here.

---

```diff
+RUN set -ex; \
+        apt-get update && apt-get install -y --no-install-recommends curl gnupg \
+        && curl -fSL -o Lightstreamer.tar.gz ${LIGHTSTREAMER_URL_DOWNLOAD} \
```

`set -e` is inert in a plain `&&` chain like this -- `&&` already ensures each command must succeed before the next runs, so there's nothing extra for `set -e` to do here. They can complement each other in more complex expressions, but not in a straight chain. The idiomatic form that actually puts `set -e` to work is `;` as the separator:

```dockerfile
RUN set -eux; \
        apt-get update; \
        apt-get install -y --no-install-recommends curl gnupg; \
        curl -fSL -o Lightstreamer.tar.gz "${LIGHTSTREAMER_URL_DOWNLOAD}"; \
        ...
```

The `-u` is worth adding too -- the `ARG`-declared variables are always in scope, but a typo in a variable name (say, `${LIGHTSTREAMER_URL_DWONLOAD}`) would silently expand to empty and pass garbage to `curl` rather than failing immediately. Defense in depth.

gfinocchiaro:
> ```diff
> +LABEL maintainer="Lightstreamer Server Development Team <support@lightstreamer.com>"
> ```
> 
> Every `Dockerfile` in the diff has this. We don't actively recommend `LABEL` in official images -- the build system already sets OCI annotations on the final published image (see [#3540 (comment)](https://github.com/docker-library/official-images/issues/3540#issuecomment-2715158488) for what that looks like; you might find you don't need any label at all). If you do want to keep it, the key name is the blocker: a bare `maintainer` key is the same problem as the deprecated `MAINTAINER` instruction. Either drop the `LABEL` entirely -- the `Maintainers:` field in the library file already captures this -- or rename the key to the proper OCI-namespaced form:
> 
> ```dockerfile
> LABEL org.opencontainers.image.authors="Lightstreamer Server Development Team <support@lightstreamer.com>"
> ```
> 
> This applies to every `Dockerfile` in the diff, so ideally it's a one-line fix in the template.

Got it. I'm going to use the OCI-namedspace form.

> 
> ```diff
> +FROM eclipse-temurin:8-jdk
> ```
> 
> (same pattern for `11-jdk`, `17-jdk`, `21-jdk`, `25-jdk` throughout)
> 
> None of the `FROM` lines pin to a specific OS release -- `eclipse-temurin:8-jdk` currently resolves to `8-jdk-resolute` (Ubuntu 25.10), but that alias will silently roll forward when the default changes. That's the square wheel: the base OS update happens under you rather than being a deliberate choice. Tags like `eclipse-temurin:8-jdk-noble` or `eclipse-temurin:8-jdk-jammy` let you control when you move between Ubuntu releases.
> 
> These `FROM` tags are unchanged from the previous Dockerfiles, so I wouldn't hold this PR over it -- worth cleaning up eventually, but not a new concern introduced here.

Ok, this will be done while relasing the next major version (8.0).

> 
> ```diff
> +RUN set -ex; \
> +        apt-get update && apt-get install -y --no-install-recommends curl gnupg \
> +        && curl -fSL -o Lightstreamer.tar.gz ${LIGHTSTREAMER_URL_DOWNLOAD} \
> ```
> 
> `set -e` is inert in a plain `&&` chain like this -- `&&` already ensures each command must succeed before the next runs, so there's nothing extra for `set -e` to do here. They can complement each other in more complex expressions, but not in a straight chain. The idiomatic form that actually puts `set -e` to work is `;` as the separator:
> 
> ```dockerfile
> RUN set -eux; \
>         apt-get update; \
>         apt-get install -y --no-install-recommends curl gnupg; \
>         curl -fSL -o Lightstreamer.tar.gz "${LIGHTSTREAMER_URL_DOWNLOAD}"; \
>         ...
> ```
> 
> The `-u` is worth adding too -- the `ARG`-declared variables are always in scope, but a typo in a variable name (say, `${LIGHTSTREAMER_URL_DWONLOAD}`) would silently expand to empty and pass garbage to `curl` rather than failing immediately. Defense in depth.

Cool. I'm going to use this form. Thanks!

tianon (approved):

tianon:
Just to make sure, are all these older lightstreamer versions still actively supported?  ie, if there's a security issue, will they have a new release?

gfinocchiaro:
> Just to make sure, are all these older lightstreamer versions still actively supported? ie, if there's a security issue, will they have a new release?

Typically, only the latest version is actively supported and will receive updates in case of security issues. However, this may vary based on specific customer needs.

tianon:
Ah, so in that case, all those older versions should be removed :sweat_smile:

As a reminder / supplementary note, removing tags here will remove them from the "Supported" section on the Hub readme (and will prevent us from spending cycles rebuilding them on the official build servers), but the tags will still be available to users who want them.  (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)

gfinocchiaro:
Thanks @tianon for clarifying. In the next PR we will remove no longer supported tags.

PR #21856, opened 2026-07-16
Odoo: update 17.0-19.0 to release 20260716

d-fence:
Hello,

here are the latest updated of Odoo supported versions.
Also, npm was removed from the image in order to reclaim space.

Thks

### DIFF
--- a/library/odoo
+++ b/library/odoo
@@ -1,16 +1,16 @@
 Maintainers: Christophe Monniez <moc@odoo.com> (@d-fence)
 GitRepo: https://github.com/odoo/docker
-GitCommit: 6f3febcf4fc91b069d707b7b1bf93c77b019d0aa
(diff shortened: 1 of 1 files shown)


yosifkit:
`DEBIAN_FRONTEND=noninteractive` shouldn't be needed for this package install, and you may want to evaluate if it is needed on the other `apt-get install` anymore. It has been a while since I've seen package installs where it is necessary.

d-fence:
> `DEBIAN_FRONTEND=noninteractive` shouldn't be needed for this package install, and you may want to evaluate if it is needed on the other `apt-get install` anymore. It has been a while since I've seen package installs where it is necessary.

Thanks for the review, it should be fixed in odoo/docker@f7f6765354127c8ca31c373f51a840d9ec79149d

yosifkit (approved):

PR #21872, opened 2026-07-20
Kapacitor v1.8.6

srebhan:
Thank you!


tianon (approved):

library/kapacitor
```diff
@@ -10,9 +10,9 @@ Directory: kapacitor/1.7
 Tags: 1.7-alpine, 1.7.7-alpine, alpine
```
tianon, line 10: Should `alpine` here move down to 1.8 too like `latest` did?

srebhan: Yeah, will put up a follow-up... See #21882.

PR #21873, opened 2026-07-20
Chronograf remove unsupported ARM32 variant for v1.11

srebhan:
The variant `arm32v7` is no longer built for Chronograf since version v1.11.0, therefore remove the architecture for the said version.


tianon (approved):
Nice, thank you!

PR #21874, opened 2026-07-20
Update CrateDB to 6.4.0

BaurzhanSakhariev:


library/crate
tianon, line 24: Does this mean that all or some of these versions are EOL now? 👀

tianon (approved):

library/crate
BaurzhanSakhariev: Right, older versions were re-added due to releasing a critical fix to older versions, 
but fine to be removed now.

I removed them in https://github.com/docker-library/official-images/pull/21883, thanks for the reminder!

PR #21877, opened 2026-07-20
Update golang

yosifkit:
Changes:

- https://github.com/docker-library/golang/commit/02d710e: Update tip to tip-20260719, commit 9f236fbe3fa00dfe1534148d84f505ddd770d3cf
- https://github.com/docker-library/golang/commit/127caaa: Update tip to commit 03845e30f7b73d1703bd8c21017297f6eecb76d6
- https://github.com/docker-library/golang/commit/6b11c5a: Update tip to commit cbb6afc18aba358641e69d10c3094837423451a6

PR #21879, opened 2026-07-20
Update pull_request_target to appease actions/checkout

yosifkit:
> Error: Refusing to check out fork pull request code from a 'pull_request_target' workflow. This workflow runs with the base repository's GITHUB_TOKEN, secrets, default-branch cache scope, and runner access. Fetching and executing a fork's code in that trusted context commonly leads to "pwn request" vulnerabilities. To opt in, review the risks at https://gh.io/securely-using-pull_request_target and set 'allow-unsafe-pr-checkout: true' on the actions/checkout step.

We were already extremely careful to not execute code from the pull request checkout; this just reverses the logic to specifically only pull `library/` and non-script `.external-pins` changes from the pull request.

### DIFF
--- a/.github/workflows/munge-pr.yml
+++ b/.github/workflows/munge-pr.yml
@@ -5,7 +5,6 @@ on:
 
 permissions:
   contents: read
-  pull-requests: write
 
 defaults:
   run:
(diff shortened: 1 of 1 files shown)


.github/workflows/munge-pr.yml
```diff
@@ -34,27 +33,27 @@ jobs:
...
         name: Affected Images
+        env:
+          PR_COMMIT: ${{ github.event.pull_request.head.sha }}
         run: |
           (set +x; echo "::stop-commands::$(echo -n ${{ github.token }} | sha256sum | head -c 64)")
-          git fetch --quiet https://github.com/docker-library/official-images.git master
-          externalPins="$(git diff --no-renames --name-only FETCH_HEAD...HEAD -- '.external-pins/*/**')"
+          # fetch the PR changes, but just library and pin updates, no scripts
+          git fetch --quiet origin "${PR_COMMIT}:"
+          git checkout FETCH_HEAD -- library/ '.external-pins/*/**'
```
tianon: We should be able to drop the `git checkout` line because the lines below use `git diff` to extract the data we need directly from the commits themselves (so the local state doesn't matter).

.github/workflows/munge-pr.yml
tianon, line 135: This checkout has the exact same problem 😅

tianon (dismissed):

.github/workflows/munge-pr.yml
yosifkit: I don't think `actions/checkout` will block it; `base.ref` is the target branch, not the pull request branch. Do we want to move this to the pull request branch and not the target branch?

tianon: Ah yeah, good catch; I guess this is fine for now, but it *would* be nice if we could test `diff-pr` changes easier given how hardened this is.  Let's punt for now.

tianon:
Nice, tested successfully in https://github.com/docker-library/official-images/pull/21877

PR #21887, opened 2026-07-21
influxdb: 3.9.11 + 3.10.5 (core + enterprise)

cannonpalms:
Updates the influxdb3 core and enterprise images to the latest patch releases:

- `3.9-core`: 3.9.6 -> **3.9.11**
- `3.9-enterprise`: 3.9.9 -> **3.9.11**
- `3.10-core` (`core`): 3.10.3 -> **3.10.5**
- `3.10-enterprise` (`enterprise`): 3.10.3 -> **3.10.5**

GitCommit `4935f4eb6b374869721872ecd7ef3e01bd228cf4` (influxdata/influxdata-docker master). Release tarballs and detached signatures for all four are published on dl.influxdata.com and verified.

### DIFF
--- a/library/influxdb
+++ b/library/influxdb
@@ -62,18 +62,22 @@ Tags: 2-alpine, 2.9-alpine, 2.9.1-alpine, alpine
 Architectures: amd64, arm64v8
 Directory: influxdb/2.9/alpine
 
-Tags: 3.9-core, 3.9.6-core
+Tags: 3.9-core, 3.9.11-core
 Architectures: amd64, arm64v8
+GitCommit: 4935f4eb6b374869721872ecd7ef3e01bd228cf4
 Directory: influxdb/3.9-core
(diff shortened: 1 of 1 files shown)


yosifkit (approved):

library/influxdb
tianon, line 7: It's not a big deal, but any particular reason you introduced a bunch of entry-specific `GitCommit` values instead of updating the global one?

praveen-influx: https://github.com/docker-library/official-images/pull/21953 - we've a new version `3.11.0` release and as part of that I've updated the patch versions (that were originally updated in this PR) and moved to using global `GitCommit`.

praveen-influx:
Could this be closed in favor of this PR https://github.com/docker-library/official-images/pull/21953 please? Thanks.

PR #21907, opened 2026-07-23
Update Apache Storm version tags and directories

reiabreu:
@rzo1

### DIFF
--- a/library/storm
+++ b/library/storm
@@ -14,7 +14,7 @@ Directory: 2.8.9-jre21
 
 Tags: 3.0.0-jre25, 3.0-jre25, 3.0.0, 3.0,latest
 GitCommit: 1e325fe56337922f3c06a5f790a5c4f7170acb4f
-Directory: 2.8.9
+Directory: 3.0.0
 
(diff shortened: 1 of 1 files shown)


library/storm
```diff
@@ -4,10 +4,18 @@ Maintainers: The Apache Storm Project <dev@storm.apache.org> (@asfbot),
...
-Tags: 2.8.8-jre21, 2.8-jre21
-GitCommit: 177a1534bf910c2271845f4eaedef7c040559fbc
-Directory: 2.8.8-jre21
+Tags: 2.8.9-jre21, 2.8-jre21
+GitCommit: d6fd2bc163a48be40cac2b5a1142c5d317d2d438
+Directory: 2.8.9-jre21
+
+Tags: 3.0.0-jre25, 3.0-jre25, 3.0.0, 3.0,latest
+GitCommit: 1e325fe56337922f3c06a5f790a5c4f7170acb4f
+Directory: 2.8.9
```
yosifkit: This should reference the correct directory:
```suggestion
Directory: 3.0.0
```

reiabreu: Thank you for catching it. I've pushed a fix.

yosifkit (approved):

PR #21911, opened 2026-07-23
Nim: Add versions for 2.0.x and a 2.2.x branches. Add i386 and arm32v7 images

moigagoo:


yosifkit:
- Since these are all just `FROM` Debian Trixie, they can stay as `apt-get dist-clean`. Nothing was uninstalled, so `autoremove` won't do anything. `apt-get clean` is done [automatically](https://github.com/debuerreotype/debuerreotype/blob/7de510f470a079b301d7f88abbb757283031d6da/scripts/debuerreotype-minimizing-config#L87-L109) in `debian` images.
  ```diff
  -  apt-get dist-clean
  +  apt-get autoremove -y; \
  +  apt-get clean; \
  +  rm -rf /var/lib/apt/lists/*
  ```

- :confused: isn't this covered by the catch-all case (`*)`) that also does `exit 1`?
  ```diff
  +  if [ "$url" = "none" ]; then \
  +    echo >&2 "error: no upstream binary for ${dpkgArch} in this Nim version"; exit 1; \
  +  fi; \
  ```

- It seems that architecture support is more important than a node runtime, so the docs need to be updated with a recommend way to do the `nim js -r main.nim` example. I assume the [`nodejs` apt package](https://packages.debian.org/trixie/nodejs) in Trixie is just too old. Perhaps the example should recommend that users create a multistage build that copies their `nim` result into a `node` image for running? :shrug:

moigagoo:
> Since these are all just FROM Debian Trixie, they can stay as apt-get dist-clean. Nothing was uninstalled, so autoremove won't do anything. apt-get clean is done [automatically](https://github.com/debuerreotype/debuerreotype/blob/7de510f470a079b301d7f88abbb757283031d6da/scripts/debuerreotype-minimizing-config#L87-L109) in debian images.

Reverted to `apt-get dist-clean`.

> 😕 isn't this covered by the catch-all case (*)) that also does exit 1?

These are two separate checks: one checks the arch, the other checks that a pre-built binary is available for a given version-arch combination.

> It seems that architecture support is more important than a node runtime, so the docs need to be updated with a recommend way to do the nim js -r main.nim example. I assume the [nodejs apt package](https://packages.debian.org/trixie/nodejs) in Trixie is just too old. Perhaps the example should recommend that users create a multistage build that copies their nim result into a node image for running? 🤷

Sent a PR to docs: https://github.com/docker-library/docs/pull/2682

moigagoo:
One check doesn't pass but that seems to a network issue unrelated to my changes.

moigagoo:
@yosifkit the tests all pass now, humbly requesting a review.

tianon:
I think what he was getting at with the `none` check is that it's wild to template a file with a `none` value there -- does your templating solution not allow you to remove that line completely instead?  Then the `case` statement would catch that issue directly, and the `if` that's added to all versions could go away completely because it's checking roughly the same thing.

moigagoo:
@tianon thanks, now I get it.

I've switched to a proper templating engine and now the Dockerfiles don't have that `none` bit: https://github.com/nim-lang/docker-images/blob/develop/dockerfiles/2.2.10/Dockerfile

library/nim
```diff
@@ -1,8 +1,60 @@
...
-Architectures: amd64, arm64v8
+Architectures: amd64, arm64v8, i386, arm32v7
 Directory: dockerfiles/2.2.10
 
+Tags: 2.2.8
+Architectures: amd64, arm64v8, i386, arm32v7
+Directory: dockerfiles/2.2.8
+
+Tags: 2.2.6
+Architectures: amd64, arm64v8, i386, arm32v7
```
yosifkit:
```suggestion
Architectures: amd64, i386, arm32v7
```

Since this is the one that can't build on `arm64v8`, the architecture should be dropped from it.

tianon (approved):
As discussed -- a one-time backfill of historical versions is totally fine. Just keeping in mind that "supported" means these would get new releases if a security issue were found, so we'll be looking forward to a follow-up to trim the list down to just the latest version of each actively supported series (2.2, etc) once everything's published. :+1:

(Removing tags here will remove them from the "Supported" section on the Hub readme and will prevent us from spending cycles rebuilding them on the official build servers, but the tags will still be available to users who want them.  See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)

PR #21934, opened 2026-07-28
Update photon image

gerrit-photon:
Update photon image


michellew-vmware:
@tianon and @yosifkit , could you please review this? Our customer is kind of waiting for it.

Thank you for your time and support,
Michelle

yosifkit (approved):

PR #21937, opened 2026-07-28
Update xwiki to 18.6.0

manuelleduc:


yosifkit:
For the 18 images: we don't yet use buildkit's automatic `TARGETARCH` in Docker Official Images. We recommend using `apk --print-arch` or `dpkg --print-architecture` as the most reliable options to retrieve the user space architecture of the image (example: [golang](https://github.com/docker-library/golang/blob/294097713ad2eaf9ee4c5e9dc57e040ac4b402d6/1.26/trixie/Dockerfile#L15-L18)).

For the `16` and `17-mariadb-tomcat` build failures, I am unsure of the cause:
```console
Step 12/25 : RUN rm -rf /usr/local/tomcat/webapps/* &&   mkdir -p /usr/local/tomcat/temp &&   mkdir -p /usr/local/xwiki/data &&   curl -fSL "${XWIKI_URL_PREFIX}/xwiki-platform-distribution-war-${XWIKI_VERSION}.war" -o xwiki.war &&   echo "$XWIKI_DOWNLOAD_SHA256 xwiki.war" | sha256sum -c - &&   unzip -d /usr/local/tomcat/webapps/ROOT xwiki.war &&   rm -f xwiki.war
 ---> Running in cc847da88c76
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0  313M    0  159k    0     0   181k      0  0:29:28 --:--:--  0:29:28  181k
  0  313M    0 1085k    0     0   856k      0  0:06:14  0:00:01  0:06:13  857k
curl: (92) HTTP/2 stream 1 was not closed cleanly: CANCEL (err 8)
```

vmassol:
@yosifkit My bad, I caused that regression.

I've opened https://jira.xwiki.org/browse/XDOCKER-421 on our side, with the PR at https://github.com/xwiki/xwiki-docker/pull/97 that I'm applying.

Regarding the other build failures, it's because we've had infra issues with maven.xwiki.org . It's now fixed and should work fine.

@manuelleduc will update this PR soon with new sha1.

Thx and sorry for that.

tianon:
A few minor suggestions for further improvement (I don't think any of this is explicitly blocking):

---

```diff
+  ln -fs $(ls -d /opt/libreoffice*) /opt/libreoffice && \
```

The `$(ls -d /opt/libreoffice*)` is unquoted, so word-splitting will break the `ln` invocation if the glob matches more than one path. There's also a cleaner option here that avoids the filesystem glob entirely -- LibreOffice installs into `/opt/libreoffice<major>.<minor>/` (e.g. `25.8.7` → `/opt/libreoffice25.8`), so you can derive the path directly from the version variable and verify it's real in the same step:

```sh
ln -fs "/opt/libreoffice${LIBREOFFICE_VERSION%.*}" /opt/libreoffice; \
test -d /opt/libreoffice/program; \
```

`${LIBREOFFICE_VERSION%.*}` strips the patch component (`25.8.7` → `25.8`), no filesystem inspection needed. The `test -d` resolves through the symlink and fails the build if `dpkg -i` somehow didn't produce the expected directory -- which is the right outcome given how load-bearing this symlink is for JODConverter's auto-discovery.

---

```diff
+RUN apt-get update && \
+  apt-get --no-install-recommends -y install \
```

and

```diff
+RUN LO_ARCH="$(dpkg --print-architecture)" && \
+  case "$LO_ARCH" in \
```

Both `RUN` blocks use `&&` chaining -- this should be `set -eux; \` with `;` separators. It's cleaner, eliminates the `&& \` noise, and makes the error-propagation semantics explicit rather than relying on exit-code chaining through each step.

---

```diff
+  echo "$LO_SHA256 /tmp/libreoffice.tar.gz" | sha256sum -c - && \
```

Minor: the `*`-prefixed binary mode is the conventional format for verifying downloaded binaries, and since this is Debian-based, `sha256sum` is from coreutils and `--strict` is available -- it causes `sha256sum` to reject malformed input lines rather than silently skipping them:

```sh
echo "$LO_SHA256 */tmp/libreoffice.tar.gz" | sha256sum --strict -c -; \
```

vmassol:
@tianon thx a lot for your review. Applied in https://github.com/xwiki/xwiki-docker/pull/100 

Do you want the sha1 for this PR updated or is it ok that it's applied only for future PRs?

Thx

tianon:
I think it's fine for that to wait for a future PR :+1:

tianon (approved):

PR #21946, opened 2026-07-29
Redis: Release 8.10.0

Peter-Sh:
Hi Docker Team!

Could you please merge this new Redis 8.10.0 release.

Changes: https://github.com/redis/redis/releases/tag/8.10.0

### DIFF
--- a/library/redis
+++ b/library/redis
@@ -56,16 +56,16 @@ GitCommit: 3769a841048d940966635f8f59471f3d350707bf
 GitFetch: refs/tags/v8.2.8
 Directory: alpine
 
-Tags: 8.10-rc2, 8.10-rc2-trixie
+Tags: 8.10.0, 8.10, 8.10.0-trixie, 8.10-trixie
 Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, riscv64, ppc64le, s390x
-GitCommit: 11c8b131eb4c9a36110d70ceffedaa7397fb6cbf
-GitFetch: refs/tags/v8.10-rc2
+GitCommit: 0a24b09f88a6cb8feaf130d4eca7bd6b9b4c73a3
+GitFetch: refs/tags/v8.10.0
 Directory: debian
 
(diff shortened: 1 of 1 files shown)


library/redis
```diff
@@ -56,16 +56,16 @@ GitCommit: 3769a841048d940966635f8f59471f3d350707bf
 GitFetch: refs/tags/v8.2.8
 Directory: alpine
```
tianon, line 69: It looks like security support for 8.2 might have ended? :eyes:

library/redis
```diff
@@ -56,16 +56,16 @@ GitCommit: 3769a841048d940966635f8f59471f3d350707bf
...
 
-Tags: 8.10-rc2, 8.10-rc2-trixie
+Tags: 8.10.0, 8.10, 8.10.0-trixie, 8.10-trixie
```
tianon: Should `latest` and similar aliases move here now, or did you want to wait for a bit before moving those?

tianon (dismissed):

library/redis
Peter-Sh: OMG, I hit this version comparison bug again :face_holding_back_tears:  Thanks for catching it, @tianon!
I updated to make 8.10 latest

library/redis
Peter-Sh: @tianon
Maybe I'm missing something, but 8.2.8 is the latest in 8.2 series and has recent security fixes https://github.com/redis/redis/releases#release-8.2.8

tianon: Ah, that's what I get for trusting https://endoflife.date/redis somewhat blindly -- that's open source data though, and a common place users look too, so I'd suggest contributing there to get more accurate data. :eyes:

tianon (approved):

library/redis
Peter-Sh: @tianon Thanks, we'll take a look, at a first glance data looks pretty incorrect

PR #21952, opened 2026-07-30
Update Eclipse Temurin images to version 11.0.32 and 17.0.20

gdams:


tianon:
https://github.com/docker-library/official-images/pull/21636#issuecomment-4675348295

> This is a great example of the exact flavor of feature creep I was worried about originally and why I pushed back on the feature in the beginning. 😅
> 
> It's not a bad feature / tool, but it's especially and particularly wild for it to be part of the `eclipse-temurin` images directly.

gdams:
> [#21636 (comment)](https://github.com/docker-library/official-images/pull/21636#issuecomment-4675348295)
> 
> > This is a great example of the exact flavor of feature creep I was worried about originally and why I pushed back on the feature in the beginning. 😅
> > It's not a bad feature / tool, but it's especially and particularly wild for it to be part of the `eclipse-temurin` images directly.

I do wonder whether we could split it out into it's own container image at some point. I'm not entirely sure what it would look like though

tianon:
Yeah, it's honestly a completely separate thing that's only barely related to Eclipse Temurin :sweat_smile:

adamfarley:
Overlaps with https://github.com/docker-library/official-images/pull/22002

gdams:
closing in favour of https://github.com/docker-library/official-images/pull/22002

PR #21968, opened 2026-08-03
Add Resolute, Update Semeru versions with 26_03 release

ramachandrabolisetti:
Added resolute for all releases.
Updated OE and CE releases 8.0.502.0,11.0.32.0,17.0.20.0,21.0.12.0,25.0.4.0,26.0.2.0.

### DIFF
--- a/library/ibm-semeru-runtimes
+++ b/library/ibm-semeru-runtimes
@@ -5,164 +5,247 @@ GitRepo: https://github.com/ibmruntimes/semeru-containers.git
 GitFetch: refs/heads/ibm
 
 #-----------------------------openj9 v8 images---------------------------------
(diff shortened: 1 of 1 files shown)


yosifkit (approved):

tianon:
As a heads-up on something pre-existing in these Dockerfiles -- not a blocker for this PR, but worth a future cleanup pass:

```dockerfile
ENV LANG='en_US.UTF-8' LANGUAGE='en_US:en' LC_ALL='en_US.UTF-8'
```

Setting all three of `LANG`, `LANGUAGE`, and `LC_ALL` feels like overkill -- `LANG` alone is probably enough. Setting `LC_ALL` in particular is a bit heavy-handed since it overrides everything else and can't be unset by users at runtime. Something to consider cleaning up when convenient. 😅

PR #21979, opened 2026-08-03
Add amazonlinux releases

ishwartg:
### Updated Packages for Amazon Linux 2023:
- amazon-linux-repo-cdn-2023.12.20260727-0.amzn2023
- system-release-2023.12.20260727-0.amzn2023

### DIFF
--- a/library/amazonlinux
+++ b/library/amazonlinux
@@ -12,19 +12,19 @@ Maintainers: Amazon Linux <amazon-linux@amazon.com> (@amazonlinux),
 GitRepo: https://github.com/amazonlinux/container-images.git
 GitCommit: cc7a1876866f4056fa73a789a5b758358151c189
 
-Tags: 2023, latest, 2023.12.20260724.0
+Tags: 2023, latest, 2023.12.20260727.0
 Architectures: amd64, arm64v8
 amd64-GitFetch: refs/heads/al2023
-amd64-GitCommit: 5a0863340c247026a6912f2e76976294cb9bf836
+amd64-GitCommit: 0c02d87a58138cbd40c56c00ca144091415cdce7
 arm64v8-GitFetch: refs/heads/al2023-arm64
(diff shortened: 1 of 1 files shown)


yosifkit (approved):

library/amazonlinux
tianon, line 32: I see `amazonlinux:2` was updated here, but https://aws.amazon.com/amazon-linux-2/ seems pretty clear that it's EOL 😅  is it still getting updates?

Also, I'm not sure how we missed that `amazonlinux:1` is very clearly not getting updates for a long time now 🙈

If these are EOL, they should be removed from this list.  As a reminder/background info, removing tags here will remove them from the "Supported" section on the Hub readme (and will prevent us from spending cycles rebuilding them on the official build servers), but the tags will still be available to users who want them.  (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)

tianon: Looking through https://github.com/docker-library/repo-info/commit/3d2cc12b84679f19ba70a699ce3472913064a496, it appears this was _just_ updating the URLs to a newer snapshot, not any package updates (which tracks, given that it's EOL).

sktomer: Let me take this feedback back to Amazon Linux leaders and decide on whether we should remove AL2 and AL1. AL1 I'm pretty sure we should.

tianon: FYI affected maintainers for the Amazon Linux 2 EOL (whether it gets removed or not, you might want to take action based on it being EOL):

- `amazoncorretto`: @corretto @lutkerd @benty-amzn @alvdavi @rgithubli @mrserb @Rudometov @elifaslan1 @chadrako @ianrichr @synecdoche @cost0much @midver @mmm-choi @satyenme
- `swift`: @tkremenek @shahmishal @hamin @swizzlr

PR #21977, opened 2026-08-03
Update Debian (notably, complete `mips64le` removal)

tianon:
This contains the inevitable result of https://lists.debian.org/debian-devel-announce/2025/11/msg00001.html -- as bookworm is now LTS, mips64le is no longer supported, and has been officially dropped from the official Debian archives now, so there is no longer _any_ active suite with support for `mips64le`.

This also means the end of `mips64le` within DOI at large (as our primary build system is running Debian Bookworm and is thus affected by this as well, but also marks the end of any base image providing `mips64le` images -- it also couldn't come at a better time, given that our build server has been just barely limping along for ages now).

PR #21980, opened 2026-08-03
Remove arm32v5 and s390x from bookworm images

tianon:
The Debian LTS hammer cometh down (https://github.com/docker-library/official-images/pull/21977)


tianon:
I expect fireworks from CI here, but hopefully the diff is really straightforward (and any `Dockerfile` changes that might've slipped in here are a bug).

tianon:
(we should restart at least the "Naughty" job here once https://github.com/docker-library/official-images/pull/21977 is merged)

yosifkit (approved):

tianon:
Adorable case of failing GitHub UX -- fully green mergeability, but if you click into the job, it clearly failed to generate a matrix.

https://github.com/docker-library/official-images/actions/runs/30960655832?pr=21980

image

image

PR #21983, opened 2026-08-04
Update ELK images for 9.5.0

vincentDeuschle:


library/elasticsearch
```diff
@@ -17,3 +17,8 @@ Tags: 9.4.4
 Architectures: amd64, arm64v8
 GitFetch: refs/heads/9.4
 GitCommit: e62732912302a282ec0de973e7ed140c69a31520
```
tianon, line 19: Are these versions still actively supported (across all three products)?

tianon (approved):

yosifkit:
Closing in favor of https://github.com/docker-library/official-images/pull/22033 (otherwise there will be conflicts over there)

PR #21987, opened 2026-08-04
library/perl: 5.40.4 -> 5.40.5, 5.42.2 -> 5.42.3

zakame:
- http://nntp.perl.org/group/perl.perl5.porters/271122
- http://nntp.perl.org/group/perl.perl5.porters/271123
- https://github.com/Perl/docker-perl/pull/215
- Update supported architectures per https://github.com/docker-library/official-images/pull/21977


tianon (approved):
Thank you! :heart:

PR #22000, opened 2026-08-06
Update Jetty versions to 12.0.38 & 12.1.12

lachlan-roberts:
- Update Jetty versions to `12.0.38` & `12.1.12`
- Drop dedicated base images for al2023 now that al2 is EOL.

### DIFF
--- a/library/jetty
+++ b/library/jetty
@@ -1,160 +1,144 @@
-Maintainers: Greg Wilkins <gregw@webtide.com> (@gregw),
-             Lachlan Roberts <lachlan@webtide.com> (@lachlan-roberts),
+Maintainers: Lachlan Roberts <lachlan@webtide.com> (@lachlan-roberts),
              Olivier Lamy <olamy@webtide.com> (@olamy),
              Joakim Erdfelt <joakim@webtide.com> (@joakime)
 GitRepo: https://github.com/eclipse/jetty.docker.git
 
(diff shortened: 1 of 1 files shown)


yosifkit:
Are you certain that you want to change `FROM amazoncorretto:11-al2023` to `FROM amazoncorretto:11`? This means that you'll be beholden to `amazoncorretto`'s next "default" tag change. Since you are tagging your images with `-al2023`, it could make sense to be just as specific with your parent image specification (and then provide "plain" tags for jetty users), but it is up to you.

You are losing the `al2023` tag combinations like `jetty:12.0.38-jdk25-al2023-amazoncorretto` where you previously had `jetty:12.0.37-jdk25-al2023-amazoncorretto`. Did you want to keep those options?

lachlan-roberts:
@yosifkit I will discuss with the team and get back to you.

lachlan-roberts:
@yosifkit I have made some updates so the amazoncorretto images are now pinned to al2023 specific tags. 

All of the same Jetty tag options such as `jetty:12.0.38-jdk25-al2023-amazoncorretto` and `jetty:12.0.38-jdk25-amazoncorretto` are still maintained, with both using the same dockerfile with the `amazoncorretto:25-al2023` base image.

yosifkit (approved):

PR #22005, opened 2026-08-06
Update haproxy

yosifkit:
Changes:

- https://github.com/docker-library/haproxy/commit/ca63f43: Update 3.5 to 3.5-dev4


tianon:
https://github.com/docker-library/haproxy/commit/ca63f43 looks good and has successful builds

PR #22009, opened 2026-08-06
Remove `mips64le` from the set of "supported architectures"

tianon:
As of Debian Bookworm going LTS, we no longer have a supported way to build images successfully for this architecture (https://github.com/docker-library/official-images/pull/21977).


yosifkit (approved):
👋

PR #22028, opened 2026-08-10
Telegraf v1.39.3

srebhan:
Thank you!


yosifkit (approved):

srebhan:
@yosifkit any reason you didn't merge?

tianon:
The logic here is a little weird, but I think it's probably fine :+1:

srebhan:
@tianon do you mean the logic in our PR? Sorry :confused:.. ;-)

tianon:
Yeah, it's kind of strange to remove root from all groups in the image just so that users can use `--group-add` without _also_ using `--user` (because their deployments will have stronger security guarantees if they use `--user` explicitly too).

Basically the logic you've added here makes it so that `--user` can be semi-implied when not specified, but in a way that preserves supplementary groups and that doesn't guarantee to the user that their runtime will run as non-root.  I think it's fine, it's just surprising.  (If I, as a user, am going to the trouble of doing `--group-add`, adding `--user` to increase the security of my runtime is an easy step :sweat_smile:)

PR #3424, opened 2017-09-07
Update openjdk to non-headless on non-slim images

yosifkit:
Add test for openjdk non-headless: uimanager-font

contains: https://github.com/docker-library/openjdk/pull/143

PR #22092, opened 2026-08-19
Update for Corretto releases: 8.504.01.1, 11.0.32.10.1, 17.0.20.10.1, 21.0.12.9.1, 25.0.4.8.1, 26.0.2.11.1

jyma22:
Update amazoncorretto for latest release. For more details, see the following:
* [corretto/corretto-docker repo](https://github.com/corretto/corretto-docker/commit/main)
* [corretto/corretto-8 release](https://github.com/corretto/corretto-8/releases/tag/8.504.01.1)
* [corretto/corretto-11 release](https://github.com/corretto/corretto-11/releases/tag/11.0.32.10.1)
* [corretto/corretto-17 release](https://github.com/corretto/corretto-17/releases/tag/17.0.20.10.1)
* [corretto/corretto-21 release](https://github.com/corretto/corretto-21/releases/tag/21.0.12.9.1)
* [corretto/corretto-25 release](https://github.com/corretto/corretto-25/releases/tag/25.0.4.8.1)
* [corretto/corretto-26 release](https://github.com/corretto/corretto-26/releases/tag/26.0.2.11.1)

### DIFF
--- a/library/amazoncorretto
+++ b/library/amazoncorretto
@@ -15,57 +15,57 @@ Maintainers: Amazon Corretto Team <corretto-team@amazon.com> (@corretto),
              Satyen Subramaniam <satyenme@amazon.com> (@satyenme)
 GitRepo: https://github.com/corretto/corretto-docker.git
 GitFetch: refs/heads/main
-GitCommit: a4a86adc91afa6ae78207899cd01bd88c9b46322
+GitCommit: 10edb080f80d1f4d9dfaa8bc8242d016c6676fd3
 
-Tags: 8-al2, 8u502-al2, 8-al2-full, 8-al2-jdk, 8-al2-generic, 8u502-al2-generic, 8-al2-generic-jdk
+Tags: 8-al2, 8u504-al2, 8-al2-full, 8-al2-jdk, 8-al2-generic, 8u504-al2-generic, 8-al2-generic-jdk
 Architectures: amd64, arm64v8
 Directory: 8/jdk/al2-generic
 
-Tags: 8-al2023, 8u502-al2023, 8-al2023-jdk, 8, 8-jdk, 8u502, latest
+Tags: 8-al2023, 8u504-al2023, 8-al2023-jdk, 8, 8-jdk, 8u504, latest
 Architectures: amd64, arm64v8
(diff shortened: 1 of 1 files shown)


tianon:
Is it intentional that `--setopt=install_weak_deps=False` wasn't added to the 11-headless image?

Also, this causes the `java-uimanager-font` test to fail which is _probably_ expected?

cost0much:
11 AL template hasn't received the same "Recommends" patch for its dependencies that 17+ have. Yes, it's intentional.

It's expected that the headless images will now have less packages installed, as previous non-required dependencies are no longer installed. Although, I was hoping that this change wouldn't have too much downstream effects, so these failing checks still warrant some concern. I assume these tests are set by dockerhub? What is dockerhub's recommendation for this change? We have some customers wanting a slimmer headless image, but simultaneously it looks like this change might potentially affect current users. It's not too late to revert it, however.

yosifkit:
> Although, I was hoping that this change wouldn't have too much downstream effects, so these failing checks still warrant some concern. I assume these tests are set by dockerhub? What is dockerhub's recommendation for this change?

We just run 3 small tests that exercise the jdk as a way to have all the different JDKs on equal ground. The specific [`java-uimanager-font` test](https://github.com/docker-library/official-images/blob/b412dcc924239a679a2448db3de28e8d05f67135/test/tests/java-uimanager-font/container.java) was added way back in https://github.com/docker-library/official-images/pull/3424 in response to https://github.com/docker-library/openjdk/issues/129. The test isn't required for images where it is expected to fail and we could adjust it to be skipped on `amazoncorretto:*headless*` images. I think it is fine if you want to decrease the size of the headless images though it may cause breakage for some users.

Rudometov:
What's the size savings here? If not too much, maybe it's not worth it.

cost0much:
After internal discussion, we're going to revert the `--setopt=install_weak_deps=False` change and re-release it at a later date. This PR can be closed in favor of the new one.

tianon:
> What's the size savings here? If not too much, maybe it's not worth it.

The JDK layer before was 295MB, and after it's 267MB (for `26-al2023-headless` specifically).

PR #22095, opened 2026-08-19
SapMachine August 2026 updates

raeessap:
This updates the SapMachine docker images to the August 2026 releases

### DIFF
--- a/library/sapmachine
+++ b/library/sapmachine
@@ -1,282 +1,282 @@
 Maintainers: Christoph Langer <sapmachine@sap.com> (@realclanger), Arno Zeller <sapmachine@sap.com> (@ArnoZeller)
 GitRepo: https://github.com/SAP/SapMachine-infrastructure.git
 
-Tags: jre-headless, jre-headless-ubuntu, 26-jre-headless, 26-jre-headless-ubuntu, 26.0.2-jre-headless, 26.0.2-jre-headless-ubuntu, jre-headless-ubuntu-noble, jre-headless-ubuntu-24.04, 26-jre-headless-ubuntu-noble, 26-jre-headless-ubuntu-24.04, 26.0.2-jre-headless-ubuntu-noble, 26.0.2-jre-headless-ubuntu-24.04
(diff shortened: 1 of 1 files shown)


tianon:
```diff
-    gpg --no-default-keyring --keyring gnupg-ring:/etc/apt/trusted.gpg.d/sapmachine.gpg --batch --keyserver hkps://keys.openpgp.org --recv-keys CACB9FE09150307D1D22D82962754C3B3ABCFE23 && \
-    chmod 644 /etc/apt/trusted.gpg.d/sapmachine.gpg && \
-    echo "deb http://dist.sapmachine.io/debian/$(dpkg --print-architecture)/ ./" > /etc/apt/sources.list.d/sapmachine.list && \
+    gpg --no-default-keyring --keyring gnupg-ring:/usr/share/keyrings/sapmachine.gpg --batch --keyserver hkps://keys.openpgp.org --recv-keys CACB9FE09150307D1D22D82962754C3B3ABCFE23 && \
+    chmod 644 /usr/share/keyrings/sapmachine.gpg && \
+    wget -qO /etc/apt/sources.list.d/sapmachine.sources \
+      https://dist.sapmachine.io/debian/sapmachine.sources && \
```

Moving the keyring off `/etc/apt/trusted.gpg.d` (system-wide -- every entry there is trusted for *all* repos, not just this one) to a scoped keyring under `/usr/share/keyrings/`, and picking up the deb822 `.sources` format with a `Signed-By:` field, is a real improvement.

But downloading that `.sources` file with `wget` and no verification of any kind is a blocker. This is the file that decides which repo URI and which key everything downstream trusts, and it's coming from an unpinned, unversioned URL -- its content can change silently between builds with nothing in this repo to show it happened, and there's not even a checksum, let alone a signature, on it. That's the same problem as pulling a config snippet from a branch-name URL: it's mutable, unauditable, and unreproducible, just applied to the one file that matters most here.

It's only 6 lines -- I'd rather see it reproduced directly in the `Dockerfile`, the same way the old `deb ...` line was inline and visible in every diff:

```dockerfile
    { \
        echo 'Types: deb'; \
        echo 'URIs: https://dist.sapmachine.io/debian'; \
        echo 'Suites: stable'; \
        echo 'Components: main'; \
        echo 'Architectures: amd64 arm64 ppc64el'; \
        echo 'Signed-By: /usr/share/keyrings/sapmachine.gpg'; \
    } > /etc/apt/sources.list.d/sapmachine.sources && \
```

That keeps the deb822/`Signed-By:` upgrade while keeping the trust anchor visible and diffable like everything else here. As a bonus, `wget` (and its install/remove pair) can be dropped entirely once this is inlined.

This applies identically across all 32 Ubuntu-based variants (headless/full, 22.04/24.04, all four Java versions) since they all pick up the same pattern.

raeessap:
@tianon thank you for your valuable input. i have made the changes accordingly. Please review.

tianon (approved):

PR #22113, opened 2026-08-20
Update openjdk

yosifkit:
Changes:

- https://github.com/docker-library/openjdk/commit/da2d75fd: Update 27 to 27


tianon:
Nice, release candidate!  (from https://openjdk.org/projects/jdk/27/, GA is expected 2026/09/15)

PR #22147, opened 2026-08-27
Update 6 to 6.61.0

### LINKED ISSUE #22081: [Ghost Docker Image] - Labeling breaking changes
We're planning on making some structural/breaking changes to the Ghost official image (https://github.com/TryGhost/docker-library-ghost) in the lead-up to Ghost 7.0. The image breaking changes are separate from any breaking changes that Ghost 7.0 itself will have, and given that 7.0 is still a ways away for Ghost itself (making a 7.0-rc image not really feasible here) we'd like to create a variant of the existing 6.x images with said breaking changes. This new variant would allow the community to test the new changes + update any tooling in preparation for Ghost 7.0 down the line.

The question I have it - what would make the most sense from a labeling perspective for this image? I'm currently thinking to use `-next` as the variant name, but didn't see any recommended pattern for this 🤔 

Also curious for feedback on some of the breaking changes we want to make (and how closely they do/don't align with the recommended official-images pattern). Ultimately the idea is to align the official image with the one Ghost currently builds from its own repo (which we're planning on making a "nightly" release) The initial list of changes is:

- removal of Ghost-CLI (small)
- move `/var/lib/ghost` (and the associated content volume) into `/home/ghost` (biggest change since it affects volume mounts)
- remove the `node` user and add `ghost` with uid 1000 (ideally should be small unless someone's doing something with USER node in a downstream image)

Ghost-Slimer:
Update 6 to 6.61.0

Generated from https://github.com/TryGhost/docker-library-ghost/commit/820b574a83ddbf062f6ad1d5e77eb455a954ebc7.

closes #22081

### DIFF
--- a/library/ghost
+++ b/library/ghost
@@ -2,7 +2,7 @@
 
 Maintainers: Austin Burdine <austin@ghost.org> (@acburdine)
 GitRepo: https://github.com/TryGhost/docker-library-ghost.git
-GitCommit: 820b574a83ddbf062f6ad1d5e77eb455a954ebc7
+GitCommit: 56d19a954c21ff611956fe9927535a3267801d19
 
 Tags: 6.61.0-bookworm, 6.61.0, 6.61-bookworm, 6.61, 6-bookworm, 6, bookworm, latest
 Directory: 6/bookworm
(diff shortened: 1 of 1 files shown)


yosifkit:
I usually reject PRs that attempt to use a multi-stage build since they are often just avoiding clean up steps. But this is a little more nuanced than that. The differences that I see compared to just building `FROM node:...`:

- renaming the `node` user to `ghost`
- not having to clean up build-time packages or caches
- not shipping yarn v1 in the image

While I would reject if it was just doing the first two since a multi-stage build just adds complication and those can be more simply accomplished without it, I am enticed by the removal of yarn. That is the default in the newest node images, but if it is reasonable to also remove `npm` (`npm uninstall -g npm` in the build stage) I could be convinced of the multi-stage build.

acburdine:
@yosifkit The original approach just copied the node binary from the base image, the reason I copied npm as well is that Ghost's adapters are often shipped as npm packages, and it's easiest to install them by fetching the tarball from npm - `npm pack <pkg>@version` has that built in.

acburdine:
also see https://github.com/docker-library/official-images/issues/22081#issuecomment-5331682389 - my understanding of this bit of @tianon's comment:

> For removing the user, you might consider some surgical COPY --from=node

was to use multi-stage, though I might have misunderstood what was meant there

tianon:
I think my conceptual was maybe more surgical than just copying everything - if you're just copying everything, renaming the user instead is pretty lightweight (and the other tool cleanup code already exists :sweat_smile:).

acburdine:
Made https://github.com/TryGhost/docker-library-ghost/pull/482 switching it back to a single-stage image. Adds about ~17mb of overhead but that's probably fine, esp since this next variant is still several hundred mb smaller than the current one overall.

yosifkit:
I think https://github.com/TryGhost/docker-library-ghost/pull/482 looks good.

yosifkit (approved):

PR #22153, opened 2026-08-27
Include both new *and* old `Maintainers:` in our comment pings

### LINKED ISSUE #20365: improve PR review comments to cc old *and* new maintainers when the maintainer list is changed
When image maintainers are added or removed from the `Maintainers:` list, we only ping the old maintainers, not the new ones, and that causes various issues like us missing typos in their GitHub handle, them not being aware it happened, etc, so we should refactor that (aging) automation to fix that.

tianon:
Previously, we only included the pre-existing maintainers, which leaves a gap on new-image PRs *and* on changed-maintainers PRs where the new maintainers don't see our pings until the first *other* PR, and that's not ideal.

- fixes https://github.com/docker-library/official-images/issues/20365

Assisted-By: "claude my eyes right out" (but I had to babysit it so badly that this might as well be original work; classic 🙃)

### DIFF
--- a/.github/workflows/munge-pr.yml
+++ b/.github/workflows/munge-pr.yml
@@ -143,14 +143,27 @@ jobs:
       - name: Gather Maintainers
         env:
           IMAGES: ${{ needs.gather.outputs.images }}
+          PR_HEAD_SHA: ${{ github.event.pull_request.head.sha }}
         run: |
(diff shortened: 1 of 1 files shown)


tianon:
If we get this into a state we're happy with, I've pushed https://github.com/infosiftr/stackbrew/compare/tianon-sso which we can use to test it out quickly after merge.

yosifkit (approved):

PR #22155, opened 2026-08-28
Add 'latest' and '8' tags back to mongo shared tags

### LINKED ISSUE #768: mongo 8 and latest dropped
https://github.com/docker-library/official-images/pull/21976 dropped the shared tags for `8` and `latest`. We now therefore end-up with `8.2.12` which is EOL. Please add them back on `8.3` as this is the currently supported version for `8` and `latest` according to documentation.

Falco20019:
Fixes https://github.com/docker-library/mongo/issues/768

### DIFF
--- a/library/mongo
+++ b/library/mongo
@@ -5,7 +5,7 @@ Maintainers: Tianon Gravi <admwiggin@gmail.com> (@tianon),
 GitRepo: https://github.com/docker-library/mongo.git
 
 Tags: 8.3.8-noble, 8.3-noble
-SharedTags: 8.3.8, 8.3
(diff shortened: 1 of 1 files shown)


yosifkit:
> `# this file is generated via https://github.com/docker-library/mongo/blob/e0800cc3a844163ece7ecc8e270da5fe6e70eec0/generate-stackbrew-library.sh`

Since the `library/mongo` file is incorrect, the script needs to be updated first: https://github.com/docker-library/mongo/pull/767. That will make sure that the automation doesn't just try to remove these tags again.

Falco20019:
Thanks for creating the PR on the script. Looking forward to see this fixed 👍

PR #22194, opened 2026-09-02
Update clojure

Quantisan:


tianon:
https://wiki.debian.org/LTS/ :eyes: (bookworm becoming LTS means no more s390x)

yosifkit (approved):

tianon:
To be clear, a lot of your tags _can_ keep s390x, if you want to, but bookworm tags in specific need overrides to remove it, if you want to keep it elsewhere. :sweat_smile:

tianon:
(but if you'd rather just drop the arch completely, that's totally fine too)

cap10morgan:
> (but if you'd rather just drop the arch completely, that's totally fine too)

I'd be shocked if anyone was using it 🤣

PR #22227, opened 2026-09-08
Add Nim 2.2.12.

moigagoo:


library/nim
yosifkit: As discussed -- a one-time backfill of historical versions is totally fine. Just keeping in mind that "supported" means these would get new releases if a security issue were found, so we'll be looking forward to a follow-up to trim the list down to just the latest version of each actively supported series (2.2, etc) once everything's published. :+1:

(Removing tags here will remove them from the "Supported" section on the Hub readme and will prevent us from spending cycles rebuilding them on the official build servers, but the tags will still be available on Docker Hub to users who want them.)

yosifkit (dismissed):

library/nim
moigagoo: Thanks for the reminder. Let's not delay this change. I've kept only the latest 2.2.x and 2.0.x versions.

yosifkit (approved):

PR #5372, opened 2019-01-30
Update websphere-liberty

arthurdm:
- We have added new configuration XML snippets which are available for customers when they're building their derived image, such as to add support for Hazelcast
- Small fix in the kernel (for user's 1001 home)

PR #22228, opened 2026-09-08
Updates for WebSphere and Open Liberty 26.0.0.9

wraschke:


tianon:
I'm really concerned by the sheer amount of Docker-image-custom shell scripts, XML files, etc that continue to grow in these images -- it's already a lot, and it's grown again here.  It's not normally a blocker by itself, but it's definitely a huge "code smell" that makes me wonder how much of this behavior is really container-specific and how much should really be part of upstream?  Are the types of things you're doing in shell scripts and extra XML files that get copied into the image unique to containers, or are they things that non-container users might _also_ find useful such that they should be part of Liberty itself or at least part of the upstream Liberty distribution packages?

leochr:
@tianon Thank you for the review and comment. The changes are container-specific. A helper script to assist our users manage packages as part of their container build for applications. The other changes are for security hardening (i.e. disable insecure HTTP port, disable dynamic config update in containers) to run Liberty securely in containers. Those configurations can't be setup upstream as it would break existing users and dev tools. We'll keep your comment in mind to continue to add only container-specific changes. Thank you.

tianon:
It's worth noting this isn't new: I raised basically the same question — how much of this is really container-specific vs. things that should live upstream — back in https://github.com/docker-library/official-images/pull/5372#issuecomment-459207701 (in 2019), and the trend just kept compounding since.

One concrete example from this PR: it adds `helpers/build/package-helper.sh` into all three `kernel-*` variant directories (both `26.0.0.9-kernel-*` and the floating `kernel-*` paths, six copies).  Its `cmd_install()` calls `microdnf` and `rpm -qa`/`-q`/`-ql` throughout, but every one of those directories builds from a `Dockerfile.ubuntu.*` file using `apt-get`. Nothing in this diff appears to actually invoke the script, so as far as I can tell it's completely dead code here.

leochr:
@tianon The new set of changes are container specific. `package-helper.sh` is a helper script that we package with Liberty containers to help users install additional packages. The users run it from their Dockerfiles. At this time, package helper supports microdnf, but upon user requests we would consider enhancing it to support other package managers. The helper scripts are common for all OS variants (UBI, Ubuntu), hence they are included and packaged from the Git repo. 

Could you please help to merge this PR as it's blocking the release and is now overdue and users are waiting for it? 
Thank you.

tianon:
The issue is that none of these images are UBI-based and the script is being copied into Ubuntu-based images -- in which it doesn't work / can't be used.

wraschke:
@tianon Leo made some changes to the Ubuntu Dockerfiles in WASdev/ci.docker.

I updated the websphere-liberty file to reference that new commit.

Assuming the jobs (still in progress) complete, we hope you will merge.

leochr:
@tianon To address your review comment, we made changes so that `package-helper.sh` is not included in the Ubuntu images. Please approve and merge. Thank you.

PR #22173, opened 2026-08-31
Update Groovy to 6.0.0-beta-3 and 5.1.1

keeganwitt:

PR #22239, opened 2026-09-09
eclipse-temurin updates for August 2026 CSPU

smlambert:


library/eclipse-temurin
```diff
@@ -7,908 +7,940 @@ GitFetch: refs/heads/main
...
-SharedTags: 8u502-b07-jdk, 8-jdk, 8
+Tags: 8u504-b01-jdk-resolute, 8-jdk-resolute, 8-resolute
+SharedTags: 8u504-b01-jdk, 8-jdk, 8
 Architectures: amd64, arm64v8, ppc64le
-GitCommit: 33e90a8e65658c51e78419e09f12fb86de0fa3e3
+GitCommit: 511f9356dc4d50932a0a5f8cfb0f87ed1aef4f07
 Directory: 8/jdk/ubuntu/resolute
 
-Tags: 8u502-b07-jdk-noble, 8-jdk-noble, 8-noble
+Tags: 8u504-b01-jdk-resolute-arm32v7, 8-jdk-resolute-arm32v7, 8-resolute-arm32v7
```
tianon, line 36: There's some discussion of this over in https://github.com/docker-library/official-images/pull/22173#issuecomment-5516735795, but I don't think this is the right answer here.  Having explicit tags just for arm32v7 pushes this problem all the way out to the actual UX of the image repo, but doesn't solve any of the underlying or follow-on problems of arm32v7 coming and going.

I'm honestly curious whether arm32v7 is really worth going to any trouble for, let alone this much trouble?  The Docker images definitely don't have a huge userbase, so I'm curious whether you see a large userbase for it upstream still?

tianon: cc @gdams @sxa (since the bot didn't manage to comment successfully given the sheer size of the diff here)

gdams: I think my primary concern here is that JDK8 is an active LTS release stream. In the past when we've suggested deprecating the binary builds we've received some push back from the community. My idea with this approach was that it wouldn't block us getting the rest of JDK8 shipped but it still allowed arm32 users to receive security updates.

yosifkit: These arch specific tags are likely a no go; as seen from the CI failures, I'm not sure where else in our DOI build system that this might fail.

I am fine if the architecture comes and goes from the regular tags. In either case, the downstream DOI consumers have to deal with parent images being removed. For users outside of DOI that need a consistently pullable `arm32v7` image, they could pull it from [`arm32v7/eclipse-temurin:8-jdk`](https://hub.docker.com/r/arm32v7/eclipse-temurin/) whether or not it is currently in the `eclipse-temurin:8-jdk` image index.

---

A possible suggestion that could remove the need to rebuild the other architectures when `arm32v7` is added back is moving to GPG signatures instead of embedded checksums in the Dockerfile. In this way, the Dockefile wouldn't change when the architecture is added back since the version bump to update the download URLs would have already happened. I don't know if it is worth the effort or not since users seem to like more rebuilds anyway.

alexanderveit: > cc @gdams @sxa (since the bot didn't manage to comment successfully given the sheer size of the diff here)

This is not the first time that this problem stops the whole release process.

alexanderveit:
Do you know if there’s any chance the images will be released anytime soon?

tianon:
Can we get a simplified version of this with just the version bumps to unblock the releases and revisit the architecture split conversation separately?

PR #22245, opened 2026-09-10
InfluxDB OSS and Enterprise v1.13.1

gwossum:
- Bump InfluxDB Enterprise to v1.13.1
- Add InfluxDB OSS v1.13.1

### DIFF
--- a/library/influxdb
+++ b/library/influxdb
@@ -4,18 +4,26 @@ Maintainers: Brandon Pfeifer <bpfeifer@influxdata.com> (@bnpfeife),
              Praveenkumar Hemakumar <pkumar@influxdata.com> (@praveen-influx),
              Trevor Hilton <thilton@influxdata.com> (@hiltontj)
 GitRepo: https://github.com/influxdata/influxdata-docker
-GitCommit: c4978d4a9ebb34f1b8aebfabd1846d294b0f3852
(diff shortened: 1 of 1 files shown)


yosifkit (approved):

library/influxdb
```diff
@@ -4,18 +4,26 @@ Maintainers: Brandon Pfeifer <bpfeifer@influxdata.com> (@bnpfeife),
...
-Tags: 1.13-data, 1.13.0-data
+Tags: 1.13, 1.13.1
+Architectures: amd64, arm64v8
+Directory: influxdb/1.13
+
+Tags: 1.13-alpine, 1.13.1-alpine
+Architectures: amd64, arm64v8
+Directory: influxdb/1.13/alpine
+
+Tags: 1.13-data, 1.13.1-data
```
yosifkit, line 17: Are the `data`, `meta`, `data-alpine`, and `meta-alpine` tags still staying on `1.12.x` or will they move at a specified future release or perhaps not move at all?

gwossum: This is intentional. We want more testing of 1.13 before making it the default for everyone. It's a low-risk release, but we always like playing a little safe with the 1.x releases. 1.13 should become the default by the time 1.13.2 is released.

PR #22248, opened 2026-09-11
Update Nextcloud

J0WI:


yosifkit (approved):

yosifkit:
Just a heads up that the s390x builds of these images will be delayed by up to a week. Our build infrastructure from IBM is offline for maintenance/replacement and is expected to be restored by 16 September 2026.

PR #22255, opened 2026-09-12
Update Storm to 3.1.0

reiabreu:
Following the 3.1.0 images added in storm-docker (apache/storm-docker#3), merged as:
https://github.com/apache/storm-docker/commit/31a83426d69a1e3c5bfc0b77fc135da2ed3a79fd

Adds `3.1.0` and `3.1.0-lite` (Java 25) and moves the `latest` tag to 3.1.0.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


library/storm
```diff
@@ -12,10 +12,18 @@ Tags: 2.8.9-jre21, 2.8-jre21
 GitCommit: d6fd2bc163a48be40cac2b5a1142c5d317d2d438
 Directory: 2.8.9-jre21
 
-Tags: 3.0.0-jre25, 3.0-jre25, 3.0.0, 3.0,latest
+Tags: 3.0.0-jre25, 3.0-jre25, 3.0.0, 3.0
 GitCommit: 1e325fe56337922f3c06a5f790a5c4f7170acb4f
 Directory: 3.0.0
 
 Tags:  3.0.0-lite-jre25, 3.0-lite-jre25, 3.0.0-lite, 3.0-lite
```
yosifkit: Any particular reason to keep the 3.0.x and 2.8.x versions listed here?  Removing it here will remove it from the "Supported" section on the Hub readme (and will prevent us from spending cycles rebuilding it on the official build server), but the tag will still be available to pull from Docker Hub for users who want it.  (See https://github.com/docker-library/official-images#library-definition-files for more detail on this.)

yosifkit (dismissed):

library/storm
rzo1: No.

rzo1: @reiabreu Don't have write access, but can be removed imho :)

reiabreu: Makes total sense, I apologize, did not consider that. Fixing

reiabreu: Fix has been pushed. Thank you

yosifkit (approved):

PR #22260, opened 2026-09-14
elixir v1.19.6

getong:


getong:
This pr is related to elixir 1.19.6., which is all passed.
But 1.16, 1.17. 1.18 does not pass, it might be some os mirror error, which is not related to this pr.

yosifkit:
Build errors in the Erlang `24` and `25` based images are related to the end of life for Debian Bullseye. The mirrors have been giving varied errors for the past week. I recommend removing `debian:bullseye` based images.

`debian:bullseye` images will likely be removed as a supported DOI soon, especially since images from it will now fail to build.

> Starting in September, Debian will not provide further security updates for Debian 11.
>
>\- https://www.debian.org/News/2026/20260831

yosifkit (approved):

yosifkit:
Merging to get the new update. The bullseye removal can be done after its removal from `debian` images.

PR #22266, opened 2026-09-15
in erlang/otp, update bookworm , deprecate bullseye

getong:


getong:
https://github.com/docker-library/official-images/pull/22260

yosifkit (approved):

PR #22270, opened 2026-09-15
Update Ubuntu 2026-09-15

ROCKsBot:
Update Ubuntu to the latest published OCI tarballs as of 2026-09-15.


tianon (approved):

tianon:
Just a heads up that the s390x builds of these images will be delayed by a few days. Our build infrastructure from IBM is offline for maintenance/replacement.

PR #22273, opened 2026-09-15
Official Convertigo 8.4.4 release!

nicolas-albert:
Update the Convertigo image to 8.4.4 (tag commit 7fd4b780d2eca81a5e48459fe547d6800e1013ea). Changelog: https://github.com/convertigo/convertigo/blob/8.4.4/CHANGELOG.md


tianon:
I'm honestly kind of concerned to see `/cacerts` here -- that feels like hefty scope creep to implement directly in the image entrypoint script (especially since logic like that inevitably will only grow more and more complex over time). :sweat_smile:

nicolas-albert:
Thanks for the feedback. I have updated the implementation to make the intent and behavior safer.

The use case comes mainly from enterprise customers running Convertigo behind corporate proxies performing TLS inspection. These environments require their private corporate root CA to be trusted for outbound HTTPS connections.

Historically, we documented how to manually add those CAs to the JVM `cacerts`. With Docker and especially Kubernetes/OpenShift, many of our customers now run the container as an arbitrary non-root `runAsUser`, so modifying the JDK truststore inside the container is neither convenient nor always possible.

The `/cacerts` mount is intended as a convenience for this specific case: an administrator can mount one or more corporate CA certificates read-only. At startup, before the JVM is launched, the entrypoint copies the standard JDK truststore to a writable temporary location and imports the additional CAs into that copy. The original JDK truststore is never modified.

I also changed the implementation so that it does not override an explicitly configured JVM truststore. If the user already provides `-Djavax.net.ssl.trustStore=...`, that configuration takes precedence and `/cacerts` is ignored.

So the resulting behavior is:

* default configuration: use the standard JDK truststore;
* `/cacerts` mounted: use the standard JDK trust anchors plus the mounted corporate CAs;
* explicit `javax.net.ssl.trustStore`: preserve and use the user's own truststore.

This lets Kubernetes/OpenShift administrators add their corporate CA with a simple read-only certificate mount, while keeping the standard Java truststore configuration available for users who need full control.

opicciotto:
Hello
Any news out this PR ? Thank you in advance ;)

tianon:
Sorry, to be clear: the extra `case`/`JAVA_OPTS` branching your last revision needed to handle the explicit-truststore edge case is exactly the kind of creeping complexity I meant in my first comment.

A related thought on this: convertigo is built `FROM tomcat:9-jdk21-temurin-jammy`, which means it's already carrying [`eclipse-temurin`'s own cacerts-import entrypoint script (`/entrypoint.sh`) in the base image](https://github.com/adoptium/containers/blob/33e90a8e65658c51e78419e09f12fb86de0fa3e3/21/jdk/ubuntu/jammy/entrypoint.sh) -- `docker-library/tomcat` didn't strip it out (yet - TBD), just reset `ENTRYPOINT []` so it doesn't run automatically (see docker-library/tomcat@55546d6). Given that, would it make sense to lean on that script instead of reimplementing the same truststore-copy-and-`keytool`-import logic here?  I'm not convinced that tomcat won't remove that file in the future (either explicitly or incidentally), but at least sharing *efforts* feels worthwhile if we're going to grow multiple disparate implementations of this same behavior in DOI (and having *separate* incompatible interfaces for them is not great either).

As an example of the types of silent and unexpected bugs: that script was disabled by default in `tomcat` specifically because it was re-execing via POSIX `sh` and silently dropping environment variables with dots in their names, which broke Spring Boot's externalized config (docker-library/tomcat#77, recurred as docker-library/tomcat#302, discussed upstream in adoptium/containers#415 -- this is an effort that has been *fraught* with problems and scope creep since it was introduced).

nicolas-albert:
Fair enough, and thanks for pointing at the Temurin script. I removed the `/cacerts` implementation entirely: the entrypoint now only delegates to `/__cacert_entrypoint.sh` when `USE_SYSTEM_CA_CERTS` is set (certificates mounted in `/certificates/*.crt`), and fails with an explicit message if the base image stops providing that script.
The explicit-truststore case needs no branching anymore, since a `-Djavax.net.ssl.trustStore` given on the command line takes precedence over the `JAVA_TOOL_OPTIONS` set by the script.
Verified on the image as root, as an arbitrary non-root UID, with and without an explicit truststore.
The documentation now points to the Temurin mechanism instead of describing our own.

tianon (approved):

PR #22280, opened 2026-09-15
Quote diff-pr's bashbrew output via jq's `@sh`

tianon:
We already practice extreme defense-in-depth here, but this adds some explicit hardening against not only explicit attacks (which would get access to nothing more interesting than they already have *better* access to via Actions on a personal repo) but also technically valid `Directory` values that *can* potentially contain wild characters like spaces.

With thanks to @the-red-ace and @allanvictor for finding and reporting this gap to us responsibly. :bow: :heart:


yosifkit (approved):

tianon:
Success: https://github.com/docker-library/official-images/pull/22281#issuecomment-5689784343 :+1:

PR #22288, opened 2026-09-17
Update mageia images

juanluisbaptiste:
Update mageia images


tianon:
Scrolling through the diff at https://github.com/docker-library/official-images/actions/runs/35181839358/attempts/1#summary-105075496277 and it's clear that installing `vim` is why the diff is so big -- did it get installed on purpose, or was that an unintentional addition?  (ie, should we roll with the big diff, which isn't a blocker by itself, or should we wait for an update?)

juanluisbaptiste:
It has always been long, vim is indirectly installed by `basesystem-minimal`:

```
# urpmq --requires-recursive basesystem-minimal|grep vim
vim-X11|vim-enhanced|vim-minimal
vim-minimal

```
So you can go ahead and merge it!

Thanks

tianon (approved):

PR #22291, opened 2026-09-17
Add distroless Telegraf image

srebhan:
This PR adds a new image without any distribution to minimize size and attack surface. We start publishing this for Telegraf v1.40, older versions will not be available.


srebhan:
This PR required #22293 fixing the tests for distribution-less images.

tianon:
Using the term "distroless" for something that isn't part of or officially related to [the distroless project](https://github.com/GoogleContainerTools/distroless) feels somewhat inappropriate; maybe instead document how to get to an image like this from another variant instead of officially supporting it here for now?

srebhan:
@tianon I'm open for another naming to signal "an image without a distribution". We are aiming at making this an "official" image because there is a demand from our users as this image has less moving parts potentially subject to security issues...

srebhan:
@tianon any update? Should I name the tag `1.40.0-scratch` instead?

tianon:
Sorry!  Yes, I think `scratch` is probably a better name (it does have some other precedence within DOI, but only `nats` :sweat_smile:).

Also, given it's the *exact* same binaries being downloaded and copied in, I'd suggest using something like `COPY --from=telegraf:xxx ...` instead of download/verifying them fresh.  All that other rootfs prep code definitely makes me nervous, but it's maybe unavoidable.

PR #22292, opened 2026-09-17
nats: Update to v2.15.0

neilalexander:
More info here: https://github.com/nats-io/nats-server/releases/tag/v2.15.0


library/nats
tianon: Looking at https://docs.nats.io/release-notes/ and 2.12 is EOL now (and thus should be removed), right? :eyes:

> Final release [v2.12.15](https://github.com/nats-io/nats-server/releases/tag/v2.12.15), 2026-08-12

tianon (dismissed):

library/nats
neilalexander: Normally we remove the tags for the just EOL'd release after the x.x.1 of the new release, hope that's OK.

neilalexander: Or I can remove them now if you prefer.

yosifkit (approved):

PR #22305, opened 2026-09-18
Update Debian (esp. 13.7 and bullseye full EOL)

tianon:


tianon:
FYI affected maintainers for the Debian Bullseye EOL/removal:

- `clojure`: @Quantisan @cap10morgan
- `haskell`: @tarleb @develop7
- `neurodebian`: @yarikoptic
- `perl`: @PeterMartini @zakame

yosifkit (approved):

PR #22307, opened 2026-09-18
library/perl: drop bullseye

zakame:
- https://github.com/Perl/docker-perl/pull/218
- cf https://github.com/docker-library/official-images/pull/22305


yosifkit (approved):
Thank you for the quick response! ❤️

zakame:
Thanks @yosifkit and @tianon ! 🚀 🎉

